PR #99, opened 2022-06-16
Fix externalRedis for correct template

xom4ek:
`externalRedis` in https://github.com/grafana/oncall/blob/dev/helm/oncall/templates/_env.tpl#L141 but here `external_redis`


shubhshekhar09:
Please commit this changes to main branch or else helm install will fail. As variable defined in tpl and value.yaml are different for externalRedis.

Konstantinov-Innokentii (approved):

PR #100, opened 2022-06-16
fix(helm): fix password key in external redis secret

adityacs:
Fix password key in external redis secret


adityacs:
Wrong upstream branch

Issue #98, opened 2022-06-16
Issue when trying to install with helm


PatMis81:
Hi, 
I have encountered an issue when trying to install the helm chart on my local kubernetes. I get the following message:

`Error: INSTALLATION FAILED: template: oncall/templates/engine/job-migrate.yaml:48:14: executing "oncall/templates/engine/job-migrate.yaml" at <include "snippet.redis.env" .>: error calling include: templa
te: oncall/templates/_env.tpl:157:12: executing "snippet.redis.env" at <include "snippet.redis.host" .>: error calling include: template: oncall/templates/_env.tpl:140:46: executing "snippet.redis.host" a
t <.Values.externalRedis.host>: nil pointer evaluating interface {}.host`

Regards,
Patrick

andy812:
Solution: change `external_redis` to `externalRedis` in values.yaml

PatMis81:
Thank you Andy! This worked!

andy812:
Be careful, there's a lot of errors: `nginx-ingress` instead of `ingress-nginx`, cert issuer issue, etc.
Please, don't close this bug until it will be fixed by Grafana team.

PatMis81:
Reopen for several bugs to be fixed by Grafana.

idv-Evgenii:
Hi, this error appears at startup and at the same time it starts everything except oncall celery and oncall engine, it initializes under oncall, but does not start

Error: INSTALLATION FAILED: failed post-install: warning: Hook post-install oncall/templates/cert-issuer.yaml failed: Internal error occurred: failed calling webhook "webhook.cert-manager.io": Post "https://******.svc:443/mutate?timeout=30s": service "test-cert-manager-webhook" not found

but nevertheless it starts grafana, but without oncall

idv-Evgenii:
Hi, this error appears at startup and at the same time it starts everything except oncall celery and oncall engine, it initializes under oncall, but does not start

Error: INSTALLATION FAILED: failed post-install: warning: Hook post-install oncall/templates/cert-issuer.yaml failed: Internal error occurred: failed calling webhook "webhook.cert-manager.io": Post "https://******.svc:443/mutate?timeout=30s": service "test-cert-manager-webhook" not found

but nevertheless it starts grafana, but without oncall

francis-cassbana:
Hi , I have disabled ingress false, eventhough its trying to install ingress in default namespace ,
[...]

PR #101, opened 2022-06-16
fix(helm): fix password key in external redis secret

adityacs:
Fix password key in external redis secret


xom4ek:
#98 one more bug with helm

iskhakov:
Thank you for the fix!

iskhakov (approved):

PR #104, opened 2022-06-16
Tweak docker-compose

mderynck:
- Remove setting for celery to exit
- Only bind required ports keep others internal

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,8 +1,3 @@
-version: "3"
-
-networks:
-  oncall:
-
 services:
   engine:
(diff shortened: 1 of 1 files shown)


Matvey-Kuk:
Why do they need their own network? :)

Matvey-Kuk (approved):

mderynck:
> Why do they need their own network? :)

We don't need it I removed that, the default network it creates suits our purposes already.
Instead I made the ports for mysql and redis not bind to the host in case the user has other instances.

Matvey-Kuk (approved):

PR #154, opened 2022-06-26
Update README.md

yasharne:
using `.env_hobby` and passing it via `--env-file .env_hobby` causes this error:
```
invalid interpolation format for services.mysql_to_create_grafana_db.command: "required variable MYSQL_PASSWORD is missing a value: err". You may need to escape any $ with another $
```
docker version: 
```
Client: Docker Engine - Community
 Version:           20.10.17
 API version:       1.41
 Go version:        go1.17.11
 Git commit:        100c701
 Built:             Mon Jun  6 23:02:46 2022
 OS/Arch:           linux/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.17
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.17.11
  Git commit:       a89b842
  Built:            Mon Jun  6 23:00:51 2022
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.6
  GitCommit:        10c12954828e7c7c9b6e0ea9b0c02b01407d3ae1
 runc:
  Version:          1.1.2
  GitCommit:        v1.1.2-0-ga916309
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```
Docker compose version:
```
Docker Compose version v2.6.0
```
OS: ubuntu 22.04

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,12 +33,12 @@ RABBITMQ_PASSWORD=rabbitmq_secret_pw
 MYSQL_PASSWORD=mysql_secret_pw
 COMPOSE_PROFILES=with_grafana  # Remove this line if you want to use existing grafana
 GRAFANA_USER=admin
-GRAFANA_PASSWORD=admin" > .env_hobby
+GRAFANA_PASSWORD=admin" > .env
 ```
 
 3. Launch services:
 ```bash
-docker-compose --env-file .env_hobby -f docker-compose.yml up --build -d
+docker-compose -f docker-compose.yml up --build -d
 ```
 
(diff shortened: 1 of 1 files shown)


Matvey-Kuk (changes requested):
Hi! We use ".env_hobby" in order to avoid clashing with ".env" used in the developer instruction. In the case both envs use the same file name it becomes hard to switch from one to another. May we think about another way of fixing it?

PR #176, opened 2022-06-29
Add configurable RabbitMQ protocol

ml-software-services:
Services like AmazonMQ hosted RabbitMQ use the AMQPS protocol, so this PR adds the option to change the default protocol for the helm deployment, but defaults to the normal AMQP protocol if not specified.


mderynck (approved):

iskhakov:
thank you for the PR, merging

PR #194, opened 2022-07-04
Manual incidents for all teams

Konstantinov-Innokentii:
- Add possibility to create manual incident for all teams, not only for General.

- Fix bugs in rendering source  links and links in web.

- Simplify manual incidents templates. Migration which reformat shape of manual incident alert's payload is introduced. 

image


Matvey-Kuk:
@Konstantinov-Innokentii could you please share a screenshot or video of how does it work?

engine/config_integrations/manual.py
```diff
@@ -12,39 +12,28 @@
...
 # Default templates
-slack_title = """{% set metadata = payload.view.private_metadata %}
-{%-if "message" in metadata -%}
-{% set title = "Message from @" + metadata.author_username %}
-{%- else -%}
-{% set title = payload.view.state["values"].TITLE_INPUT.FinishCreateIncidentViewStep.value %}
-{%- endif -%}
-*<{{ grafana_oncall_link }}|#{{ grafana_oncall_incident_id }} {{ title }}>* via {{ integration_name }}
+slack_title = """{{ payload.oncall.title }}
+ *<{{ grafana_oncall_link }}|#{{ grafana_oncall_incident_id }} {{ title }}>* via {{ integration_name }}
```
Ferril, line 16: I think it's better to start title from incident_id, like it was before. BTW what is {{ title }} here?
```suggestion
slack_title = """*<{{ grafana_oncall_link }}|#{{ grafana_oncall_incident_id }} {{ payload.oncall.title }}>*
 via {{ integration_name }}
```

Ferril (approved):

engine/apps/alerts/migrations/0004_auto_20220630_1343.py
```diff
@@ -0,0 +1,54 @@
...
+
+    return payload
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ('alerts', '0003_grafanaalertingcontactpoint_datasource_uid'),
+    ]
+
+    operations = [
```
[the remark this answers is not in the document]
iskhakov: The reasoning is that not only us running the service and OSS users can have troubles with migrations

Also I think we should start writing the migrations that are reversible

engine/apps/slack/scenarios/manual_incident.py
```diff
@@ -0,0 +1,658 @@
...
+MANUAL_INCIDENT_ORG_SELECT_ID = "manual_incident_org_select"
+MANUAL_INCIDENT_ROUTE_SELECT_ID = "manual_incident_route_select"
+MANUAL_INCIDENT_TITLE_INPUT_ID = "manual_incident_title_input"
+MANUAL_INCIDENT_MESSAGE_INPUT_ID = "manual_incident_message_input"
+
+DEFAULT_TEAM_VALUE = "default_team"
+
+
+# StartCreateIncidentFromMessage triggers creation of a manual incident from the slack message via submenu
+class StartCreateIncidentFromMessage(scenario_step.ScenarioStep):
```
iskhakov, line 21: Let's put comments under the class like we have in other places, e.g.
```
class EscalationDeliveryStep(scenario_step.ScenarioStep):
    """
    used for user group and channel notification in slack
    """
```

engine/apps/slack/scenarios/ancillary.py
```diff
@@ -0,0 +1,42 @@
+import logging
```
iskhakov, line 1: not sure if ancillary is a good name for the file, it should reflect the purpose of the scenario

engine/apps/slack/scenarios/manual_incident.py
```diff
@@ -0,0 +1,658 @@
...
+
+        channel_id = private_metadata["channel_id"]
+
+        input_id_prefix = private_metadata["input_id_prefix"]
+        selected_organization = _get_selected_org_from_payload(payload, input_id_prefix)
+        selected_team = _get_selected_team_from_payload(payload, input_id_prefix)
+        selected_route = _get_selected_route_from_payload(payload, input_id_prefix)
+
+        user = slack_user_identity.get_user(selected_organization)
+        alert_receive_channel = AlertReceiveChannel.get_or_create_manual_integration(
```
iskhakov, line 77: We'll create the new manual integration for each team. May be reflect that in the integration name. Just in case both teams will configure the same slack channel for their alerts

engine/apps/alerts/migrations/0004_auto_20220630_1343.py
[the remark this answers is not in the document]
Konstantinov-Innokentii: This migration doesn't change data, only adds new field. It is needed not because of teams, but because of fact that our current template for manual integration is impossible to understand for the ordinary user.

PR #200, opened 2022-07-05
Add possibility to modify default route on integration create

Konstantinov-Innokentii:
Context: https://raintank-corp.slack.com/archives/CRUKW54N5/p1656946147548309


Ferril (approved):
LGTM

PR #212, opened 2022-07-07
Fix integrations giving the wrong URL when using a path prefix

mderynck:
In the case where a reverse proxy is using a path prefix to the OnCall API the path prefix is being removed when giving the link back.

To fix it BASE_URL should be terminated with / and the path being joined should not start with /

urljoin behavior for reference:
```
urljoin("http://localhost:8080/test123/", "/destination") => http://localhost:8080/destination
urljoin("http://localhost:8080/test123/", "destination") => http://localhost:8080/test123/destination
urljoin("http://localhost:8080/test123", "/destination") => http://localhost:8080/destination
urljoin("http://localhost:8080/test123", "destination") => http://localhost:8080/destination
```
Currently we return URLs from the API in 3 ways:
- urljoin
- directly appending strings
- directly appending result of reverse()


matiasb (approved):
LGTM, +1 to unify a criteria to build URLs (and having a shared helper implementing this criteria for all our use cases sounds good to me)

mderynck:
Added a utility method for creating URLs so that we can handle the path prefix without breaking existing configurations by requiring them to have a certain format for BASE_URL.

Matvey-Kuk (approved):

Issue #140, opened 2022-06-23
Shift timestamps rendered incorrectly in Slack on mobile


mderynck:
![Screenshot_20220623-084450_Slack](https://user-images.githubusercontent.com/28077050/175336328-7c4bd41a-266b-4246-8625-2c4df17a16fb.jpg)

Times for shift notification are being rendered as ms instead of normal format. 
Screenshot is from Slack app, Android 10 on Samsung Galaxy S10e

vadimkerr:
Renders correctly on iOS 15, so seems like the issue is Android specific.

mderynck:
Docs:
https://api.slack.com/reference/surfaces/formatting#date-formatting
To recreate:
https://app.slack.com/block-kit-builder/T01TZ502XCH#%7B%22blocks%22:%5B%7B%22type%22:%22section%22,%22text%22:%7B%22type%22:%22mrkdwn%22,%22text%22:%22%3C!date%5E1392734382%5E%7Bdate_short%7D%20%7Btime%7D%7CFail%20to%20parse%3E%22%7D%7D%5D%7D

PR #225, opened 2022-07-12
Use formatted UTC time for fallback in slack message

mderynck:
Format timestamp as UTC instead of displaying integer in cases where slack is failing to format the timestamp at the client.

https://github.com/grafana/oncall/issues/140


Matvey-Kuk (approved):
Cool catch! Didn't realize there is a fallback

vadimkerr (approved):

PR #231, opened 2022-07-13
Fix missing integration icons

mderynck:
Some icons were showing up as webhook instead of their intended icon.


Matvey-Kuk (approved):

mderynck:
@Konstantinov-Innokentii Current PR doesn't fix that one,
It should be able to be fixed by adding an entry for it, what icon should it be for Manual Incident?

iskhakov (approved):

PR #236, opened 2022-07-14
Improve installation error handling

mderynck:
Improve feedback on a couple issues that could be encountered during install:

- Add a message to restart Grafana for the 404 case when plugin is newly installed
- Fix null exception on 502 so page gives feedback instead of printing exception to console.


Matvey-Kuk (approved):
@mderynck I like how user-friendly is this! LGTM.

iskhakov (approved):

PR #241, opened 2022-07-18
Add version mismatch warning

vadimkerr:


grafana-plugin/src/containers/DefaultPageLayout/DefaultPageLayout.tsx
```diff
@@ -88,14 +89,33 @@ const DefaultPageLayout: FC<DefaultPageLayoutProps> = observer((props) => {
...
-        {currentTeam?.limits.show_limits_warning && !getItem(currentTeam.limits.warning_text) && (
-          <Alert
-            className={styles.alert}
-            severity="warning"
-            title={currentTeam?.limits.warning_text}
-            onRemove={getRemoveAlertHandler(currentTeam?.limits.warning_text)}
-          />
+        {store.backendVersion && plugin?.version && store.backendVersion !== plugin?.version && (
+          <Alert className={styles.alert} severity="warning" title={'Version mismatch!'}>
+            Please make sure you have the same versions of Grafana OnCall plugin and Grafana OnCall engine, otherwise
```
Matvey-Kuk, line 94: of **the** Grafana OnCall

engine/apps/api/serializers/organization.py
```diff
@@ -109,7 +110,26 @@ def get_banner(self, obj):
...
+            text = (
+                "Version mismatch! Please make sure you have the same versions of Grafana OnCall plugin "
+                "and Grafana OnCall engine, "
+                "otherwise there could be issues with your Grafana OnCall installation! "
+                f"Current plugin version: {frontend_version}, current engine version: {backend_version}. "
+                "To install the latest Grafana OnCall plugin version, "
+                "visit the plugin configuration page and click Update. "
+                "To update Grafana OnCall engine, re-pull the grafana/oncall image and restart (or see the docs)."
+            )
+            version_warning = {"period_title": "Version mismatch", "show_limits_warning": True, "warning_text": text}
```
Matvey-Kuk, line 130: @vadimkerr are you sure it should be a part of `get_limits` method?

Matvey-Kuk (changes requested):
@vadimkerr I think that it may be confusing to give partial instructions about how to update the system in banners. I propose to make a paragraph here: https://grafana.com/docs/grafana-cloud/oncall/open-source/ called "How to update" where we will point users to different instructions for updating helm, docker-compose etc. And link to this paragraph from the product banners.

engine/apps/api/serializers/organization.py
vadimkerr: It shouldn't, but the idea is that this warning should pop up on new backend version and old frontend version. So we need to show a warning without changing the frontend, it's the only place where we can show a warning generated on backend I think. Are there any other places on backend that I forgot about?

vadimkerr:
> @vadimkerr I think that it may be confusing to give partial instructions about how to update the system in banners. I propose to make a paragraph here: https://grafana.com/docs/grafana-cloud/oncall/open-source/ called "How to update" where we will point users to different instructions for updating helm, docker-compose etc. And link to this paragraph from the product banners.

Yes, makes sense, let me add a paragraph on it in the docs and reference it in the warning banner.

engine/apps/api/serializers/organization.py
Konstantinov-Innokentii: I think for now it's ok, but we need to refactor it later.

maskin25 (approved):

vadimkerr:
Here's a PR with docs on OSS update: https://github.com/grafana/oncall/pull/242

Matvey-Kuk (approved):

PR #242, opened 2022-07-18
Add docs on OSS update

vadimkerr:


README.md
```diff
@@ -56,6 +56,20 @@ Grafana Url: http://grafana:3000
 6. Enjoy! Check our [OSS docs](https://grafana.com/docs/grafana-cloud/oncall/open-source/) if you want to set up Slack, Telegram, Twilio or SMS/calls through Grafana Cloud. 
 
 
+## Update
+To update your Grafana OnCall hobby environment:
```
iskhakov, line 60: they also need to 
1. Stop the existing docker-compose using the old file
2. Re-download the latest docker-compose file and start using the new version
https://github.com/grafana/oncall/pull/242/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R23

vadimkerr: I think docker compose can do this automatically:
1. docker-compose should restart any containers with updated images
2. docker-compose uses the `grafana/oncall` image, so if new image is pulled it will be updated automatically as well

iskhakov: ok, sounds good, the only problem if we change names, add, or remove the containers
they can get into the orphaned stage. We should take care of that

vadimkerr: Considering that `docker-compose.yml` doesn't change on update, the only case when things break is if we delete the `grafana/oncall` repo on docker hub. I think we won't do that in foreseeable future. So the orphans shouldn't be an issue here.

README.md
```diff
@@ -56,6 +56,20 @@ Grafana Url: http://grafana:3000
 6. Enjoy! Check our [OSS docs](https://grafana.com/docs/grafana-cloud/oncall/open-source/) if you want to set up Slack, Telegram, Twilio or SMS/calls through Grafana Cloud. 
 
 
+## Update
```
Matvey-Kuk, line 59: @vadimkerr may be "Update version"?

Matvey-Kuk (approved):

iskhakov (approved):

README.md
vadimkerr: Added an issue on re-downloading `docker-compose.yml` as well: https://github.com/grafana/oncall/issues/244

PR #251, opened 2022-07-19
[helm] enable annotations for external service

k1rk:
add ability to specify annotations for external service


Matvey-Kuk:
@k1rk thank you for the contribution!

PR #262, opened 2022-07-20
Bring back cached_render_for_web

vadimkerr:
This PR brings back `cached_render_for_web` and `active_cache_for_web_calculation_id` fields for the AlertGroup model.
The 0004 migration on alerts app introduced in 1.0.7 can be very slow and even fail in some cases on some DBs (e.g. MySQL).
The idea behind this PR is to drop the last 0004 migration on alerts app to make sure that migrations can be run on write heavy alertgroup table, but at the same time recreate these fields in case the 0004 migration was already applied.


mderynck (approved):

iskhakov (approved):

Konstantinov-Innokentii:
I don't like that with :
```
...
except OperationalError:
            pass
```
this migration will be performed even with no connection to the database

vadimkerr:
> I don't like that with :
> 
> ```
> ...
> except OperationalError:
>             pass
> ```
> 
> this migration will be performed even with no connection to the database

I don't like this either, but there are no generic error codes for different databases, e.g. MySQL and PostgreSQL will have different error codes, so I didn't want to check manually for all the DBs.
I checked this, so migration won't start at all if there's no db connection. Also it only can affect 1.0.7 users in some extremely rare cases.

Matvey-Kuk (approved):

PR #288, opened 2022-07-25
Create new oncall shift on update

Ferril:
Check if the updated shift has already started. If it has, set `until` date of this shift to `now` and create a new shift with the updated data.
Do the same on delete without creation the new shift.
Update shifts that haven't started, as usual.


engine/apps/schedules/models/custom_on_call_shift.py
```diff
@@ -250,12 +281,16 @@ def convert_to_ical(self, time_zone="UTC"):
...
+                rotation_ical = ""
                 start = self.get_next_start_date(event_ical)
                 if not start:  # means that rotation ends before next event starts
                     break
                 for user_counter, user in enumerate(users, start=1):
                     event_ical = self.generate_ical(user, start, user_counter, counter, time_zone)
-                    result += event_ical
+                    rotation_ical += event_ical
+                # if event has already been started, add rotation ical to final ical result
+                if start >= self.rotation_start:
```
matiasb, line 292: Couldn't we add the event anyway? (even if there is no instances occurring for a while) Wondering how we would eventually add it to the calendar otherwise (ie. how we would trigger the addition when needed)

engine/apps/api/serializers/on_call_shifts.py
```diff
@@ -36,7 +36,7 @@ class Meta:
         fields = [
             "id",
             "organization",
-            "name",
+            "title",
```
matiasb, line 39: I guess we will need to include the `updated_shift` field too, read-only, to indicate when there is an updated version of the current shift (eventually, frontend should decide to show a warning or "redirect" the user to edit the latest version of the shift)

Ferril: makes sense

engine/apps/schedules/models/custom_on_call_shift.py
Ferril: Anyway I reverted this changes because I think it's better to do updates for ical generation in another PR.

matiasb (approved):
Looks good!

engine/apps/api/serializers/on_call_shifts.py
```diff
@@ -136,16 +126,13 @@ def _validate_rotation_start(self, shift_start, rotation_start):
             raise serializers.ValidationError({"rotation_start": ["Incorrect rotation start date"]})
 
     def _validate_until(self, rotation_start, until):
-        if until is not None and until < rotation_start:
-            raise serializers.ValidationError({"until": ["Incorrect rotation end date"]})
+        if until is not None:
+            if until < rotation_start:
```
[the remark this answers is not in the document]
Ferril: Yes, thanks. I added another condition here and removed it later so this change no longer makes sense

vadimkerr (approved):

PR #309, opened 2022-07-29
Autoresolve condition select is linked with alert template tab

Ukochka:
- URLs for tabs have been added
- Check for invalid autoresolving condition at settings tab has been added
- Link from Autoresolving select to autoresolving condition has been added


Matvey-Kuk:
I checked this PR, merging.

Issue #167, opened 2022-06-28
Slack notifications not displaying properly on mobile


mderynck:
On mobile notifications from slack show contents as "This content can't be displayed".

A similar issue: 
https://github.com/grafana/grafana/issues/42768
was fixed in Grafana with this PR:
https://github.com/grafana/grafana/pull/45750

We likely need to pass through and add a top level text field to the slack messages we are sending where it is missing.

skeilson:
@mderynck we have at least 2 reports of this problem. Is there any update on this and if not, could an ETA be provided?

Matvey-Kuk:
@skeilson this issue is not in the short-term backlog yet. Will let you know once we have any updates. 

Any contributors willing to fix this are very welcome!

Bahus:
> this issue is not in the short-term backlog yet

@Matvey-Kuk this confuses me a lot. 

As far I remember slack always was a paramount integration in amixer/grafana on call, so it's very strange to hear that most of the product users (especially paying ones) still have to click on a notification to open slack and see full alert message. Same on a phone and a desktop.

It's very very annoying, counterproductive and (from my point of view) unacceptable for a paid service.

In addition I don't think this going to be an easy fix, since a title can include text, links, markup, can be shot/long etc, so it will require some testing on desktop/mobile devices and I'm sure one have to make some compromises.

rs:
Agreed. As a paying customer, this is making oncall barely usable and very inconvenient.

I would suggest a special field to set the notification text in the oncall configuration UI.

iskhakov:
We decided to prioritize this bugfix, thanks for the feedback. 
It's in our short-term backlog now

iskhakov:
We were able to reproduce the issue on the android devices. On ios and on the desktop push notifications were able to display the information about the alert. We'll apply the hotfix soon and will consider adding the separate optional template to configure the text you see in the push notification

rs:
We can definitely reproduce the issue on iOS and macOS.

Matvey-Kuk:
@mderynck 🎉

PR #323, opened 2022-08-02
Fix slack notifications not displaying properly

mderynck:
#167 

Use text fallback field in all places where attachments and blocks are used. Replace attachments with blocks in some cases so text is not repeated in both.

Blocks do not support having the colored line on the left.  Alternatively we could use a colored dot emoji.

From slack docs:
> Any content displayed within attachments may be wrapped, truncated, or hidden behind a "show more" style option by Slack clients. This isn't the case with the top-level blocks field.

### DIFF
--- a/engine/apps/alerts/tasks/call_ack_url.py
+++ b/engine/apps/alerts/tasks/call_ack_url.py
@@ -38,20 +38,18 @@ def call_ack_url(ack_url, alert_group_pk, channel, http_method="GET"):
             "block_id": "alert",
             "text": {
                 "type": "mrkdwn",
-                "text":  text,
-            }
+                "text": text,
+            },
         },
-        {
-            "type": "divider"
-        },
+        {"type": "divider"},
         {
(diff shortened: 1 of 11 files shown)


iskhakov:
Added block or text field to each postMessage and update

mderynck:
I tested the most recent set of changes on Ubuntu and Android. All the messages displayed properly when displaying as notifications or mentions & reactions.

Matvey-Kuk:
@mderynck could you please provide screenshots of how Slack messages look after update?

mderynck:
Current (Android Mentions & Reactions) - I think the slack client was updated so they appear blank instead of "Content cannot be displayed"

Before_Mention

Current (Android Thread)

Before_Thread

After (Android Mentions & Reactions) - Ignore the bot icon, mine is customized so I can tell different versions apart

After_Mention

After (Android Thread)

After_Thread

iskhakov:
Merging

PR #330, opened 2022-08-03
Update generating ical for on-call shifts

Ferril:
Update generating ical for on-call shifts with respect to `rotation_start`


matiasb (approved):
Looks good :+1:

vadimkerr (approved):

PR #337, opened 2022-08-04
Wrong team error for alert groups page

vadimkerr:
This PR adds a 403 error for cases when alert group belongs to a team which is not user's currently selected team.

Example for a user that is in teams "General" and "test". When the user tries to see an alert group of team test while their current team is General: 
Screenshot 2022-08-04 at 16 28 23

Example for a user that's not in the team that owns the alert group:
Screenshot 2022-08-04 at 16 29 05


engine/common/api_helpers/mixins.py
```diff
@@ -174,6 +177,69 @@ def get_object(self):
...
+
+            if obj_team is None or obj_team in self.request.user.teams.all():
+                if obj_team is None:
+                    obj_team = Team(public_primary_key=None, name="General", email=None, avatar_url=None)
+
+                return Response(
+                    data={"error_code": "wrong_team", "owner_team": TeamSerializer(obj_team).data},
+                    status=status.HTTP_403_FORBIDDEN,
+                )
+
```
matiasb, line 210: Wondering if it would be simpler (to read) to remove the team filter in `AlertGroupView` and override here to check team belonging only?

matiasb (approved):
Looks good, it seems a bit complex, but makes sense (for future reusability).
Inline thought added.

engine/common/api_helpers/mixins.py
```diff
@@ -174,6 +177,69 @@ def get_object(self):
         return obj
 
 
+class TeamFilteringMixin:
+    """
+    This mixin returns 403 and {"error_code": "wrong_team", "owner_team": {"name", "id", "email", "avatar_url"}}
+    in case a requested instance doesn't belong to user's current_team.
```
Matvey-Kuk, line 183: ❤️

engine/apps/api/tests/test_team.py
```diff
@@ -85,3 +85,198 @@ def test_list_teams_permissions(
...
+    alert_receive_channel = make_alert_receive_channel(organization)
+    alert_group = make_alert_group(alert_receive_channel)
+
+    # escalation_chain = make_escalation_chain(organization)
+    # schedule = make_schedule(organization, schedule_class=OnCallScheduleCalendar)
+    # webhook = make_custom_action(organization)
+
+    for endpoint, instance in (
+        ("alertgroup", alert_group),
+        # todo: implement team filtering for other resources
```
Matvey-Kuk, line 123: Would be nice to file an issue about this ToDo.

Matvey-Kuk (approved):

Issue #118, opened 2022-06-21
Calculate OnCall schedule quality


Matvey-Kuk:
Alice, Bob, Sam are on call. I think it would be nice to show some "schedule quality" to help users keep schedules in a good shape.

```
100% = ((50% * balance) + (50% * working_hours_balance)) * ( 1 / gaps_rate)
```

We're trying to balance on call hours between engineers on call. Balance = 1 is good. Balance = 0 is bad.
```
balance = abs (min(Alice, Bob) hours / max(Alice, Bob) hours) * .... all pairs
```

We're trying to avoid people being on call outside of working hours. At least their on call time outside of working hours should be equal.
```
working_hours_balance = abs (min(Alice, Bob) hours outside of working hours / max (Alice, Bob) hours outside of working hours) * .... all pairs
```

We want to avoid gaps in on call schedule. Shift with no users is equal to gap:
```
gaps_rate = gaps hours / total observed period hours 
```

Konstantinov-Innokentii:
@Matvey-Kuk you want to calculate schedule's score or per-user score?

Matvey-Kuk:
@Konstantinov-Innokentii schedule's score

vadimkerr:
@Matvey-Kuk I propose to use a slightly different formula for balance score calculation:
```
balance = (min(Alice, Bob) hours / max(Alice, Bob) hours + .... all pairs) / number of pairs
```

It should eliminate the issue with hours distributions like `[7, 8, 8, 8, 8, 8]`, where the formula you provided returns around 50% score, but `[7, 8, 8, 8, 8, 8]` distribution is pretty balanced and should score more than 50% (returns around 95% with the formula I proposed).

maskin25:
@teodosii design [here](https://www.figma.com/file/e8B4keoGxQRlenH8rYTCpN/Grafana-OnCall-Calendar?node-id=1038%3A23688)

vadimkerr:
The backend part is now merged in https://github.com/grafana/oncall/pull/338, @teodosii will be working on the frontend part in a separate PR.

teodosii:
Frontend is implemented but waiting for a minor backend change to differentiate good remarks for warnings within the api response. Once that is done we can unblock the frontend too, because right now on frontend we cannot differentiate info messages (like 'Good schedule') from warnings ('Gaps')

PR #338, opened 2022-08-04
Schedule score backend

vadimkerr:
This PR adds an endpoint returning a schedule quality score, overloaded users and comments on the existing issues (e.g. balance issues or gaps).

## Limitations
- Since working hours editor is not implemented yet, there are only two scores taken into account: balance score and a score representing the ratio of time when someone is on-call to the whole time period.
- Time period is now set to be constant (90 days from today), so **in some cases the results will be inaccurate** (when rotations don't align with the time period)
- It only takes primary rotations into account (overrides are ignored)

## Usage
`GET /api/internal/v1/schedules/<pk>/quality?date=<TOMORROW_DATE>`

Note that `date` should be tomorrow date, because we can only be sure about changing tomorrow's shifts (some of the shifts for current day could be "deleted" but still show up in the UI).

## Example response
```json
{
  "total_score": 90,
  "comments": ["Schedule has no gaps", "Schedule is well-balanced, but still can be improved"],
  "overloaded_users": ["USSZ5WRH2CUA9", "U74XJZSSQGBIH"]
}
```

Issue: #118


engine/apps/schedules/ical_utils.py
```diff
@@ -315,6 +315,57 @@ def list_users_to_notify_from_ical_for_period(schedule, start_datetime, end_date
     return users_found_in_ical
 
 
+def get_schedule_quality(schedule, date, days):
+    OnCallSchedule = apps.get_model("schedules", "OnCallSchedule")
+
+    shifts = (
+        list_of_oncall_shifts_from_ical(
```
matiasb, line 322: maybe you can switch to use `.filter_events` method from schedule now?

engine/apps/api/views/schedule.py
```diff
@@ -280,6 +282,17 @@ def related_escalation_chains(self, request, pk):
...
 
+    @action(detail=True, methods=["get"])
+    def quality(self, request, pk):
+        schedule = self.original_get_object()
+        user_tz, date = self.get_request_timezone()
+        days = int(self.request.query_params.get("days", 90))  # todo: check if days could be calculated more precisely
+
+        events = schedule.filter_events(user_tz, date, days=days, with_empty=True, with_gap=True)
+
+        schedule_score = get_schedule_score(events, days)
```
matiasb, line 293: Wondering if this could be a method of the base schedule model, and do the call to filter_events there (so you would only call here to `schedule.get_score(days)`)

engine/apps/schedules/quality.py
```diff
@@ -0,0 +1,322 @@
...
+
+
+def get_balance_score(events: list[dict]) -> float:
+    users_to_events_map = {}
+    for event in events:
+        user_pks = [user["pk"] for user in event["users"]]
+
+        for user_pk in user_pks:
+            if user_pk in users_to_events_map:
+                users_to_events_map[user_pk].append(event)
```
matiasb, line 178: I guess you could use a defaultdict here? Also, maybe you can calculate duration_map directly, accumulating events duration in each iteration?

engine/apps/schedules/quality.py
```diff
@@ -0,0 +1,322 @@
...
+                users_to_events_map[user_pk] = [event]
+
+    duration_map = {}
+    for user_pk, events in users_to_events_map.items():
+        duration_map[user_pk] = timedelta_sum(event_duration(event) for event in events)
+
+    score = 0
+    number_of_pairs = 0
+    for user_1 in duration_map:
+        for user_2 in duration_map:
```
matiasb, line 189: If I understand correctly, maybe you can use [itertools combinations](https://docs.python.org/3/library/itertools.html#itertools.combinations) here?

engine/apps/schedules/quality.py
```diff
@@ -0,0 +1,322 @@
...
+
+    return balance_score
+
+
+def get_balance_outside_working_hours(events: list[dict], users: dict[str, User]) -> float:
+    users_to_events_map = {}
+    for event in events:
+        user_pks = [user["pk"] for user in event["users"]]
+
+        for user_pk in user_pks:
```
matiasb, line 208: If this will be calculated for several functions, it could make sense to have a helper function at the module level?

engine/apps/schedules/quality.py
```diff
@@ -0,0 +1,322 @@
...
+        for event in users_to_events_map[user_pk]:
+            start, end = event["start"], event["end"]
+            if event["all_day"]:
+                start = get_day_start(start)
+                end = get_day_end(end)
+
+            user = users[user_pk]
+            spans = spans_inside_working_hours(start, end, user)
+            spans_duration = timedelta_sum(span[1] - span[0] for span in spans)
+            outside_working_hours_duration += event_duration(event) - spans_duration
```
matiasb, line 227: Maybe it makes sense to calculate inside/outside working hours duration in the same forloop? (so, having a helper returning both mappings, splitting/iterating events one time?)

matiasb (commented):
Changes look good. Added a few comments/suggestions, feel free to ignore them if they don't make sense :-)

engine/apps/schedules/quality.py
```diff
@@ -0,0 +1,322 @@
...
+def get_inside_working_hours_score(events: list[dict], users: dict[str, User]):
+    inside_working_hours_duration = datetime.timedelta(seconds=0)
+
+    for event in events:
+        user_pks = [user["pk"] for user in event["users"]]
+
+        for user_pk in user_pks:
+            start, end = event["start"], event["end"]
+            if event["all_day"]:
+                start = get_day_start(start)
```
Ferril, line 154: If I understand correctly, here it makes sense to calculate `start` and `end` for `all_day` events with respect to calendar timezone, because `all_day` event for different timezones will have different `start` and `end` time

engine/apps/schedules/quality.py
```diff
@@ -0,0 +1,322 @@
...
+                users_to_events_map[user_pk] = [event]
+
+    outside_working_hours_duration_map = {}
+    for user_pk in users_to_events_map:
+        outside_working_hours_duration = datetime.timedelta(seconds=0)
+
+        for event in users_to_events_map[user_pk]:
+            start, end = event["start"], event["end"]
+            if event["all_day"]:
+                start = get_day_start(start)
```
Ferril, line 221: the same about tz for `all_day` events

engine/apps/schedules/quality.py
vadimkerr: 👍

engine/apps/schedules/quality.py
vadimkerr: 👍 , removed this completely

engine/apps/schedules/quality.py
vadimkerr: thanks, yes itertools.combinations is definetely the way to go

teodosii:
It's on my list, I need some more time to finish current work and then I can get on this one.

vadimkerr:
Some additional backend chages are needed to support [this design](https://github.com/grafana/oncall/issues/118#issuecomment-1310227452), so converting this back to draft PR.

vadimkerr:
Merging the backend part for now, @teodosii let's have a separate PR for the frontend.

Issue #343, opened 2022-08-08
Add node selection logic in the helm chart


vsmanu:
The oncall helm chart is missing basic node selection logic. It would be nice to have a granular control over the node selection.

vsmanu:
I have created a pull request, at least for engine https://github.com/grafana/oncall/pull/344

PR #344, opened 2022-08-08
343: node selection for helm chart - engine only - Initial commit

vsmanu:


Matvey-Kuk:
Hi! Thank you for the contribution! Could you please provide more description about what's this update about? Thank you.

vsmanu:
> Hi! Thank you for the contribution! Could you please provide more description about what's this update about? Thank you.

Hi,
That was a pull request from my side for the Issue https://github.com/grafana/oncall/issues/343. 
If there are better alternative maybe you could suggest. My opinion is to have something which can also address the dependent bitanmi charts as well, as of now the commit only covers the engine.

vadimkerr (approved):
seems good to me, @iskhakov can you please check this out as well?

PR #345, opened 2022-08-08
Add postgresql support for development/testing

matiasb:


DEVELOPER.md
```diff
@@ -45,6 +47,9 @@ pip install -U pip wheel
 # Copy and check .env file.
 cp .env.example .env
 
+# NOTE: if you want to use the PostgreSQL db backend add DB_BACKEND=postgresql to your .env file;
```
Matvey-Kuk, line 50: ❤️

docker-compose-developer-pg.yml
```diff
@@ -0,0 +1,74 @@
+version: '3.2'
+
+services:
+
+  postgres:
```
iskhakov, line 5: Another approach might be using compose profile like here 
https://github.com/grafana/oncall/blob/dev/docker-compose.yml#L173

But I'm fine with both approaches

iskhakov (approved):

Matvey-Kuk:
Would be nice to mention possibility of using PG as a DB here: https://grafana.com/docs/oncall/latest/open-source/#production-environment, also highlight that it's not a preferred option.

docker-compose-developer-pg.yml
matiasb: I was looking at the `profile` option, it seems tricky to implement in this case because we still need the mysql service for grafana (and setup differs between the compose files), so it sounds like we would need one mysql profile besides a postgresql one, requiring to specify one or the other. I think I prefer to keep things simple for our default use case for now.

matiasb:
> Would be nice to mention possibility of using PG as a DB here: https://grafana.com/docs/oncall/latest/open-source/#production-environment, also highlight that it's not a preferred option.

My idea was to add something there once I prepare a PR updating the helm chart to support the setup using an external postgresql.

PR #372, opened 2022-08-16
On call shifts update logic

Ferril:
Use `rotation_start` and `until` dates for calculation of event start and end dates for web schedules


engine/apps/schedules/ical_utils.py
```diff
@@ -171,8 +177,14 @@ def get_shifts_dict(calendar, calendar_type, schedule, datetime_start, datetime_
...
-                start = event[ICAL_DATETIME_START].dt
-                end = event[ICAL_DATETIME_END].dt
+                if isinstance(schedule, OnCallScheduleWeb):
+                    start = max(event[ICAL_DATETIME_START].dt, event[ICAL_DATETIME_STAMP].dt.date())
+                    end = event[ICAL_DATETIME_END].dt
+                    if event.get(ICAL_RRULE, {}).get(ICAL_UNTIL):
+                        end = min(event[ICAL_DATETIME_END].dt, event[ICAL_RRULE][ICAL_UNTIL][0].date())
+                else:
+                    start = event[ICAL_DATETIME_START].dt
+                    end = event[ICAL_DATETIME_END].dt
```
matiasb, line 187: Wondering if this logic above could be refactored out somehow to use it here and reuse below?

engine/apps/schedules/models/custom_on_call_shift.py
```diff
@@ -301,13 +301,16 @@ def convert_to_ical(self, time_zone="UTC"):
                     if not start:  # means that rotation ends before next event starts
                         all_rotation_checked = True
                         break
-                    elif start >= self.rotation_start:  # event has already started, generate iCal for each user
+                    elif (
+                        self.source == CustomOnCallShift.SOURCE_WEB and start + self.duration > self.rotation_start
```
matiasb, line 305: To be sure, do we need to check for `SOURCE_WEB` here? wouldn't all other shifts have `start >= rotation_start`? (assuming all other shifts got `rotation_start` set to `shift_start`)

matiasb (approved):
LGTM, minor comments/questions added.

engine/apps/schedules/models/custom_on_call_shift.py
Ferril: We don't set `rotation_start` to `shift_start` because in this case we will break repetition rules for future events. We keep in iCal theirs original start date and recalculate theirs start with respect to `rotation_start` in other place. And only for `SOURCE_WEB`, we assume that the event can start after the original start, that's why we're checking event type here

engine/apps/schedules/ical_utils.py
Ferril: makes sense

PR #381, opened 2022-08-17
Fix for final event calculation when splitting events

matiasb:


engine/apps/schedules/models/custom_on_call_shift.py
```diff
@@ -351,7 +353,8 @@ def get_rotation_date(self, event_ical, get_next_date=False):
         current_event = Event.from_ical(event_ical)
         # take shift interval, not event interval. For rolling_users shift it is not the same.
         interval = self.interval or 1
-        current_event["rrule"]["INTERVAL"] = interval
+        if "rrule" in current_event:
```
Ferril, line 356: I would describe in a comment the case when the event doesn't have `rrule`

Ferril (approved):
LGTM, minor comment

PR #383, opened 2022-08-18
Add calculation for rotation index

Ferril:
Save rotation index to keep rotation order after editing the shift, change event start date on update to avoid recalculation of rotation order if count of rotation user groups was changed.


matiasb (approved):
Looks good, one minor comment added

Issue #63, opened 2022-06-14
Add raw alert content view


Matvey-Kuk:
We show raw content of last alert in the template editor (payload example):
<img width="1010" alt="Screenshot 2022-06-14 at 16 04 57" src="https://user-images.githubusercontent.com/3284841/173584039-8984d341-8eb0-4b6c-a419-029044176166.png">

Users are missing the ability to check raw content of some specific alerts in the history. May be add the button here: ?
<img width="656" alt="Screenshot 2022-06-14 at 16 05 09" src="https://user-images.githubusercontent.com/3284841/173584352-e223a84c-7b04-4314-a6e9-5d290009b3c0.png">

maskin25:
good proposal, let's add toggler that allows see raw content

Konstantinov-Innokentii:
@Matvey-Kuk I have a question here. We want to have this toggler for each alert in the group? 
<img width="780" alt="Снимок экрана 2022-08-18 в 16 01 53" src="https://user-images.githubusercontent.com/20221722/185379914-6ad7d06d-7ca5-485f-99a6-57858df26691.png">

Matvey-Kuk:
@Konstantinov-Innokentii yep, that's my proposal.

raphael-batte:
Sketch:
![image](https://user-images.githubusercontent.com/102970268/186622227-e496f1de-76fa-46b7-b829-808e88c5a032.png)

![image](https://user-images.githubusercontent.com/102970268/186622367-2311a880-c912-46e5-bc9c-0265afddfeec.png)

Konstantinov-Innokentii:
@raphael-batte since there are several alerts in the group we should have the show/hide button near each one, right?
<img width="782" alt="Снимок экрана 2022-08-25 в 14 27 27" src="https://user-images.githubusercontent.com/20221722/186628786-4a5aca73-3242-4b99-b3b0-685acd7c53e0.png">

raphael-batte:
At this case something like this 
![image](https://user-images.githubusercontent.com/102970268/186637506-bd8f3640-d183-44e8-aca5-a032742d76fe.png)

raphael-batte:
more complex view 
![image](https://user-images.githubusercontent.com/102970268/186860567-db99e24b-5a42-4b8b-9784-cc26a1195cd2.png)

PR #392, opened 2022-08-23
63 raw response content

teodosii:
New feature - #63 

Changes
- Added button next to group alert to visualize the alert's payload in a modal
- Refactored components within `IncidentPage` - we now have 3 separate components `Incident`, `GroupedIncidentsList` and `AttachedIncidentsList` instead of having everything in a class


grafana-plugin/src/models/alertgroup/alertgroup.ts
```diff
@@ -298,6 +298,12 @@ export class AlertGroupStore extends BaseStore {
     });
   }
 
+  @action
```
maskin25, line 301: actually we don't need `action` here, " An action is any piece of code that modifies the state."

grafana-plugin/src/models/alertgroup/alertgroup.ts
```diff
@@ -298,6 +298,12 @@ export class AlertGroupStore extends BaseStore {
...
+  async getPayloadForIncident(pk: Alert['pk']) {
+    const result = await makeRequest(`/alerts/${pk}`, {});
```
maskin25, line 303: `return await makeRequest(`/alerts/${pk}`, {});`

maskin25 (approved):

PR #393, opened 2022-08-23
Set DATA_UPLOAD_MAX_MEMORY_SIZE to 1mb

Konstantinov-Innokentii:


iskhakov (approved):

engine/settings/base.py
```diff
@@ -451,7 +451,8 @@
 
 GRAFANA_INCIDENT_STATIC_API_KEY = os.environ.get("GRAFANA_INCIDENT_STATIC_API_KEY", None)
 
-DATA_UPLOAD_MAX_MEMORY_SIZE = 5242880
+# Based on 99th percentile of content-length.
```
iskhakov, line 454: this comment doesn't make sense in oss version code

engine/settings/base.py
```diff
@@ -451,7 +451,7 @@
 
 GRAFANA_INCIDENT_STATIC_API_KEY = os.environ.get("GRAFANA_INCIDENT_STATIC_API_KEY", None)
 
-DATA_UPLOAD_MAX_MEMORY_SIZE = 5242880
+DATA_UPLOAD_MAX_MEMORY_SIZE = 1048576
```
vadimkerr, line 454: Probably worth adding a comment that it's 10MB

vadimkerr (approved):

engine/settings/base.py
vadimkerr: Not sure, I thought it's 10MB from the first glance. So probably worth adding a comment that it's 1MB.

vadimkerr: Or probably make it even `1 * 1024 * 1024` to be super verbose

abodacs: we can add underscore based on PEP 515
so 1048576 will be 1_048_576

Issue #387, opened 2022-08-19
[pagerduty-migrator] Error performing migration


gilbertobr:
I came across the following error when using MODE="migrate"

This error usually appears when the process is almost over.

Note: If you run MODE="migrate" again, it breaks elsewhere.
Saying that there is already a "schedule" registered.
FIle: https://github.com/grafana/oncall/blob/dev/tools/pagerduty-migrator/migrator/oncall_api_client.py

In my view... the behavior should be:
If it already exists, it jumps to the next one.

This Field "Start API CALL" and "End API CALL" I put it to make it easier to identify the error.

![image](https://user-images.githubusercontent.com/20375245/185642858-ba4b45a1-f8d0-4830-9d98-b24458d6b498.png)

gilbertobr:
If I add an elif it breaks somewhere else...
![image](https://user-images.githubusercontent.com/20375245/185660512-16380018-7eb1-4f72-a21d-3654796571ac.png)

```
def api_call(method: str, path: str, **kwargs) -> requests.Response:
    url = urljoin(ONCALL_API_URL, path)

    response = requests.request(
        method, url, headers={"Authorization": ONCALL_API_TOKEN}, **kwargs
    )

    try:
        response.raise_for_status()
    except HTTPError as e:
        if e.response.status_code == 429:
            cooldown_seconds = int(e.response.headers["Retry-After"])
            sleep(cooldown_seconds)
            return api_call(method, path, **kwargs)
        elif e.response.status_code == 400:
            print(response.text)
        else:
            raise

    return response
```

Schedule Value:

`{..., 
'oncall_schedule': {'detail': 'Schedule with this name already exists'}, 'unmatched_users': []}`

vadimkerr:
Thank you for opening the issue! After taking a look on this, I discovered that the script deletes a Grafana OnCall integration with the same title before migration, please see: https://github.com/grafana/oncall/blob/dev/tools/pagerduty-migrator/migrator/resources/integrations.py#L43.
I couldn't reproduce this as well, so I'm closing the issue.

Issue #378, opened 2022-08-16
Integration_report function error


gilbertobr:
After locally applying this suggested fix
https://github.com/grafana/oncall/issues/377

I have received the following error.
![image](https://user-images.githubusercontent.com/20375245/184967097-3a223852-f42b-4cb0-aa55-64a659ac1f72.png)

gilbertobr:
I don't know if this would be the best scenario, but I managed to solve it by replacing the key integration["oncall_type"] for integration["name"]

https://github.com/grafana/oncall/blob/dev/tools/pagerduty-migrator/migrator/report.py


Code Adjustment:

```
TAB = " " * 4
SUCCESS_SIGN = "✅"
ERROR_SIGN = "❌"


def format_user(user: dict) -> str:
    result = "{} ({})".format(user["name"], user["email"])

    if user["oncall_user"]:
        result = "{} {}".format(SUCCESS_SIGN, result)
    else:
        result = "{} {} — no Grafana OnCall user found with this email".format(
            ERROR_SIGN, result
        )

    return result


def format_schedule(schedule: dict) -> str:
    if schedule["unmatched_users"]:
        result = "{} {} — schedule references unmatched users".format(
            ERROR_SIGN, schedule["name"]
        )
    else:
        result = "{} {}".format(SUCCESS_SIGN, schedule["name"])

    return result


def format_escalation_policy(policy: dict) -> str:
    if policy["unmatched_users"] and policy["flawed_schedules"]:
        result = "{} {} — policy references unmatched users and schedules with unmatched users".format(
            ERROR_SIGN, policy["name"]
        )
    elif policy["unmatched_users"]:
        result = "{} {} — policy references unmatched users".format(
            ERROR_SIGN, policy["name"]
        )
    elif policy["flawed_schedules"]:
        result = "{} {} — policy references schedules with unmatched users".format(
            ERROR_SIGN, policy["name"]
        )
    else:
        result = "{} {}".format(SUCCESS_SIGN, policy["name"])

    return result


def format_integration(integration: dict) -> str:
    result = integration["service"]["name"] + " - " + integration["name"]

    if not integration["name"]:
        result = (
            "{} {} — cannot find appropriate Grafana OnCall integration type".format(
                ERROR_SIGN, result
            )
        )
    elif integration["is_escalation_policy_flawed"]:
        policy_name = integration["service"]["escalation_policy"]["summary"]
[...]

PR #401, opened 2022-08-23
Pagerduty migrator/patch pack

gilbertobr:
I created a PR where it contains all the corrections related to the reported ERROR
https://github.com/grafana/oncall/issues/387 is https://github.com/grafana/oncall/issues/378

Including a script to speed up the admin insert all pagerduty users into grafana via API

### DIFF
--- a/tools/pagerduty-migrator/migrator/oncall_api_client.py
+++ b/tools/pagerduty-migrator/migrator/oncall_api_client.py
@@ -5,7 +5,7 @@ import requests
 from requests import HTTPError
 
 from migrator.config import ONCALL_API_TOKEN, ONCALL_API_URL
-
+TAB = " " * 4
 
 def api_call(method: str, path: str, **kwargs) -> requests.Response:
     url = urljoin(ONCALL_API_URL, path)
@@ -21,6 +21,12 @@ def api_call(method: str, path: str, **kwargs) -> requests.Response:
             cooldown_seconds = int(e.response.headers["Retry-After"])
             sleep(cooldown_seconds)
             return api_call(method, path, **kwargs)
+        elif e.response.status_code == 400:
+            json_res = response.json()
+            print(TAB + ">>> " + json_res["detail"])
+        elif e.response.status_code == 500:
+            json_res = response.json()
+            print(TAB + ">>> " + json_res["detail"])
         else:
             raise
 
--- a/tools/pagerduty-migrator/migrator/report.py
+++ b/tools/pagerduty-migrator/migrator/report.py
@@ -47,9 +47,13 @@ def format_escalation_policy(policy: dict) -> str:
 
 
 def format_integration(integration: dict) -> str:
-    result = integration["service"]["name"] + " - " + integration["name"]
+    result = integration["service"]["name"] + " " + str(integration["name"])
 
-    if not integration["oncall_type"]:
+    if not "oncall_type" in integration.keys():
+        result = "{} {} — integration type not found. Manual webhook inserted.".format(
+                SUCCESS_SIGN, result
+        )
+    elif not integration["oncall_type"]:
         result = (
             "{} {} — cannot find appropriate Grafana OnCall integration type".format(
                 ERROR_SIGN, result
@@ -128,12 +132,12 @@ def integration_report(integrations: list[dict]) -> str:
 
     for integration in sorted(
         integrations,
-        key=lambda i: bool(i["oncall_type"] and not i["is_escalation_policy_flawed"]),
+        key=lambda i: bool(i["name"] and not i["is_escalation_policy_flawed"]),
         reverse=True,
     ):
         result += f"\n" + TAB + format_integration(integration)
         if (
-            integration["oncall_type"]
+            integration["name"]
             and not integration["is_escalation_policy_flawed"]
             and integration["oncall_integration"]
         ):
--- a/tools/pagerduty-migrator/migrator/resources/integrations.py
+++ b/tools/pagerduty-migrator/migrator/resources/integrations.py
@@ -15,19 +15,20 @@ def match_integration(integration: dict, oncall_integrations: list[dict]) -> Non
 
 
 def match_integration_type(integration: dict, vendors: list[dict]) -> None:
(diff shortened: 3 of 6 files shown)


tools/pagerduty-migrator/scripts/README.md
```diff
@@ -0,0 +1,9 @@
+# Grafana Oncall Scripts
```
vadimkerr, line 1: Great idea to make a user migrator from PD to Grafana! I propose to make this as a separate PR though.

tools/pagerduty-migrator/migrator/resources/integrations.py
```diff
@@ -43,9 +44,14 @@ def migrate_integration(integration: dict, escalation_policies: list[dict]) -> N
 
     oncall_name = "{} - {}".format(integration["service"]["name"], integration["name"])
 
+    if "oncall_type" in integration.keys():
+        oncall_type = integration["oncall_type"]
+    else:
+        oncall_type = "webhook"
```
vadimkerr, line 50: Not really sure about falling back to the "webhook" integration type, probably we could make this as a feature flag in an env variable? Something like `FALLBACK_TO_WEBHOOK_INTEGRATION_TYPE` and make this as a separate PR? The idea behind the integration type matching inititally was to match them 1 to 1 without any guessing or fallbacks.

tools/pagerduty-migrator/migrator/resources/integrations.py
```diff
@@ -15,19 +15,20 @@ def match_integration(integration: dict, oncall_integrations: list[dict]) -> Non
...
-
-    integration["oncall_type"] = PAGERDUTY_TO_ONCALL_VENDOR_MAP.get(vendor_name)
+    if integration["vendor"] != None:
+        vendors_map = {vendor["id"]: vendor for vendor in vendors}
+    
+        if integration["type"] not in [
+            "generic_events_api_inbound_integration",
+            "events_api_v2_inbound_integration",
+        ]:
+            integration["name"] = None
```
vadimkerr, line 25: `integration["name"]` is really meant to hold the integration name, not the type/vendor, so I propose to change this back to `integration["oncall_type"]`.

tools/pagerduty-migrator/migrator/resources/integrations.py
```diff
@@ -15,19 +15,20 @@ def match_integration(integration: dict, oncall_integrations: list[dict]) -> Non
...
-        "events_api_v2_inbound_integration",
-    ]:
-        integration["oncall_type"] = None
-        return
-
-    vendor_id = integration["vendor"]["id"]
-    vendor_name = vendors_map[vendor_id]["name"]
-
-    integration["oncall_type"] = PAGERDUTY_TO_ONCALL_VENDOR_MAP.get(vendor_name)
+    if integration["vendor"] != None:
```
vadimkerr, line 18: Maybe we could just set the `integration["oncall_type"]` to `None` in case `integration["vendor"] is None`? I think it would simplify things a lot.

tools/pagerduty-migrator/migrator/report.py
```diff
@@ -47,9 +47,13 @@ def format_escalation_policy(policy: dict) -> str:
 
 
 def format_integration(integration: dict) -> str:
-    result = integration["service"]["name"] + " - " + integration["name"]
+    result = integration["service"]["name"] + " " + str(integration["name"])
 
-    if not integration["oncall_type"]:
+    if not "oncall_type" in integration.keys():
+        result = "{} {} — integration type not found. Manual webhook inserted.".format(
```
vadimkerr, line 53: Not really sure about falling back to the "webhook" integration type, probably we could make this as a feature flag in an env variable? Something like FALLBACK_TO_WEBHOOK_INTEGRATION_TYPE and make this as a separate PR? The idea behind the integration type matching inititally was to match them 1 to 1 without any guessing or fallbacks.

tools/pagerduty-migrator/migrator/oncall_api_client.py
```diff
@@ -21,6 +21,12 @@ def api_call(method: str, path: str, **kwargs) -> requests.Response:
             cooldown_seconds = int(e.response.headers["Retry-After"])
             sleep(cooldown_seconds)
             return api_call(method, path, **kwargs)
+        elif e.response.status_code == 400:
```
vadimkerr, line 24: Good idea to add some exception logging! Slight improvement could be
```suggestion
        else:
            print(TAB + ">>> " + response.json()["detail"])
            raise
```
to log any unhandled exceptions and raise further.

vadimkerr (changes requested):
Thank you for the PR! Here's my general feedback:
1. I think we need to revert the changes related to `integration["name"]` and `integration["oncall_type"]`, because it mixes up two different things (integration verbal name and integration type) and could introduce some unexpected behaviour.
2. I'm not sure we should fallback to "webhook" OnCall integration type by default. Probably worth adding an env variable for this.
3. Adding a script for migrating users from PD is a nice idea, let's open a separate PR for it.
4. The issue with the vendor being empty we could easily solve by setting the `oncall_type` to `None` (meaning that we can't match it to any of the Grafana OnCall integration types).
5. I couldn't reproduce the bug with HTTP 400 with duplicate integration name because the script actually deletes the integration with the same name before migrating in the `migrate_integration` method (`tools/pagerduty-migrator/migrator/resources/integrations.py:33`). Could you provide some more info on how to reproduce it?

Thank you so much for your contribution and please let me know if you have any questions or feedback!

gilbertobr:
Thanks, I'll do some tests following what was passed.

tools/pagerduty-migrator/scripts/README.md
gilbertobr: I just created the PR: https://github.com/grafana/oncall/pull/403

tools/pagerduty-migrator/migrator/resources/integrations.py
vadimkerr: Please see https://github.com/grafana/oncall/pull/483.

PR #405, opened 2022-08-24
Generate unique distinction for alert groups with broken templates / unrenderable distinction

vadimkerr:
There's a `unique_together` constraint on the `AlertGroup` model for fields (`channel_id`, `channel_filter_id`, `distinction`, `is_open_for_grouping`). We rely on this uniqueness when grouping the alert groups, but in some cases when the distinction cannot be rendered due to the templating errors, the `distinction` field is set to `None`. In this case there could exist multiple alert groups with `distinction=None` and it breaks the expectation of uniqueness for this list of fields.
This PR makes sure that every alert group has a distinction not equal to `None`. In case there's an error in templating, a unique distinction will be generated, so every alert will end up in different alert groups for channels with broken templates / unrenderable distinction.

### DIFF
--- a/engine/apps/alerts/models/alert.py
+++ b/engine/apps/alerts/models/alert.py
@@ -196,8 +196,8 @@ class Alert(models.Model):
         if grouping_id_template is not None:
             group_distinction, _ = apply_jinja_template(grouping_id_template, raw_request_data)
 
-        # Insert demo uuid to prevent grouping of demo alerts.
-        if is_demo:
+        # Insert demo uuid to prevent grouping of demo alerts or alerts with group_distinction=None
+        if is_demo or not group_distinction:
             group_distinction = cls.insert_demo_uuid(group_distinction)
 
         if group_distinction is not None:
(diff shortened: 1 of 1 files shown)


engine/apps/alerts/models/alert.py
```diff
@@ -196,8 +196,8 @@ def render_group_data(cls, alert_receive_channel, raw_request_data, is_demo=Fals
         if grouping_id_template is not None:
             group_distinction, _ = apply_jinja_template(grouping_id_template, raw_request_data)
 
-        # Insert demo uuid to prevent grouping of demo alerts.
-        if is_demo:
+        # Insert demo uuid to prevent grouping of demo alerts or alerts with group_distinction=None
+        if is_demo or not group_distinction:
             group_distinction = cls.insert_demo_uuid(group_distinction)
```
Konstantinov-Innokentii, line 201: rename **insert_demo_uuid** to **insert_random_uuid**

Konstantinov-Innokentii (approved):

PR #414, opened 2022-08-25
Remove auto-recreating logic for UserNotificationPolicy

Konstantinov-Innokentii:
It's removed to get rid of select_for_update on User on each notify_user_task.

Now we checking presence of UserNotificationPolicies on each notification and if they are deleted we create set of default ones. For that logic select_for_update is needed to prevent duplicates of these sets, but it causes Deadlocks. (31 last day) So, I decided to remove this logic and just forbid to delete last UserNotificationPolicy.

### DIFF
--- a/engine/apps/base/models/user_notification_policy.py
+++ b/engine/apps/base/models/user_notification_policy.py
@@ -184,7 +184,7 @@ class UserNotificationPolicy(OrderedModel):
             return "Not set"
 
     def delete(self):
-        if self.order == 0:
+        if UserNotificationPolicy.objects.filter(important=self.important, user=self.user).count() == 1:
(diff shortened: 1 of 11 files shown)


engine/apps/base/models/user_notification_policy.py
```diff
@@ -69,23 +70,8 @@ def validate_channel_choice(value):
...
-        if notification_policies.exists():
-            return notification_policies
-
-        if important:
-            policies = self.create_important_policies_for_user(user)
-        else:
-            policies = self.create_default_policies_for_user(user)
-
-        return policies
+    def get_user_policies(self, user: User, important: bool) -> "QuerySet[UserNotificationPolicy]":
```
vadimkerr, line 73: probably worth deleting `get_user_policies` and just use filter?

vadimkerr (approved):

PR #423, opened 2022-08-29
Add unlink slack account option to profile

### LINKED ISSUE #342: Allow user to unlink slack account
There is no way to change or unlink a slack account in the user's profile.

matiasb:
Fixes #342

### DIFF
--- a/engine/apps/api/views/user.py
+++ b/engine/apps/api/views/user.py
@@ -357,6 +357,13 @@ class UserView(
         user = self.get_object()
         user.slack_user_identity = None
         user.save(update_fields=["slack_user_identity"])
+        write_chatops_insight_log(
+            author=request.user,
+            event_name=ChatOpsEvent.USER_UNLINKED,
+            chatops_type=ChatOpsType.SLACK,
+            linked_user=user.username,
+            linked_user_id=user.public_primary_key,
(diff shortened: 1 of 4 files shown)


engine/apps/api/views/user.py
```diff
@@ -350,6 +352,13 @@ def get_telegram_verification_code(self, request, pk):
 
         return Response({"telegram_code": str(new_code.uuid), "bot_link": bot_link}, status=status.HTTP_200_OK)
 
+    @action(detail=True, methods=["post"])
+    def unlink_slack(self, request, pk):
+        user = self.get_object()
+        user.slack_user_identity = None
```
vadimkerr, line 358: Should we also delete the `SlackUserIdentity` instance in case there are no users associated with it?

vadimkerr (approved):
LGTM

engine/apps/api/views/user.py
matiasb: I based this on the existing reset slack process (`ResetSlackView`), but also I think we don't want to lose track of any potential data (messages, etc) associated to the slack user?

vadimkerr: Makes sense 👍

PR #432, opened 2022-08-30
Fix events start date calculation

Ferril:
Revert changes in events start date calculation, fix bug with getting wrong events start date


matiasb (approved):
LGTM

PR #434, opened 2022-08-30
Add drone step for creating & pushing docker manifest list + cleanup

vadimkerr:


mderynck (approved):
nit: maybe we do something about the pipeline names? I feel like these will easily be mixed up
```
OSS engine release (amd64)
OSS engine release (arm64)
```
Otherwise lgtm

vadimkerr:
> nit: maybe we do something about the pipeline names? I feel like these will easily be mixed up
> 
> ```
> OSS engine release (amd64)
> OSS engine release (arm64)
> ```
> 
> Otherwise lgtm

Thanks, I'll merge it now to test things out and will try to make `arm64` vs `amd64` more distinguishable. Not sure how to do it though, probably upper case `ARM64` vs `AMD64` looks better?

PR #448, opened 2022-08-31
Add external postgresql support to helm chart

matiasb:


iskhakov:
fixed in https://github.com/grafana/oncall/pull/661/files

PR #488, opened 2022-09-06
Alert group search

vadimkerr:


engine/apps/public_api/views/integrations.py
```diff
@@ -58,17 +59,27 @@ def get_object(self):
...
-            instance=serializer.instance,
+            instance=instance,
             author=self.request.user,
             event=EntityEvent.UPDATED,
             prev_state=prev_state,
             new_state=new_state,
         )
 
+        if new_web_title_template != prev_web_title_template:
+            update_verbose_name_for_alert_receive_channel.delay(instance.pk)
```
Matvey-Kuk, line 81: Nice idea about delay using PK.

engine/apps/alerts/models/alert_group.py
```diff
@@ -899,7 +899,7 @@ def wipe_by_user(self, user: User) -> None:
             self.resolve(resolved_by=AlertGroup.WIPED)
             self.stop_escalation()
             self.distinction = ""
-            self.verbose_name = "Wiped incident"
+            self.verbose_name = None
```
Matvey-Kuk, line 902: By this PR you change the purpose of `verbose_name` field of AlertGroup. It used to store rendered name for UI, now it acts more as an `search_cache`. It worth mentioning next to the model field definition. Also worth mentioning why exactly we don't want to migrate.

P.S. rock-n-roll could be renaming the field to smth self-explanory, but anchoring using `db_column='verbose_name'` to avoid migrations.

engine/apps/alerts/tasks/alert_group_verbose_name.py
```diff
@@ -0,0 +1,70 @@
...
+# BATCH_SIZE is how many alert groups will be processed per second (for every individual alert receive channel)
+BATCH_SIZE = 1000
+
+
+def batch_ids(queryset, cursor):
+    return list(queryset.filter(id__gt=cursor).order_by("id").values_list("id", flat=True)[:BATCH_SIZE])
+
+
+@shared_dedicated_queue_retry_task
+def update_verbose_name_for_alert_receive_channel(alert_receive_channel_pk):
```
Matvey-Kuk, line 17: Some comments are missing: Why do we need to do this? When?

engine/apps/alerts/tasks/alert_group_verbose_name.py
```diff
@@ -0,0 +1,70 @@
...
+    while ids:
+        update_verbose_name.apply_async((alert_receive_channel_pk, ids), countdown=countdown)
+
+        cursor = ids[-1]
+        ids = batch_ids(queryset, cursor)
+        countdown += 1
+
+
+@shared_dedicated_queue_retry_task
+def update_verbose_name(alert_receive_channel_pk, alert_group_pks):
```
Matvey-Kuk, line 34: It's silent in terms of logs. Will it be comfortable for release engineer to roll it out?

engine/settings/prod_without_db.py
```diff
@@ -139,6 +139,8 @@ def on_uwsgi_worker_exit():
     "apps.schedules.tasks.drop_cached_ical.drop_cached_ical_for_custom_events_for_organization": {"queue": "critical"},
     "apps.schedules.tasks.drop_cached_ical.drop_cached_ical_task": {"queue": "critical"},
     # LONG
+    "apps.alerts.tasks.alert_group_verbose_name.update_verbose_name_for_alert_receive_channel": {"queue": "long"},
+    "apps.alerts.tasks.alert_group_verbose_name.update_verbose_name": {"queue": "long"},
```
Matvey-Kuk, line 143: 👍

Matvey-Kuk (changes requested):
Using pre-rendered templates stored in DB we face invalidation problem. That's ok, but we don't give users any idea about it in the UI. We discussed that, but I see this PR is not in a draft mode so looks like it's forgotten.

engine/apps/public_api/views/integrations.py
Matvey-Kuk: I don't like re-rendering all old Alert Groups even in the background. It will cause pretty unpredictable impact on infrastructure. Didn't we agree on leaving them with old cache and speaking about it in the UI?

engine/apps/alerts/models/alert_group.py
vadimkerr: 👍 , I think we can just rename, should be pretty cheap in terms of migration

engine/apps/alerts/tasks/alert_group_verbose_name.py
vadimkerr: Added some additional debug logging

engine/apps/public_api/views/integrations.py
vadimkerr: Removed cache invalidation, let's try this without it.

engine/apps/alerts/tasks/alert_group_web_title_cache.py
```diff
@@ -0,0 +1,87 @@
...
+
+@shared_dedicated_queue_retry_task
+def update_web_title_cache(alert_receive_channel_pk, alert_group_pks):
+    """
+    Update the web_title_cache field for alert groups with pk in alert_group_pks,
+    for alert receive channel with pk = alert_receive_channel_pk.
+    """
+    task_logger.debug(
+        f"Starting update_web_title_cache, alert_receive_channel_pk: {alert_receive_channel_pk}, "
+        f"first alert_group_pk: {alert_group_pks[0]}, last alert_group_pk: {alert_group_pks[-1]}"
```
Matvey-Kuk, line 49: ❤️

Matvey-Kuk (approved):

Issue #135, opened 2022-06-22
Display a sample command when creating API token


mderynck:
On the Grafana OnCall->Settings page when you create a new API token it would be nice if it also displayed an example curl command to make use of the token for testing instead of needing to lookup in the docs.  This would match the behavior that Grafana has when you create an API key in its settings page.

Konstantinov-Innokentii:
Good idea, we have /info endpoint in public api. But I am wondering which url wi should use, BASE_URL from settings I guess?

Matvey-Kuk:
We added this task to the core team's backlog.

raphael-batte:
Some concept 

![Image](https://user-images.githubusercontent.com/102970268/177986735-13079746-4b4c-42b9-a64d-7311929b8a60.png)

PR #503, opened 2022-09-07
135 - token form

teodosii:
Fix for #135

### DIFF
--- a/grafana-plugin/src/components/SourceCode/SourceCode.module.scss
+++ b/grafana-plugin/src/components/SourceCode/SourceCode.module.scss
@@ -2,28 +2,27 @@
   position: relative;
   width: 100%;
 
-  &:hover .button {
+  &:hover .copyButton,
+  &:hover .copyIcon {
     opacity: 1;
   }
 }
 
 .scroller {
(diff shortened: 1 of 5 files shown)


grafana-plugin/src/components/SourceCode/SourceCode.module.scss
```diff
@@ -1,29 +1,29 @@
...
   opacity: 0;
   transition: opacity 0.2s ease;
 }
 
-.root:hover .button {
-  opacity: 1;
-}
-
-.root pre {
+pre {
```
maskin25, line 26: don't add global styles to particular module.css please, the only place to add global styles is `grafanaGlobalStyles.css`

maskin25 (changes requested):

grafana-plugin/src/components/SourceCode/SourceCode.module.scss
teodosii: Hmm, that's a mistake I see. it should've been `.root pre` because it was already existing

maskin25 (approved):

PR #509, opened 2022-09-08
Users from search results have been added to the state of options

Ukochka:
The problem was that when we search in dropdown we did not add search result to the options and that is why we could not add them to values.


grafana-plugin/src/containers/RemoteSelect/RemoteSelect.tsx
```diff
@@ -58,14 +56,24 @@ const RemoteSelect = inject('store')(
       }));
     };
 
+    function mergeOptions(oldOptions: SelectableValue[], newOptions: SelectableValue[]) {
+      const existedValues = oldOptions.map((o) => o.value);
+      return newOptions.filter(({ value }) => !existedValues.includes(value)).concat(oldOptions);
```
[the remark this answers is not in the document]
maskin25: I agree

grafana-plugin/src/containers/RemoteSelect/RemoteSelect.tsx
```diff
@@ -58,14 +56,24 @@ const RemoteSelect = inject('store')(
...
     useEffect(() => {
       makeRequest(href, {}).then((data) => {
         setOptions(getOptions(data.results || data));
       });
     }, []);
 
     const loadOptionsCallback = useCallback((query: string) => {
-      return makeRequest(href, { params: { search: query } }).then((data) => getOptions(data.results || data));
+      return makeRequest(href, { params: { search: query } }).then((data) => {
+        setOptions(getOptions(data.results || data));
```
maskin25, line 74: this `loadOptionsCallback` should return old options +  new options (merged together), but it works in some other weird way

maskin25 (changes requested):

grafana-plugin/src/containers/RemoteSelect/RemoteSelect.tsx
Ukochka: Had a call with @maskin25 and decided to leave it as it is

maskin25 (approved):

Issue #86, opened 2022-06-15
Add arm64 docker image for grafana/oncall


vadimkerr:
When running the following command from README.md: `docker-compose --env-file .env_hobby -f docker-compose.yml up --build -d`, uWSGI fails to start on my M1 Mac.

uWSGI logs:

```
[uWSGI] getting INI configuration from uwsgi.ini
[log-encoder] registered format ${strftime:%Y-%m-%d %H:%M:%S} ${msgnl}
2022-06-15 08:06:33 *** Starting uWSGI 2.0.20 (64bit) on [Wed Jun 15 08:06:33 2022] ***
2022-06-15 08:06:33 compiled with version: 11.2.1 20220219 on 14 June 2022 18:16:43
2022-06-15 08:06:33 os: Linux-5.10.104-linuxkit #1 SMP PREEMPT Thu Mar 17 17:05:54 UTC 2022
2022-06-15 08:06:33 nodename: cfcc13d1c266
2022-06-15 08:06:33 machine: x86_64
2022-06-15 08:06:33 clock source: unix
2022-06-15 08:06:33 pcre jit disabled
2022-06-15 08:06:33 detected number of CPU cores: 4
2022-06-15 08:06:33 current working directory: /etc/app
2022-06-15 08:06:33 writing pidfile to /tmp/project-master.pid
2022-06-15 08:06:33 detected binary path: /usr/local/bin/uwsgi
2022-06-15 08:06:33 uWSGI running as root, you can use --uid/--gid/--chroot options
2022-06-15 08:06:33 setgid() to 2000
2022-06-15 08:06:33 setuid() to 1000
2022-06-15 08:06:33 chdir() to /etc/app
2022-06-15 08:06:33 your memory page size is 4096 bytes
2022-06-15 08:06:33 detected max file descriptor number: 1048576
2022-06-15 08:06:33 lock engine: pthread robust mutexes
2022-06-15 08:06:33 unable to set PTHREAD_PRIO_INHERIT
```

When building the `grafana/oncall` image manually on my machine and using it in the docker-compose file, all works as expected. My proposal is to add a docker image for arm64 to DockerHub, so the instruction works for Apple Silicon/arm64 users.

Konstantinov-Innokentii:
Strange, I ran this on m1 pro, everything worked.

iskhakov:
Agree, 
let's add another env variable, that use can configure during docker-compose up, to use arm64 image
and add the ci job to build such an image

[...]

PR #513, opened 2022-09-08
Add instructions for running on arm64

scubbo:
See [issue](https://github.com/grafana/oncall/issues/86)

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,10 @@ We prepared multiple environments: [production](https://grafana.com/docs/grafana
 curl -fsSL https://raw.githubusercontent.com/grafana/oncall/dev/docker-compose.yml -o docker-compose.yml
 ```
 
+If you are attempting to run Oncall on a Raspberry Pi or other ARM64 architecture, make the following changes throughout:
+* `mysql:5.7` -> `arm64v8/mysql`
+* `platform: linux/x86_64` -> `platform: linux`
+
 2. Set variables:
 ```bash
 echo "DOMAIN=http://localhost:8080


README.md
```diff
@@ -25,6 +25,10 @@ We prepared multiple environments: [production](https://grafana.com/docs/grafana
 curl -fsSL https://raw.githubusercontent.com/grafana/oncall/dev/docker-compose.yml -o docker-compose.yml
 ```
 
+If you are attempting to run Oncall on a Raspberry Pi or other ARM64 architecture, make the following changes throughout:
+* `mysql:5.7` -> `arm64v8/mysql`
```
vadimkerr, line 29: Hmm, I don't see a `5.7` tag on `arm64v8/mysql`, the earliest `arm64v8/mysql` version is `8.x.x`. Grafana OnCall could potentially run on MySQL 8 (I'm not sure, haven't tested that), but I wouldn't recommend this since it's only tested on `5.7`. I could propose to create a separate issue on updating docker-compose images and helm chart to use mysql 8, as well as changing this in our CI.
Then, after updating to mysql 8 we wouldn't need to use `arm64v8/mysql`, the official mysql docker repo has arm64 images for versions `8.x.x`.

vadimkerr (changes requested):
Thank you for your contribution! 
I'm not sure we want to include this change at the moment, but it raises a valid question about adding support for modern MySQL versions.

README.md
scubbo: Sounds good - thanks for your feedback! I'll cut that issue, and resolve this PR.

scubbo:
[Issue](https://github.com/grafana/oncall/issues/607).

Issue #472, opened 2022-09-02
Add wrong team error for all team-specific resources


vadimkerr:
Currently there's only one page ("Alert Groups") that is capable of showing a human readable error when user tries to access a resource from a different team:

<img width="482" alt="Screenshot 2022-08-04 at 16 28 23" src="https://user-images.githubusercontent.com/20116910/182886919-639e0fd7-9753-476c-bf66-1c924f159ccc.png">

There are other resources that can benefit from this as well, e.g. integrations, schedules, etc.

Matvey-Kuk:
For example: for "escalation chain".

![Image](https://user-images.githubusercontent.com/3284841/188407134-15e961bb-3658-481c-a23c-57b604f9148f.png)

Matvey-Kuk:
How to reproduce:
- Open any specific object (Escalation chain, Maintenance,  Schedule...)
- Copy url link in the browser
- Switch team
- Past the link to the browser
- You'll see error message

raphael-batte:
Probably we can use here the same UI as we have at Alerts. We just need to change descriptions

PR #520, opened 2022-09-12
472 wrong team

teodosii:
Fix for #472


engine/apps/api/views/user.py
```diff
@@ -245,6 +249,26 @@ def retrieve(self, request, *args, **kwargs):
         serializer = self.get_serializer(instance, context=context)
         return Response(serializer.data)
 
+    def wrong_team_response(self):
+        """
+        This method returns 403 and {"error_code": "wrong_team", "owner_team": {"name", "id", "email", "avatar_url"}}.
```
vadimkerr, line 254: probably worth adding a comment why TeamFilteringMixin cannot be used here

vadimkerr (approved):

grafana-plugin/src/models/base_store.ts
```diff
@@ -37,10 +41,13 @@ export default class BaseStore {
   }
 
   @action
-  async getById(id: string) {
-    return await makeRequest(`${this.path}${id}/`, {
+  async getById(id: string, skipErrorHandling = false, fromOrganization = false) {
+    let path = `${this.path}${id}`;
+    if (fromOrganization) {path = path.concat('?from_organization=true');}
+
+    return await makeRequest(path, {
```
maskin25, line 48: `params: { from_organization:fromOrganization? true: undefined}`

grafana-plugin/src/models/schedule/schedule.ts
```diff
@@ -25,6 +25,18 @@ export class ScheduleStore extends BaseStore {
     this.path = '/schedules/';
   }
 
+  @action
+  async loadItem(id: Schedule['id'], skipErrorHandling = false): Promise<Schedule> {
```
maskin25, line 29: great that you add typings for promises!

maskin25:
I don't like that you copypasted 

```
notFound?: boolean;
  wrongTeamError?: boolean;
  teamToSwitch?: { name: string; id: string };
  wrongTeamNoPermissions?: boolean;
```
to many pages, it would be good to wrap this state, requests, logic and so on into separate service/top level component and manage this stuff in one place to avoid copypaste

maskin25 (changes requested):

maskin25 (approved):

PR #522, opened 2022-09-12
Refresh ical files on custom shift updates instead of dropping them.

Ferril:


matiasb (approved):
LGTM

Ferril:
Close pr because refreshing ical triggers slack notification too often

Issue #249, opened 2022-07-18
Improve OnCall plugin initialization latency


mderynck:
When you refresh the browser OnCall will run a synchronization process to ensure it knows about all users in the Grafana instance.  Currently this is implemented as a task in celery with the front-end checking for completion.  This was done to reduce request latency.  If you have recently synced and nothing has changed with users this step is unnecessary.

Unfortunately we cannot cache this based on time because in the case you add a new user to Grafana the plugin won't be able to find out about until the cache expires or the user is given a path to force refresh.

Currently plugin sync tries to sync by waiting 2s at a time, we could reduce this to 0.5s, 1s, 2s to try to improve latency.

mderynck:
Another spot for improvement here is switching teams triggers a reload/initialization.  If possible switching teams should be more like navigating between tabs and not trigger a plugin reload.

PR #529, opened 2022-09-14
Increase num of `getPluginSyncStatus` retries to 10

mirismaili:
Already 5 retries (5 * 2 sec = 10 sec) is not enough in our cases. Usually, we have 15 seconds of latency.


mderynck:
Ideally if we used an exponential backoff here instead of a fixed time interval we could improve latency for fast connections (first request at 100ms), accommodate longer times while limiting the number of requests we are issuing.  This would be able to improve this issue as well https://github.com/grafana/oncall/issues/249

mderynck (approved):

PR #537, opened 2022-09-17
Allow customized SMS message, fix #197

### LINKED ISSUE #197: Allow customized SMS message without surrounding "Grafana OnCall" text
Currently SMS are not customizable except for a string in the middle of the fixed message:

> You are invited to check an incident #515 with title "This is a test title" in Grafana OnCall organization: "this is a test organization", alert channel: Grafana Alerting - Grafana Alerting, alerts registered: 1, https://thisisatestorganization.grafana.net/a/grafana-oncall-app/?page=incident&id=I3H1WHSAJ8HWF
> Your Grafana OnCall <3

The users on our end do not universally speak English, nor do they interface with Grafana OnCall directly.

Therefore we would like to be able to change the _whole_ message content without surrounding fixed references to _Grafana Oncall_, links and the ascii heart.

The current [sms renderer](https://github.com/grafana/oncall/blob/b0d98e772eb87ed75e35bd6ca015402c601457b5/engine/apps/alerts/incident_appearance/renderers/sms_renderer.py#L25) could be changed to leave out that additional fixed text.

ohidurbappy:
#### What's changed
- Allows Customized SMS messages for Oncall with a Fixed prefix - "**Grafana OnCall alert:**" as suggested by @Matvey-Kuk to avoid system misuse of any kind. 
- The default content of the sms_title is changed to a template that is very similar to the current SMS format. 
- A new `grafana_onlink_organization` variable is added to the reserved template variable to be used within a message.

### DIFF
--- a/engine/apps/alerts/incident_appearance/renderers/sms_renderer.py
+++ b/engine/apps/alerts/incident_appearance/renderers/sms_renderer.py
@@ -18,15 +18,6 @@ class AlertGroupSmsRenderer(AlertGroupBaseRenderer):
     def render(self):
         templated_alert = self.alert_renderer.templated_alert
         title = str_or_backup(templated_alert.title, DEFAULT_BACKUP_TITLE)
-        if self.alert_group.channel.organization.slack_team_identity and (permalink := self.alert_group.permalink):
-            incident_link = permalink
-        else:
-            incident_link = self.alert_group.web_link
         return (
-            f"You are invited to check an incident #{self.alert_group.inside_organization_number} with title "
-            f'"{title}" in Grafana OnCall organization: "{self.alert_group.channel.organization.stack_slug}", '
-            f"alert channel: {self.alert_group.channel.short_name}, "
-            f"alerts registered: {self.alert_group.alerts.count()}, "
-            f"{incident_link}\n"
-            f"Your Grafana OnCall <3"
+            f"Grafana OnCall alert : {title}"
         )
--- a/engine/config_integrations/webhook.py
+++ b/engine/config_integrations/webhook.py
@@ -32,7 +32,18 @@ web_message = """\
 
 web_image_url = slack_image_url
 
-sms_title = web_title
+sms_title = """\
+{{ payload.message }}, 
+Incident #
+{{ grafana_oncall_incident_id }}, 
+Organization: 
+{{ grafana_oncall_organization }}, 
+Alert channel: 
+{{ integration_name }},
+Link: 
+{{ grafana_oncall_link }}
+
+"""
 
 phone_call_title = sms_title
 
--- a/engine/apps/alerts/incident_appearance/templaters/alert_templater.py
+++ b/engine/apps/alerts/incident_appearance/templaters/alert_templater.py
@@ -167,6 +167,7 @@ class AlertTemplater(ABC):
             context = {
                 "grafana_oncall_incident_id": self.incident_id,
                 "grafana_oncall_link": self.link,
+                "grafana_oncall_organization": channel.organization.stack_slug,
                 "integration_name": channel.verbal_name,
                 "source_link": templated_alert.source_link,
                 "amixr_incident_id": self.incident_id,  # TODO: decide on variable names
--- a/grafana-plugin/src/components/AlertTemplates/AlertTemplatesForm.tsx
+++ b/grafana-plugin/src/components/AlertTemplates/AlertTemplatesForm.tsx
@@ -205,9 +205,9 @@ const AlertTemplatesForm = (props: AlertTemplatesFormProps) => {
                 {activeGroup === 'telegram' && ', html'}
                 {' supported. '}
                 Reserved variables available: <Text keyboard>payload</Text>, <Text keyboard>grafana_oncall_link</Text>,{' '}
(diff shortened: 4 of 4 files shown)


engine/apps/alerts/incident_appearance/renderers/sms_renderer.py
```diff
@@ -18,15 +18,6 @@ def alert_renderer_class(self):
...
-            incident_link = self.alert_group.web_link
         return (
-            f"You are invited to check an incident #{self.alert_group.inside_organization_number} with title "
-            f'"{title}" in Grafana OnCall organization: "{self.alert_group.channel.organization.stack_slug}", '
-            f"alert channel: {self.alert_group.channel.short_name}, "
-            f"alerts registered: {self.alert_group.alerts.count()}, "
-            f"{incident_link}\n"
-            f"Your Grafana OnCall <3"
-        )
+            f"Grafana OnCall alert : {title}"
```
vadimkerr, line 22: Let's remove this "Grafana OnCall alert : " and just return `title` so the SMS title is fully customizable

engine/config_integrations/webhook.py
```diff
@@ -32,7 +32,18 @@
 
 web_image_url = slack_image_url
 
-sms_title = web_title
+sms_title = """\
```
vadimkerr, line 35: I propose to keep this closer to the current version of SMS (probably can drop the alert count, @Matvey-Kuk wdyt?), for example:
```python
"""\
You are invited to check an alert group #{{ grafana_oncall_incident_id }}
Organization: {{ grafana_oncall_organization }}
Integration: {{ integration_name }}
Title: {{ this could be different based on integration }}
Link: {{ grafana_oncall_slack_link or grafana_oncall_link }}

Your Grafana OnCall <3
"""
```

Probably would be a good idea to add `grafana_oncall_slack_link` to jinja2 context as well, so the there's a link to Slack message as it is currently.

Also please note that there are multiple integrations in [engine/config_integrations](https://github.com/grafana/oncall/tree/dev/engine/config_integrations), so we must make sure to set sms_title on all of them so it can be used not only for webhook integration, but others as well (e.g. alertmanager)

vadimkerr (changes requested):
Thank you so much for your contibution! This is definitely a useful feature.
There are a couple of points to consider:
1. We can add one more variable to jinja2 context (`grafana_oncall_slack_link`) so we can replicate the current version of SMS closer. I think some users find the Slack message link more useful than a general link, so let's keep it this way by default.
2. It's important to specify default SMS title template for other integration types as well (see `slack_title` and `slack_message` templates as a reference)

Please let me know if you need any help with this.

engine/apps/alerts/incident_appearance/renderers/sms_renderer.py
Matvey-Kuk: I think it's important to keep hardcoded "Grafana OnCall" in the SMS to avoid misuse

engine/config_integrations/webhook.py
Matvey-Kuk: I like seeing `grafana_oncall_link` link here. We want to make it les slack-specific. Also nice to add `grafana_oncall_link_slack` as @vadimkerr mentioned.

vadimkerr:
Closing this as stale, @ohidurbappy feel free to reopen 🙂

yokhahn:
@vadimkerr I didn't write this patch, but I don't think the issue is stale.. this is something I'd like as well.. why was it stalled instead of applied?  
This makes it so that sms phone notifications don't show enough quick information to allow somebody to make a quick decision.
Why not just say "OnCall:" at the front of the message if you want to have some kind of standard prefix?

Matvey-Kuk:
@yokhahn 13 days after this PR was opened, Vadim performed the review, left a few comments and proposed help to drive this PR forward and make it merged. Not receiving any response he closed the PR. 

We don't merge PR's until team members are happy with their quality, they provide help with reveiws and providing context, but they don't polish PR's for contributors. 

@yokhahn if you want to make this PR merged, I would suggest re-opening it and working together with a team in order to merge it, I don't think the current approach will work.

engine/apps/alerts/incident_appearance/renderers/sms_renderer.py
yokhahn: Being in the SMS is a separate question from keeping it in the title. As a result, I think the answer you gave didn't answer the question asked.   I also question what you mean by SMS misuse.  Are you worried about somebody using their SMS account to send abusive SMS messages (they could do that straight with the SMS account no need to go through oncall for that), or are you more worried about specifically the Grafana cloud instance being abused, which sounds more like a config question on the cloud version?

I think any progress on this PR would require answer what is wanted/required?  (but I think having this closed won't promote that discussion.  But maybe I'm misunderstanding something.

yokhahn:
> @yokhahn 13 days after this PR was opened, Vadim performed the review, left a few comments and proposed help to drive this PR forward and make it merged. He closed the PR after waiting for some time for a response from the contributor.
> 
> We don't merge PR's until team members are happy with their quality, they provide help with reveiws and providing context, but they don't polish PR's for contributors.
> 
> @yokhahn if you want to make this PR merged, I would suggest re-opening it and working together with a team in order to merge it, I don't think the current approach will work.

13 days to stale?  Why?

PR #538, opened 2022-09-19
Changed Redis Connection String possibilities

Timmypedia:
Hello,
We wan't to use our existing Redis in AWS. For that we need the redis-secure protocol and the possiblity to add the username. It would be much flexibler if we do the same, as you guys did for RabbitMQ.

I made the changes for backward compatibility optional.

Best Regards


joeyorlando (approved):

nu-vi:
Hi, will the REDIS_PROTOCOL env variable be able to be set via the helm values in the future?

Timmypedia:
> Hi, will the REDIS_PROTOCOL and REDIS_USERNAME env variables be able to be set via the helm values in the future?

I made a PR (#694)  to fix this. Last time i forgot to edit the Helm Chart, cause i did it without. Now we need this as well.

nu-vi:
I understand some elasticache configurations require the use of a user, but having the redis_protocol env variable be set of Helm chart is still valuable. 

It's possible to have elasticache configured without the use of an user, as well as other external redis instances where one would want to use the 'rediss' protocol.

PR #543, opened 2022-09-21
#539 - add slack permalink to alert group public API response

### LINKED ISSUE #539: Add link to slack to alert_groups HTTP API response
The UI for oncall exposes a handy button called "View in Slack" when looking at an alert group. I wanted to use this slack link in script I was writing.

Unfortunately, it's not exposed in the documented `GET /api/v1/alert_groups/` endpoint. The UI gets it by making a request to Grafana at `GET /oncall/api/internal/v1/alertgroups`, which I assume gets mapped to `GET api/internal/v1/alertgroups`. 

It would be nice to also expose this link on the public too.

joeyorlando:
This PR closes #539. 

It makes the following changes:
- update the `AlertGroup` public API serializer to include `permalinks` property, a dictionary representing associated well.. permalinks. Right now it only has one key, `slack` which represents a Slack message URL (if the alertgroup has an associated Slack message)
- add `.nvmrc` file to `/grafana-plugin` that is pegged to `14.17.0` (same version we seem to be using on CI). This allows seamless integration for developers using [nvm](https://github.com/nvm-sh/nvm) (and for those whom don't use it, it can simply be ignored)
- add `.python-version` to root `.gitignore`. This allows for developers using [pyenv](https://github.com/pyenv/pyenv)/[pyenv-virtualenv](https://github.com/pyenv/pyenv-virtualenv) to simply place a `.python-version` file in the project root dir to simplify switching/activating of their project's virtualenv
- update `DEVELOPER.md`. I ran into a few small snags so wanted to update based on my experience for the next developer setting the project up for the first time (mostly related to the need to have `postgres` as a prerequisite dependency; there was also some markdown autoformatter changes that my IDE made, I can revert these if preferred)

**Local testing**
Screen Shot 2022-09-21 at 17 57 52

**Updated documentation**
Screen Shot 2022-09-22 at 10 10 52


engine/apps/public_api/tests/test_incidents.py
```diff
@@ -38,6 +38,7 @@ def construct_expected_response_from_incidents(incidents):
                 "resolved_at": resolved_at,
                 "acknowledged_at": acknowledged_at,
                 "title": None,
+                "permalink": None,
```
joeyorlando, line 41: wondering if it is worthwhile adding a test case here to make sure that this is equal to the slack message URL when the alert group's `slack_message` is not `None`?

CHANGELOG.md
```diff
@@ -1,106 +1,138 @@
 # Change Log
 
+## v1.0.37 (2022-09-21)
+
+- Add `permalink` to `AlertGroup` public API response schema
+- Update `DEVELOPER.md` to reflect successfully running project locally for the first time
```
vadimkerr, line 6: Not sure if we want to include this to changelog since we only include user-facing changes there

Matvey-Kuk (changes requested):
Congrats on the first PR!

engine/apps/public_api/serializers/incidents.py
```diff
@@ -29,6 +29,7 @@ class Meta:
             "resolved_at",
             "acknowledged_at",
             "title",
+            "permalink",
```
vadimkerr, line 32: I sense a n+1 issue here since permalink depends on `SlackMessage` model, please check if select_related / prefetch_related is helpful here.

engine/apps/public_api/tests/test_incidents.py
vadimkerr: Yeah I guess we can add a test on this, not so critical though

vadimkerr (changes requested):

DEVELOPER.md
```diff
@@ -96,20 +106,22 @@ yarn watch
...
 
 5. Open Grafana in the browser http://localhost:3000 (login: oncall, password: oncall) notice OnCall Plugin is not enabled, navigate to Configuration->Plugins and click Grafana OnCall
 
 6. Some configuration fields will appear be available. Fill them out and click Initialize OnCall
+
 ```
-OnCall API URL: 
-http://host.docker.internal:8080
+OnCall API URL:
+http://host.docker.internal:8000
```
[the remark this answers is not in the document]
vadimkerr: This is `DEVELOPER.md` so probably it's not breaking the instruction. I personally use port 8000 on dev so it doesn't clash with oss instalation running on port 8080, so I recommended Joey to use 8000 when I was helping him with his local developer setup. Being able to use non-dev docs with 8080 can be useful though, so now I'm not so sure about changing dev to 8000. wdyt @Matvey-Kuk?

engine/apps/public_api/tests/test_incidents.py
vadimkerr: Nice idea to add links for multiple messengers @Matvey-Kuk!

CHANGELOG.md
Matvey-Kuk: @vadimkerr it's public API change, so sounds pretty user-facing for me :)

joeyorlando: just to clarify, you mean just the line regarding the `DEVELOPER.md` change, correct?

DEVELOPER.md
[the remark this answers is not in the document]
joeyorlando: actually after retrying things by running `engine` on `8080` I was able to get things working smoothly, I'll revert this port change in the documentation (must've been some other unrelated issue I was running into during initial setup)

engine/apps/public_api/serializers/incidents.py
joeyorlando: Good point! I noticed in the "internal" api serializer we don't add `permalink` to either `SELECT_RELATED` or `PREFETCH_RELATED`. Should the public api serializer deviate from this?

engine/apps/public_api/tests/test_incidents.py
joeyorlando: good suggestion @Matvey-Kuk! will refactor to take this into consideration

CHANGELOG.md
vadimkerr: yup, only the second point about DEVELOPER.md

engine/apps/public_api/serializers/incidents.py
vadimkerr: Yes, the internal api is a little bit different and only returns permalinks for requests to particular alert group. The `AlertGroupListSerializer` doesn't have a `permalink` field so it's not in `SELECT_RELATED` or `PREFETCH_RELATED`. For public api we only have one serialiser so we probably should add it here.

engine/apps/public_api/tests/test_incidents.py
joeyorlando: Renamed the field in the public API `AlertGroup` serializer from `permalink` to a dict named `permalinks`.

*Note*: I think it eventually would make sense to rename the `permalink` property on the `AlertGroup` model (which right now represents the slack message permalink). I avoided doing that in this PR to avoid making a bigger/out-of-scope PR and instead just added a TODO note in the relevant code. Thoughts?

vadimkerr (approved):
lgtm!
Couple things to consider:
- Please reflect this change in public API docs: https://github.com/grafana/oncall/blob/3db4038a4fdec4b01c98b78069e90bf9017cc577/docs/sources/oncall-api-reference/alertgroups.md
- Consider adding more permalinks (telegram, msteams), I propose to do it in another PR.

engine/settings/dev.py
```diff
@@ -9,6 +9,8 @@
 
     pymysql.install_as_MySQLdb()
 
+DEBUG = True
```
vadimkerr, line 12: 👍

joeyorlando:
> lgtm! Couple things to consider:
> 
> * Please reflect this change in public API docs: https://github.com/grafana/oncall/blob/3db4038a4fdec4b01c98b78069e90bf9017cc577/docs/sources/oncall-api-reference/alertgroups.md
> * Consider adding more permalinks (telegram, msteams), I propose to do it in another PR.

@vadimkerr see [this commit](https://github.com/grafana/oncall/pull/543/commits/b8117edf30c71bdc8aabf447e75e7f4b548c8e6a) which updates the public API docs

vadimkerr:
> @vadimkerr see [this commit](https://github.com/grafana/oncall/pull/543/commits/b8117edf30c71bdc8aabf447e75e7f4b548c8e6a) which updates the public API docs

Oh, I see, thanks! Haven't noticed that on my first scan.

grafana-plugin/.nvmrc
```diff
@@ -0,0 +1 @@
+14.17.0
```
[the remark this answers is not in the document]
joeyorlando: @teodosii confirmed the version in Slack [here](https://raintank-corp.slack.com/archives/C0229FD3CE9/p1663843944952399?thread_ts=1663748429.944139&cid=C0229FD3CE9)

engine/apps/public_api/tests/test_incidents.py
Matvey-Kuk: @joeyorlando migrating AlertGroup model is dangerous because it's HUGE and may cause long downtime while migrating, but it's good not to collect tech debt. So I'm ok with any decision you make :)

PR #547, opened 2022-09-21
Use phonenumbers for validation instead of regex

mderynck:
Use phonenumbers package instead of regex to validate live settings to support EU number for Twilio


mderynck:
https://github.com/grafana/oncall-private/issues/1078

matiasb (approved):

PR #548, opened 2022-09-21
Rework queries for alert groups bulk actions

matiasb:
Avoid OR queries when filtering root vs dependent alert groups, switching to split queries/updates instead.

### DIFF
--- a/engine/apps/alerts/models/alert_group.py
+++ b/engine/apps/alerts/models/alert_group.py
@@ -978,18 +978,8 @@ class AlertGroup(AlertGroupSlackRenderingMixin, EscalationSnapshotMixin, models.
         self.delete()
 
     @staticmethod
-    def bulk_acknowledge(user: User, alert_groups: "QuerySet[AlertGroup]") -> None:
+    def _bulk_acknowledge(user: User, alert_groups_to_acknowledge: "QuerySet[AlertGroup]") -> None:
         AlertGroupLogRecord = apps.get_model("alerts", "AlertGroupLogRecord")
-        root_alert_groups_to_acknowledge = alert_groups.filter(
-            ~Q(acknowledged=True, resolved=False),  # don't need to ack acknowledged incidents once again
-            root_alert_group__isnull=True,
-            maintenance_uuid__isnull=True,  # don't ack maintenance incident
-        )
-        # Find all dependent alert_groups to update them in one query
-        dependent_alert_groups_to_acknowledge = AlertGroup.all_objects.filter(
-            root_alert_group__in=root_alert_groups_to_acknowledge
-        )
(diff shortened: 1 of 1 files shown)


engine/apps/alerts/models/alert_group.py
```diff
@@ -1098,41 +1081,39 @@ def bulk_resolve(user: User, alert_groups: "QuerySet[AlertGroup]") -> None:
...
+        root_alert_groups_to_resolve = alert_groups.filter(
             resolved=False,
-            acknowledged=True,
             root_alert_group__isnull=True,
-            maintenance_uuid__isnull=True,  # don't restart maintenance incident
+            maintenance_uuid__isnull=True,
         )
-        dependent_alert_groups_unack = AlertGroup.all_objects.filter(root_alert_group__in=root_alert_groups_unack)
-        alert_groups_to_restart_unack = root_alert_groups_unack | dependent_alert_groups_unack
+        if root_alert_groups_to_resolve.count() == 0:
```
vadimkerr, line 1095: probably `.exists` will be more efficient here
```suggestion
        if not root_alert_groups_to_resolve.exists():
```

engine/apps/alerts/models/alert_group.py
```diff
@@ -1042,31 +1032,24 @@ def bulk_acknowledge(user: User, alert_groups: "QuerySet[AlertGroup]") -> None:
...
-        dependent_alert_groups_to_resolve = AlertGroup.all_objects.filter(
-            root_alert_group__in=root_alert_groups_to_resolve
+        # Find all dependent alert_groups to update them in one query
+        # convert qs to list to prevent changes by update
+        root_alert_group_pks = list(root_alert_groups_to_acknowledge.values_list("pk", flat=True))
+        dependent_alert_groups_to_acknowledge = AlertGroup.unarchived_objects.filter(
+            root_alert_group__pk__in=root_alert_group_pks
         )
-        alert_groups_to_resolve = root_alert_groups_to_resolve | dependent_alert_groups_to_resolve
+        AlertGroup._bulk_acknowledge(user, root_alert_groups_to_acknowledge)
```
vadimkerr, line 1047: we can probably run these two queries in a transaction so we don't have any inconsistencies between state of root and dependent alert groups

vadimkerr (approved):
lgtm

engine/apps/alerts/models/alert_group.py
```diff
@@ -1042,31 +1032,24 @@ def bulk_acknowledge(user: User, alert_groups: "QuerySet[AlertGroup]") -> None:
...
+        AlertGroup._bulk_acknowledge(user, dependent_alert_groups_to_acknowledge)
```
vadimkerr, line 1048: we can probably run these two queries in a transaction so we don't have any inconsistencies between state of root and dependent alert groups

PR #551, opened 2022-09-22
Feat 549 - add telegram permalink to alert groups http response

### LINKED ISSUE #549: Add telegram permalink to alert_groups HTTP API response

joeyorlando:
This PR closes #549.

It does the following:
- renames/refactors the `AlertGroup` model's `permalink` property to `slack_permalink`
- adds a `telegram` key to the `permalinks` object in the alert group's public API HTTP response object

**Tested locally**
Screen Shot 2022-09-22 at 18 09 46

### DIFF
--- a/engine/apps/alerts/models/alert_group.py
+++ b/engine/apps/alerts/models/alert_group.py
@@ -407,10 +407,8 @@ class AlertGroup(AlertGroupSlackRenderingMixin, EscalationSnapshotMixin, models.
 
     @property
     def telegram_permalink(self) -> typing.Optional[str]:
-        main_telegram_message = self.telegram_messages.all()
-
-        if main_telegram_message:
-            return main_telegram_message[0].link
+        if hasattr(self, "prefetched_telegram_messages"):
+            return self.prefetched_telegram_messages[0].link if self.prefetched_telegram_messages else None
         return None
 
     @property
--- a/engine/apps/alerts/models/alert_group.py
+++ b/engine/apps/alerts/models/alert_group.py
@@ -406,9 +406,18 @@ class AlertGroup(AlertGroupSlackRenderingMixin, EscalationSnapshotMixin, models.
             return self.slack_message.permalink
 
     @property
+    def telegram_permalink(self) -> typing.Optional[str]:
+        main_telegram_message = self.telegram_messages.all()
+
+        if main_telegram_message:
+            return main_telegram_message[0].link
+        return None
+
+    @property
     def permalinks(self) -> Permalinks:
         return {
             "slack": self.slack_permalink,
+            "telegram": self.telegram_permalink,
         }
 
     @property
--- a/engine/apps/alerts/models/alert_group.py
+++ b/engine/apps/alerts/models/alert_group.py
@@ -401,15 +401,14 @@ class AlertGroup(AlertGroupSlackRenderingMixin, EscalationSnapshotMixin, models.
         raise NotImplementedError
 
     @property
-    def permalink(self):
+    def slack_permalink(self):
         if self.slack_message is not None:
             return self.slack_message.permalink
 
     @property
     def permalinks(self) -> Permalinks:
-        # TODO: refactor 'permalink' property (maybe 'slack_permalink'?) once we add the next permalink
         return {
-            "slack": self.permalink,
+            "slack": self.slack_permalink,
         }
 
     @property
--- a/engine/apps/alerts/models/alert_group.py
+++ b/engine/apps/alerts/models/alert_group.py
@@ -408,19 +408,21 @@ class AlertGroup(AlertGroupSlackRenderingMixin, EscalationSnapshotMixin, models.
     @property
     def telegram_permalink(self) -> typing.Optional[str]:
         """
-            This property will attempt to access an attribute, `prefetched_telegram_messages`, representing a list of
-            prefetched telegram messages. If this attribute does not exist, it falls back to performing a query.
+        This property will attempt to access an attribute, `prefetched_telegram_messages`, representing a list of
+        prefetched telegram messages. If this attribute does not exist, it falls back to performing a query.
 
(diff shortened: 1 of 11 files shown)


joeyorlando:
A question I had was whether it or it makes sense to return just one, or multiple telegram permalinks? As I understand an `AlertGroup` has a one-to-many relationship with a `TelegramMessage`.

vadimkerr:
> A question I had was whether it or it makes sense to return just one, or multiple telegram permalinks? As I understand an `AlertGroup` has a one-to-many relationship with a `TelegramMessage`.

I would say we only want to show the main alert group message in the channel (not DMs). The query would be something like this:

```python
alert_group.telegram_messages.filter(Q(chat_id__startswith="-") & Q(message_type=TelegramMessage.ALERT_GROUP_MESSAGE).first()
```

If `chat_id` starts with "-" it means it's a channel message and not a DM.

engine/apps/alerts/models/alert_group.py
```diff
@@ -401,15 +401,25 @@ def is_alert_a_resolve_signal(self, alert):
...
-    def permalink(self):
+    def slack_permalink(self):
         if self.slack_message is not None:
             return self.slack_message.permalink
 
+    @property
+    def telegram_permalink(self) -> typing.Optional[str]:
+        from apps.telegram.models.message import TelegramMessage
+
+        main_telegram_message = self.telegram_messages.filter(
```
vadimkerr, line 412: Please check if "telegram_messages" should be added to prefetch_related (probably using django's `Prefetch` to filter out only needed messages?).

joeyorlando: hmm.. it's a tough call as I'm not yet very familiar with how `prefetch_related` works in Django. It actually looks like by adding it to `prefetch_related` it results in 1 additional SQL query?

See below from some local test results when fetching two alert groups that both have a discord message associated with them.

Without `telegram_messages` in `IncidentSerializer.PREFETCH_RELATED`:
without

With `telegram_messages` in `IncidentSerializer.PREFETCH_RELATED`:
with

joeyorlando: I'll try scripting a quick way to generate a bunch of alert groups and telegram messages. I think the difference may become evident then with a large amount of data.

matiasb: hm.. from there, it seems the prefetched data is not being used, since you still have the 2 queries to get the telegram messages (maybe because the data is not used in the serializer instance but inside a called method? the idea behind prefetch is to get all the rows you will need in one query, and then use that prefetched data instead of triggering new queries in this case you will be introducing an extra/unused query that could be expensive).

joeyorlando: @matiasb those screenshots are a bit outdated. @vadimkerr and I paired for a bit and it looks like the following results in only one additional query to `telegram_telegrammessage`: 
```python3
PREFETCH_RELATED = [
        "alerts",
        Prefetch(
            "telegram_messages",
            TelegramMessage.objects.filter(chat_id__startswith="-", message_type=TelegramMessage.ALERT_GROUP_MESSAGE),
        ),
    ]
```

engine/apps/alerts/models/alert_group.py
```diff
@@ -401,15 +401,23 @@ def is_alert_a_resolve_signal(self, alert):
...
+    def slack_permalink(self):
         if self.slack_message is not None:
             return self.slack_message.permalink
 
+    @property
+    def telegram_permalink(self) -> typing.Optional[str]:
+        main_telegram_message = self.telegram_messages.all()
+
+        if main_telegram_message:
+            return main_telegram_message[0].link
```
vadimkerr, line 413: This could potentially return a wrong telegram message if invoked directly (without any prefetching).
One of the ways to solve it would be to prefetch to a custom attribute in the Prefetch and check if this attribute is set in this method.

So the prefetch would be something like this `Prefetch("telegram_messages", ..., to_attr="prefetched_telegram_messages")`
And in the telegram_permalink property add a check for it: `if hasattr(self, "prefetched_telegram_messages"):`

But there's also a method that could be even prettier: https://stackoverflow.com/a/19651840. I have not tried this myself, but looks promising in this case.

engine/apps/alerts/models/alert_group.py
```diff
@@ -401,15 +401,23 @@ def is_alert_a_resolve_signal(self, alert):
...
 
     @property
-    def permalink(self):
+    def slack_permalink(self):
         if self.slack_message is not None:
             return self.slack_message.permalink
 
+    @property
+    def telegram_permalink(self) -> typing.Optional[str]:
+        main_telegram_message = self.telegram_messages.all()
```
vadimkerr, line 410: Probably this should be main_telegram_message**s**?

vadimkerr (changes requested):

engine/apps/alerts/models/alert_group.py
joeyorlando: @vadimkerr thanks for pointing this out 😄

I believe [this commit](https://github.com/grafana/oncall/pull/551/commits/fd2d815d5823f61f3fc95dd303340d6494660c55) should address this concern.

engine/apps/alerts/models/alert_group.py
```diff
@@ -401,15 +401,21 @@ def is_alert_a_resolve_signal(self, alert):
...
-    def permalink(self):
+    def slack_permalink(self):
         if self.slack_message is not None:
             return self.slack_message.permalink
 
+    @property
+    def telegram_permalink(self) -> typing.Optional[str]:
+        if hasattr(self, "prefetched_telegram_messages"):
+            return self.prefetched_telegram_messages[0].link if self.prefetched_telegram_messages else None
+        return None
```
vadimkerr, line 412: In case it's not prefetched I think we should make a query in this method
```suggestion
        return self.telegram_messages.filter(chat_id__startswith="-", message_type=TelegramMessage.ALERT_GROUP_MESSAGE).first()

```

vadimkerr (approved):

PR #553, opened 2022-09-22
Use create_engine_url to add prefix to previous/next links

mderynck:
Fix API responses not using path prefix in links returning for previous/next pages during pagination.


mderynck:
https://github.com/grafana/oncall-private/issues/1386

mderynck:
https://github.com/grafana/support-escalations/issues/3994

vadimkerr (approved):

PR #567, opened 2022-09-27
Fully specify ingressClassName

scubbo:
As per [here](https://kubernetes.io/docs/concepts/services-networking/ingress/#deprecated-annotation), the annotation-based approach of setting Ingress Class was deprecated in Kubernetes v1.18, which is presumably why there is a [check](https://github.com/grafana/oncall/blob/7deb6fb9206f7372be36c7f0c1e06880dbe83772/helm/oncall/templates/ingress-regular.yaml#L4) to only add the annotation for Kubernetes versions lower than this. However, there was no corresponding logic to add an explicit `ingressClassName` for later versions. This change adds that.

### DIFF
--- a/helm/oncall/templates/ingress-regular.yaml
+++ b/helm/oncall/templates/ingress-regular.yaml
@@ -25,7 +25,7 @@ metadata:
 spec:
   {{- if and .Values.ingress.className (semverCompare ">=1.18-0" .Capabilities.KubeVersion.GitVersion) }}
   ingressClassName: {{ .Values.ingress.className }}
-  {{- end}}
+  {{- end }}
   {{- if .Values.ingress.tls }}
   tls:
       {{- tpl (toYaml .Values.ingress.tls) . | nindent 4 }}
--- a/helm/oncall/templates/ingress-regular.yaml
+++ b/helm/oncall/templates/ingress-regular.yaml
@@ -23,7 +23,10 @@ metadata:
     {{- toYaml . | nindent 4 }}
   {{- end }}
 spec:
-  {{- if .Values.ingress.tls }} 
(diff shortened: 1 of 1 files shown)


scubbo:
**NOTE** that `.Capabilities.KubeVersion.GitVersion` is [deprecated as of Helm 3.x](https://github.com/helm/charts/issues/20918). The usages of `[...].GitVersion` in this file should probably be replaced with `[...].Version` (or, for full
compatibility, declare a variable like:
```
{{- if (semverCompare ">=3.x" .Capabilities.HelmVersion.Version) -}}
{{- $kube_version := .Capabilities.KubeVersion.Version -}}
{{- else -}}
{{- $kube_version := .Capabilities.KubeVersion.GitVersion -}}
{{- end -}}
```
) - but, since I'm _extremely_ new to Helm charts, I wanted to keep this commit as small as possible to minimize the chances of me misunderstanding something or making a mistake. If that's the correct change to make (or if someone can point me in the right direction), I'd be happy to make that change too.

helm/oncall/templates/ingress-regular.yaml
```diff
@@ -23,7 +23,10 @@ metadata:
     {{- toYaml . | nindent 4 }}
   {{- end }}
 spec:
-  {{- if .Values.ingress.tls }} 
+  {{- if and .Values.ingress.className (semverCompare ">=1.18-0" .Capabilities.KubeVersion.GitVersion) }}
+  ingressClassName: {{ .Values.ingress.className }}
+  {{- end}}
```
iskhakov, line 28:
```suggestion
  {{- end }}
```

iskhakov:
I agree with https://github.com/grafana/oncall/pull/567#issuecomment-1259916948, changing GitVersion can be done in the different PR

iskhakov (approved):

PR #578, opened 2022-09-29
Simplify README.md commands 

vadimkerr:
1. Rename `.env_hobby` to `.env` so the commands are more consice in README.md
2. Remove GRAFANA_USER and GRAFANA_PASSWORD from instruction


Matvey-Kuk:
.env will clash with dev environment. I suggest renaming dev's .env to .env_dev to avoid collision

vadimkerr:
> .env will clash with dev environment. I suggest renaming dev's .env to .env_dev to avoid collision

Renamed `.env` to `.env.dev` as well as `.env.example` to `.env.dev.example`. Also updated DEVELOPER.md to reflect that.

Matvey-Kuk (approved):

PR #587, opened 2022-09-30
Catch exception for when oncall user is out of sync with slack user

mderynck:
In the situation where a slack user id for the author of a message is no longer associated with a user in OnCall (User deletion in OnCall or Slack) catch the exception and display a message instead of 500 server error.


matiasb (approved):
LGTM

PR #600, opened 2022-10-03
schedules fix

teodosii:


grafana-plugin/src/pages/schedules/Schedules.tsx
```diff
@@ -94,12 +100,13 @@ class SchedulesPage extends React.Component<SchedulesPageProps, SchedulesPageSta
       }
 
       const schedules = store.scheduleStore.getSearchResult();
-      const scheduleId = schedules && schedules.find((res) => res.id === id)?.id;
-      if (scheduleId || id === 'new') {
-        this.setState({ scheduleIdToEdit: id });
-      } else {
-        openErrorNotification(`Schedule with id=${id} is not found. Please select schedule from the list.`);
-      }
+      scheduleId = schedules && schedules.find((res) => res.id === id)?.id;
```
maskin25, line 103: `const schedule = await store.scheduleStore .loadItem(id, true)` returns schedule data if no errors happened, so I suppose you don't need search
`scheduleId = schedules && schedules.find((res) => res.id === id)?.id;`

maskin25 (approved):

PR #611, opened 2022-10-04
Add tiny deployment using SQLite and Redis

### LINKED ISSUE #85: "Tiny" deployment version for those who want a small deployment footprint?
A few HN commenters raised a topic of a large deployment footprint: https://news.ycombinator.com/item?id=31740902

We deliver OnCall with redis + rabbit + mariadb + workers + web server because we focus on reliability a lot. For some users it's not a priority. We could think about some "tiny" deployment environment based on sqlite. 

- Django UWSGI web server -> SQLite
- Celery workers -> One celery worker for all queues using Celery-SQLAlchemy -> SQLite (https://docs.celeryq.dev/en/v5.2.7/getting-started/backends-and-brokers/index.html#sqlalchemy). It's marked as not stable in docs btw. Also our celery version is 4.3.
- Redis -> We could go without cache at all or with in-memory Django engine cache. Probably impacted parts:
   - Alert Group page in web.
   - Grouping performance for heavy alert storm from Alertmanager. Typical K8S-generated alert storm.
   - Google Calendar parser relies on caching a lot.
   - Anything else?
- Rabbit -> Could be removed.
- MySQL -> Could be removed.

In total we'll need uWSGI + celery. Both could be packed in the one docker container.

Also we should make users absolutely aware of the performance/reliability limitations if they choose such a deployment.

Any thoughts?

vadimkerr:
- Add a new default docker compose file `docker-compose.yml` that uses SQLite3 as a DB and Redis as a celery broker
- Rename the old `docker-compose.yml` to `docker-compose-mysql-rabbitmq.yml` to reflect that it's using MySQL as a DB and RabbitMQ as a celery broker

Should resolve #85


joeyorlando (approved):
Overall LGTM!

One small suggestion I would have would be to consider [extending docker compose services](https://docs.docker.com/compose/extends/#extending-services) in a common `.yml` file to avoid having to duplicate service configurations across the four files (things like `engine`, `oncall_db_migration`, and `grafana`). Up to you if you think it is worthwhile 😃

iskhakov:
I don't think we need to keep `docker-compose-mysql-rabbitmq.yml` at all. I propose to have `docker-compose.yaml` for hobby version as a single file, which is easy to download and understand. 

All other versions of docker compose with different components may be kept for development purposes

vadimkerr:
> I don't think we need to keep `docker-compose-mysql-rabbitmq.yml` at all. I propose to have `docker-compose.yaml` for hobby version as a single file, which is easy to download and understand.
> 
> All other versions of docker compose with different components may be kept for development purposes

The idea here is to give examples of different deployment options using different database and broker types, so I thought we could even add smth like `docker-compose-postgresql-rabbitmq.yml` at some point, to have as many deployment examples as possible so it can be used as a reference. I don't think removing `docker-compose-mysql-rabbitmq.yml` is a good idea at this point because we don't have any other sources showing which env variables to set when using MySQL.
Also I beleive that `docker-compose-mysql-rabbitmq.yml` should be still the default way for hobby deployment (since sqlite has many limitations while it doesn't simplify the README.md instructions at all). Probably we can keep different examples of docker-compose files, but somehow keeping the main README.md clean and simple. wdyt @iskhakov?

docker-compose-sqlite-redis.yml
```diff
@@ -0,0 +1,98 @@
...
+          memory: 500m
+          cpus: '0.5'
+    healthcheck:
+      test: ["CMD", "redis-cli", "ping"]
+      timeout: 5s
+      interval: 5s
+      retries: 10
+
+  grafana:
+    image: "grafana/grafana:9.0.0-beta3"
```
Konstantinov-Innokentii, line 76: Probably we can use not beta version here?

vadimkerr: I used the same version as in the existing docker-compose file for hobby deployment: https://github.com/grafana/oncall/blob/dev/docker-compose.yml#L134. Not sure about bumping the grafana version, so I guess we can do this in another PR if we wanted to?

matiasb (approved):
Some thoughts:
- I would keep a default `docker-compose.yml` file with our preferred config (ie. mysql, rabbitmq), which can be explicited in our docs
- Refactoring to avoid duplication would be nice, but I think we still want this `docker-compose.yml` to be the only file to download when setting up the hobby env, right?

iskhakov:
I still think we should have the default way to install hobby version, without learning our infrastructure components and their combinations. Also renaming the file during the download may confuse the users
```
curl -fsSL https://raw.githubusercontent.com/grafana/oncall/dev/docker-compose-mysql-rabbitmq.yml -o docker-compose.yml
```

vadimkerr:
> I still think we should have the default way to install hobby version, without learning our infrastructure components and their combinations. Also renaming the file during the download may confuse the users
> 
> ```
> curl -fsSL https://raw.githubusercontent.com/grafana/oncall/dev/docker-compose-mysql-rabbitmq.yml -o docker-compose.yml
> ```

So probably we can just rename `docker-compose-mysql-rabbitmq.yml` back to `docker-compose.yml`?
So there will be a "default" `docker-compose.yml` with mysql and rabbitmq, and `docker-compose-sqlite-redis.yml` with sqlite and redis, as well as a mention about this tiny version in `README.md`, wdyt @iskhakov?

iskhakov:
I think sqlite should be fine for hobby version unless it doesn't have critical limitations. If it does have critical limitations we should not use at all.

joeyorlando:
I'm suggesting we don't merge this PR. See Slack conversation [here](https://raintank-corp.slack.com/archives/C0229FD3CE9/p1666182402356199?thread_ts=1666181024.260149&cid=C0229FD3CE9) for more context.

PR #613, opened 2022-10-04
Improve Plugin Initialization Latency

mderynck:
For https://github.com/grafana/oncall/issues/249

Use exponential backoff for wait times between attempts when syncing during configuration or loading instead of a fixed 2 seconds.  This will result in much quicker response when loading.  This will also cover cases where you have longer latency as the later steps will wait much longer (For https://github.com/grafana/oncall/pull/529).

### DIFF
--- a/grafana-plugin/src/state/plugin.ts
+++ b/grafana-plugin/src/state/plugin.ts
@@ -31,6 +31,11 @@ export async function startPluginSync() {
   return await makeRequest('/plugin/sync', { method: 'POST' });
 }
 
+
+export const SYNC_STATUS_RETRY_LIMIT = 10;
+
+export const syncStatusDelay = retryCount => new Promise(resolve => setTimeout(resolve, 10 * 2 ** retryCount));
+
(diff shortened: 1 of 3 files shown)


grafana-plugin/src/state/plugin.ts
```diff
@@ -31,6 +31,11 @@ export async function startPluginSync() {
   return await makeRequest('/plugin/sync', { method: 'POST' });
 }
 
+
+export const SYNC_STATUS_RETRY_LIMIT = 10;
+
+export const syncStatusDelay = retryCount => new Promise(resolve => setTimeout(resolve, 10 * 2 ** retryCount));
```
teodosii, line 37: Looks good but I wonder why do we need to calculate it based on `retryCount` and instead not relying on a default number? Previously we had `2000ms`. It feels slightly odd to have a sequence of timeouts - `20, 40, 80, 160...` and so on, add to that that timeouts lower than `200-300ms` aren't really noticeable.

teodosii (approved):

PR #617, opened 2022-10-05
New schedules 2nd bunch of fixes

maskin25:


grafana-plugin/src/components/Table/Table.tsx
```diff
@@ -33,28 +33,35 @@ export interface Props<RecordType = unknown> extends TableProps<RecordType> {
...
-    expandable.expandIcon = ({ expanded, record }) => {
-      return (
-        <div className={cx('expand-icon', { [`expand-icon__expanded`]: expanded })}>
-          <ExpandIcon />
-        </div>
-      );
-    };
-  }
+  const expandable = useMemo(() => {
+    return expandableProp
```
teodosii, line 41: Just a suggestion to help with naming as I see you've renamed `expandable` to `expandableProp`. There's a naming convention you can append `fn` to functions to let the reader know it's a function, therefore you can have `expandableFn` which is the memo, and `expandable` as a prop.

grafana-plugin/src/components/TimelineMarks/TimelineMarks.tsx
```diff
@@ -17,6 +17,8 @@ const cx = cn.bind(styles);
 const TimelineMarks: FC<TimelineMarksProps> = (props) => {
   const { startMoment, debug } = props;
 
+  const currentMoment = useMemo(() => dayjs(), []);
```
teodosii, line 20: Wouldn't this cause an issue if let's say you navigated on this page at 11:59PM and then the clock changes to the next day? Shouldn't this be reflected in the `currentMoment`?

grafana-plugin/src/components/Table/Table.tsx
maskin25: fixed

grafana-plugin/src/components/TimelineMarks/TimelineMarks.tsx
maskin25: yep, but we don't care about this now, our calendar is not realtimic now, realtimic behavior should be done from all the sides

Issue #119, opened 2022-06-21
Celery workers graceful shutdown & restart


geowatson:
For now celery workers are using SIGTERM shutdown within the container itself, thus increasing Restarts value each time.

It would be better to use kubeapi for such task to ensure graceful restart for the service, e.g: [client-go example](https://github.com/kubernetes/client-go/tree/master/examples/create-update-delete-deployment)

Matvey-Kuk:
<img width="384" alt="Screenshot 2022-07-05 at 16 26 37" src="https://user-images.githubusercontent.com/3284841/177338726-1723dc14-af83-45e4-9615-6ded15f52784.png">
Fair comment from our slack community. Tons of pod restarts look weird.

Matvey-Kuk:
We added this task to the core team's backlog.

iskhakov:
@Matvey-Kuk I'm sure that the probes will restart the container it it doesn't reply to celery ping, it will work for any worker (pod). We can figure out something similar for hobby if necessary.

I don't thing we should restart the pod through kubeapi from the application as proposed in this issue.

Matvey-Kuk:
@iskhakov agree, thank you :)

PR #623, opened 2022-10-07
Allow disabling autorestarts in helm deployment

iskhakov:
This PR allows to disable to default restarts from values.yaml.
This feature increases the k8s pod restart counter, but we do it intentionally, as a precaution additional to probes.


iskhakov:
Fix for https://github.com/grafana/oncall/issues/119

Issue #452, opened 2022-09-01
One-time installation key UI/UX


Matvey-Kuk:
Proposal:
1) Get rid of one-time installation token. Make backend to allow registering the first frontend only.
2) We still need URLS for "Frontend -> OnCall Backend" and "OnCall Backend -> Grafana. Make sure we provide proper error messages and explanation what is where.
3) Draw schematic image of network connectivity and put it right on the configuration page. To clear up what URL's do we ask for.

raphael-batte:
One more case with same problem https://github.com/grafana/oncall/issues/468

iskhakov:
https://github.com/grafana/oncall/issues/122

joeyorlando:
note to self: update the frontend configuration error message to not mention the need to regenerate a new token

joeyorlando:
Closing as this was implemented in #624

PR #624, opened 2022-10-07
Get rid of installation token + add a bunch of tests

Ferril:
**What this PR does**:
Gets rid of the need to supply "one time invite token" and "grafana url" in the OSS configuration form.

https://www.loom.com/share/d865b2a260a345f7b7b0742002da5f65

This is done by requiring to supply the grafana API URL as an environment variable on the backend. Additionally/optionally, an OnCall API URL environment variable can be passed in to the frontend (this basically allows completely skipping the need to configure anything). Other stuff:
- Deduplicated a lot of the sync logic on the frontend + made error message more useful and consistent
- Split `PluginConfigPage` component into several subcomponents (making it easier to test each individual component)
- Moved `RootWithLoader` (from `plugin/GrafanaPluginRootPage`) into its own subcomponent (making it easier to test)
- Added tests for pre-existing components that were touched:
  - `PluginConfigPage` component (and its new subcomponents)
  - `state/plugin` and `state/rootBaseStore` functions
  - `apps.grafana_plugin` django app

#452


engine/apps/grafana_plugin/helpers/client.py
```diff
@@ -73,7 +76,7 @@ def __init__(self, api_url: str, api_token: str):
         super().__init__(api_url, api_token)
 
     def check_token(self) -> Tuple[Optional[Response], dict]:
-        return self.api_get("api/org")
+        return self.api_head("api/org")
```
joeyorlando, line 79: no need to do a `GET` here, `HEAD` gives us the info we need (ie. 404, 401, 403) and is faster/transmits less data

engine/apps/grafana_plugin/apps.py
```diff
@@ -0,0 +1,33 @@
...
+            Organization = apps.get_model("user_management", "Organization")
+            has_existing_org = Organization.objects.first() is not None
+
+            # only enforce the following for new setups - if no organization exists in the database
+            # and the OSS_GRAFANA_API_URL env var is not specified, exit the application
+            if has_existing_org is False and settings.SELF_HOSTED_SETTINGS["GRAFANA_API_URL"] is None:
+                logger.error(
+                    f"For OSS installations, OSS_GRAFANA_API_URL is a required environment variable. Please set it and restart the application."
+                )
+                sys.exit()
```
joeyorlando, line 33: I took a bit of an opinionated approach to this but I think we should crash the app on OSS startup if certain conditions are not met, like this one. ([`AppConfig.ready` docs](https://docs.djangoproject.com/en/4.1/ref/applications/#django.apps.AppConfig.ready))

engine/apps/grafana_plugin/views/status.py
```diff
@@ -42,7 +33,6 @@ def get(self, request: Request) -> Response:
         return Response(
             data={
                 "is_installed": is_installed,
-                "grafana_connection_ok": connected_to_grafana,
```
joeyorlando, line 45: this attribute is not consumed by the frontend. therefore I propose to get rid of it ([repo search results](https://github.com/grafana/oncall/search?q=grafana_connection_ok))

grafana-plugin/src/plugin.json
```diff
@@ -117,11 +117,7 @@
       "headers": [
         {
           "name": "X-Instance-Context",
-          "content": "{ \"grafana_token\": \"{{ .SecureJsonData.grafanaToken }}\", \"grafana_url\": \"{{ .JsonData.grafanaUrl }}\" }"
-        },
-        {
-          "name": "Authorization",
-          "content": "{{ .SecureJsonData.onCallInvitationToken }}"
+          "content": "{ \"grafana_token\": \"{{ .SecureJsonData.grafanaToken }}\" }"
```
joeyorlando, line 120: we no longer need to send the grafana URL to OnCall API, for the self-hosted install route, because this is now being configured directly in OnCall API

grafana-plugin/src/components/Avatar/Avatar.test.tsx
```diff
@@ -1,12 +1,9 @@
 import React from 'react';
 
-import { describe, expect, test } from '@jest/globals';
```
joeyorlando, line 3: this is automatically "injected" into test files, not need to import this. Additionally, `@testing-library/jest-dom` is now imported in `jest.setup.ts` so there's no need to import this in every test file.

grafana-plugin/src/containers/DefaultPageLayout/DefaultPageLayout.tsx
```diff
@@ -1,11 +1,11 @@
 import plugin from '../../../package.json'; // eslint-disable-line
 import React, { FC, useEffect, useState, useCallback } from 'react';
 
-import { AppRootProps } from '@grafana/data';
 import { getLocationSrv } from '@grafana/runtime';
 import { Alert } from '@grafana/ui';
 import cn from 'classnames/bind';
 import { observer } from 'mobx-react';
+import { AppRootProps } from 'types';
```
joeyorlando, line 8: use the new strongly typed `AppRootProps`. See `types.ts` for more details, in short:
```ts
export type OnCallPluginMetaJSONData = {
  stackId: number;
  orgId: number;
  onCallApiUrl: string;
  license: string;
};

export type AppRootProps = BaseAppRootProps<OnCallPluginMetaJSONData>;
```

grafana-plugin/src/containers/PluginConfigPage/parts/ConfigurationForm/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+      setItem(ONCALL_API_URL_LOCAL_STORAGE_KEY, onCallApiUrl);
+      onSuccessfulSetup();
+    } else {
+      setSetupErrorMsg(errorMsg);
+      setFormLoading(false);
+    }
+  }, []);
+
+  return (
+    <Form<FormProps> defaultValues={{ onCallApiUrl: getItem(ONCALL_API_URL_LOCAL_STORAGE_KEY) }} onSubmit={setupPlugin}>
```
joeyorlando, line 82: `Form` from `@grafana/ui` has a lot of nice utilities for error handling, validation, and managing input state. So we no longer need to manually handle this.

grafana-plugin/src/state/rootBaseStore.ts
```diff
@@ -192,52 +142,75 @@ export class RootBaseStore {
...
-      if (!syncStartStatus.allow_signup) {
+    } else if (!is_installed) {
+      if (!allow_signup) {
         this.signupAllowedForPlugin = false;
         return;
       }
-      const userRole = await this.getUserRole();
-      if (userRole !== 'Admin') {
+
+      if (!contextSrv.hasRole(OrgRole.Admin)) {
```
joeyorlando, line 183: the `contextSrv` class already provides a utility for this (basically provides access to the currently authenticated grafana user, and some helper methods), which doesn't require making any additional network calls.

See [here](https://github.com/grafana/grafana/blob/main/public/app/core/services/context_srv.ts#L103-L109) for more details.

grafana-plugin/src/containers/PluginConfigPage/PluginConfigPage.tsx
```diff
@@ -1,342 +1,178 @@
-import React, { useCallback, useEffect, useState } from 'react';
```
joeyorlando, line 1: the main changes in this file were:
- made error messaging more clear/consistent across `PluginConfigPage` and `GrafanaPluginRootPage` by:
  - backend now will distinctly return 400, 401, or 403, with a specific error message
  - there was a lot of duplication of sync related functionality between these two components, this has been moved to `state/plugin`
- extracted a few portions out into separate components (ex. `ConfigurationForm`) as it was easier to test them this way

grafana-plugin/src/state/rootBaseStore/index.ts
```diff
@@ -198,52 +147,75 @@ export class RootBaseStore {
...
-      if (!syncStartStatus.allow_signup) {
+    } else if (!is_installed) {
+      if (!allow_signup) {
         this.signupAllowedForPlugin = false;
         return;
       }
-      const userRole = await this.getUserRole();
-      if (userRole !== 'Admin') {
+
+      if (!contextSrv.hasRole(OrgRole.Admin)) {
```
joeyorlando, line 188: this replaces `rootBaseStore.getUserRole`

grafana-plugin/src/state/rootBaseStore/index.ts
```diff
@@ -143,50 +136,6 @@ export class RootBaseStore {
...
-  async getUserRole() {
-    const user = await getBackendSrv().get('/api/user');
-    const userRoles = await getBackendSrv().get('/api/user/orgs');
-    const userRole = userRoles.find(
-      (userRole: { name: string; orgId: number; role: string }) => userRole.orgId === user.orgId
-    );
-    return userRole.role;
-  }
-
-  async finishSync(get_sync_response: any) {
```
joeyorlando, line 155: all of this sync logic has been deduplicated/moved to `state/plugin`

engine/apps/grafana_plugin/urls.py
```diff
@@ -11,9 +11,9 @@
 app_name = "grafana-plugin"
 
 urlpatterns = [
-    re_path(r"self-hosted/install/?", SelfHostedInstallView().as_view()),
-    re_path(r"status/?", StatusView().as_view()),
-    re_path(r"install/?", InstallView().as_view()),
-    re_path(r"sync_organization/?", SyncOrganizationView().as_view()),
-    re_path(r"sync/?", PluginSyncView().as_view()),
+    re_path(r"self-hosted/install/?", SelfHostedInstallView().as_view(), name="self-hosted-install"),
```
Matvey-Kuk, line 14: ❤️

engine/apps/grafana_plugin/views/self_hosted_install.py
```diff
@@ -1,43 +1,48 @@
...
-
-    def post(self, request: Request) -> Response:
-        token_string = get_authorization_header(request).decode()
+    def post(self, _request: Request) -> Response:
+        """
+        We've already validated that settings.OSS_GRAFANA_API_URL is set (in apps.grafana_plugin.GrafanaPluginConfig)
+        The user is now trying to finish plugin installation. We'll take the Grafana API url that they specified +
+        the token that we are provided and first verify them. If all is good, upsert the organization in the database,
+        and provision the plugin.
+        """
```
Matvey-Kuk, line 20: ❤️

engine/apps/grafana_plugin/views/status.py
Matvey-Kuk: @maskin25

Matvey-Kuk (approved):
I checked mostly backend part, LGTM so far!

helm/oncall/templates/_env.tpl
```diff
@@ -23,6 +23,8 @@
   value: "1024"
 - name: BROKER_TYPE
   value: {{ .Values.broker.type | default "rabbitmq" }}
+- name: OSS_GRAFANA_API_URL
+  value: "TODO:"
```
joeyorlando, line 27: @iskhakov had some questions around this

CHANGELOG.md
```diff
@@ -1,4 +1,22 @@
...
+## [Unreleased]
+
+### Added
+
+### Changed
+
+- For OSS installations of OnCall, initial configuration is now simplified. When running for local development, you no longer need to configure the plugin via the UI. This is achieved through passing one environment variable to both the backend & frontend containers, both of which have been preconfigured for you in `docker-compose-developer.yml`.
+  - The Grafana API URL **must be** passed as an environment variable, `GRAFANA_API_URL`, to the OnCall backend (and can be configured by updating this env var in your `./dev/.env.dev` file)
+  - The OnCall API URL can optionally be passed as an environment variable, `ONCALL_API_URL`, to the OnCall UI. If the environment variable is found, the plugin will "auto-configure", otherwise you will be shown a simple configuration form to provide this info.
+- For Helm installations TODO:
```
joeyorlando, line 17: to finish

helm/oncall/values.yaml
```diff
@@ -269,6 +271,9 @@ grafana:
   plugins:
     - grafana-oncall-app
 
+externalGrafana:
+  url:
```
joeyorlando, line 275: this is the main change in this file, the rest is `yaml` IDE autoformatter changes

helm/oncall/templates/_env.tpl
```diff
@@ -111,6 +113,15 @@
...
 {{- end -}}
 
+{{- define "snippet.grafana.url" -}}
+{{- if .Values.externalGrafana.url -}}
+{{- .Values.externalGrafana.url | quote }}
+{{- else if .Values.grafana.enabled -}}
+http://{{ include "oncall.grafana.fullname" . }}
+{{- else -}}
+{{- required "externalGrafana.url is required when not grafana.enabled" .Values.externalGrafana.url | quote }}
+{{- end -}}
```
joeyorlando, line 123: @iskhakov would be good to get your 👀 on these helm related changes

docker-compose-developer.yml
```diff
@@ -29,6 +30,8 @@ services:
       context: ./grafana-plugin
       dockerfile: Dockerfile.dev
       labels: *oncall-labels
+    environment:
+      ONCALL_API_URL: http://host.docker.internal:8080
```
iskhakov, line 34: just checking if it should be necessarily forwarded through the host network (because of the browser) or can be just http://engine:8080?

engine/apps/grafana_plugin/views/self_hosted_install.py
```diff
@@ -1,43 +1,48 @@
...
+                "error"
+            ] = f"You are not authorized to communicate with the specified Grafana API - {grafana_url}"
+            return Response(data=provisioning_info, status=status.HTTP_400_BAD_REQUEST)
 
         organization = Organization.objects.filter(stack_id=stack_id, org_id=org_id).first()
         if organization:
             organization.revoke_plugin()
-            organization.grafana_url = self.instance_context["grafana_url"]
-            organization.api_token = self.instance_context["grafana_token"]
+            organization.grafana_url = grafana_url
```
iskhakov, line 44: if we take the grafana url from settings.SELF_HOSTED_SETTINGS, which http://grafana:3000 by default (https://github.com/grafana/oncall/pull/624/files#diff-f5d10b03472abe3719098ae8a8855468e92524ebe790c39a34d2c632f3f0486dR23), the rendered urls to grafana will be broken. If self.instance_context["grafana_url"] returns public url we should use it, otherwise we should specify public url in settings.SELF_HOSTED_SETTINGS

iskhakov:
Nice done! We display the warning if there is the version mismatch between the plugin and backend. However if the plugin is not connected yet to the backend user will never be able to connect. Can we add the note for the user to check the versions during the installation

helm/oncall/templates/NOTES.txt
```diff
@@ -34,11 +34,6 @@
...
   🔗 Connect Grafana OnCall Plugin to Grafana OnCall backend:
 
-     Issue the one-time token to connect Grafana OnCall backend and Grafana OnCall plugin by running these commands:
-
-        export POD_NAME=$(kubectl get pods --namespace {{ .Release.Namespace }} -l "app.kubernetes.io/name={{ include "oncall.name" . }},app.kubernetes.io/instance={{ .Release.Name }},app.kubernetes.io/component=engine" -o jsonpath="{.items[0].metadata.name}")
-        kubectl exec -it $POD_NAME --namespace {{ .Release.Namespace }} -- bash -c "python manage.py issue_invite_for_the_frontend --override"
-
      Fill the Grafana OnCall Backend URL:
 
           http://{{ include "oncall.engine.fullname" . }}:8080
```
iskhakov, line 39: TODO: remove the lines above as well

helm/oncall/values.yaml
```diff
@@ -21,7 +21,8 @@ service:
 # Engine pods configuration
 engine:
   replicaCount: 1
-  resources: {}
+  resources:
+    {}
```
iskhakov, line 25: this is the common practice for helm charts, let's keep on the same line for readability

grafana-plugin/src/plugin/GrafanaPluginRootPage.tsx
```diff
@@ -45,54 +45,44 @@ export const GrafanaPluginRootPage = (props: AppRootProps) => (
...
 
   if (store.appLoading) {
-    let text = 'Initializing plugin...';
-
-    if (!store.pluginIsInitialized) {
-      text = '🚫 Plugin has not been initialized';
-    } else if (!store.correctProvisioningForInstallation) {
-      text = '🚫 Plugin could not be initialized due to provisioning error';
-    } else if (!store.correctRoleForInstallation) {
-      text = '🚫 Admin must sign on to setup OnCall before a Viewer can use it';
```
joeyorlando, line 63: all of the plugin initialization error messages have been moved to `store.initializationError`. If there was an error it is non-null, otherwise null. See `rootBaseStore.setupPlugin` for more details.

helm/oncall/values.yaml
joeyorlando: sounds good 👍 (not sure why my IDE autoformatted them, will revert this change)

engine/apps/grafana_plugin/views/self_hosted_install.py
```diff
@@ -1,43 +1,48 @@
...
-                    "keys": [],
-                }
-            },
-        )[0]
-        self_hosted_settings.json_value["keys"].remove(token)
-        self_hosted_settings.save(update_fields=["json_value"])
-
-    def post(self, request: Request) -> Response:
-        token_string = get_authorization_header(request).decode()
+    def post(self, _request: Request) -> Response:
```
mderynck, line 14: As permission_classes is removed should we check settings license here and return 403 if it is cloud configured?

engine/apps/grafana_plugin/views/status.py
mderynck: It was going to provide status in cloud for if the URL from plugin provisioning wasn't working but since we aren't using it might as well remove it.

engine/apps/grafana_plugin/views/sync.py
```diff
@@ -49,19 +51,20 @@ def post(self, request: Request) -> Response:
             },
         )
 
-    def get(self, request: Request) -> Response:
+    def get(self, _request: Request) -> Response:
```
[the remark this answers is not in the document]
mderynck: This handler is for the front-end to check when the celery task has completed synchronizing users and teams.

engine/apps/grafana_plugin/views/sync.py
```diff
@@ -49,19 +51,20 @@ def post(self, request: Request) -> Response:
...
         stack_id = self.instance_context["stack_id"]
         org_id = self.instance_context["org_id"]
-
         token_ok = False
+
         try:
             organization = Organization.objects.get(stack_id=stack_id, org_id=org_id)
-            if organization.api_token_status == Organization.API_TOKEN_STATUS_PENDING:
```
mderynck, line 59: What happened to this check? In the post handler above `plugin_sync_organization_async.apply_async` is started and will be in PENDING state until complete. How do we know users and teams are populated?

engine/apps/grafana_plugin/views/sync.py
```diff
@@ -49,19 +51,20 @@ def post(self, request: Request) -> Response:
...
-                return Response(status=status.HTTP_202_ACCEPTED)
```
mderynck, line 60: Looking at this I see why it was collapsed but I wonder if we should instead handle all 3 cases both here and the plugin.  The plugin should continue polling on 202 instead of checking if the token_ok field is present.  Here we should return an error code here if token_status == FAILED.  As it stands this get request will give us the same answer for PENDING and FAILED.

grafana-plugin/src/state/rootBaseStore/index.ts
```diff
@@ -57,25 +50,7 @@ export class RootBaseStore {
...
-
-  @observable
-  initializationError = '';
-
-  @observable
-  retrySync = false;
-
-  @observable
-  isUserAnonymous = false;
+  initializationError = null;
```
joeyorlando, line 53: this has been simplified. all of the above attributes that were removed have been collapsed into `initializationError`, see `grafana-plugin/src/plugin/PluginSetup/index.tsx` for more details

grafana-plugin/src/state/plugin/index.ts
```diff
@@ -0,0 +1,273 @@
+import { getBackendSrv } from '@grafana/runtime';
```
joeyorlando, line 1: there was a lot of duplicated logic/function calls between `rootBaseStore` and `src/PluginConfigPage/PluginConfigPage.tsx`. This has all been centralized/tested here. I also decided to make these methods as `static` methods on an ES6 class (this made mocking/testing intra-module calls way easier; alternative was dependency-injection or some other dirty hacks)

grafana-plugin/src/plugin/PluginSetup/index.tsx
```diff
@@ -0,0 +1,58 @@
+import React, { FC, PropsWithChildren, useCallback, useEffect } from 'react';
```
joeyorlando, line 1: extracted this from `src/plugin/PluginConfigPage/PluginConfigPage.tsx`, this made testing this component much easier.

joeyorlando:
> we display the warning if there is the version mismatch between the plugin and backend. However if the plugin is not connected yet to the backend user will never be able to connect. Can we add the note for the user to check the versions during the installation

@iskhakov can we chat more about this next week? I'm not sure I follow 100%

engine/apps/grafana_plugin/views/self_hosted_install.py
joeyorlando: good idea! added this condition + test case

engine/apps/grafana_plugin/views/sync.py
```diff
@@ -49,19 +51,20 @@ def post(self, request: Request) -> Response:
...
-            elif organization.api_token_status == Organization.API_TOKEN_STATUS_OK:
-                token_ok = True
+            token_ok = organization.api_token_status == Organization.API_TOKEN_STATUS_OK
         except Organization.DoesNotExist:
             logger.info(f"Organization for stack {stack_id} org {org_id} was not found")
+        except Exception as e:
```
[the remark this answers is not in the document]
joeyorlando: good catch, I'll remove this `except` branch

README.md
```diff
@@ -44,23 +44,13 @@ SECRET_KEY=my_random_secret_must_be_more_than_32_characters_long" > .env
...
-4. Issue one-time invite token:
-
-```bash
-docker-compose run engine python manage.py issue_invite_for_the_frontend --override
-```
-
-**Note**: if you remove the plugin configuration and reconfigure it, you will need to generate a new one-time invite token for your new configuration.
-
-5. Go to [OnCall Plugin Configuration](http://localhost:3000/plugins/grafana-oncall-app), using log in credentials as defined above: `admin`/`admin` (or find OnCall plugin in configuration->plugins) and connect OnCall _plugin_ with OnCall _backend_:
+4. Go to [OnCall Plugin Configuration](http://localhost:3000/plugins/grafana-oncall-app), using log in credentials as defined above: `admin`/`admin` (or find OnCall plugin in configuration->plugins) and connect OnCall _plugin_ with OnCall _backend_:
```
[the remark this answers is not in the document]
joeyorlando: ohhh I see. so we would provide a `yaml` file to the `grafana` container which would already container this info in the `jsonData` object?

helm/oncall/templates/NOTES.txt
joeyorlando: you mean to remove the "Fill the Grafana URL:" section, correct? Because I think you would still need to configure the "Grafana OnCall Backend URL" input as there would be no way to provide that in env var in the UI build for helm releases. Let's chat more about this on Monday.

docker-compose-developer.yml
joeyorlando: I tested this locally. It complains because our container names in `docker-compose-developer.yml` use underscores instead of dashes (eg. `oncall_engine`). This isn't RFC 1738 compliant and the plugin-proxy complains. I propose we follow up in a future PR and make this enhancement.

engine/apps/grafana_plugin/views/self_hosted_install.py
joeyorlando: I think I understand what you mean. The only way to get this to work then would to be something like this:
```diff
x-env-vars: &oncall-env-vars
  BROKER_TYPE: ${BROKER_TYPE}
-  GRAFANA_API_URL: http://grafana:3000
+  GRAFANA_API_URL: http://localhost:3000
```
this would generate public URLs that point to the actual local grafana, but the problem is that the OnCall containers cannot communicate w/ the `grafana` container @ `localhost:3000`. Maybe there is some docker networking magic that I can add to the oncall containers to make `localhost:3000` point to `grafana:3000`?

engine/apps/grafana_plugin/views/sync.py
joeyorlando: fair point, I'll add this condition back 👍 

Currently though, the UX is the same, whether or not this condition exists. I think the only benefit of updating the UI to poll based on HTTP 202, is that if the sync fails (and `token_ok` is False) then the sync polling would exit earlier, rather than looping for the hardcoded amount of time. Am I stating that correctly?

if you think it's worthwhile implementing, I propose we create a separate issue and address it there.

engine/apps/grafana_plugin/views/self_hosted_install.py
joeyorlando: so one solution that works that I just tested is to modify `/etc/hosts` to add the following:
```
127.0.0.1 grafana
```
I could modify `make start` to idempotently add this record to that file, thoughts?

README.md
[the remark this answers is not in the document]
iskhakov: yep, you can add the file to grafana e.g.
```
cat << EOF > /plugins/grafana-oncall-self-hosted.yml
        apiVersion: 1
        apps:
          - type: grafana-oncall-app
            jsonData:
              onCallApiUrl: http://host.docker.internal:8000
              grafanaUrl: http://grafana:3000
            secureJsonData:
              onCallInvitationToken: $$TOKEN
        EOF
```

engine/apps/grafana_plugin/views/self_hosted_install.py
joeyorlando: @iskhakov see [this commit](https://github.com/grafana/oncall/pull/624/commits/e39d116444f22bed12abb8dc35e9308de1be5a92), lmk your thoughts. I know it's not ideal having to modify `/etc/hosts`, but the [`hostess` tool](https://github.com/cbednarski/hostess) I introduced in the `Makefile` makes it quite simple and cross-platform compatible.

Alternative solution, like we discussed, is to use `http://localhost:3000` as the value for `GRAFANA_API_URL` and _somehow_ modify the `oncall_engine`/`oncall_celery` containers such that within those containers `http://localhost:3000` points to `http://grafana:3000`. I have yet to find a solution for this.

helm/oncall/README.md
```diff
@@ -284,3 +275,9 @@ redis-data-release-oncall-redis-replicas-1 redis-data-release-oncall-redis-repli
 ```bash
 kubectl delete secrets certificate-tls release-oncall-cert-manager-webhook-ca release-oncall-ingress-nginx-admission
 ```
+
+## Troubleshooting
+
+### Issues during initial configuration
+
+In the event that you run into issues during initial configuration, it is possible that mismatching versions between your OnCall backend and UI is the culprit. Ensure that the versions match, and if not consider updating your `helm` deployment.
```
joeyorlando, line 283: @iskhakov added a small note as we discussed

Makefile
```diff
@@ -62,13 +62,30 @@ define run_engine_docker_command
...
+		echo "installing hostess"; \
+		git clone https://github.com/cbednarski/hostess "${HOME}/hostess"; \
+		cd "${HOME}/hostess"; \
+		make install; \
+	fi
+
+	@if ! hostess has grafana; then \
+		sudo hostess add grafana 127.0.0.1; \
+	fi
+
```
joeyorlando, line 88: this is a notable change to point out in how `make start` will operate. It will now check if the user has `hostess` installed, if not install it, and then idempotently add a record to `/etc/hosts` so that http://grafana:3000 can be accessed from your browser

grafana-plugin/package.json
```diff
@@ -101,8 +104,6 @@
     "node": ">=14"
   },
   "dependencies": {
-    "@types/query-string": "^6.3.0",
-    "@types/react-transition-group": "^4.4.5",
```
joeyorlando, line 105: `@types/**` packages should be listed under `devDependencies`

joeyorlando:
**TODO**
- [x] handle error msg case where user puts in a completely bogus URL that causes a CORS error (ex. https://grafana.com)
- [x] address PR comments
- [x] figure out `host.docker.internal` vs `oncall-engine` hostname
- [x] finish updating `CHANGELOG.md`
- [x] finish updating `helm` documetation
- [x] handle page reload thingy for initial installation
- [x] final round of local testing
- [x] finish tests for `state/plugin` and `state/rootBaseStore`
- [x] add env vars to helm deployment
- [x] Tests updated
- [x] Documentation added

PR #625, opened 2022-10-07
fix: Fix postgresql engine initial migrations

sergeyshevch:
Followup from #262. 
Migration 0005 will fail if the database engine is PostgreSQL. I faced such an issue in my initial migrations setup.

In my setup, I just fixed the database manually but it shouldn't block users from migrations


sergeyshevch:
@vadimkerr I guess you should look on it

vadimkerr:
Hey @sergeyshevch, thank you for the PR! I will take a look at it today.

engine/apps/alerts/migrations/0005_alertgroup_cached_render_for_web.py
```diff
@@ -12,7 +12,7 @@ class AddFieldIfNotExists(migrations.AddField):
     def database_forwards(self, app_label, schema_editor, from_state, to_state):
         try:
             super().database_forwards(app_label, schema_editor, from_state, to_state)
-        except OperationalError:
+        except (OperationalError, ProgrammingError):
```
vadimkerr, line 15: Could you please also add a comment, so it's clear that:
```
# duplicate coulumn error raises OperationalError for MySQL and SQLite, and ProgrammingError for PostgreSQL
```

vadimkerr (changes requested):
I reproduced the issue you described, definetely a bug for setups using PostgreSQL as the database. The fix you prepared should generally solve the issue, but after applying the patch I'm getting the following error when running migrations:
```
django.db.utils.InternalError: current transaction is aborted, commands ignored until end of transaction block
```
Seems like we need to add `atomic = False` for this migration so these two operations in the migration are independent and catching `ProgrammingError` is enough. 

So I propose adding `atomic = False` as a class attribute for this particular `Migration`.

sergeyshevch:
Ok. I will fix it on Monday

vadimkerr:
Closing this as already fixed by https://github.com/grafana/oncall/pull/660

PR #630, opened 2022-10-07
Add simple Makefile

matiasb:
Setting up dev environment using Makefile:

```
# setup virtualenv and deps
$ make bootstrap
# start docker services (db, rabbit, etc)
$ make docker-services-start
# setup db (re-run to update when there are pending migrations)
$ make migrate
```

Running the services:
```
# run the oncall engine
$ make run
# run celery workers
$ make start-celery
# run celery beat
$ make start-celery-beat
# build and watch for plugin changes
$ make watch-plugin
```

Other available targets:
```
$ make dbshell
$ make shell
$ make test
$ make manage ARGS="<django-command> [args]"
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,72 @@
+include .env.dev
+
+ENV = $(CURDIR)/venv
+CELERY = $(ENV)/bin/celery
+PRECOMMIT = $(ENV)/bin/pre-commit
+PIP = $(ENV)/bin/pip
+PYTHON3 = $(ENV)/bin/python3
+PYTEST = $(ENV)/bin/pytest
+
+DOCKER_FILE ?= docker-compose-developer.yml
+
+define setup_engine_env
+	export `grep -v '^#' .env | xargs -0` && cd engine
+endef
+
+$(ENV):
+	python3.9 -m venv venv
+
+bootstrap: $(ENV)
+	$(PIP) install -U pip wheel
+	cp .env.dev.example .env.dev
+	cd engine && $(PIP) install -r requirements.txt
+	@touch $@
+
+migrate:
+	$(setup_engine_env) && $(PYTHON3) manage.py migrate
+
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -0,0 +1,72 @@
...
+	until $$(nc -z -v -w30 localhost 3306); do sleep 1; done;
+
+docker-services-restart:
+	docker-compose -f $(DOCKER_FILE) restart
+
+docker-services-stop:
+	docker-compose -f $(DOCKER_FILE) stop
+
+watch-plugin:
+	cd grafana-plugin && yarn install && yarn && yarn watch
```
vadimkerr, line 70: If I remember correctly `yarn` and `yarn install` are the same command

Makefile
```diff
@@ -0,0 +1,72 @@
+include .env.dev
+
+ENV = $(CURDIR)/venv
```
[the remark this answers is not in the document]
matiasb: Maybe we can make it (optionally) configurable via an env var? Then you could set it in your `.env.dev` file (like I do for `DOCKER_FILE`)

Makefile
matiasb: So you mean I could just do `yarn` here? (fwiw, these are the steps mentioned in the [`DEVELOPER.md`](https://github.com/grafana/oncall/blob/dev/DEVELOPER.md#frontend-setup) docs)

vadimkerr: Seems like we can just do `yarn`: https://stackoverflow.com/a/60729594, probably there's a mistake in DEVELOPER.md. Not a big deal though, just nitpicking 😄

Makefile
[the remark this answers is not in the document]
vadimkerr: Yes, this makes sense

vadimkerr (approved):

PR #639, opened 2022-10-11
Allow non-JSON custom webhook templates, provided they render as valid JSON

### LINKED ISSUE #638: Post Json Data via webhooks
Hello I define a webhooks for rocketchat and i want to post data to.
bet there is an extra validation in backend code that prevent me to send data
the problem is in this file:
engine/apps/api/serializers/custom_button.py
in validate_data method

this is my data that i want to send in post request:
```
{
   "receiver":"rocketchat-default",
   "status":"firing",
   "alerts":[
      {
         "status":"firing",
         "labels": {{ alert_payload.labels }},
         "annotations": {"description": "This alert was sent by user for the demonstration purposes"},
         "startsAt":"{{ alert_payload.startsAt }}",
         "endsAt":"{{ alert_payload.endsAt }}",
         "generatorURL":"{{ alert_payload.generatorURL }}"
      }
   ],
   "commonAnnotations":{
      "description":"{{ alert_payload.annotations.description }}",
      "summary":"{{ alert_group_id }}"
   },
   "externalURL":"http://vmalertmanager-darkube-alertmanager-0:9093"
}
```

I want to send a json dict for labels in alert body but i cant

```
 "labels": {{ alert_payload.labels }},

```

if i comment json.loads(data) in mentioned method the problem has been solved

sd2k:
Previously, both the provided template _and_ the rendered template had
to be valid JSON in order for validation to pass. This was unnecessarily
restrictive: really, only the rendered template needs to be valid JSON.
It also disallowed using templates such as:

    {
      "labels": {{ alert_payload.labels | tojson }}
    }

even though this would be valid JSON after rendering.

This commit relaxes the validation of custom webhook templates so that
they don't need to be valid JSON, provided that the rendered template
_is_ valid JSON. This is checked using a dummy dictionary of render
params, which use a constant string for the `alert_group_id` field
and a `defaultdict(dict)` for the `alert_payload` field. This should
permit templates like the one above, but still deny templates such as

    {
      "labels": {{ alert_payload.labels }}
    }

which would otherwise fail later if `labels` is not valid JSON.

This should resolve #638.

### DIFF
--- a/engine/apps/api/serializers/custom_button.py
+++ b/engine/apps/api/serializers/custom_button.py
@@ -69,7 +69,7 @@ class CustomButtonSerializer(serializers.ModelSerializer):
                     # If we instead used a `defaultdict(dict)` or `defaultdict(lambda: 1)` we
                     # would accidentally accept templates such as `{"name": {{ alert_payload.name }}}`
                     # which would then fail at the true render time due to the
-                    "alert_payload": defaultdict(lambda: ""),
+                    "alert_payload": defaultdict(str),
                     "alert_group_id": "abcd",
                 }
             )
--- a/engine/apps/api/serializers/custom_button.py
+++ b/engine/apps/api/serializers/custom_button.py
@@ -69,6 +69,8 @@ class CustomButtonSerializer(serializers.ModelSerializer):
                     # If we instead used a `defaultdict(dict)` or `defaultdict(lambda: 1)` we
                     # would accidentally accept templates such as `{"name": {{ alert_payload.name }}}`
                     # which would then fail at the true render time due to the
+                    # lack of explicit quotes around the template variable; this would render
+                    # as `{"name": some_alert_name}` which is not valid JSON.
                     "alert_payload": defaultdict(str),
                     "alert_group_id": "abcd",
                 }
--- a/engine/apps/api/serializers/custom_button.py
+++ b/engine/apps/api/serializers/custom_button.py
@@ -1,4 +1,5 @@
 import json
+from collections import defaultdict
 
 from django.core.validators import URLValidator, ValidationError
 from jinja2 import Template, TemplateError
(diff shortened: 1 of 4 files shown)


engine/apps/api/serializers/custom_button.py
```diff
@@ -51,15 +52,31 @@ def validate_data(self, data):
...
+                    # We don't know what the actual payload will be, so we use a defaultdict
+                    # so that attribute access within a template will never fail
+                    # (provided it's only one level deep - we won't accept templates that attempt
+                    # to do nested attribute access).
+                    # Every attribute access should return a string to ensure that users are
+                    # correctly using `tojson` or wrapping fields in strings.
+                    # If we instead used a `defaultdict(dict)` or `defaultdict(lambda: 1)` we
+                    # would accidentally accept templates such as `{"name": {{ alert_payload.name }}}`
+                    # which would then fail at the true render time due to the
+                    "alert_payload": defaultdict(lambda: ""),
```
vadimkerr, line 72: Just curious if there's some particular reason to use `defaultdict(lambda: "")` instead of `defaultdict(str)` here?

engine/apps/api/serializers/custom_button.py
```diff
@@ -51,15 +52,31 @@ def validate_data(self, data):
...
+                    # Validate that the template can be rendered with a JSON-ish alert payload.
+                    # We don't know what the actual payload will be, so we use a defaultdict
+                    # so that attribute access within a template will never fail
+                    # (provided it's only one level deep - we won't accept templates that attempt
+                    # to do nested attribute access).
+                    # Every attribute access should return a string to ensure that users are
+                    # correctly using `tojson` or wrapping fields in strings.
+                    # If we instead used a `defaultdict(dict)` or `defaultdict(lambda: 1)` we
+                    # would accidentally accept templates such as `{"name": {{ alert_payload.name }}}`
+                    # which would then fail at the true render time due to the
```
vadimkerr, line 71: I think something's missing at the end of the comment line after `Due to the`?

vadimkerr (changes requested):
Thank you for the contribution! 
Very nice PR overall, can you please also update [`apps.public_api.serializers.action.ActionCreateSerializer.validate_data`](https://github.com/grafana/oncall/blob/allow-non-json-webhook-templates/engine/apps/public_api/serializers/action.py#L53) so the public API is consistent with the changes you made to the internal API?

engine/apps/api/serializers/custom_button.py
sd2k: Oh, nope, I think I had a `lambda: somethingelse` halfway through and just replaced the `somethingelse` but yours is much nicer. Thanks! Fixed in eabc402.

engine/apps/api/serializers/custom_button.py
sd2k: Good spot, fixed in 1aeee93.

vadimkerr (approved):

sd2k:
> can you please also update [apps.public_api.serializers.action.ActionCreateSerializer.validate_data](https://github.com/grafana/oncall/blob/allow-non-json-webhook-templates/engine/apps/public_api/serializers/action.py#L53) so the public API is consistent with the changes you made to the internal API?

I've done this in b569c32, but I haven't yet added tests for the public API. I'll look into that next.

Matvey-Kuk (approved):
Great PR! @vadimkerr please merge if you feel comfortable.

PR #646, opened 2022-10-12
Use markdown formatting for web based resolution notes

sd2k:
We would like to include some more sophisticated content in our
resolution notes (specifically URLs), so need the resolution notes to be
rendered as Markdown for this.

This PR changes _all_ resolution note text blocks to be markdown, but
perhaps it would be better to allow the caller to specify that they
would prefer markdown to plain text (this would also allow backwards
compatibility). Let me know if you'd prefer that and I can try and
figure it out!


vadimkerr (approved):
Look good to me, @Konstantinov-Innokentii can you please take a look?

iskhakov (approved):

PR #661, opened 2022-10-18
Helm chart: add support for PostgreSQL

DavidSpek:
**What this PR does**:
This PR adds support for using PostgreSQL instead of MySQL (MariaDB) too the Helm chart. The chart works as it is, but there are some style choices that could be changed if those are wanted. For example, rename the existing mysql helpers to generic `db` helpers and avoid duplication.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added
- [ ] `CHANGELOG.md` updated


helm/oncall/values.yaml
```diff
@@ -91,13 +91,13 @@ env: []
...
-    - hosts:
-        - "{{ .Values.base_url }}"
-      secretName: certificate-tls
+  annotations: {}
+    # kubernetes.io/ingress.class: "nginx"
+    # cert-manager.io/issuer: "letsencrypt-prod"
+  tls: []
+    # - hosts:
+    #     - "{{ .Values.base_url }}"
+    #   secretName: certificate-tls
```
DavidSpek, line 100: These were disabled since it was not possible to set the `className` with the `kubernetes.io/ingress.class: "nginx"` annotation being present.

DavidSpek:
@vadimkerr maybe you could have a quick look at this.

DavidSpek:
@iskhakov could you also have a look at this PR.

helm/oncall/values.yaml
iskhakov: Removing this lines could break the single click installation, 
how do you think if we add the new ingress definition called `ingress-custom`, which would not have these annotations

iskhakov (commented):
Thank you for the contribution! Finally we'll have the Postgres support.
I left some minor comments, also could you please add the Postgres block to the readme. Please let me know if any assistance needed.

helm/oncall/templates/_env.tpl
```diff
@@ -136,6 +136,74 @@
 {{- end -}}
 {{- end -}}
 
+{{- define "snippet.postgresql.env" -}}
+- name: DATABASE_TYPE
+  value: "postgresql"
```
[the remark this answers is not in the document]
DavidSpek: Sure, sounds good.

DavidSpek:
@iskhakov I will remove the changes to the ingress in the `values.yaml`. But note that setting the `kubernetes.io/ingress.class` annotation breaks the `ingress.className` so it cannot be used. This is because you can't use the annotation together with the `className` and there is no way to remove the annotation for people using this chart as a dependency.

DavidSpek:
@iskhakov this should be ready for merging.

iskhakov:
Thank you for the contribution, merging

iskhakov (approved):

Issue #518, opened 2022-09-09
[Helm] Env vars are not merged


th30nlyw4y:
Hello!
I've faced the problem with merging env vars, since they're defined and rendered as a list and helm doesn't support lists merge (and most likely will never support https://github.com/helm/helm/issues/3486). 

E.g. in following config:
### values.yaml
```
  env:
    - name: GRAFANA_CLOUD_NOTIFICATIONS_ENABLED
      value: False
    - name: SEND_ANONYMOUS_USAGE_STATS
      value: False
    - name: FEATURE_TELEGRAM_INTEGRATION_ENABLED
      value: False
```
### secret.values.yaml
```
  env:
    - name: SLACK_CLIENT_OAUTH_ID
      value: "***"
    - name: SLACK_CLIENT_OAUTH_SECRET
      value: "***"
    - name: SLACK_INSTALL_RETURN_REDIRECT_HOST
      value: "***"
    - name: SLACK_SIGNING_SECRET
      value: "***"
```
After deploying using `helm upgrade -i --namespace "${NAMESPACE}" "${RELEASE}" -f values.yaml -f secret.values.yaml .` we only have secret values in deployment. I think that it's better to use smth like it is proposed [here](https://github.com/helm/helm/issues/3486#issuecomment-700365277) instead of current template

Matvey-Kuk:
@iskhakov could you please check this one?

th30nlyw4y:
I may prepare PR as well :)

Matvey-Kuk:
@th30nlyw4y it would be nice!

th30nlyw4y:
I'm sorry, been busy last weeks. Going to prepare some fix in a near future

th30nlyw4y:
Hi!
PR is prepared, waiting for discussion and approval :)

th30nlyw4y:
Hi!
Any news here ?

PR #674, opened 2022-10-19
Add ability to set more env vars via config & change env template & f…

th30nlyw4y:
**What this PR does**:
- Adds ability to provide Twilio and two more Slack env values through helm values
- Adds ability to specify env vars by array (for backwards compatibility) or by map
- Fixes spacing of env vars (in engine & celery deployment -- containers + initContainers):
    before:
    ```yaml
    env:
                - name: BASE_URL
                   value: https://example.com
    ```
    after:
    ```yaml
    env:
        - name: BASE_URL
           value: https://example.com
    ```
**Which issue(s) this PR fixes**: #518


th30nlyw4y:
Done rebasing, should be fine now

th30nlyw4y:
Ok, let's add it. I see no problems here

th30nlyw4y:
HI, @iskhakov !
Sorry for bothering you, but could you, please, check this PR ?
I think these improvements would really simplify env vars configuration

iskhakov (approved):
Looks good, thank you for the contribution! 

One small note, these values are going to be applied only to the new installations, they will not work for the "upgraded" installations, because live settings (same variables stored in the database and available from web interface on `Env Variables` page) are already populated. We need to change that logic and prefer env variables if the live settings are set to default values.

Approving because it doesn't break anything

helm/oncall/values.yaml
```diff
@@ -96,13 +98,27 @@ oncall:
     password: ~
     tls: ~
     fromEmail: ~
+  twilio:
+    # Twilio account SID/username to allow OnCall to send SMSes and make phone calls
+    accountSid: ~
+    # Twilio password to allow OnCall to send SMSes and make calls
+    authToken: ~
+    # Number from which you will receive calls and SMS
+    phoneNumber: ~
```
iskhakov, line 107: Please specify that this _value must be in quotes_ otherwise it has int type and rendered as ` 1.234567e+10`

iskhakov (changes requested):

helm/oncall/values.yaml
th30nlyw4y: Sorry, forgot about that thing! (We've faced the same issue :))

iskhakov (approved):

th30nlyw4y:
@iskhakov sorry for bothering you again (i hope that's the last time :)), but should i wait while someone merges this or do smth else by myself ? As i can see, a can't merge it even after all checks (and i'm not sure that i should be able to do that)

th30nlyw4y:
Hi, @Matvey-Kuk !
Maybe you could help here ?)

iskhakov:
Thank you for the contribution. Merging and releasing shortly

PR #675, opened 2022-10-19
Rework schedule escalation chains/policies count as a subquery

matiasb:


engine/apps/api/serializers/schedule_base.py
```diff
@@ -76,7 +76,8 @@ def get_on_call_now(self, obj):
     def get_number_of_escalation_chains(self, obj):
         # num_escalation_chains param added in queryset via annotate. Check ScheduleView.get_queryset
         # return 0 for just created schedules
-        return getattr(obj, "num_escalation_chains", 0)
+        num = getattr(obj, "num_escalation_chains", 0)
+        return num if num is not None else 0
```
vadimkerr, line 80: Wondering if this is more elegant and easy to understand?
```suggestion
        return num or 0
```

vadimkerr (approved):

engine/apps/api/serializers/schedule_base.py
matiasb: I think the original is more explicit, but I guess it is subjective :-) Both work, I can apply your suggestion.

PR #694, opened 2022-10-24
added Redis variables to be configured in helm chart

Timmypedia:
**What this PR does**:
Added the redis connection properties, defined in base.py,  to be configureable in the helm chart.
No issue directly, but its a fix for something that is clearly missing. also closes a request at #538

**Checklist**
- [ ] Tests updated
- [ ] Documentation added
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/engine/settings/base.py
+++ b/engine/settings/base.py
@@ -140,11 +140,11 @@ if DATABASE_TYPE == DatabaseTypes.MYSQL:
     pymysql.install_as_MySQLdb()
 
 # Redis
+REDIS_PROTOCOL = os.getenv("REDIS_PROTOCOL", "redis")
 REDIS_USERNAME = os.getenv("REDIS_USERNAME", "")
 REDIS_PASSWORD = os.getenv("REDIS_PASSWORD")
 REDIS_HOST = os.getenv("REDIS_HOST")
 REDIS_PORT = os.getenv("REDIS_PORT", 6379)
-REDIS_PROTOCOL = os.getenv("REDIS_PROTOCOL", "redis")
 
 REDIS_URI = os.getenv("REDIS_URI")
 if not REDIS_URI:
--- a/helm/oncall/templates/_env.tpl
+++ b/helm/oncall/templates/_env.tpl
@@ -202,11 +202,35 @@
 {{- end -}}
 {{- end -}}
 
+{{- define "snippet.redis.protocol" -}}
+{{- if and (not .Values.redis.enabled) .Values.externalRedis.protocol -}}
+{{ .Values.externalRedis.protocol | quote }}
+{{- else -}}
+"redis"
+{{- end -}}
+{{- end -}}
+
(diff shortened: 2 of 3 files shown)


Timmypedia:
i didn't write or updated the existing tests, just tried to use it with helm locally. So if this is neccessary that have to be done by someone more experienced in this project

Timmypedia:
It seems like, django can't use a Username to authenticate the User at a Redis. This is a breaking change for using OnCall with ElasticCache in AWS.  I Would close this MR until it's possible to really profit from giving the Username and Password to the Helm Chart.
https://django-redis-cache.readthedocs.io/en/latest/advanced_configuration.html#password

iskhakov (approved):

Timmypedia:
Opened an Issue for that #696

nu-vi:
I understand some elasticache configurations require the use of a user, but having the redis_protocol env variable be set of Helm chart is still valuable.

It's possible to have elasticache configured without the use of an user, as well as other external redis instances where one would want to use the 'rediss' protocol.

I think it would still make sense to merge these changes (excluding perhaps the externalRedis.username) without first handling the username issue.

lucasfcnunes:
@nu-vi
I'm enabling the redis from the chart because it's not possible to use our production "rediss://..."

Is there any work around?

nu-vi:
If you want to keep using the helm chart for the installation, the simplest solution would be to manually add the env variable `REDIS_PROTOCOL: "rediss"` to the celery and engine deployments, and restart. 

Of course, anytime you do an helm upgrade you would have to do this again.
That's why it makes sense to add this feature to the helm chart.

I have since moved to install oncall via Kubernetes manifests directly instead of relying on the helm chart.

lucasfcnunes:
Another issue is how to select a specific redis db e.g. `/5`
i.e. `rediss://:mypassword@redis.cache.example:6380/5`

rknightion:
> If you want to keep using the helm chart for the installation, the simplest solution would be to wait for the helm deploy, and then, manually add the env variable `REDIS_PROTOCOL: "rediss"` to the celery and engine deployments (with kubectl edit), and restart.

A better way is to use the "env" option in the chart. This is how I got elasticache working for me:

```
env:
  - name: REDIS_USERNAME
    value: ""
  - name: REDIS_PASSWORD
    value: ""
  - name: REDIS_HOST
    value: "whatever.use1.cache.amazonaws.com"
  - name: REDIS_PORT
    value: "6379"
  - name: REDIS_PROTOCOL
    value: "redis"
```
Did the trick for me

lucasfcnunes:
@rknightion 

What about logical DB <dbid> selection? 0 is the default but it's already being used...

*Option in python redis lib: https://github.com/redis/redis-py/blob/1596ac6a0bbc1e00e57c6a48e255a8f917498e39/redis/client.py#L148

PR #717, opened 2022-10-26
Remove grafana-cloud absolute aliases and add canonical URLs

jdbaldry:


alyssawada (approved):
Thanks a bunch, Jack!

PR #718, opened 2022-10-26
Helm chart: allow using existing secrets for oncall secrets, slack, and telegram

chinnonae:
**What this PR does**:
Allow using existing secrets for these parts
- oncall secrets (SECRET_KEY, MIRAGE_SECRET_KEY, ~~MIRAGE_CIPHER_IV~~)
- slack integration (SLACK_CLIENT_OAUTH_ID, SLACK_CLIENT_OAUTH_SECRET, SLACK_SIGNING_SECRET)
- telegram integration (TELEGRAM_TOKEN)

**Checklist**
- [ ] Tests updated
- [ ] Documentation added
- [ ] `CHANGELOG.md` updated


chinnonae:
I've found that MIRAGE_CIPHER_IV generated by this chart is too long (40 bytes). I already committed the changes to the branch in this PR. Is this okay?

DavidSpek:
You should probably add a key value when using an existing secret. You can see how I did this for Postgres in https://github.com/grafana/oncall/pull/661.

helm/oncall/templates/_env.tpl
```diff
@@ -4,15 +4,18 @@
...
       key: SECRET_KEY
 - name: MIRAGE_SECRET_KEY
   valueFrom:
     secretKeyRef:
-      name: {{ template "oncall.fullname" . }}
+      name: {{ template "snippet.oncall.secret.name" . }}
       key: MIRAGE_SECRET_KEY
 - name: MIRAGE_CIPHER_IV
-  value: "1234567890abcdef"
+  valueFrom:
```
iskhakov, line 15: Not sure if we can change this value on the fly as it can break existing installations. Let's keep the default value https://github.com/luojilab/django-mirage-field#mirage_cipher_iv
This field can be considered as non secret

chinnonae: Okay. I've restore the default and put warning in the values file.

iskhakov (approved):

Issue #633, opened 2022-10-10
Better support for non-continuous shift schedules


andy-paine-numan:
We have some schedules that we would only like to be Monday-Friday, 9am-5pm as they are for non-urgent issues. We assign someone to these shifts for each working day (rather than one person for the whole week) which seems impossible to represent with the current `rolling_users` shift type as a daily rotation cannot be filtered with `by_day`.

In addition to this, we get Slack notifications for shifts that are unfilled between 5pm-9am that we do not intend to fill which creates a bit of unnecessary noise.

The rolling users shifts and Slack integration work extremely well otherwise so we would love the extra flexibility for this use case

raphael-batte:
Our upcoming calendar will have this rotation type.

iskhakov:
closing as this is implemented

PR #720, opened 2022-10-26
Enable daily/by-day shifts option for web schedules

matiasb:
**What this PR does**:
- Enable daily/by-day shifts option for web schedules
- A few minor fixes for schedules backend (to handle some ical/terraform cases)
- Update Makefile

Related to #633 and #733

### DIFF
--- a/engine/apps/api/views/schedule.py
+++ b/engine/apps/api/views/schedule.py
@@ -1,3 +1,5 @@
+import datetime
+
 import pytz
 from django.core.exceptions import ObjectDoesNotExist
 from django.db.models import Count, OuterRef, Subquery
@@ -289,7 +291,11 @@ class ScheduleView(
         users = {u: None for u in schedule.related_users()}
         for e in events:
             user = e["users"][0]["pk"] if e["users"] else None
-            if user is not None and users.get(user) is None and e["end"] > now:
+            event_end = e["end"]
+            if not isinstance(event_end, datetime.datetime):
+                # all day events end is a date, make it a datetime for comparison
+                event_end = datetime.datetime.combine(event_end, datetime.datetime.min.time(), tzinfo=pytz.UTC)
+            if user is not None and users.get(user) is None and event_end > now:
                 users[user] = e
 
         result = {"users": users}
--- a/engine/apps/schedules/models/custom_on_call_shift.py
+++ b/engine/apps/schedules/models/custom_on_call_shift.py
@@ -277,31 +277,44 @@ class CustomOnCallShift(models.Model):
         return is_finished
 
     def _daily_by_day_to_ical(self, time_zone, start, users_queue):
-        """Create ical weekly shifts to distribute user groups combining daily + by_day."""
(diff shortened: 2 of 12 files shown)


engine/apps/schedules/models/custom_on_call_shift.py
```diff
@@ -274,6 +276,56 @@ def event_is_finished(self):
 
         return is_finished
 
+    def _daily_by_day_to_ical(self, time_zone, start, users_queue):
+        """Create ical weekly shifts to distribute user groups combining daily + by_day."""
```
vadimkerr, line 280: Probably it makes sense to provide an example here in the comment so it's more clear what the method does

engine/apps/schedules/models/custom_on_call_shift.py
```diff
@@ -274,6 +276,56 @@ def event_is_finished(self):
...
+        day_by_day_rrule = copy.deepcopy(self.event_ical_rules)
+        day_by_day_rrule["interval"] = 1
+        for users in cycle_users:
+            for i in range(self.interval):
+                if not start:  # means that rotation ends before next event starts
+                    all_rotations_checked = True
+                    break
+                last_start = start
+                day = CustomOnCallShift.ICAL_WEEKDAY_MAP[start.weekday()]
+                if (users, day, i) in combinations:
```
vadimkerr, line 299: Could there be any issues if there are identical user groups in `users_queue`? Seems like in this case `(users, day, i)` could be identical and the loop would exit earlier than expected?

engine/apps/api/views/schedule.py
```diff
@@ -289,7 +291,11 @@ def next_shifts_per_user(self, request, pk):
         users = {u: None for u in schedule.related_users()}
         for e in events:
             user = e["users"][0]["pk"] if e["users"] else None
-            if user is not None and users.get(user) is None and e["end"] > now:
+            event_end = e["end"]
+            if not isinstance(event_end, datetime.datetime):
```
vadimkerr, line 295: Will this be still needed after https://github.com/grafana/oncall/pull/742 is merged and [`end` is always `datetime.datetime`](https://github.com/grafana/oncall/blob/e103ec9152f0945deab760bf944c9d8a11975389/engine/apps/schedules/models/on_call_schedule.py#L230)?

vadimkerr (approved):

engine/apps/api/views/schedule.py
matiasb: This endpoint is available for any type of schedule, so it may still break depending on how we call it. Eventually, if we unify the backend representation we should be able to remove this check (there are a few other places where we could also clean this up later).

vadimkerr: Hmm, I thought if it uses `final_events` which always passes `all_day_datetime=True` we should be good without this check here?

```
events = schedule.final_events(user_tz, starting_date, days=30)
```

matiasb: Oh, I see what you mean, you are right. With the other PR already merged, will update here and review the other occurrences I took note of :+1:

PR #760, opened 2022-11-02
One startup command to rule them all

joeyorlando:
**What this PR does**:
This PR simplifies running the project locally.

- running the entire project locally for development can now be done with `make start`. "Hot-reloading" on both the frontend and backend still works as before. Also, adds `grafana-plugin/Dockerfile.dev` to be able to containerize the frontend (frontend needed `WATCHPACK_POLLING=true` env var to properly re-build the plugin on changes).
- fixes the issue, when running the project locally, where you had to restart the `grafana` container in order for it to recognize your local OnCall setup. This was caused because if the `/dist` directory didn't exist, Grafana would just use the remote latest version of OnCall. It gets around this by running `yarn build:dev` _before_ running `docker-compose`.
- parses a [`COMPOSE_PROFILES` env var](https://docs.docker.com/compose/profiles/) which handles which docker services to start, and which `.env` files to load
- allows specifying a `GRAFANA_VERSION` env var to change the docker-compose image tag of the `grafana` container
- deduplicates `docker-compose-developer.yml` and `docker-compose-developer-pg.yml` configuration files into a single `docker-compose-developer.yml`
- most of the pre-existing `Makefile` commands have also been "dockerized"

**Other smaller stuff**
- creates a `dev` directory which stores the dev specific `.env` files. In addition moves `DEVELOPER.md` -> `./dev/README.md` and updates references to this file
- removes old instructions in `DEVELOPER.md` on running frontend outside of Docker (should be no reason to not run it in `docker-compose`?)
- Extracts `regex==2021.11.2` and `uwsgi` from `engine/Dockerfile` and instead puts them in `engine/requirements.txt` (`uwsgi` was already specified in both spots anyways..)
- pins `redis` docker images used to `7.0.5` (latest version). Currently we were not specifying a tag which implicitly means "take the latest image". This should be avoided so that we don't run into unexpected issues when later versions of redis are published

https://github.com/grafana/oncall-private/milestone/16

**Checklist**
- [x] Documentation added
- [x] `CHANGELOG.md` updated

**TODO**
- [x] figure out why `oncall_ui` continually prints out `Compiling... No errors found` (already have a discussion going w/ the team responsible for `@grafana/toolkit`.. update: it's related to [this](https://github.com/grafana/grafana/issues/57862))


.dockerignore
```diff
@@ -1,9 +1,5 @@
 venv/*
 venv2.7/*
 .DS_Store
-frontend/node_modules
-frontend/build
-package-lock.json
```
joeyorlando, line 6: these three ignore lines are no longer relevant

.dockerignore
```diff
@@ -1,9 +0,0 @@
-venv/*
-venv2.7/*
-.DS_Store
-frontend/node_modules
-frontend/build
-package-lock.json
-./engine/extensions
-.env
-.env-hobby
```
joeyorlando, line 9: this seemed fairly outdated. In any case, the specified `context` for all of our published images is `engine/` (see pipelines in `.drone.yml`), thereby I don't think this file would be picked up

.drone.yml
```diff
@@ -104,6 +104,7 @@ steps:
       repo: us.gcr.io/kubernetes-dev/oncall
       dockerfile: engine/Dockerfile
       context: engine/
+      target: prod
```
joeyorlando, line 107: specifying `target` of `prod` simply means "don't install the sqlite/mysql/postgres clients into the image"

.env.dev
```diff
@@ -15,7 +16,7 @@ TWILIO_AUTH_TOKEN=
 TWILIO_NUMBER=
 
 DJANGO_SETTINGS_MODULE=settings.dev
-SECRET_KEY=jkashdkjashdkjh
```
joeyorlando, line 18: ran across a few errors at some points complaining about the secret being < 30 characters

.env.dev
```diff
@@ -1,4 +1,5 @@
 RUNSERVER_PORT=8080
+COMPOSE_PROFILES=with_grafana
```
joeyorlando, line 2: this was referenced in `README.md`. It mentioned to manually add this to a `.env`. I added a note about how to disable this in `DEVELOPER.md`.

.gitignore
```diff
@@ -29,12 +24,10 @@ node_modules
...
 # misc
 .DS_Store
-.env.local
-.env.development.local
-.env.test.local
-.env.production.local
 .swp
 
+.env
+
```
joeyorlando, line 30: bit of cleanup of some redundant rules (or rules that seem outdated)

DEVELOPER.md
```diff
@@ -1,188 +1,81 @@
...
+    - [Setting environment variables](#setting-environment-variables)
+    - [`COMPOSE_PROFILES`](#compose-profiles)
   - [Slack application setup](#slack-application-setup)
   - [Update drone build](#update-drone-build)
-- [Troubleshooting](#troubleshooting)
-  - [ld: library not found for -lssl](#ld-library-not-found-for--lssl)
-  - [Could not build wheels for cryptography which use PEP 517 and cannot be installed directly](#could-not-build-wheels-for-cryptography-which-use-pep-517-and-cannot-be-installed-directly)
-  - [django.db.utils.OperationalError: (1366, "Incorrect string value ...")](#djangodbutilsoperationalerror-1366-incorrect-string-value-)
-  - [Empty queryset when filtering against datetime field](#empty-queryset-when-filtering-against-datetime-field)
-- [Hints](#hints)
```
joeyorlando, line 13: Sections under "Hints"/"Troubleshooting" should no longer be relevant now that things are run all in docker

README.md
```diff
@@ -18,60 +18,7 @@ Developer-friendly incident response with brilliant Slack integration.
...
-```shell
-# Update Docker image
-docker-compose pull engine
-
-# Re-deploy
-docker-compose up -d
-```
-
-After updating the engine, you'll also need to click the "Update" button on the [plugin version page](http://localhost:3000/plugins/grafana-oncall-app?page=version-history).
-See [Grafana docs](https://grafana.com/docs/grafana/latest/administration/plugin-management/#update-a-plugin) for more info on updating Grafana plugins.
```
joeyorlando, line 74: personal nitpick here.. I felt that the instructions under "Getting Started" conflicted/duplicated instructions in `DEVELOPER.md`. I would propose just having the below message to point users to `DEVELOPER.md` if their goal is to run the project locally.

engine/settings/base.py
```diff
@@ -92,11 +92,11 @@ class DatabaseTypes:
     },
 }
 
-DATABASE_NAME = os.getenv("DATABASE_NAME") or os.getenv("MYSQL_DB_NAME")
-DATABASE_USER = os.getenv("DATABASE_USER") or os.getenv("MYSQL_USER")
-DATABASE_PASSWORD = os.getenv("DATABASE_PASSWORD") or os.getenv("MYSQL_PASSWORD")
-DATABASE_HOST = os.getenv("DATABASE_HOST") or os.getenv("MYSQL_HOST")
```
joeyorlando, line 98: the `MYSQL_*` env vars are not referenced anywhere in the codebase. Lets simplify things by just using the `DATABASE_*` env vars.

engine/settings/dev.py
```diff
@@ -17,15 +17,7 @@
         "PORT": DATABASE_PORT or DATABASE_DEFAULTS[DATABASE_TYPE]["PORT"],
     }
 
-if BROKER_TYPE == BrokerTypes.RABBITMQ:
-    CELERY_BROKER_URL = "pyamqp://rabbitmq:rabbitmq@localhost:5672"
-elif BROKER_TYPE == BrokerTypes.REDIS:
-    CELERY_BROKER_URL = "redis://localhost:6379"
```
joeyorlando, line 23: this was redundant with what was specified in `engine/settings/base.py`

engine/settings/hobby.py
```diff
@@ -1,7 +0,0 @@
-from .prod_without_db import *  # noqa: F403
```
joeyorlando, line 1: AFAICT this file was not referenced anywhere?

joeyorlando:
one thing that I still couldn't figure out is how to have the `grafana` container "wait" until the oncall UI container has finished building, this way you do not need to restart the `grafana` container.

For now I figured this is a nice-to-have

README.md
mderynck: I disagree with this.  I think the landing page for the project should have the most basic setup to run OnCall without following another link.  Most of the projects I use tell me how to get started on their landing page, having to go to another page is just an opportunity to lose people.  If I want to just try things out, reading my options are production or local development tells me this project isn't for me.

joeyorlando:
add back `settings.hobby`

DEVELOPER.md
```diff
@@ -1,188 +1,84 @@
 - [Developer quickstart](#developer-quickstart)
```
joeyorlando, line 1: add section on using own IDE, maybe give options to disable certain docker-compose containers (ie. profiles)

docker-compose.yml
```diff
@@ -1,98 +1,233 @@
...
+    command: "python manage.py runserver 0.0.0.0:8080"
+    env_file: *oncall-env-files
+    environment: *oncall-env-vars
     ports:
       - "8080:8080"
-    command: >
-      sh -c "uwsgi --ini uwsgi.ini"
-    environment: *oncall-environment
     volumes:
-      - oncall_data:/var/lib/oncall
```
iskhakov, line 27: we cannot get rid of docker-compose.yaml and[docker-compose-mysql-rabbitmq.yml completely because they are used for hobby installations.
Hobby installation can be done without cloning git repo. We need to keep named volumes for such installation. I propose to have two different versions of docker-compose files: for hobby deployments (from docker hub) and for development (build from local context)

I want to keep hobby yaml as simple as possible as it will increase conversion to installation

README.md
joeyorlando: discussed w/ @mderynck. Kept the basic instructions on `README.md` and modified them w/ the new startup instructions

engine/Dockerfile
```diff
@@ -1,23 +1,24 @@
-FROM python:3.9-alpine
-RUN apk add bash python3-dev build-base linux-headers pcre-dev mariadb-connector-c-dev openssl-dev libffi-dev git
-RUN pip install uwsgi
+FROM python:3.9-alpine AS prod
 
 WORKDIR /etc/app
-COPY ./requirements.txt ./
```
joeyorlando, line 6: `COPY ./requirements.txt ./` was redundant as there is a `COPY ./ ./` statement down on line 10

engine/Dockerfile
```diff
@@ -1,23 +1,24 @@
...
-RUN pip install regex==2021.11.2
```
joeyorlando, line 7: removed `regex==2021.11.2` from the `Dockerfile` and instead placed it in `requirements.txt`. `uwsgi` was already specified in `requirements.txt` so I removed it from this file.

.env.dev
```diff
@@ -15,7 +15,7 @@ TWILIO_AUTH_TOKEN=
 TWILIO_NUMBER=
```
joeyorlando, line 15: I think by not having to have users copy `.env.dev.example` -> `.env.dev` we can avoid the case where a new change in `.env.dev.example` doesn't get reflected in their `.env.dev` or `.env` because they forgot to copy any new additions.

The docker containers are started w/ 3 env files:
- `.env.dev`
- `.env.${DB}.dev`
- `.env` (gitignored; variables set in  this file get the highest priority)

DEVELOPER.md
```diff
@@ -1,188 +1,87 @@
...
+### Running the project
 
-- [isort](https://github.com/PyCQA/isort), [black](https://github.com/psf/black) and [flake8](https://github.com/PyCQA/flake8) are used to format backend code
-- [eslint](https://eslint.org) and [stylelint](https://stylelint.io) are used to format frontend code
-- To run formatters and linters on all files: `pre-commit run --all-files`
-- To install pre-commit hooks: `pre-commit install`
+1. Firstly, ensure that you have `docker` [installed](https://docs.docker.com/get-docker/) and running on your machine.
+2. Run `make start`. By default this will run everything in Docker, using SQLite as the database and Redis as the message broker/cache. See [Running in Docker](#running-in-docker) below for more details on how to swap out/disable which components are run in Docker.
+3. Open Grafana in a browser [here](http://localhost:3000/plugins/grafana-oncall-app) (login: `oncall`, password: `oncall`).
+4. You should now see the OnCall plugin configuration page. Fill out the configuration options as follows:
```
joeyorlando, line 21: you no longer need to restart your `grafana` container for it to recognize local OnCall. This further simplifies these instructions a bit.

joeyorlando:
`@grafana/toolkit` doesn't seem to want to support hot-reloading for the frontend. There's a ongoing Slack discussion on how to resolve this

.env.dev
vadimkerr: I propose to keep the `.env` untouched for dev environment, since `.env` is used for `docker-compose.yml` by default. Would be great to keep them separate, and use only `.env.dev` and `.env.${DB}.dev` for dev environment.

.env.mysql.dev
```diff
@@ -0,0 +1,12 @@
+DATABASE_USER=root
```
vadimkerr, line 1: I like the idea of having different env files for different DBs on dev environment, but don't like that it's in the root directory. Probably we can create a new `dev` directory to keep all of that stuff there?

.gitignore
```diff
@@ -1,17 +1,12 @@
 # Backend
 */db.sqlite3
-engine/oncall_dev.db
+engine/*.db
 *.pyc
 venv
 .python-version
-.env
-.env_hobby
-.env.dev
```
vadimkerr, line 9: I think `.env.dev` should be ignored too, so `.env` is for running OSS hobby, `.env.dev` for running local dev environment

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Change Log
 
+## v1.1.0 (YYYY-MM-DD)
+
+- Modify `docker-compose-developer` configuration files, and `Makefile`, to support running everything in containers for local development
```
vadimkerr, line 5: Not sure if we need to include this to the change log, given these are not user facing changes

DEVELOPER.md
```diff
@@ -1,188 +1,87 @@
...
-  - [Setup using Makefile](#setup-using-makefile)
+  - [Running the project](#running-the-project)
+  - [Running in Docker](#running-in-docker)
+    - [`COMPOSE_PROFILES`](#compose_profiles)
+  - [Useful `make` commands](#useful-make-commands)
+  - [Setting environment variables](#setting-environment-variables)
   - [Slack application setup](#slack-application-setup)
   - [Update drone build](#update-drone-build)
-- [Troubleshooting](#troubleshooting)
-  - [ld: library not found for -lssl](#ld-library-not-found-for--lssl)
```
vadimkerr, line 9: I would imagine some of these could be still useful when `engine` is running not in docker.
For example, I couldn't get my PyCharm connected to the Python interpreter in the engine service, so for now I would probably use docker-compose without the `engine` profile, so I can connect to the local interpreter for debugging purposes.

Makefile
```diff
@@ -1,73 +1,89 @@
...
+	DB=mysql
+else ifeq ($(findstring mysql,$(COMPOSE_PROFILES)),mysql)
+	DB=postgres
+else
+	DB=sqlite
+endif
+
+# conditionally assign BROKER_TYPE based on what is present in COMPOSE_PROFILES
+ifeq ($(findstring redis,$(COMPOSE_PROFILES)),redis)
+	BROKER_TYPE=redis
```
vadimkerr, line 23: `redis` is always needed (at least for caching), so seems like it would be a good idea to allow overriding `BROKER_TYPE` to `rabbitmq` even if `redis` is in `COMPOSE_PROFILES`

Makefile
```diff
@@ -1,73 +1,89 @@
...
+else
+	BROKER_TYPE=rabbitmq
+endif
+
+define run_engine_docker_command
+    docker run -it \
```
vadimkerr, line 29: Probably it's better to use `docker-compose run` here? So there would be no need to specify the volume, env and network info.

Makefile
```diff
@@ -1,73 +1,89 @@
...
+	fi
+endif
 
-shell: bootstrap
-	$(setup_engine_env) && $(PYTHON3) manage.py shell $(ARGS)
+	COMPOSE_PROFILES=$(COMPOSE_PROFILES) DB=$(DB) BROKER_TYPE=$(BROKER_TYPE) docker-compose -f $(DOCKER_COMPOSE_FILE) up --remove-orphans -d
 
-test: bootstrap
-	$(setup_engine_env) && $(PYTEST) --ds=settings.dev $(ARGS)
+stop:
```
vadimkerr, line 53: Probably we could also introduce a `purge` command that removes volumes as well?

docker-compose-developer.yml
```diff
@@ -1,80 +1,232 @@
...
-      - "3000:3000"
-    depends_on:
-      mysql:
-        condition: service_healthy
+    profiles:
+      - grafana
+
+volumes:
+  redisdata:
+  grafana_data:
```
vadimkerr, line 225: nit: all other volume names don't have an underscore, consider making this consistent.
Also we could probably add a `_dev` postfix for these volumes so they're not mixed up with the volumes used for hobby OSS environment.

engine/Dockerfile
vadimkerr: This is intentional: https://stackoverflow.com/a/34399661

engine/settings/base.py
vadimkerr: Let's keep `MYSQL_*` vars for now, because this is needed for backwards compatibility. 
I should've add some comments here when I was working on this, my bad.

engine/settings/base.py
```diff
@@ -106,7 +106,7 @@ class DatabaseTypes:
 DATABASE_CONFIGS = {
     DatabaseTypes.SQLITE3: {
         "ENGINE": DATABASE_ENGINE,
-        "NAME": DATABASE_NAME or "/var/lib/oncall/oncall.db",
```
vadimkerr, line 109: I think `or "/var/lib/oncall/oncall.db"` should stay so there's a sensible default value and there's no need to specify `DATABASE_NAME` when using sqlite.
Probably this also would break existing hobby instances that don't declare `DATABASE_NAME` explicitly.

vadimkerr (changes requested):
This will simplify things so much, amazing job @joeyorlando!

Makefile
joeyorlando: great suggestion 👍

Makefile
joeyorlando: do you mean a `docker system prune` helper command that removes oncall dev resources?

engine/Dockerfile
joeyorlando: good catch, I'll revert this and add a comment

Makefile
joeyorlando: I think by inversing the logic to this, it _should_ work:
```Makefile
# conditionally assign BROKER_TYPE based on what is present in COMPOSE_PROFILES
# if the user specifies both rabbitmq and redis, we'll make the assumption that rabbitmq is the broker
ifeq ($(findstring rabbitmq,$(COMPOSE_PROFILES)),rabbitmq)
	BROKER_TYPE=rabbitmq
else
	BROKER_TYPE=redis
endif
```

Test cases:
```shell
❯ COMPOSE_PROFILES=redis,rabbitmq make stop
COMPOSE_PROFILES=redis,rabbitmq DB=sqlite BROKER_TYPE=rabbitmq docker-compose -f docker-compose-developer.yml down

❯ COMPOSE_PROFILES=rabbitmq make stop
COMPOSE_PROFILES=rabbitmq DB=sqlite BROKER_TYPE=rabbitmq docker-compose -f docker-compose-developer.yml down

❯ COMPOSE_PROFILES=redis make stop
COMPOSE_PROFILES=redis DB=sqlite BROKER_TYPE=redis docker-compose -f docker-compose-developer.yml down
```

I would argue that it's safe to assume that if the user specifies both `redis` and `rabbitmq` in `COMPOSE_PROFILES` that `rabbitmq` should be the broker

.env.dev
joeyorlando: good point 👍

.env.mysql.dev
joeyorlando: good suggestion on `dev` directory, I think this is a lot cleaner, I'm going to move the `.env.dev*` files there and considering also moving `DEVELOPER.md` there as well, thoughts?

vadimkerr: Makes sense, probably also can rename it to `README.md` so it shows up when you visit the directory on GH

Makefile
vadimkerr: Yes, this looks nice

Makefile
vadimkerr: Hmm, I was thinking more like adding a `--volumes` flag to the `down` command: https://docs.docker.com/engine/reference/commandline/compose_down/#options

But we could probably do `docker prune` stuff as well

engine/Dockerfile
```diff
@@ -1,19 +1,26 @@
...
 
+FROM base AS dev
+
+# these are needed for the django dbshell command
+RUN apk add sqlite mysql-client postgresql-client
+
+FROM base AS prod
 
 # This is required for prometheus_client to sync between uwsgi workers
 RUN mkdir -p /tmp/prometheus_django_metrics;
```
joeyorlando, line 26: having it setup this way does two things:
- doesn't affect current image builds in our CI pipelines (not specifying a build target builds the final target, in this case it would skip installing the database clients which is likely what we want in our prod image)
- we simply specify the `dev` build target in our `docker-compose-developer.yml` and this will add the db clients locally

vadimkerr (approved):

PR #761, opened 2022-11-02
Helm: allow for using existing secret for RabbitMQ

DavidSpek:
**What this PR does**:
This PR adds the ability to use an existing secret for external RabbitMQ. It a key in the existing secret can be specified for both the password and the username. The ability to specify the username key is required when an operator is used to create RabbitMQ credentials. We are using the [RabbitMQ Messaging Topology Kubernetes Operator](https://github.com/rabbitmq/messaging-topology-operator) to create the vhost, user and permissions for OnCall, where both the username and password are auto-generated and placed into a secret.

**Checklist**
- [ ] Tests updated
- [x] Documentation added
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/helm/oncall/Chart.yaml
+++ b/helm/oncall/Chart.yaml
@@ -8,7 +8,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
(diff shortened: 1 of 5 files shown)


iskhakov (approved):

iskhakov:
@DavidSpek just FYI, I've applied a small fix to the RABBITMQ_USERNAME part

PR #767, opened 2022-11-03
Rares/jest fix packages

teodosii:
**What this PR does**:

- Adds `React` as dependency
- Correct Jest configuration to use Babel and import React
- Rewrites correct `yarn.lock` file


grafana-plugin/src/components/CardButton/CardButton.test.tsx
```diff
@@ -1,4 +1,4 @@
-import 'jest/matchMedia.ts';
+import 'src/jest/matchMedia.ts';
```
maskin25, line 1: just curious why do we need `src` prefix here?

maskin25 (approved):

grafana-plugin/src/components/CardButton/CardButton.test.tsx
teodosii: At some moment it complained on that import, removed it now.

PR #786, opened 2022-11-05
Improve documentation on local testing

scubbo:
Four main changes:
* Correct the port number from `8000` to `8080`.
* Clarify that, in fact, only _public_ API endpoints (not internal ones) are available at this point. Unless I've misunderstood `engine/apps/auth_token/auth.py:PluginAuthentication` (which seems to apply to most non-public APIs), authentication requires an `X-Instance-Context` header, which itself requires an `stackId`, which doesn't seem to be accessible to the user. My intuition is that Public APIs (that is - those authenticated with `APITokenAuthentication`) are intended to be called by the UI (or by developers testing/experimenting), whereas internal APIs (those authenticated with `PluginAuthentication`) should only be called by the Grafana instance (via `/api/plugin-proxy`) or by the plugin calling itself. It's entirely possible that I've misunderstood this, though, so please do feel free to correct me!
* Give guidance on how to find an Authentication token. I don't want to admit to how long I spent clicking around trying to reverse-engineer how to construct a token before checking the obvious place!
* Guidance on how to find the url for a given API - or, equivalently, how to find the set of available APIs.

Also fix a double-space.


scubbo:
Closing this pull request as it's been superceded by later changes.

PR #793, opened 2022-11-07
Check for access deleted because it is not yet loaded

Ukochka:
**What this PR does**:

**Checklist**
- [ ] Tests updated
- [ ] Documentation added
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/grafana-plugin/src/pages/users/Users.tsx
+++ b/grafana-plugin/src/pages/users/Users.tsx
@@ -75,10 +75,6 @@ class Users extends React.Component<UsersProps, UsersState> {
     const { usersFilters, page } = this.state;
     const { userStore } = store;
 
-    if (!store.isUserActionAllowed(UserAction.ViewOtherUsers)) {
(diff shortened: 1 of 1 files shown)


grafana-plugin/src/pages/users/Users.tsx
```diff
@@ -75,10 +75,6 @@ class Users extends React.Component<UsersProps, UsersState> {
     const { usersFilters, page } = this.state;
     const { userStore } = store;
 
-    if (!store.isUserActionAllowed(UserAction.ViewOtherUsers)) {
```
maskin25, line 78: but if user really doesn't have such permissions then he/she will able to see users list...it seems we need to solve this issue in some other way

maskin25 (changes requested):

joeyorlando:
hi there @Ukochka 👋
does this PR need a review/should it be merged? or should we close it?

Issue #715, opened 2022-10-25
Cannot send email notification via no-auth SMTP server


vonschtirlitz:
Hello,

We were really excited for the email capability rolled out the other day, however, it does not appear to properly work with our internal SMTP relay server, which does not have authentication. We cannot provide a null value for password in the helm template, k8s secret, and GUI, and the celery logs report a `SMTPNotSupportedError('SMTP AUTH extension not supported by server.')` when giving it some dummy values. Is there a workaround to support sending out emails with no authentication, similar to how Grafana does it?

Thanks!

vadimkerr:
Hi @vonschtirlitz, thank you for opening the issue!
Could you please try to set both `EMAIL_HOST_USER` and `EMAIL_HOST_PASSWORD` to be empty strings and see if this helps?

acr-varonis:
@vadimkerr if we set both the EMAIL_HOST_USER and EMAIL_HOST_PASSWORD how can we specify the FROM field of the email message?

vadimkerr:
Seems like this wouldn't work as is, I'm working on introducing a new optional env variable `EMAIL_FROM` that can be used to specify the "from" email address.

acr-varonis:
@vadimkerr an EMAIL_FROM sound like it would be a good solution.

Also I'm unable to unset the user/pass I get "Value: This field may not be null."

Thanks!

vadimkerr:
@acr-varonis I see, I'll take a look at it as well. Thank you for raising this!

vonschtirlitz:
@vadimkerr sorry for the late reply, yes as @acr-varonis mentioned, you cannot pass a null or empty value for the email_host_user or password through the web interface for grafana.

However, we are using a helm chart and pass our values through there. If defining values from the helm chart, you can leave the `oncall.smtp.username` value as null, but chart will fail to start if you make `oncall.smtp.password` value null, as password is required by the secret template. Email is still enabled with the null username, but you get a `ValueError('Invalid address ""')` in the celery logs. You also cannot edit the value from the env variables page since there is nothing there to edit.

Looking at the code and the docs for the email sending framework, our workaround for this was to simply force a null password value in the connection values in the [email tasks code](https://github.com/grafana/oncall/blob/dev/engine/apps/email/tasks.py#L86), and rebuild the image, and pass some random value through helm. Django will not attempt authentication if there is no username or password. It also seems that the from email is the same as the email_host_user, so it looks like it skipped authentication, but didn't know who to send from.

[...]

PR #797, opened 2022-11-07
helm: allow empty smtp password, bump chart version

vadimkerr:
**What this PR does**:
Allow empty SMTP password for helm configuration

https://github.com/grafana/oncall/issues/715


iskhakov:
do we need the secret if the password is empty? May be not create that variable at all if password is not set

iskhakov (approved):

Issue #136, opened 2022-06-23
FEATURE: Use Twilio API keys instead of Auth Tokens


Monkman08:
Can we use Twilio API Credentials instead of Auth Tokens? This is an issue for us as there is only one set of Auth Tokens per Twilio account.

Matvey-Kuk:
I see https://github.com/grafana/oncall/pull/664/files is merged, just checking if it requires documentation & ENV page update?

mderynck:
This feature is still being tested, currently auth token is still required for call status verification.

joeyorlando:
Closing as this was introduced in #664

PR #820, opened 2022-11-10
Twilio add simple check if using API key instead of AUTH token

mderynck:
**What this PR does**:
When using a Twilio API key instead of an Auth token to send messages and make calls simply check the account SID is matching since the request validatation only supports auth tokens ([docs](https://www.twilio.com/docs/usage/security)).  The actual request handling code ensures that the Call SID matches a call we have.

If this isn't secure enough we can either scrap the use of the API key and force only using Auth token or we can leave it as a less secure option.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/engine/apps/twilioapp/views.py
+++ b/engine/apps/twilioapp/views.py
@@ -19,12 +19,17 @@ class AllowOnlyTwilio(BasePermission):
     def has_permission(self, request, view):
         # https://www.twilio.com/docs/usage/tutorials/how-to-secure-your-django-project-by-validating-incoming-twilio-requests
         # https://www.django-rest-framework.org/api-guide/permissions/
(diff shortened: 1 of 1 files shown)


joeyorlando (approved):
LGTM.

I see that we deploy OnCall (for Grafana cloud) using `TWILIO_AUTH_TOKEN`. I assume `TWILIO_ACCOUNT_SID` was an OSS feature request?

mderynck:
> I see that we deploy OnCall (for Grafana cloud) using `TWILIO_AUTH_TOKEN`. I assume `TWILIO_ACCOUNT_SID` was an OSS feature request?

Yes https://github.com/grafana/oncall/issues/136

PR #825, opened 2022-11-10
Cleanup to reduce Helm chart noise

ksa-real:
- Multiple '---' and new lines could be previously produced by secrets.yaml
- Some duplication removed from _env.tpl
- telegram.webhookUrl defaults to base_url

**Checklist**
- [ ] Tests updated
- [ ] Documentation added
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/helm/oncall/templates/_env.tpl
+++ b/helm/oncall/templates/_env.tpl
@@ -52,9 +52,9 @@ MIRAGE_SECRET_KEY
 {{- end -}}
 
 {{- define "snippet.oncall.slack.env" -}}
-{{- if .Values.oncall.slack.enabled -}}
 - name: FEATURE_SLACK_INTEGRATION_ENABLED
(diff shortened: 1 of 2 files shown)


ksa-real:
This doesn't include items in the checklist. Please review the changes first and if it is fine with you, I'll add the rest.

ksa-real:
This original issue has been fixed by another commit. Not that my approach was to define a snippet:
```
{{/* Single env variable. Context: list of var_name: string, value: <map or other> */}}
{{- define "snippet.env" -}}
- name: {{ first . }}
{{- $value := index . 1 }}
{{- if kindIs "map" $value }}
  {{- $value | toYaml | nindent 2 }}
{{- else }}
  value: {{ $value | quote }}
{{- end }}
{{- end }}
```

And then define an environmental variable like this:
```

ksa-real:
I left just the cleanup part in my PR

PR #831, opened 2022-11-11
Add additional validation for route's position field

Konstantinov-Innokentii:
https://github.com/grafana/oncall-private/issues/1466

**Checklist**
- [x] Tests updated

### DIFF
--- a/engine/common/api_helpers/mixins.py
+++ b/engine/common/api_helpers/mixins.py
@@ -154,10 +154,23 @@ class OrderedModelSerializerMixin:
         if order is not None and (self.instance is None or self.instance.order != order):
             last_instance = self.Meta.model.objects.filter(**filter_kwargs).order_by("order").last()
             max_order = last_instance.order if last_instance else -1
+            print(f"MAX ORDER {max_order}")
             if self.instance is None:
                 max_order += 1
             if order > max_order:
                 raise BadRequest(detail="Invalid value for position field")
+
+    def _validate_manual_order(self, order):
+        """
+        For manual ordering validate just that order is valid PositiveIntegrer.
+        User of manual ordering is responsible for correct ordering.
+        However, manual ordering not intended for use somewhere, except terraform provider.
+        """
+
+        # https://docs.djangoproject.com/en/4.1/ref/models/fields/#positiveintegerfield
+        MAX_POSITIVE_INTEGER = 2147483647
+        if order is not None and order < 0 or order > MAX_POSITIVE_INTEGER:
+            raise BadRequest(detail="Invalid value for position field")
 
 
 class PublicPrimaryKeyMixin:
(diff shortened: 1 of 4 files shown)


engine/common/api_helpers/mixins.py
```diff
@@ -154,11 +154,24 @@ def _validate_order(self, order, filter_kwargs):
         if order is not None and (self.instance is None or self.instance.order != order):
             last_instance = self.Meta.model.objects.filter(**filter_kwargs).order_by("order").last()
             max_order = last_instance.order if last_instance else -1
+            print(f"MAX ORDER {max_order}")
```
joeyorlando, line 157:
```suggestion
```

engine/common/api_helpers/mixins.py
```diff
@@ -159,6 +159,18 @@ def _validate_order(self, order, filter_kwargs):
...
+        """
+        For manual ordering validate just that order is valid PositiveIntegrer.
+        User of manual ordering is responsible for correct ordering.
+        However, manual ordering not intended for use somewhere, except terraform provider.
+        """
+
+        # https://docs.djangoproject.com/en/4.1/ref/models/fields/#positiveintegerfield
+        MAX_POSITIVE_INTEGER = 2147483647
+        if order is not None and order < 0 or order > MAX_POSITIVE_INTEGER:
+            raise BadRequest(detail="Invalid value for position field")
```
joeyorlando, line 172: could we just set the `manual_order` attribute in  the `ChannelFilter` model to something like this?
```python3
class ChannelFiler(django.models.Model):
    manual_order = django.models.PositiveIntegerField(**<rest_of_attributes>)     
```

this way we wouldn't need this `_validate_manual_order` method, nor need to hardcode 2147483647. Maybe give it a try and see if your tests still pass.

Konstantinov-Innokentii: It works a little bit different. He have an **order** field in model (It comes from OrderedModel). In serializer that field is called **position** and by default passing that in PUT request will reorder all other routes. 
But it didn't work for terraform (lots of reasons). We introduced a hidden manual_order boolean param to the public channel filter api, to not to reorder all routes automatically when the requests comes from terraform provider. On the other hand we gave an opportunity to pass "-1" as position to move route to the very bottom of list.

joeyorlando: ok ok I see the note in the test about -1 being valid in certain scenarios, so using a built-in range validator doesn't make sense 👍
```
# Test negative value. Note, that for "manual_order"=False, -1 is valud option (It will move route to the bottom)
```

Is there no way to include this validation logic into the `ChannelFilterSerializer.validate` method? This way you do not need to manually invoke it in `ChannelFilterUpdateSerializer.update` and `ChannelFilterSerializer.create`?

Other than that my only comments would be to import the min/max value for `PositiveIntegerField` rather than hardcode:
```python
from django.db.backends.base.operations import BaseDatabaseOperations

min_allowed, max_allowed = BaseDatabaseOperations.integer_field_ranges["PositiveIntegerField"]
```

and update the `BadRequest` detail to something like:
```python
f"Invalid value for position field. Must be >= {min_allowed} and <= {max_allowed}"
```

Konstantinov-Innokentii: It makes sense to move all the validation in one place, but it will require some further refactoring, since after the validation of the order we are updating it via calling **_change_position** method. My original aim was just to fix that 500 error which bothered me with alerts. Our validation system is a little bit messy with its parts across View's and Serializer's methods, but I don't want to make a decision on a new approach yet.

Issue #835, opened 2022-11-11
OnCall bot doesn't show warning pop-up on slash command for disconnected user in Slack


Ferril:
If user with disconnected slack account tries to use `/oncall` slash command in Slack, they see error "dispatch_failed"
<img width="396" alt="Screenshot 2022-11-01 at 08 36 58" src="https://user-images.githubusercontent.com/22730923/201315862-b3d50e80-2200-4d76-b7d5-1b8951e1972f.png">

instead of OnCall warning pop-up
<img width="514" alt="Screenshot 2022-11-11 at 11 04 11" src="https://user-images.githubusercontent.com/22730923/201316919-bf5d6b76-02b8-4af1-96dc-9adca524aa72.png">

Konstantinov-Innokentii:
Fixed

PR #836, opened 2022-11-11
Fix warning pop-up on action in slack

Ferril:
**What this PR does**:
Add warning pop-up on action in slack if user is not connected to OnCall

https://github.com/grafana/oncall/issues/835

### DIFF
--- a/engine/apps/slack/views.py
+++ b/engine/apps/slack/views.py
@@ -272,9 +272,8 @@ class SlackEventApiEndpointView(APIView):
                 return Response(status=200)
         elif not slack_user_identity.users.exists():
             # Means that slack_user_identity doesn't have any connected user
-            warning_text = "Permission denied. Please connect your Slack account to OnCall."
             # Open pop-up to inform user why OnCall bot doesn't work if any action was triggered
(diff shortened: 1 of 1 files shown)


engine/apps/slack/views.py
```diff
@@ -270,6 +270,12 @@ def post(self, request):
                 # Open pop-up to inform user why OnCall bot doesn't work if any action was triggered
                 self._open_warning_window_if_needed(payload, slack_team_identity, warning_text)
                 return Response(status=200)
+        elif not slack_user_identity.users.exists():
+            # Means that slack_user_identity doesn't have any connected user
+            warning_text = "Permission denied. Please connect your Slack account to OnCall."
+            # Open pop-up to inform user why OnCall bot doesn't work if any action was triggered
+            self._open_warning_window_if_needed(payload, slack_team_identity, warning_text)
```
vadimkerr, line 277: Probably we can use `_open_warning_for_unconnected_user` here?

vadimkerr (approved):

engine/apps/slack/views.py
Ferril: Makes sense

Issue #841, opened 2022-11-11
helm chart: make ttlSecondsAfterFinished of job-migrate configurable


kudiyarov:
Setting spec.ttlSecondsAfterFinished tells built-in Job controller to delete Job resource in 20s after completion (current setting):
https://github.com/grafana/oncall/blob/41f8d4afb19ef1217663bbd94180602d9b3767eb/helm/oncall/templates/engine/job-migrate.yaml#L9

In case of using fluxcd for deploy oncall to k8s and after changing some helm values that are present in the "job-migrate" template (i.e. resources/image), fluxcd/helm-controller runs the Job and after some time I get the following error at the helm level:
`Upgrade "oncall" failed: jobs.batch "oncall-engine-migrate-2022-11-09-18-55-09" not found`

Looks like Job controller removes Job resource while fluxcd/helm-controller is expecting to see it's state.

Making ttlSecondsAfterFinished configurable allows me to disable this setting, therefore allows the helm-controller to delete the old Job resource in case of Job's parameters changes

3rikas:
+1 on this. I am running into the same exact problem using ArgoCD.

kudiyarov:
done by @erikastr in https://github.com/grafana/oncall/pull/957 , thanks

PR #843, opened 2022-11-11
parameterize job-migrate ttlSecondsAfterFinished in helm chart

kudiyarov:
**What this PR does**:
parameterize ttlSecondsAfterFinished for job-migrate resource in helm chart

**Which issue(s) this PR fixes**: https://github.com/grafana/oncall/issues/841


kudiyarov:
I've rewritten job-migrate to act as a chart hook by default. ArgoCD also should be able to recognize these annotations.
In case when deployment method doesn't support "hooks", it is possible to specify ttlSecondsAfterFinished.

kudiyarov:
it has some not tested cases

PR #844, opened 2022-11-11
Add documentation for issue make when CDPATH is set unusually

scubbo:
Note - you may be wondering "how could your system possibly function if `.` is not part of `CDPATH`? Testing suggests that behaviour is inconsistent between shells - `sh` will entirely ignore the current directory if `.` is absent, but `zsh` will still attempt to search the current directory even if `.` is not in `CDPATH`


joeyorlando (approved):
thank you for enhancing the documentation!

Issue #846, opened 2022-11-12
Cannot start backend services with `make start` unless _all_ services are started


scubbo:
System information:
* Macbook Pro, Mac OS 12.6, Intel Core i7
* Docker compose version 2.7.0

Steps to reproduce:
```
$ git clone git@github.com:grafana/oncall.git
Cloning into 'oncall'...
[...]

$ cd oncall

$ git show --pretty=oneline -s
41f8d4afb19ef1217663bbd94180602d9b3767eb (HEAD -> dev, origin/dev, origin/HEAD) PD migrator: add support for email notifications (#839)

$ make init
[...]
✔ Compiling...
✨  Done in 45.96s.

$ make start
COMPOSE_PROFILES=engine,oncall_ui,redis,grafana DB=sqlite BROKER_TYPE=redis SQLITE_DB_FILE=./engine/oncall.db docker compose -f docker-compose-developer.yml up --remove-orphans -d
[+] Running 5/6
 ⠿ Container grafana              Created                                                  0.1s
 ⠿ Container oncall_ui            Started                                                 22.9s
 ⠿ Container redis                Waiting                                                 23.0s
 ⠿ Container oncall_db_migration  Created                                                  0.1s
 ⠿ Container oncall_celery        Created                                                  0.1s
 ⠿ Container oncall_engine        Created                                                  0.1s
no such service: postgres
make: *** [start] Error 1
# Note - I can repeat this command and it will error on either `postgres` or `mysql`, apparently randomly

$ make start COMPOSE_PROFILES=engine,oncall_ui,redis,grafana,postgres
[...]
no such service: mysql
make: *** [start] Error 1
make start COMPOSE_PROFILES=engine,oncall_ui,redis,grafana,postgres,mysql
COMPOSE_PROFILES=engine,oncall_ui,redis,grafana,postgres,mysql DB=mysql BROKER_TYPE=redis docker compose -f docker-compose-developer.yml up --remove-orphans -d
[+] Running 10/10
 ⠿ Container oncall_ui                      Running                                        0.0s
 ⠿ Container mysql                          Healthy                                        1.0s
 ⠿ Container redis                          Running                                        0.0s
 ⠿ Container mysql_to_create_grafana_db     Star...                                        1.8s
 ⠿ Container postgres                       Healthy                                        1.0s
 ⠿ Container postgres_to_create_grafana_db  S...                                           1.9s
 ⠿ Container oncall_db_migration            Recreated                                      0.2s
[...]

Issue #239, opened 2022-07-17
Docker-compose "ERROR: Encountered errors while bringing up the project."


Gufderald:
### Discussed in https://github.com/grafana/oncall/discussions/237

<div type='discussions-op-text'>

<sup>Originally posted by **Gufderald** July 15, 2022</sup>
Hi!
While trying to start hobby playground environment, I've faced with next error on `docker-compose --env-file .env_hobby -f docker-compose.yml up --build -d` stage:
```
Creating network "oncall_default" with the default driver
Creating volume "oncall_dbdata" with default driver
Creating volume "oncall_rabbitmqdata" with default driver
Creating oncall_mysql_1    ... done
Creating oncall_redis_1    ... done
Creating oncall_rabbitmq_1 ... done
Creating oncall_oncall_db_migration_1        ... done
Creating oncall_mysql_to_create_grafana_db_1 ... done
Creating oncall_grafana_1                    ... done

ERROR: for engine  Container "5da314125e65" exited with code 1.

ERROR: for celery  Container "5da314125e65" exited with code 1.
ERROR: Encountered errors while bringing up the project.
```
I can also see `kombu.exceptions.OperationalError: [Errno 111] Connection refused` in `oncall_db_migration` logs.

Howewer, after manual restart of `oncall_db_migration` and one more `docker-compose --env-file .env_hobby -f docker-compose.yml up --build -d`, everything works fine. Am I missing something?
</div>

Matvey-Kuk:
@Gufderald thank you for reporting this! Sounds like `oncall_db_migration ` starts before rabbit. We have healthcheck for MySQL, let me see if I can make smth like this for rabbit.

Matvey-Kuk:
@Gufderald should be fixed, thank you for reporting!

Gufderald:
Should I open another issue for that?

Gufderald:
Actually yes. Previously I've tested it both on our main node (64 threads/32GB ram) and VM(2 threads, 4GB ram), and they had same behaviour. Now node is buisy, so I tested only on VM. I'll be able to try it later, and report it here. Thank you for your reply!

Gufderald:
@Matvey-Kuk Hi again! Tested it on a powerful machine with updated `docker-compose.yml`, and got almost same behaviour as at the beginning of this tread: 
```
Creating oncall_redis_1    ... done
Creating oncall_rabbitmq_1 ... done
Creating oncall_mysql_1    ... done
[...]

PR #868, opened 2022-11-21
Add Makefile command to make migrations

scubbo:
Apologies if I've missed it, but I couldn't find an existing command to create/update the `migrations/` directory contents based on changes to the Django models.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -158,6 +158,9 @@ backend-bootstrap:
 backend-migrate:
 	$(call backend_command,python manage.py migrate)
 
+make-migrations:
+	$(call backend_command,python manage.py makemigrations)
+
 run-backend-server:
 	$(call backend_command,python manage.py runserver 0.0.0.0:8080)
 
--- a/dev/README.md
+++ b/dev/README.md
@@ -92,6 +92,7 @@ make start # start all of the docker containers
 make stop # stop all of the docker containers
 make restart # restart all docker containers
 make build # rebuild images (e.g. when changing requirements.txt)
+make make-migrations # generate Django migration files from model definitions
 
 # this will remove all of the images, containers, volumes, and networks
 # associated with your local OnCall developer setup


iskhakov (approved):

joeyorlando:
@scubbo do you mind just adding this command to the list of `make` commands in `./dev/README.md`? 😄

scubbo:
README updated! :)

I should highlight that there's [something wrong the Make setup on my local environment](https://github.com/grafana/oncall/issues/239#issuecomment-1317830381) - I have been able to successfully run the `make-migrations` command in this PR at least once, but I've also had it fail (on apparently identical setup) as well. This might be to do with my Make setup problems, or because I'm just doing something wrong related to the migrations (this is my first time working with Django) - either way, please do check for yourself on your own setups that this command works as-expected before accepting it, don't just take my word for it!

joeyorlando:
> do check for yourself on your own setups that this command works as-expected before accepting it

This looks good to me. `make backend-migrate` is working successfully for me locally (after resolving a `pip` dependency M1 Mac specific issue in #877).

Come to think of it though, we _may_ want to do something like this instead:
```Makefile
backend-manage-command:
	$(call backend_command,python manage.py $(CMD))
```

and invoking as such:
```shell
make backend-manage-command CMD="makemigrations"
```

Reason being that we would eventually end up with one `make` command for each Django manage command. This approach would allow the `make` command to be a bit more dynamic. AFAIK `make` does not support doing something like `make backend-manage-command makemigrations`, hence the `CMD` argument.

joeyorlando:
@scubbo just out of curiosity, and to gather feedback on the local setup, what is missing with the `docker-compose` setup that you run the API/celery workers outside of docker?

For example, I know some folks use PyCharm and connect the debugging tools to their server/celery workers (PyCharm _does_ have support to connect to a remote `docker-compose` Python interpreter, but there is currently a [JetBrains bug filed](https://youtrack.jetbrains.com/issue/PY-52173) which prevents it from working with `COMPOSE_PROFILES`, which we are using).

scubbo:
> Come to think of it though, we may want to do something like this instead:

Ahhh! Yeah, that makes a lot of sense. Initially I was looking for a `make` command to run arbitrary `manage.py` commands, but I couldn't figure out how to plumb parameters (that is, command-names) to `make` itself (I've only recently left a BigTech company where all the Builder Tools were proprietary, so I'm learning Open Source tools as I go. Env variables were, in hindsight, an obvious choice 🙃)

I'll go ahead and make that change.

> just out of curiosity, and to gather feedback on the local setup, what is missing with the `docker-compose` setup that you run the API/celery workers outside of docker?

1. Nothing's missing from docker-compose - I just find that wrapping common `docker-compose` commands in a single `make` command is simpler (and, my reading of the `dev/README.md` file suggested that `make` was the "blessed" approach). But, I think I've miscommunicated somewhere, because...
2. ...I'm not (intentionally?) running API/celery workers outside of docker - they're still started up with `make start` which, AFAICT, does start them in docker (that is - `oncall_celery` shows up in `docker ps`). Though, as I noted [here](https://github.com/grafana/oncall/issues/846), I have had to set `COMPOSE_PROFILES` to start _all_ services, since the `depends_on` behaviour didn't seem to actually work with my older version of `docker-compose`. This meant that `oncall_celery` gets started by `make start`, while it doesn't appear to with the default `COMPOSE_PROFILES`

So, to cut through those complications and local-issues - my ideal situation would be that I can run `make start` (or `make init start` the first time around) which would result in everything starting up (in Docker) that I need.

joeyorlando:
closing as this has been implemented in #886. Thanks for your contribution @scubbo!

PR #871, opened 2022-11-21
Add alert flow integration test

iskhakov:
**What this PR does**:
Adds pytest integration test which
Tests celery tasks related to the alert lifecycle
This test is celery specific, see https://docs.celeryq.dev/en/stable/userguide/testing.html
It uses in memory broker and runs celery worker in separate thread

**Checklist**
- [ ] Tests updated
- [ ] Documentation added
- [ ] `CHANGELOG.md` updated


engine/apps/alerts/tests/integration_test_alert_flow.py
```diff
@@ -0,0 +1,69 @@
...
+from apps.alerts.models import AlertGroup, AlertReceiveChannel
+from apps.integrations.tasks import create_alert
+
+pytest_plugins = ["celery.contrib.pytest"]
+
+# Uncomment this block if you want to test specific queues
+# @pytest.fixture(scope='session')
+# def celery_worker_parameters():
+#     return {
+#         'queues':  ('celery', 'default', 'retry', 'critical'),
```
joeyorlando, line 15: is this commented out block still required?

engine/apps/alerts/tests/integration_test_alert_flow.py
```diff
@@ -0,0 +1,69 @@
...
+#     }
+
+
+@pytest.fixture(scope="session")
```
joeyorlando, line 19: `scope="session"` means that this fixture is auto-used once for **all** tests, correct? could this maybe be scoped to this file (`scope="module"`)? or if it makes sense to keep it to `scope="session"`, maybe move it to `engine/conftest.py`?

joeyorlando (approved):
Nice 👍 

out of curiosity, since the file is named `integration_test_*`, will this test file be picked up when running `pytest` or does it need to be run separately?

iskhakov:
not relevant anymore

PR #872, opened 2022-11-21
Refactor create/update contact points for Alerting integration

Ferril:
**What this PR does**:
- Keep grafana version on create/update contact points to avoid multiple requests to alerting 
- Add retry limit on create contact point async
- Fix bugs related on create contact point
- Update logs on create/update contact point, make them more clear
- Avoid unnecessary requests to Grafana Alerting

**Checklist**
- [x] `CHANGELOG.md` updated


engine/apps/alerts/tasks/create_contact_points_for_datasource.py
```diff
@@ -45,57 +45,57 @@ def create_contact_points_for_datasource(alert_receive_channel_id, datasource_li
...
-    if datasources_to_create:
-        schedule_create_contact_points_for_datasource(alert_receive_channel_id, datasources_to_create)
+    if datasources_to_create and retry_counter < RETRY_LIMIT:
+        logger.debug(
+            f"Create CP task: Retry to create contact points for integration {alert_receive_channel_id}, "
+            f"retry counter: {retry_counter}, datasource list {len(datasource_list)}"
+        )
+        retry_counter += 1
+        schedule_create_contact_points_for_datasource(alert_receive_channel_id, datasources_to_create, retry_counter)
     else:
```
matiasb, line 91: Wondering if we should have more visibility (logs or exception) about an exhausted retries scenario?

engine/apps/alerts/tasks/create_contact_points_for_datasource.py
```diff
@@ -18,22 +17,23 @@ def get_cache_key_create_contact_points_for_datasource(alert_receive_channel_id)
 
 
 @shared_dedicated_queue_retry_task
-def schedule_create_contact_points_for_datasource(alert_receive_channel_id, datasource_list):
+def schedule_create_contact_points_for_datasource(alert_receive_channel_id, datasource_list, retry_counter=0):
```
matiasb, line 20: I guess we have this custom retry handling (instead of using celery retries) because a retry in the create task is really triggering the schedule task (ie. a different task) and we want to keep the count of tasks since the initially scheduled create task, right? Maybe adding some comment would be useful.

matiasb (approved):
LGTM

Konstantinov-Innokentii (approved):

engine/apps/alerts/grafana_alerting_sync_manager/grafana_alerting_sync.py
```diff
@@ -102,50 +117,84 @@ def create_contact_points(self) -> None:
...
+        )
+        if not config_to_update:  # failed to get alerting config for this datasource
+            return None, response_info
+
+        updated_config_from_alerting, response_info = self.add_contact_point_to_grafana_alerting(
+            datasource, is_grafana_datasource, config, config_to_update
+        )
+
+        if not updated_config_from_alerting:
+            return None, response_info
```
Konstantinov-Innokentii, line 162: Why we are returning response_info here? I didn't find any usage of second return value

engine/apps/alerts/tasks/create_contact_points_for_datasource.py
Ferril: I improved logging here to get more information about every step and I also I changed 'restart' task to 'retry' for better visibility.

PR #874, opened 2022-11-21
Mobile app messaging backend

vadimkerr:
This PR moves all the mobile app related backend features to a dedicated messaging backend

- Create a new `mobile_app` Django app
- Move notification logic to the messaging backend `MobileAppBackend`
- Remove critical push notifications for now for simplicity
- Move mobile-specific API/authentication code and models to `mobile_app`
- Rename `MOBILE_APP_PUSH_NOTIFICATIONS_ENABLED` -> `FEATURE_MOBILE_APP_INTEGRATION_ENABLED` for consistency
- Make sure this still works for the existing users of the hackathon app


engine/apps/mobile_app/tasks.py
```diff
@@ -0,0 +1,95 @@
...
+        UserNotificationPolicyLogRecord.objects.create(
+            author=user,
+            type=UserNotificationPolicyLogRecord.TYPE_PERSONAL_NOTIFICATION_FAILED,
+            notification_policy=notification_policy,
+            alert_group=alert_group,
+            reason="Mobile push notification error",
+            notification_step=notification_policy.step,
+            notification_channel=notification_policy.notify_by,
+        )
+        logger.error(f"Error while sending a mobile push notification: user {user_pk} has no devices set up")
```
vadimkerr, line 52: `logger.info`

engine/apps/alerts/tasks/notify_user.py
```diff
@@ -348,47 +346,6 @@ def perform_notification(log_record_pk):
...
-                "orgName": f"{alert_group.channel.organization.stack_slug}",
-                "incidentId": f"{alert_group.public_primary_key}",
-                "status": f"{alert_group.status}",
-                "aps": {
-                    "alert": f"Critical page: {message}",
-                    "interruption-level": "critical",
-                    "sound": "ambulance.aiff",
-                },
-            },
-        )
```
joeyorlando, line 391: 👏

engine/apps/auth_token/migrations/0001_squashed_initial.py
```diff
@@ -48,7 +48,7 @@ class Migration(migrations.Migration):
                 ('digest', models.CharField(max_length=128)),
                 ('created_at', models.DateTimeField(auto_now_add=True)),
                 ('revoked_at', models.DateTimeField(null=True)),
-                ('expire_date', models.DateTimeField(default=apps.auth_token.models.mobile_app_verification_token.get_expire_date)),
+                ('expire_date', models.DateTimeField(default=timezone.now)),
```
joeyorlando, line 51: this would mean the token would expire immediately? this is `get_expire_date`:
```python3
def get_expire_date():
    return timezone.now() + timezone.timedelta(seconds=MOBILE_APP_AUTH_VERIFICATION_TOKEN_TIMEOUT_SECONDS)
```

engine/apps/mobile_app/tasks.py
```diff
@@ -0,0 +1,95 @@
...
+        return
+
+    message = get_push_notification_message(alert_group)
+    thread_id = f"{alert_group.channel.organization.public_primary_key}:{alert_group.public_primary_key}"
+
+    if critical:
+        aps = {
+            "alert": f"Critical page: {message}",
+            "interruption-level": "critical",
+            "sound": "ambulance.aiff",
```
joeyorlando, line 62: are these sounds from an internal list in the `push_notifications` library?

engine/apps/mobile_app/auth.py
```diff
@@ -0,0 +1,45 @@
...
+            return None
+        return user, auth_token
+
+    def authenticate_credentials(self, token_string: str) -> Tuple[Optional[User], Optional[MobileAppAuthToken]]:
+        try:
+            auth_token = self.model.validate_token_string(token_string)
+        except InvalidToken:
+            return None, None
+
+        return auth_token.user, auth_token
```
joeyorlando, line 45: what is the difference between a verification and an auth token?
verification == initial signup and auth == subsequent api calls?

joeyorlando (approved):
LGTM! just some minor questions

engine/apps/auth_token/migrations/0001_squashed_initial.py
vadimkerr: Data from `MobileAppVerificationToken` won't be migrated on purpose, here `timezone.now` is used just to get rid of the reference to `apps.auth_token.models.mobile_app_verification_token.get_expire_date` which is now deleted.

engine/apps/mobile_app/auth.py
vadimkerr: Yup, that's correct. A verification code can be used to get an auth token used for "regular" API calls.

PR #875, opened 2022-11-21
Update selected days to UTC before submit

matiasb:
**What this PR does**:
Fix issue when setting up by-day shifts in a non-UTC timezone but shift start "converts" selected days to different ones when submitting as UTC

### DIFF
--- a/grafana-plugin/src/pages/schedule/Schedule.helpers.ts
+++ b/grafana-plugin/src/pages/schedule/Schedule.helpers.ts
@@ -20,14 +20,14 @@ export const getDateTime = (date: string) => {
 };
 
 export const getUTCByDay = (dayOptions: SelectOption[], by_day: string[], moment: dayjs.Dayjs) => {
-  if (by_day.length > 0 && moment.day() !== moment.utc().day()) {
+  if (by_day.length && moment.day() !== moment.utc().day()) {
     // when converting to UTC, shift starts on a different day,
     // so we need to update the by_day list
     // depending on the UTC side, move one day before or after
+    let offset = moment.utcOffset();
     let UTCDays = [];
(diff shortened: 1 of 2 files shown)


grafana-plugin/src/pages/schedule/Schedule.helpers.ts
```diff
@@ -18,6 +19,30 @@ export const getDateTime = (date: string) => {
   return dayjs(date);
 };
 
+export const getUTCByDay = (dayOptions: SelectOption[], by_day: string[], moment: dayjs.Dayjs) => {
+  if (by_day.length > 0 && moment.day() !== moment.utc().day()) {
```
teodosii, line 23: You can use just `by_day.length`, no need to check for 0

grafana-plugin/src/pages/schedule/Schedule.helpers.ts
```diff
@@ -18,6 +19,30 @@ export const getDateTime = (date: string) => {
...
+    // when converting to UTC, shift starts on a different day,
+    // so we need to update the by_day list
+    // depending on the UTC side, move one day before or after
+    let UTCDays = [];
+    let byDayOptions = [];
+    Object.entries(dayOptions).forEach(([, option]) => byDayOptions.push(option.value));
```
teodosii, line 29: By looking at the `dayOptions` type I see they're just an array (`SelectOption[]`), nothing fancy, therefore I don't think you need `Object.entries`, you could get away just by simply iterating

`dayOptions.forEach(({ value }) => byDayOptions.push(value))`

teodosii (changes requested):

teodosii (approved):

PR #606, opened 2022-10-03
Scaffolding mattermost notification backend

matiasb:
Mattermost messaging backend implementation.

- [x] Initial base messaging backend app
- [ ] Update alert rendering to use Mattermost formatting/style
- [ ] Implement backend user model/linking
- [ ] Implement backend channel model/linking
- [ ] Implement tasks triggering notifications and updates
- [ ] Support interactions in the notifications (ack, resolve, silence)
- [ ] Implement frontend UI: user profile
- [ ] Implement frontend UI: chatops settings
- [ ] Implement frontend UI: integration: post to channel
- [ ] Implement Mattermost [app](https://developers.mattermost.com/integrate/apps/) (or [plugin](https://developers.mattermost.com/integrate/plugins/)?) adding slash commands for linking user and channels

References:
 - [Telegram implementation](https://github.com/grafana/oncall/tree/dev/engine/apps/telegram) (for user and channel verification)
 - [Demo backend implementation](https://github.com/grafana/oncall/pull/566) (for general ideas)
 - Mattermost [interactive messages](https://developers.mattermost.com/integrate/plugins/interactive-messages/)
 - Mattermost [slash commands](https://developers.mattermost.com/integrate/slash-commands/)

PR #566, opened 2022-09-27
Example messaging backend implementation

matiasb:
Requires additional deps installed in the env:
```
$ pip install plyer dbus-python
```
(you may also need to: `apt install libnotify-bin`)

For demonstration purposes only.

PR #876, opened 2022-11-21
Proof-of-concept of Alerting via Matrix

scubbo:
This commit demonstrates a minimal system for delivery of alerts via [Matrix](https://matrix.org/). Given login credentials (userId, password, and homeserver) of an existing Matrix user to send from (the "Bot User"), and given per-(Oncall-)user specifications of (Matrix) userId and roomName, this change permits the delivery of alerts via a Matrix message in the given roomName, sent by the Bot User, tagging the given userId.

**How to use**

* If necessary, create a Matrix server (I've found that the Helm chart `k8s-at-home/dendrite` from [here](https://github.com/k8s-at-home/charts) works just fine, despite being deprecated). Create a user from which to send alerts (see [this gist](https://gist.github.com/scubbo/820df6ef6e8fbe91265762598f30a505)), and (optionally) a user to route alerts to (the former is your "bot", and the latter is "you").
* Start up Oncall with environment variable `FEATURE_MATRIX_INTEGRATION_ENABLED=True`
* For an Oncall user:
  * Set their Matrix User Identity (userId and room in which to be alerted) from the "Matrix Connection" tab in User Settings
  * Add a notification step routing to "Matrix"
* Trigger an alert, observe Matrix message

(If no message is sent, double-check that the Matrix User Identity was completely set in the database - the logic I set to update it on every keystroke in the settings page sometimes gets into race conditions. See discussion below)

This change is _not_ ready for deployment - it's submitted as a proof-of-concept and as a focus for discussion, rather than as a complete feature. In particular:
* There is minimal failure-case checking - failures to log in, to join the target room, or to send a message, are all logged but are not surfaced to the user. Similarly, there is no checking at the time of updating a user's alerting configuration that the room actually exists.
* There's no testing whatsoever. I'm a big fan of tests - but I wanted to get a prototype in front of the Oncall team and ensure that this is a hypothetically-acceptable change and "correctly shaped" before I spend time finalizing, polishing, and comprehensively testing it.
* There are various UX improvements that should be introduced before finalizing this change:
  * Instead of persisting the sign-in credentials for the Bot User in environment variables, it would be better to present a sign-in dialog and then persist only the accessToken.
  * The Homeserver should be optional for creating the MatrixClient - if absent, default to the homeserver defined in the bot userId (need to check the Matrix spec to see if it's actually possible for a userId `@foo:example.org` to sign in to a homeserver other than `example.org`). Similarly, when defining the `matrix_user_identity` for an Oncall user, referencing users or rooms on the same homeserver as the Bot user should be possible without specifying the homeserver (that is, if the Bot user is `@bot-user:example.org`, then defining a `matrix_user_identity` sending to `@target-user:example.org` should only require typing `@target-user`).
* The use of async methods ("[red functions](http://journal.stuffwithstuff.com/2015/02/01/what-color-is-your-function/)") everywhere is kinda awkward. Async calls seem to be the assumed Best Practice in example of Matrix-nio code, and I ran into more intractable problems when trying to get a toy example working synchonously than async; but since we're (effectively) calling synchronously it means there are a bunch of `async/await`s for no real advantage. If this is throwing up some danger signals or is a strong style-clash, I'd be happy to work on retrofitting the client to be synchronous (or even support both types of calls) - it's much easier to get from one working implemntation to another, than to attempt to create one from scratch!
* I'm not aware of any choices I made that deliberately _sacrificed_ security, but I also didn't keep it at the forefront of my mind; and, because I'm unfamiliar with the codebase, there might well be vulnerabilities or considerations that I'm not aware of.
* It is an understatement to say that I'm not a front-end engineer - this is the first time I've worked with Django. I'm sure there's plenty of unidiomatic code in here that could be improved. In particular, updating the `matrix_user_identity` on every keystroke in `MatrixTab.tsx` feels wrong, but I don't know of a better way to implement auto-saving (this may become irrelevant if the "sign in then persist the AccessToken" approach is adopted"). I guess it would be possible to implement some logic that checks for no keystrokes in a given time period (or the dialog being closed) and only then sends an update but, again, I wanted to prioritize getting a working and discussable prototype. Styling and content of the `MatrixTab` dialog also probably needs a lot of work!

**Future improvements**

Even if this change is polished as described above, there are plenty of further improvements possible:
* An option to "notify via DM" would be pretty simple - DMs in Matrix are "just" rooms ([ref](https://github.com/matrix-org/matrix-js-sdk/issues/653) - _Direct Message is not a real concept in Matrix, everything happens inside Rooms, Direct Messages are just rooms with just two users_), so this logic would need to check for an existing "DM" room with the target user, create it if absent, and continue as normal.
* The list of existing rooms could be presented in a drop-down, rather than requiring freeform entry.
* Logic to create a Bot User when setting up the integration, rather than having to create it in advance
* Full-on ChatOps - let Matrix users respond to the bot's message to interact with the alert. From a brief investigation, it looks like Matrix doesn't offer true clickable buttons like Slack does - but a workaround could be to watch for emoji-reacts.

**Questions/Notes/Uncertainties**

* From `email/tasks.py`, it seems like the `reason` for a `UserNotificationPolicyLogRecord` should be pretty generic, and the detail should be in a `logger.error` call - I've copied that in my own setup, I just wanted to confirm that that's correct?
* When I tried setting (e.g.) `value={user_matrix_identity == null ? null : user_matrix_identity.user_id}` for the Input used to set user_id, it became uneditable. Is that intentional? The current "placeholder" behaviour isn't great, as it means the whole value gets deleted if you type anything - correcting typos means you have to type it all out again.
* What's the significance of `_render_for` in subclasses of `AlertTemplater`?
* What's the significance of `templater` and `template_fields` in `backend.py` (copied from the equivalent in `email.backend.py`)? In particular, since the Templater class is explicitly set in `apps.matrix.alert_rendering.build_message`, why is it also necessary to refer to the Templater class in this string variable?
* Why does `email/alert_rendering.py` check `if message` and then `if templated_alert.message`? Couldn't lines 35-36 be removed and line 37 dedented?
* My implementation uses `str_or_backup(html_message, raw_message)` rather than `str_or_backup(message, "")` as in the email templater, but I _think_ they're equivalent (given that the only case in which `html_message` will be absent is if `raw_message` is Falsey)
* Note the use of the earlier version of `matrix-nio` (`v0.10.0`, as opposed to the recommended `v0.20.0`), since the recommended version caused a clash with existing packages.
* It feels really weird to _always_ set `MATRIX_USER_ID/_PASSWORD/_HOMESERVER` in `engine/settings/base.py` (rather than to only set them if `FEATURE_MATRIX_INTEGRATION_ENABLED` is true) - but, if I don't do so, when accessing the site with `FEATURE_MATRIX_INTEGRATION_ENABLED=False`,I get a stacktrace related to their absence from `LiveSettings`


matiasb:
Hi @scubbo!

First of all, thanks a lot for taking the time to make this PoC, and add useful information in the PR and in the code :clap: 

Disclaimer: I haven't set up a matrix server (yet), I haven't used matrix before.

The idea seems to go in the right direction, but there a few things I would change:
(note I will comment mostly on the backend side since I'm not an expert for the frontend part)

- Ideally we shouldn't need to modify the `user_management.User` model (in an ideal world, a messaging backend app would be installable as a separated python/django package). In this case you probably want to add the FK to the User model in the `MatrixUserIdentity` model. Changes outside the matrix app should be as minimal as possible.

- It would be nice to associate the Matrix user id with an account by messaging the bot (as we do for telegram); similar to link channels and make them available to choose when setting up notifications (besides making sure we get/store the room ID and avoid resolving to it later?)

- async and django could be tricky; you may want to use django's [`async-to-sync`](https://docs.djangoproject.com/en/4.1/topics/async/#async-to-sync) instead of `asyncio.run`?

- Eventually, message sending should be handled by a celery task

- Double-checking, have you looked at the [messaging backend example](https://github.com/grafana/oncall/pull/566) and/or the WIP/plans for the [mattermost integration](https://github.com/grafana/oncall/pull/606) as inspiration on how to make this into incremental changes? (thinking on moving this forward, it would be nice to get sequential/self-contained PRs for reviewing/merging)

---

Also, let me try to answer your questions:

> From email/tasks.py, it seems like the reason for a UserNotificationPolicyLogRecord should be pretty generic, and the detail should be in a logger.error call - I've copied that in my own setup, I just wanted to confirm that that's correct?

Correct, the log record is mostly for tracking purposes, and it is displayed in the alert group log too, so `reason` should be generic/human-readable but also give an idea of what/where is the issue.

> When I tried setting (e.g.) value={user_matrix_identity == null ? null : user_matrix_identity.user_id} for the Input used to set user_id, it became uneditable. Is that intentional? The current "placeholder" behaviour isn't great, as it means the whole value gets deleted if you type anything - correcting typos means you have to type it all out again.

Hm.. This seems a frontend question, not really sure.

> What's the significance of _render_for in subclasses of AlertTemplater?

This is something that should eventually be cleaned up; it would just return an id to identify the backend (most of the time it should just return the lowercase version of the backend id).

> What's the significance of templater and template_fields in backend.py (copied from the equivalent in email.backend.py)? In particular, since the Templater class is explicitly set in apps.matrix.alert_rendering.build_message, why is it also necessary to refer to the Templater class in this string variable?

`templater` references the class that implements the templating logic, which allows, for example, using markdown for your templates and then render the output as expected by the respective backend (e.g. slack card); `template_fields` are the fields for which you allow templating (possible values are: `title`, `message`, `image_url`).
Templating is available in an integration configuration. About referencing the templater class in the backend definition, if I remember correctly, this is to know what to use for preview purposes in the web UI (yeah, there are a few rough edges and pending work to make the messaging backends integration cleaner and more general).

> Why does email/alert_rendering.py check if message and then if templated_alert.message? Couldn't lines 35-36 be removed and line 37 dedented?

Sounds like you may be right.

> My implementation uses str_or_backup(html_message, raw_message) rather than str_or_backup(message, "") as in the email templater, but I think they're equivalent (given that the only case in which html_message will be absent is if raw_message is Falsey)

The idea is to fallback to some default message (or indication that there was an error if that makes sense)

> Note the use of the earlier version of matrix-nio (v0.10.0, as opposed to the recommended v0.20.0), since the recommended version caused a clash with existing packages.

Ok :+1: 

> It feels really weird to always set MATRIX_USER_ID/_PASSWORD/_HOMESERVER in engine/settings/base.py (rather than to only set them if FEATURE_MATRIX_INTEGRATION_ENABLED is true) - but, if I don't do so, when accessing the site with FEATURE_MATRIX_INTEGRATION_ENABLED=False,I get a stacktrace related to their absence from LiveSettings

Agree, that should be fixable.

scubbo:
Awesome, thanks a ton for the feedback @matiasb! Good to know that I'm not heading off in a wildly-incorrect direction :)

I'll close this PR so it doesn't pollute your incoming queue, and work through those comments to get this into a better incrementally-acceptable state (I wasn't aware of the messaging backend example and the Mattermost scaffolding, that's very helpful! I'd mostly been working from the existing Slack and Email backends as ~~copy-paste sources~~ inspiration). One point I'm a little confused on, though:

> Eventually, message sending should be handled by a celery task

I thought it was? `notify_user_via_matrix` is annotated with `@shared_dedicated_queue_retry_task`, and the associated logs show up in `docker logs oncall_celery`'s output. It's possible I've misunderstood something there, though!

matiasb:
>>    Eventually, message sending should be handled by a celery task

> I thought it was? notify_user_via_matrix is annotated with @shared_dedicated_queue_retry_task, and the associated logs show up in docker logs oncall_celery's output. It's possible I've misunderstood something there, though!

Note that you are calling the function directly, that avoids queuing a task but instead does a normal function call. To queue the task and run it asynchronously via celery you should use [`.apply_async` or `.delay`](https://docs.celeryq.dev/en/stable/userguide/calling.html) (as you can see in the email backend). In this case with the async definition in the middle some refactoring may be needed (the decorator is to force task retries to a specific queue).

FollieHiyuki:
Sorry for commenting on this closed PR!

@scubbo are you still interested in finishing the implementation of Matrix alert support for Grafana's Oncall? I'd love to have this feature.

scubbo:
Literally - yes, I'm certainly interested in it.

Practically - given that I am now employed again (and so have limited time), and that the official [Oncall App](https://grafana.com/docs/oncall/latest/mobile-app/) is now launched (and so alternative methods of notification are less important to me), I can't in all honesty say it's likely that I'll find the time or motivation to work on it anytime soon. Sorry :/

I'd certainly be happy to chat (email/Signal/voice/whatever) to talk you through (what I can remember of) my work so far if you wanted to pick it up, though.

FollieHiyuki:
Thanks for the suggestion, but sadly I'm not confident enough in my programming skill to continue the work.

PR #884, opened 2022-11-22
Improve Jinja Template feedback and error handling

mderynck:
**What this PR does**:
When editing jinja templates for alert presentation or outgoing webhooks the UI was returning vague errors back to the console making it difficult for the user to correct their template.  Also if errors were encountered during runtime exceptions were being raised which blocked the alert from being displayed in the UI.  This PR improves that experience.

#### General Template Changes
- Templates are limited in size (50000 characters), larger templates will result in `JinjaTemplateError`
- Rendered template results are truncated to a size limit (50000 characters)
- `range` function can no longer be used
- All errors in compiling and running templates are categorized as `JinjaTemplateError` or `JinjaTemplateWarning`
  - `JinjaTemplateError` should block saving of the template
  - `JinjaTemplateWarning` should be displayed but not block saving.  The purpose of this is when templates are saved we don't always have a complete representation of the data that will be used, this allows for accessing fields in the data which may be present at runtime but are not present when the template is being written.
- Any template resulting in a SecurityError is logged

#### Alert Group Template Editor
- Preview area now displays errors/warnings to assist in fixing the template
- On pressing `Save Templates` a notification will popup if any errors block saving (Errors will block saving, Warnings will not)
  -  _In some cases errors are allowed to be saved due to the template code causing the error to not execute in all cases (inside a conditional)_

#### Alert Groups
- Titles rendered in templates are truncated to a size limit (500 characters)
- Each alert body is truncated to a size limit (50000 characters), this also improves performance for both rendering and processing as it avoids urlize needing to process as much data (https://github.com/grafana/oncall-private/issues/1445)
- If processing an alert template results in an error/warning it will be displayed as the content instead of raising an exception.
 
#### Outgoing Webhooks
- Errors/Warnings display in a notification popup to better assist correcting templates.
- Validation of the result from the `Data` template has been relaxed.
  - No longer enforcing that the result is JSON
  - Templates resulting in `JinjaTemplateWarning` are allowed to be saved to allow data that is only available at runtime (https://github.com/grafana/oncall/issues/890)
 - If processing the `Data` template results in an error/warning it will be delivered as the content of the webhook instead of raising an exception.

https://github.com/grafana/oncall/issues/890
https://github.com/grafana/oncall-private/issues/1446
https://github.com/grafana/oncall-private/issues/1445
https://github.com/grafana/oncall-private/issues/1431

**Checklist**
- [x] Tests updated
- [ ] Documentation added
- [x] `CHANGELOG.md` updated


matiasb (approved):
LGTM

PR #892, opened 2022-11-23
[Hotfix] UI not handling invalid grafana tokens properly

joeyorlando:
fix issue where frontend would not properly handle scenario where Grafana API token was not yet available/valid

### DIFF
--- a/grafana-plugin/src/state/rootBaseStore/index.ts
+++ b/grafana-plugin/src/state/rootBaseStore/index.ts
@@ -129,6 +129,9 @@ export class RootBaseStore {
    *
    * Otherwise, get the plugin connection status from the OnCall API and check a few pre-conditions:
    * - plugin must be considered installed by the OnCall API
+   * - token_ok must be true
+   *   - This represents the status of the Grafana API token. It can be false in the event that either the token
+   *   hasn't been created, or if the API token was revoked in Grafana.
    * - user must be not "anonymous" (this is determined by the plugin-proxy)
    * - the OnCall API must be currently allowing signup
    * - the user must have an Admin role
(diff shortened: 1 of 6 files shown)


grafana-plugin/src/state/rootBaseStore/index.ts
```diff
@@ -151,12 +151,12 @@ export class RootBaseStore {
...
     }
 
-    const { allow_signup, is_installed, is_user_anonymous } = pluginConnectionStatus;
+    const { allow_signup, is_installed, is_user_anonymous, token_ok } = pluginConnectionStatus;
     if (is_user_anonymous) {
       return this.setupPluginError(
         '😞 Unfortunately Grafana OnCall is available for authorized users only, please sign in to proceed.'
       );
-    } else if (!is_installed) {
+    } else if (!is_installed || !token_ok) {
```
iskhakov, line 159: may be add more clear explanation of the error?

iskhakov (approved):
Approving to hofix the error. Please review that with Michael after he is online

Makefile
```diff
@@ -166,3 +166,6 @@ run-backend-celery:
 
 backend-command:
 	$(call backend_command,$(CMD))
+
+backend-manage-command:
```
vadimkerr, line 170: Probably we should remove this from the RP? I propose to only keep stuff that's relevant for the `install` issue

grafana-plugin/src/state/rootBaseStore/index.ts
joeyorlando: good call. updated the comment for the `setupPlugin` method just a few lines above 👍

Issue #851, opened 2022-11-14
Incorrect message in phone call


greatvovan:
### Preparation
1. Set up user notifications to receive phone calls (default or important).
2. Add notification of the user to a chain (making it default or important as per the previous step).
3. Create a _Webhook_ integration, copy the URL, click _Open Escalation Settings_.
4. Click pencil and change name to "My beautiful integration".
5. Assign the chain from step `#2` to the integration through the default route.
6. Click _Change alert template and grouping_.
7. Set web title template to `{{ payload.title }}`.
8. Set phone title template to `{{ payload.title }}`.
9. Click _Save Templates_.

### Repro steps
1. Initiate an alert group creation with title "My wonderful alert group" through the integration from step `#3`.
`curl <URL_from_prep_step_3> -H 'Content-Type: Application/json' -d '{"title": "My wonderful alert group"}'`
2. Accept and listen the phone call.

### Expected behaviour
The message correctly identifies parameters of the alert group.

### Actual behaviour
The voice message: "You are invited to check an incident from Grafana OnCall. Alert [illegible - fire?] _My beautiful integration_, web hook with title _My wonderful alert group_ triggered one time. Press 1 to acknowledge, 2 to resolve, 3 to silence for 30 minutes".
**Titles of the alert group and integration are confused.**

### Environment
Grafana Cloud

greatvovan:
After listening again, it might have been "Alert **via** ..., web hook with title ... triggered one time" (correction in bold). With that, it makes sense to put integration name here, but it means that the message must be corrected anyway, because both placeholders are prepared for the name of integration. An example of corrected message:

"Alert via `integration name` titled `alert group title`"

I also not sure what is the purpose of adding "triggered one time", and what are the situations when it can be more than "one".

Konstantinov-Innokentii:
@greatvovan you can check phone_call template [here](https://github.com/grafana/oncall/pull/897/files#diff-502df564d89a3bec4a988e1041943a898421afe32df99ed7957418f29d0a9db7R16).

greatvovan:
Based on the [linked template](https://github.com/grafana/oncall/blob/8a361253aaf523ee27ff5fd93628c5abdc81a264/engine/apps/alerts/incident_appearance/renderers/phone_call_renderer.py#L16), I am realizing I was wrong with the assumption that is is the template that is bad (maybe it is bad, too, but it is not the root cause). Let's check it carefully again:

The integration name:
[...]

PR #897, opened 2022-11-24
Add dot between integration name and alert title in phone call

Konstantinov-Innokentii:
**What this PR does**:
Add dot between integration name and alert title in phone call, to make it more clear

https://github.com/grafana/oncall/issues/851

### DIFF
--- a/engine/apps/alerts/incident_appearance/renderers/phone_call_renderer.py
+++ b/engine/apps/alerts/incident_appearance/renderers/phone_call_renderer.py
@@ -13,7 +13,7 @@ class AlertPhoneCallRenderer(AlertBaseRenderer):
 class AlertGroupPhoneCallRenderer(AlertGroupBaseRenderer):
     TEMPLATE = (
         "You are invited to check an incident from Grafana OnCall. "
-        "Alert via {integration_name} with title {title} triggered {alert_count} times"
+        "Alert via {integration_name}. With title {title} triggered {alert_count} times"
     )
 
     @property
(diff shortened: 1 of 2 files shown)


iskhakov (approved):

iskhakov:
Assuming this is for the additional delay
May be we can change the speed of the whole text in twilio?

greatvovan (commented):
Looking at the template, I see that the fields are not confused as I originally assumed in #851, but there is probably another factor that contributes to the strange message we hear. What could be an example of `integration_name`? Can it be not raw name as specified in the UI but contain dots or commas?

greatvovan:
In this example:
https://soundcloud.com/fun-with-no-end/grafana-oncall-sample-notification
The only variant to disassemble the message is seemingly this:
`integration_name` = Test web hook
`title` = Congratulations
`alert_count` = 1

But how to explain the value of `integration_name`? Can it come as "Test, web hook" or "Test. Web hook" by chance? How otherwise explain the pause between "test" and "web hook"?

Matvey-Kuk (approved):
Approve this change or here is my proposal:

"That's Grafana OnCall. Escalation (title) from integration (integration name) triggered (alert count) times."

greatvovan:
As per discussion in #851, a part of the problem is that `integration_name` is not really integration name, but (integration name + type). In my opinion, they should not be concatenated and must be passed into the template as independent fields, to have a freedom of composing the message.

greatvovan:
For example, I would like it to hear as "Alert via {integration_type} integration {integration_name}, title {title}, triggered {alert_count} times". Currently we do not have such a freedom. Also no accounting for singular/plural form of "time".

Templated example: "Alert via Prometheus integration Production cluster, title CPU usage, triggered 1 time".

Konstantinov-Innokentii:
Stale

PR #905, opened 2022-11-24
Do not include event repetitions terminated before start (until rrule)

matiasb:
Related to https://github.com/grafana/oncall-private/issues/1491


engine/apps/schedules/ical_events/adapter/amixr_recurring_ical_events_adapter.py
```diff
@@ -80,7 +80,8 @@ def add_event(event):
                 continue
             repetitions = self.RepeatedEvent(event, span_start)
             for repetition in repetitions:
-                if compare_greater(repetition.start, span_stop):
+                if compare_greater(repetition.start, span_stop) or compare_greater(repetition.start, repetition.stop):
```
iskhakov, line 83: this original lib received a lot of updates, please check if we can reuse it without rewriting

PR #914, opened 2022-11-29
don't show user tabs they aren't authorized to see

joeyorlando:
**What this PR does**:
Previously the top tab bar (eg. "Alert Groups", "Users", etc) would show user tabs that they weren't authorized to see. This PR will hide tabs that user's don't have access to.

In addition it:
- removes `RootBaseStore.isUserActionAllowed` as this was simply a reference to the `isUserActionAllowed` function from `utils/authorization`. When I wanted to access this method in `pages/index.ts` I needed to have an instance of a `RootBaseStore` which seemed awkward. Refactored all references to `RootBaseStore.isUserActionAllowed`.
- removed `NavMenuItem` from `pages/routes.tsx` and `RootBaseStore.navMenuItem` attribute. Neither of these seemed referenced in `oncall` and `oncall-private` repos. I wanted to remove `NavMenuItem` initially because I saw it made a reference to Admin/Editor/Viewer..

**Checklist**
- [x] Tests updated
- [x] Documentation added (N/A)
- [x] `CHANGELOG.md` updated (this is related to recent RBAC work,  no need to update)

### DIFF
--- a/grafana-plugin/src/pages/index.tsx
+++ b/grafana-plugin/src/pages/index.tsx
@@ -3,7 +3,7 @@ import { NavModelItem } from '@grafana/data';
 import { isTopNavbar } from 'plugin/GrafanaPluginRootPage.helpers';
 import { AppFeature } from 'state/features';
 import { RootBaseStore } from 'state/rootBaseStore';
-import { UserActions, UserAction } from 'utils/authorization';
+import { UserActions, UserAction, isUserActionAllowed } from 'utils/authorization';
 
 export const PLUGIN_URL_PATH = '/a/grafana-oncall-app';
 
@@ -124,7 +124,7 @@ export const pages: { [id: string]: PageDefinition } = [
     role: 'Admin',
     hideFromTabsFn: (store: RootBaseStore) => {
       const hasLiveSettings = store.hasFeature(AppFeature.LiveSettings);
-      return isTopNavbar() || window.grafanaBootData.user.orgRole !== 'Admin' || !hasLiveSettings;
+      return isTopNavbar() || !hasLiveSettings;
     },
     path: getPath('live-settings'),
     action: UserActions.OtherSettingsRead,
@@ -136,7 +136,7 @@ export const pages: { [id: string]: PageDefinition } = [
     role: 'Admin',
     hideFromTabsFn: (store: RootBaseStore) => {
       const hasCloudFeature = store.hasFeature(AppFeature.CloudConnection);
-      return isTopNavbar() || window.grafanaBootData.user.orgRole !== 'Admin' || !hasCloudFeature;
+      return isTopNavbar() || !hasCloudFeature;
     },
     path: getPath('cloud'),
     action: UserActions.OtherSettingsWrite,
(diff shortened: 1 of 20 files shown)


grafana-plugin/src/pages/index.tsx
```diff
@@ -136,7 +136,7 @@ export const pages: { [id: string]: PageDefinition } = [
     role: 'Admin',
     hideFromTabsFn: (store: RootBaseStore) => {
       const hasCloudFeature = store.hasFeature(AppFeature.CloudConnection);
-      return isTopNavbar() || window.grafanaBootData.user.orgRole !== 'Admin' || !hasCloudFeature;
```
joeyorlando, line 139: same here

grafana-plugin/src/pages/index.tsx
```diff
@@ -156,16 +156,18 @@ export const pages: { [id: string]: PageDefinition } = [
...
-    ...current,
-    getPageNav: () =>
-      ({
-        text: isTopNavbar() ? '' : current.text,
-        parentItem: current.parentItem,
-        hideFromBreadcrumbs: current.hideFromBreadcrumbs,
-        hideFromTabs: current.hideFromTabs,
-      } as NavModelItem),
-  };
+  if (!current.action || (current.action && isUserActionAllowed(current.action))) {
```
joeyorlando, line 159: this line is the main purpose of this PR

grafana-plugin/src/pages/routes.tsx
```diff
@@ -17,21 +14,6 @@ import LiveSettingsPage from 'pages/settings/tabs/LiveSettings/LiveSettingsPage'
...
-  page: string;
-  grafanaUser: {
-    orgRole: 'Viewer' | 'Editor' | 'Admin';
-  };
-  enableLiveSettings: boolean;
-  enableCloudPage: boolean;
-  enableNewSchedulesPage: boolean;
-  backendLicense: string;
-  onNavChanged: any;
-}
```
joeyorlando, line 33: I removed this because I saw a reference to `grafanaUser.orgRole`. I then searched in both `oncall` and `oncall-private` repos and saw no reference to `NavMenuItem`, therefore I think it is safe/makes sense to remove it.

grafana-plugin/src/state/rootBaseStore/index.ts
```diff
@@ -73,9 +72,6 @@ export class RootBaseStore {
   @observable
   onCallApiUrl: string;
 
-  @observable
-  navMenuItem: NavMenuItem;
```
joeyorlando, line 77: this was the only spot where `NavMenuItem` was referenced. However, `navMenuItem` is not referenced in either codebase, I think it is safe to remove?

grafana-plugin/src/state/rootBaseStore/index.ts
```diff
@@ -199,8 +195,6 @@ export class RootBaseStore {
     this.appLoading = false;
   }
 
-  isUserActionAllowed = isUserActionAllowed;
```
joeyorlando, line 202: no need to have this

teodosii (approved):

PR #923, opened 2022-11-30
Mobile app FCM support

vadimkerr:
**What this PR does**:
- Add ability to send mobile push notification using FCM
- Add `FCMRelayView` that allows sending push notifications from OSS instances

### DIFF
--- a/engine/apps/mobile_app/backend.py
+++ b/engine/apps/mobile_app/backend.py
@@ -1,6 +1,7 @@
 import json
 
 from django.conf import settings
+from push_notifications.models import APNSDevice, GCMDevice
 
 from apps.base.messaging import BaseMessagingBackend
 from apps.mobile_app.tasks import notify_user_async
@@ -32,6 +33,10 @@ class MobileAppBackend(BaseMessagingBackend):
 
         token = MobileAppAuthToken.objects.get(user=user)
         token.delete()
+
+        # delete push notification related info for user
+        APNSDevice.objects.filter(user=user).delete()
+        GCMDevice.objects.filter(user=user).delete()
 
     def serialize_user(self, user):
         from apps.mobile_app.models import MobileAppAuthToken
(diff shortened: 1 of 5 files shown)


engine/apps/mobile_app/backend.py
```diff
@@ -33,6 +34,10 @@ def unlink_user(self, user):
         token = MobileAppAuthToken.objects.get(user=user)
         token.delete()
 
+        # delete push notification related info for user
+        APNSDevice.objects.filter(user=user).delete()
```
vadimkerr, line 38: This is unrelated to FCM, this line just makes sure that when user is unlinked, their push notification info is deleted as well as the mobile auth token.

engine/apps/mobile_app/urls.py
```diff
@@ -1,13 +1,15 @@
...
 router = OptionalSlashRouter()
 
 router.register("apns", APNSDeviceAuthorizedViewSet, basename="apns")
-router.register("gcm", GCMDeviceAuthorizedViewSet, basename="gcm")
+router.register("fcm", FCMDeviceAuthorizedViewSet, basename="fcm")
 
 urlpatterns = [
     *router.urls,
     optional_slash_path("auth_token", MobileAppAuthTokenAPIView.as_view(), name="auth_token"),
+    optional_slash_path("fcm_relay", FCMRelayView.as_view(), name="fcm_relay"),
```
joeyorlando, line 14: should this route be disabled for OSS installations to avoid confusion?

joeyorlando (approved):

engine/apps/mobile_app/urls.py
vadimkerr: Good call! Added a new env var `FCM_RELAY_ENABLED` that defaults to False in https://github.com/grafana/oncall/pull/923/commits/25fae4933d7bfdafb78b4f8b279d13c733d7e56b

PR #937, opened 2022-12-02
helm chart v1.0.13 - update appVersion to v1.1.5 + change image.pullPolicy to "Always"

joeyorlando:
# What this PR does
Instructs helm deployments to always pull the latest `grafana/oncall` image.

This resolves some issues reported by the community where the plugin was not able to be configured. I was able to reproduce this locally by running 1.1.5 of the plugin in the UI, and `grafana/oncall:1.1.0` (the default that my helm deployment decided to use). With this version combination I was unable to successfully configure the plugin (likely due to some changes in [v1.1.3](https://github.com/grafana/oncall/releases/tag/v1.1.3)).

When I switched the engine image pull policy to "Always" and updated my helm deployment, this fixed the issue.

## Checklist

- [ ] Tests updated (N/A)
- [ ] Documentation added
- [x] `CHANGELOG.md` updated

### DIFF
--- a/helm/oncall/Chart.yaml
+++ b/helm/oncall/Chart.yaml
@@ -2,25 +2,23 @@ apiVersion: v2
 name: oncall
 description: Developer-friendly incident response with brilliant Slack integration
 
-
 type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 1.0.12
+version: 1.0.13
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
 # It is recommended to use it with quotes.
-appVersion: "v1.1.0"
+appVersion: "v1.1.5"
 dependencies:
   - name: cert-manager
     version: v1.8.0
     repository: https://charts.jetstack.io
     condition: cert-manager.enabled
-
   - name: mariadb
     version: 11.0.10
     repository: https://charts.bitnami.com/bitnami
--- a/helm/oncall/values.yaml
+++ b/helm/oncall/values.yaml
@@ -7,7 +7,7 @@ base_url: example.com
 image:
   # Grafana OnCall docker image repository
   repository: grafana/oncall
-  tag: latest
+  tag:
   pullPolicy: Always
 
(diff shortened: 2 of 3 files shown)


helm/oncall/values.yaml
```diff
@@ -7,8 +7,8 @@ base_url: example.com
 image:
   # Grafana OnCall docker image repository
   repository: grafana/oncall
-  tag:
-  pullPolicy: IfNotPresent
+  tag: latest
+  pullPolicy: Always
```
joeyorlando, line 11: this is the main change in this file.. rest is IDE autoformatter changes

helm/oncall/values.yaml
```diff
@@ -7,8 +7,8 @@ base_url: example.com
 image:
   # Grafana OnCall docker image repository
   repository: grafana/oncall
-  tag:
-  pullPolicy: IfNotPresent
+  tag: latest
```
vadimkerr, line 10: I wonder if it's possible to use `appVersion` from `Chart.yml` as the tag here? Would it make more sense?

joeyorlando: good call.

Actually I reverted `tag: latest` because I realized that the engine/celery worker container specs do this:
```yaml
image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
```

so by bumping the `appVersion` in this PR I'm effectively doing what you're mentioning

helm/oncall/Chart.yaml
```diff
@@ -2,25 +2,23 @@ apiVersion: v2
...
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 1.0.12
+version: 1.0.13
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
 # It is recommended to use it with quotes.
-appVersion: "v1.1.0"
+appVersion: "v1.1.5"
```
joeyorlando, line 16: bump to latest published image version

Matvey-Kuk:
It may lead to unexpected updates, we want helm to be stable and predictable & roll versions there with some delay

Matvey-Kuk (changes requested):
It may lead to unexpected updates, we want helm to be stable and predictable & roll versions there with some delay

Matvey-Kuk (approved):

vadimkerr (approved):

PR #947, opened 2022-12-05
Introduce org uuid

Konstantinov-Innokentii:
# What this PR does
-  Introduce org uuid to ensure we have some unique identifier between clusters for routing chatops requests
- Temporary remove autoconnect personal telegram account - it doesn't work since the verification code became too long. It will be restored later ( Approved by @Matvey-Kuk )


engine/apps/telegram/models/verification/channel.py
```diff
@@ -25,7 +25,7 @@ def is_active(self) -> bool:
 
     @property
     def uuid_with_org_id(self) -> str:
```
joeyorlando, line 27: nitpick; maybe rename this property to `uuid_with_org_uuid`?

engine/apps/telegram/models/verification/personal.py
```diff
@@ -23,7 +23,7 @@ def is_active(self) -> bool:
 
     @property
     def uuid_with_org_id(self) -> str:
```
joeyorlando, line 25: same here

engine/apps/telegram/renderers/keyboard.py
```diff
@@ -83,10 +83,10 @@ def _render_button(self, text: str, action: Action, action_data: Optional[Union[
         callback_data_args = [self.alert_group.pk, action.value]
         if action_data is not None:
             callback_data_args.append(action_data)
-        # Add org id with 'x-oncall-org-id' prefix to callback data.
+        # Add org id with 'oncall-uuid' prefix to callback data.
         # It's a workaroung to pass org_id to the oncall-gateway while proxying requests.
         # TODO: switch to json str instead of ':' separated string.
```
joeyorlando, line 88: nitpick; is this last TODO here still relevant?

engine/common/oncall_gateway/oncall_gateway_client.py
```diff
@@ -71,8 +71,8 @@ def post_slack_connector(self, slack_id: str, backend: str) -> tuple[SlackConnec
         response = self._post(url=self._slack_connectors_url, json=d)
         response_data = response.json()
         return (
-            OnCallConnector(
-                response_data["oncall_org_id"],
```
joeyorlando, line 75: what is this change related to?

engine/apps/telegram/updates/update_handlers/button_press.py
```diff
@@ -71,10 +71,16 @@ def _get_action_context(data: str) -> ActionContext:
         action_name = args[1]
         action = Action(action_name)
 
-        action_data = args[2] if len(args) >= 3 and not args[2].startswith("x-oncall-org-id") else None
+        action_data = args[2] if len(args) >= 3 and not args[2].startswith("oncall-uuid") else None
```
joeyorlando, line 74:
```suggestion
        action_data = args[2] if len(args) >= 3 and not cls._is_oncall_identifier(args[2]) else None
```

joeyorlando:
> Temporary remove autoconnect personal telegram account - it doesn't work since the verification code became too long. It will be restored later

could we add the org uuid as a request header instead of in the verification code itself?

engine/common/oncall_gateway/oncall_gateway_client.py
Konstantinov-Innokentii: It's just a small bugfix related to the interaction betwenn oncall and oncall-gateway

Konstantinov-Innokentii:
> > Temporary remove autoconnect personal telegram account - it doesn't work since the verification code became too long. It will be restored later
> 
> could we add the org uuid as a request header instead of in the verification code itself?

When doing autoconnect oncall redirects user to telegram with code in query params and when you open telegram it uses code from query param. I think we can not control this telegram behaviour

joeyorlando (approved):

PR #956, opened 2022-12-06
Mobile app screen tweaks

teodosii:
# What this PR does
- Added auto refresh on QR (every 50s) + blur/loading display while we fetch a new QR

[#1508](https://github.com/grafana/oncall-private/issues/1508)

### DIFF
--- a/grafana-plugin/src/containers/MobileAppVerification/MobileAppVerification.test.tsx
+++ b/grafana-plugin/src/containers/MobileAppVerification/MobileAppVerification.test.tsx
@@ -207,9 +207,12 @@ describe('MobileAppVerification', () => {
 
     render(<MobileAppVerification userPk={USER_PK} />);
 
-    await waitFor(() => {
-      expect(loadUserMock).toHaveBeenCalled();
-    });
+    await waitFor(
+      () => {
+        expect(loadUserMock).toHaveBeenCalled();
+      },
+      { timeout: 6000 }
+    );
   });
 
   test('it polls loadUser after disconnect', async () => {
@@ -231,8 +234,11 @@ describe('MobileAppVerification', () => {
     await user.click(button); // click the disconnect button, which opens the modal
     await user.click(screen.getByText('Remove')); // click the confirm button within the modal, which actually triggers the callback
 
-    await waitFor(() => {
-      expect(loadUserMock).toHaveBeenCalled();
-    });
+    await waitFor(
+      () => {
+        expect(loadUserMock).toHaveBeenCalled();
+      },
+      { timeout: 6000 }
+    );
   });
 });
--- a/grafana-plugin/src/containers/MobileAppVerification/MobileAppVerification.test.tsx
+++ b/grafana-plugin/src/containers/MobileAppVerification/MobileAppVerification.test.tsx
@@ -208,7 +208,7 @@ describe('MobileAppVerification', () => {
     render(<MobileAppVerification userPk={USER_PK} />);
 
     await waitFor(() => {
-      expect(loadUserMock).toHaveBeenCalledTimes(1);
+      expect(loadUserMock).toHaveBeenCalled();
     });
   });
 
@@ -232,7 +232,7 @@ describe('MobileAppVerification', () => {
     await user.click(screen.getByText('Remove')); // click the confirm button within the modal, which actually triggers the callback
 
     await waitFor(() => {
-      expect(loadUserMock).toHaveBeenCalledTimes(1);
+      expect(loadUserMock).toHaveBeenCalled();
     });
   });
 });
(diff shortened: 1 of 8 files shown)


grafana-plugin/src/containers/MobileAppVerification/MobileAppVerification.test.tsx
```diff
@@ -231,8 +234,11 @@ describe('MobileAppVerification', () => {
...
     await user.click(screen.getByText('Remove')); // click the confirm button within the modal, which actually triggers the callback
 
-    await waitFor(() => {
-      expect(loadUserMock).toHaveBeenCalledTimes(1);
-    });
+    await waitFor(
+      () => {
+        expect(loadUserMock).toHaveBeenCalled();
+      },
+      { timeout: 6000 }
```
joeyorlando, line 241: these tests take > 1s (default value of `timeout`)?

grafana-plugin/src/containers/MobileAppVerification/MobileAppVerification.tsx
```diff
@@ -138,17 +150,40 @@ const MobileAppVerification = observer(({ userPk }: Props) => {
     </div>
   );
 
+  function clearTimeouts(): void {
+    clearTimeout(userTimeoutId);
+    clearTimeout(refreshTimeoutId);
+  }
+
+  function triggerTimeouts(): void {
+    setTimeout(() => queueRefreshQR(), INTERVAL_QUEUE_QR);
```
joeyorlando, line 159:
```suggestion
    setTimeout(queueRefreshQR, INTERVAL_QUEUE_QR);
```

grafana-plugin/src/containers/MobileAppVerification/MobileAppVerification.tsx
```diff
@@ -138,17 +150,40 @@ const MobileAppVerification = observer(({ userPk }: Props) => {
     </div>
   );
 
+  function clearTimeouts(): void {
```
joeyorlando, line 153: total nitpick; do we have a guideline/decision in terms of `function` vs arrow function? It could be nice to setup an `eslint` rule like [this](https://github.com/TristonJ/eslint-plugin-prefer-arrow) just to keep things consistent.

grafana-plugin/src/containers/MobileAppVerification/MobileAppVerification.tsx
```diff
@@ -116,13 +127,14 @@ const MobileAppVerification = observer(({ userPk }: Props) => {
           Sign In
         </Text>
         <Text type="primary">Open Grafana IRM mobile application and scan this code to sync it with your account.</Text>
-        <div className="u-width-100 u-flex u-flex-center">
-          <QRCode value={QRCodeValue} />
+        <div className={cx('u-width-100', 'u-flex', 'u-flex-center', 'u-position-relative')}>
+          <QRCode className={cx({ blurry: isQRBlurry })} value={QRCodeValue} />
+          {isQRBlurry && (
```
joeyorlando, line 132: what is the blur intended to do?

locally I just get a very fast flicker. only if I throttle my Chrome's network activity to "Slow 3G" do I see the blurred/loading state 🤔

joeyorlando (commented):
Overall LGTM, just had some questions regarding the blur state.

grafana-plugin/src/containers/MobileAppVerification/MobileAppVerification.tsx
joeyorlando: again a nitpick comment, but I would argue in favour of enabling an eslint rule like [`no-use-before-define`](https://eslint.org/docs/latest/rules/no-use-before-define) (this is likely just because I am coming from mostly a Python/clojure background where using a function before it is defined would not compile, but I've also not seen this done very often in JS/TS)

Ukochka (approved):

grafana-plugin/src/containers/MobileAppVerification/MobileAppVerification.tsx
teodosii: Function hoisting is part of the JS environment so they're pretty normal, it's not an anti pattern, but I understand your point. The idea is that if you navigate to a component, 9 out of 10 times you do it to see what's inside the render part, and each time you go to a file you need to scroll to that part. If you move out the helpers (mostly arrow functions from your functional component) after the render part, you'll eliminate this hassle and there's no need for infinite scrolling to find what you're looking for, because most of the times you want to see how props are being passed, or what classes are in there, or whatever DOM structure is there - and that's inside the render. Very rarely you will go to a file to actually change the logic (that's like at most 2-3 out of 10 cases most of the time)

I also find it a bit more cleaner to do it this way, because each function is a separate block, while with arrow functions you just add one after another basically in the same block, but functions would be a bit different because they're hoisted and they make the code look slightly more organized.

teodosii: One other thing we could do is to declare them outside the functional component, but then you'd need to pass in all the required data.

PR #970, opened 2022-12-09
hide mobile app verification tab in the user settings modal for unauthed users

joeyorlando:
# What this PR does
We currently show the tab to all users. However, the [`unlink_backend` endpoint](https://github.com/grafana/oncall/blob/dev/engine/apps/api/views/user.py#L128) requires the `USER_SETTINGS_WRITE` RBAC permission (or "Editor" role when RBAC is disabled). If a viewer uses the QR code to link their mobile app, and then clicks disconnect, they will get an unauthorized error.

Therefore we should just do a quick check that the current user is authorized to see this, to conditionally display this tab.

Tested locally w/ an Admin and Viewer user and is now working as intended.

## Checklist

- [ ] Tests updated (N/A)
- [x] `CHANGELOG.md` updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,16 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## v1.2.1 (TBD)
-
-### Added
-
-### Fixed
-
-- Mobile App Verification tab in the user settings modal is now hidden for users that do not have proper
-  permissions to use it
-
-## v1.2.0 (TBD)
+## v1.2.0 (2022-12-12)
 
 ### Added
 
@@ -30,6 +21,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Got 500 error when saving Outgoing Webhook ([#890](https://github.com/grafana/oncall/issues/890))
 - v1.0.13 helm chart - update the OnCall backend pods image pull policy to "Always" (and explicitly set tag to `latest`).
   This should resolve some recent issues experienced where the frontend/backend versions are not aligned.
+- Mobile App Verification tab in the user settings modal is now hidden for users that do not have proper
+  permissions to use it
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


grafana-plugin/src/pages/users/Users.tsx
```diff
@@ -162,8 +157,6 @@ class Users extends React.Component<UsersProps, UsersState> {
       });
 
     const { count, results } = userStore.getSearchResult();
-    const showMobileAppScreen: boolean =
-      id !== undefined && id !== 'me' && id === userStore.currentUserPk && store.hasFeature(AppFeature.MobileApp);
```
joeyorlando, line 166: this `id !== 'me'` check would hide the modal for me. When I would go to the user's page, then click "View my profile", this would append an `id=me` query parameter, thereby making `showMobileAppScreen` false.

In addition, I decided to move the `showMobileAppScreen` logic to the `UserSettings` component since we are already passing `userPkToEdit` down to this component

CHANGELOG.md
```diff
@@ -5,6 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## v1.2.1 (TBD)
```
Matvey-Kuk, line 8: Isn't it a mistake? Latest is 1.1.6

Matvey-Kuk (changes requested):

CHANGELOG.md
joeyorlando: (check a few lines below)

I believe @Konstantinov-Innokentii merged `dev` -> `main` earlier today, but hasn't deployed/released these changes (which would be `v1.2.0`). I suspect this commit won't make it into `v1.2.0`?

teodosii (approved):

grafana-plugin/src/pages/users/Users.tsx
teodosii: Yeah, that's a good point. Hmm I didn't notice the issue of `id !== 'me'`. Also I think it's best that you moved it back there, I didn't spot the `isCurrent` which helps in this case. :+1:

CHANGELOG.md
joeyorlando: or I can include it under the changes for `v1.2.0` and we'll just need to remember to do another `dev` -> `main` merge before doing a full release

joeyorlando: decided to just include it under `v1.2.0`, I'll be sure to do another `dev` -> `main` merge before releasing on Monday morning

PR #989, opened 2022-12-14
modify push notification settings + use fcm-django library

joeyorlando:
# What this PR does

- swaps out `django-push-notifications` for [`fcm-django`](https://github.com/grafana/fcm-django). Again.. this is a fork of the parent repo for exactly the same reason.. the migrations point to `auth_user` without letting us use our own user model, this has been patched in the `grafana` fork. The reason why we are using `fcm-django` vs `django-push-notifications` is that the latter does not support the new FCM API, only the "legacy" API. The legacy FCM API does not support certain push notification settings that we would like to use.
- modifies the iOS/Android specific push notification settings
- adds a `flower` pod in the `docker-compose-developer.yml`, useful for debugging tasks locally
- sets the mobile app verification token TTL to 5 minutes when developing locally. The default of 1 minute makes working with device emulators really tricky..

## Other stuff

This PR also swaps out the base image in `engine/Dockerfile` from `python:3.9-alpine3.16` to `python:3.9-slim-buster`.

As to why.. in short, with the introduction of the `fcm-django` library there is now a peer-dependency on [`grpcio`](https://github.com/grpc/grpc) (which is used by `firebase_admin`.. which I am using in this PR to interact directly with Firebase Cloud Messaging (FCM)). `grpcio` does not publish wheels (read: compiled binaries) for the Alpine distro. It does publish wheels for Debian and hence `pip install -r requirements.txt` does not need to build this library from the source distribution.

This is a [known "issue"](https://github.com/grpc/grpc/issues/22815#issuecomment-1107874367) and the recommended solution in the community is to.. not use alpine.

These were the numbers, when building the image locally, in terms of image size and build time:

|                                             | Local image size (uncompressed  | Build time |
| ------------------------- | -------------------------------------- | ---------- |
| `python:3.9-alpine3.16`   | 785MB  | 320s |
| `python:3.9-slim-buster` | 1.05GB  | 90s   |

### DIFF
--- a/.github/workflows/helm_tests.yml
+++ b/.github/workflows/helm_tests.yml
@@ -10,10 +10,10 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
 
-      - name: Set up Docker Buildx # We need this step for docker caching
+      - name: Set up Docker Buildx  # We need this step for docker caching
         uses: docker/setup-buildx-action@v2
 
-      - name: Build docker image locally # using github actions docker cache
+      - name: Build docker image locally  # using github actions docker cache
         uses: docker/build-push-action@v2
         with:
           context: ./engine
@@ -40,7 +40,7 @@ jobs:
         with:
           workloads: "" # all
           namespace: "" # default
-          timeout: 1000
+          timeout: 300
           max-restarts: 0
 
       - name: Bootstrap organization and integration
--- a/.github/workflows/helm_tests.yml
+++ b/.github/workflows/helm_tests.yml
@@ -10,10 +10,10 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
 
-      - name: Set up Docker Buildx  # We need this step for docker caching
+      - name: Set up Docker Buildx # We need this step for docker caching
         uses: docker/setup-buildx-action@v2
 
-      - name: Build docker image locally  # using github actions docker cache
+      - name: Build docker image locally # using github actions docker cache
         uses: docker/build-push-action@v2
         with:
           context: ./engine
@@ -40,7 +40,7 @@ jobs:
         with:
           workloads: "" # all
           namespace: "" # default
-          timeout: 300
+          timeout: 1000
           max-restarts: 0
 
       - name: Bootstrap organization and integration
--- a/engine/apps/mobile_app/tasks.py
+++ b/engine/apps/mobile_app/tasks.py
@@ -52,32 +52,60 @@ def notify_user_async(user_pk, alert_group_pk, notification_policy_pk, critical)
 
     message = get_push_notification_message(alert_group)
     thread_id = f"{alert_group.channel.organization.public_primary_key}:{alert_group.public_primary_key}"
-
-    if critical:
-        aps = {
-            "alert": f"Critical page: {message}",
-            "interruption-level": "critical",
-            "sound": "ambulance.aiff",
-        }
-    else:
-        aps = {
-            "alert": message,
-            "sound": "bingbong.aiff",
-        }
+    alert_title = f"Critical page: {message}" if critical else message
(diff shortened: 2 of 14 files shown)


imtoori (approved):

engine/apps/mobile_app/tasks.py
```diff
@@ -47,38 +58,76 @@ def notify_user_async(user_pk, alert_group_pk, notification_policy_pk, critical)
...
-            "alert": f"Critical page: {message}",
-            "interruption-level": "critical",
-            "sound": "ambulance.aiff",
-        }
-    else:
-        aps = {
-            "alert": message,
-            "sound": "bingbong.aiff",
-        }
+    # TODO: we should update this to check if FCM_RELAY is set and conditionally make a call here..
```
joeyorlando, line 69: right now the `fcm_relay` endpoint won't work properly, but we are not using this yet. Let's come back in a follow-up PR and patch this.

joeyorlando:
note: once this is deployed we should drop the tables associated with the `push_notifications` django app as they're no longer needed

.github/workflows/helm_tests.yml
```diff
@@ -40,7 +40,7 @@ jobs:
         with:
           workloads: "" # all
           namespace: "" # default
-          timeout: 300
+          timeout: 1000
```
joeyorlando, line 43: for whatever reason, installing `grpcio` takes forever so I needed to increase this timeout.. (`grpcio` is used by the `firebase_admin` package)

engine/requirements.txt
```diff
@@ -33,7 +33,7 @@ django-log-request-id==1.6.0
 django-polymorphic==3.0.0
 django-rest-polymorphic==0.1.9
 pre-commit==2.15.0
-https://github.com/grafana/django-push-notifications/archive/refs/tags/3.0.0-fix-migration.tar.gz
+https://github.com/grafana/fcm-django/archive/refs/tags/v1.0.12r1.tar.gz
```
joeyorlando, line 36: I again needed to make a fork because `fcm-django` had the same issue as `django-push-notifications` in that it wouldn't properly allow specifying the foreign key relationship for the `user_id` column (would automatically set this to `auth_user`.. which isn't appropriate here). The forked version has this issue patched.

engine/apps/mobile_app/views.py
```diff
@@ -1,25 +1,15 @@
...
 
 
-class FCMDeviceAuthorizedViewSet(GCMDeviceAuthorizedViewSet):
-    class FCMDeviceSerializer(GCMDeviceSerializer):
-        """
-        GCMDevice has cloud_message_type equal to "GCM" by default, in this serializer cloud_message_type is always set
-        to "FCM" no matter what was provided in the request.
-        """
-
-        cloud_message_type = HiddenField(default="FCM")
```
joeyorlando, line 19: this is no longer necessary because `fcm_dango` only works with.. FCM and thereby doesn't have a `cloud_message_type` field

engine/apps/mobile_app/tasks.py
```diff
@@ -47,38 +58,76 @@ def notify_user_async(user_pk, alert_group_pk, notification_policy_pk, critical)
...
 
-    extra = {
-        "orgId": alert_group.channel.organization.public_primary_key,
-        "orgName": alert_group.channel.organization.stack_slug,
-        "alertGroupId": alert_group.public_primary_key,
-        "status": alert_group.status,
-        "aps": aps,
-    }
+    message = Message(
```
joeyorlando, line 71: this was the main focus of this PR, adjusting the push notification settings to include some iOS and Android specific things. `django-push-notifications` didn't allow this because it was using the "legacy" FCM API whereas `fcm-django` uses the latest API, which supports these features.

PR #998, opened 2022-12-15
modify push notification settings + use fcm-django library

joeyorlando:
# What this PR does

- swaps out `django-push-notifications` for [`fcm-django`](https://github.com/grafana/fcm-django). Again.. this is a fork of the parent repo for exactly the same reason.. the migrations point to `auth_user` without letting us use our own user model, this has been patched in the `grafana` fork. The reason why we are using `fcm-django` vs `django-push-notifications` is that the latter does not support the new FCM API, only the "legacy" API. The legacy FCM API does not support certain push notification settings that we would like to use.
- modifies the iOS/Android specific push notification settings
- adds a `flower` pod in the `docker-compose-developer.yml`, useful for debugging tasks locally
- sets the mobile app verification token TTL to 5 minutes when developing locally. The default of 1 minute makes working with device emulators really tricky..

## Other stuff

This PR also swaps out the base image in `engine/Dockerfile` from `python:3.9-alpine3.16` to `python:3.9-slim-buster`.

As to why.. in short, with the introduction of the `fcm-django` library there is now a peer-dependency on [`grpcio`](https://github.com/grpc/grpc) (which is used by `firebase_admin`.. which I am using in this PR to interact directly with Firebase Cloud Messaging (FCM)). `grpcio` does not publish wheels (read: compiled binaries) for the Alpine distro. It does publish wheels for Debian and hence `pip install -r requirements.txt` does not need to build this library from the source distribution.

This is a [known "issue"](https://github.com/grpc/grpc/issues/22815#issuecomment-1107874367) and the recommended solution in the community is to.. not use alpine.

These were the numbers, when building the image locally, in terms of image size and build time:

|                                             | Local image size (uncompressed  | Build time (may differ based on your network speed) |
| ------------------------- | -------------------------------------- | ---------- |
| `python:3.9-alpine3.16`   | 785MB  | 320s |
| `python:3.9-slim-buster` | 1.05GB  | 90s   |

### DIFF
--- a/engine/apps/mobile_app/fcm_relay.py
+++ b/engine/apps/mobile_app/fcm_relay.py
@@ -1,4 +1,5 @@
-from push_notifications.gcm import send_message
+# from firebase_admin.messaging import Message
+# from fcm_django.models import FCMDevice
 from rest_framework import status
 from rest_framework.response import Response
 from rest_framework.views import APIView
@@ -6,6 +7,7 @@ from rest_framework.views import APIView
 REQUIRED_FIELDS = {"registration_ids", "notification", "data"}
 
 
+# TODO: update thie
 class FCMRelayView(APIView):
     def post(self, request):
         """
@@ -17,10 +19,11 @@ class FCMRelayView(APIView):
         if not REQUIRED_FIELDS.issubset(request.data.keys()):
             return Response(status=status.HTTP_400_BAD_REQUEST)
 
-        registration_ids = request.data["registration_ids"]
-        data = {
-            **request.data["data"],
-            **request.data["notification"],
-        }
+        # registration_ids = request.data["registration_ids"]
+        # data = {
+        #     **request.data["data"],
+        #     **request.data["notification"],
+        # }
 
-        return send_message(registration_ids=registration_ids, data=data, cloud_type="FCM")
+        # return FCMDevice.objects.send_message(Message(), False, ["registration_ids"])
+        return "TODO:"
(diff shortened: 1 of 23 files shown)


joeyorlando:
opened a new PR because I _think_ the caching for the old PR was causing the helm e2e tests to fail (despite them passing locally)

iskhakov:
> swaps out django-push-notifications for [fcm-django](https://github.com/grafana/fcm-django). Again.. this is a fork of the parent repo for exactly the same reason.. the migrations point to auth_user without letting us use our own user model, this has been patched in the grafana fork. The reason why we are using fcm-django vs django-push-notifications is that the latter does not support the new FCM API, only the "legacy" API. The legacy FCM API does not support certain push notification settings that we would like to use.

I agree with that, fcm-django is the fork, so we can return to the original lbrary

engine/apps/mobile_app/fcm_relay.py
```diff
@@ -1,11 +1,13 @@
...
+# from firebase_admin.messaging import Message
+# from fcm_django.models import FCMDevice
 from rest_framework import status
 from rest_framework.response import Response
 from rest_framework.views import APIView
 
 REQUIRED_FIELDS = {"registration_ids", "notification", "data"}
 
 
+# TODO: update thie
```
iskhakov, line 10: we need more explanation here

iskhakov (approved):

imtoori:
A comment on my changes

I removed the `notification` payload from the android field because we're gonna use `data-only` messages so that we can execute some dart code to change the DND (Do Not Disturb) and volume settings before showing a critical notification.

For more information on data-only messages https://firebase.google.com/docs/cloud-messaging/concept-options

imtoori (approved):

PR #1007, opened 2022-12-15
[pagerduty-migrator] If there is a schedule, continue with the process

gilbertobr:
The code is breaking, with some duplicate schedules.
Note: Not all

To fix I added an elif

image

### DIFF
--- a/tools/pagerduty-migrator/migrator/oncall_api_client.py
+++ b/tools/pagerduty-migrator/migrator/oncall_api_client.py
@@ -21,6 +21,8 @@ def api_call(method: str, path: str, **kwargs) -> requests.Response:
             cooldown_seconds = int(e.response.headers["Retry-After"])
             sleep(cooldown_seconds)
             return api_call(method, path, **kwargs)
+        elif e.response.status_code == 400:
+            pass
         else:
             raise
 


tools/pagerduty-migrator/migrator/oncall_api_client.py
```diff
@@ -21,6 +21,8 @@ def api_call(method: str, path: str, **kwargs) -> requests.Response:
             cooldown_seconds = int(e.response.headers["Retry-After"])
             sleep(cooldown_seconds)
             return api_call(method, path, **kwargs)
+        elif e.response.status_code == 400:
+            pass
```
matiasb, line 25: I think it would be better to handle this at the schedule level instead, being more specific/explicit (here it seems like you will be passing on any 400 API response, which may not be what we want).

gilbertobr: The way it is today when it already exists, it has broken the code.
My opinion would be.

If it already exists, skip to the next one.

vadimkerr:
Hey @gilbertobr 👋 
I managed to reproduce the error, this happens due to some case-sensitivity issue and uniqueness of schedule names.
E.g. this happens when there's a schedule named "test" in Grafana OnCall and a schedule named "TEST" in PagerDuty, when trying to migrate. I'll take a closer look at this next week, I think we should solve the root cause of the issue instead of ignoring 400s here.
Thank you for opening the PR!

vadimkerr:
The cause of this issue should be fixed by https://github.com/grafana/oncall/pull/1320 and https://github.com/grafana/oncall/pull/1327, so closing this one. 
@gilbertobr feel free to open an issue if you face this again 🙂

PR #1013, opened 2022-12-16
Grafana Faro integration

teodosii:
# What this PR does

- Added Faro integration
- Added support for `.env` files on frontend

# How to test

- Set up `grafana-agent` to pass data to `Loki`, `Tempo` and `Prometheus`
- Create your `.env` file in `grafana-plugin` - pass in `FARO_URL` and `FARO_API_KEY` as defined in grafana-agent config
 
```
FARO_URL=http://localhost:12345/collect
FARO_API_KEY=secret
FARO_ENABLED=true
```
- Run `grafana-agent`
- Run `yarn watch`
- Go to grafana instance and see logs/traces in the matching data sources

Example of grafana agent config file

```
logs:
  configs:
  - clients:
    - basic_auth:
        password: 
        username: 343538
      url: https://logs-prod-017.grafana.net/loki/api/v1/push
    name: integrations
    positions:
      filename: /tmp/positions.yaml
    scrape_configs:
    - job_name: integrations/agent
      journal:
        labels:
          instance: <hostname>
          job: integrations/agent
        max_age: 24h
      pipeline_stages:
      - match:
          action: drop
          drop_counter_reason: only_keep_grafana_agent_logs
          selector: '{unit!="grafana-agent.service"}'
      - regex:
          expression: (level=(?P<log_level>[\s]*debug|warn|info|error))
      - labels:
          level: log_level
      relabel_configs:
      - source_labels:
        - __journal__systemd_unit
        target_label: unit
    - job_name: integrations/agent
      pipeline_stages:
      - json:
          expressions:
            message: message
      - regex:
          expression: (level=(?P<log_level>[\s]*debug|warn|info|error))
      - labels:
          level: log_level
      - output:
          source: message
      relabel_configs:
      - source_labels:
        - computer
        target_label: instance
      - replacement: grafana-agent.service
        target_label: unit
      windows_events:
        bookmark_path: ./bookmark-application.xml
        eventlog_name: Application
        labels:
          job: integrations/agent
        use_incoming_timestamp: true
        xpath_query: '*[System[Provider[@Name=''Grafana Agent'']]]'
    target_config:
      sync_period: 10s
metrics:
  configs:
  - name: integrations
    remote_write:
    - basic_auth:
        password: 
        username: 689135
      url: https://prometheus-us-central1.grafana.net/api/prom/push
  global:
    scrape_interval: 60s
  wal_directory: /tmp/grafana-agent-wal
traces:
  configs:
    - name: integrations
      remote_write:
        - endpoint: tempo-us-central1.grafana.net:443
          basic_auth:
            username: 340051
            password: 
      receivers:
        otlp:
          protocols:
            grpc:
            http:
              cors:
                allowed_origins:
                  - http://localhost:12345
                  - http://localhost:3000
                max_age: 7200
integrations:
  app_agent_receiver:
      autoscrape:
        enable: true
        metrics_instance: integrations
      instance: 'frontend'
      logs_instance: integrations
      traces_instance: integrations
      server:
        host: 0.0.0.0
        port: 12345
        api_key: 'secret'
        cors_allowed_origins:
          - http://localhost:3000
      logs_labels:
        app: frontend
        kind:
      logs_send_timeout: 5000
      sourcemaps:
        download: true
```


maskin25 (approved):

grafana-plugin/package.json
```diff
@@ -40,25 +40,7 @@
...
-    "@babel/plugin-proposal-optional-chaining": "^7.18.9",
-    "@babel/plugin-syntax-decorators": "^7.18.6",
-    "@babel/plugin-syntax-dynamic-import": "^7.8.3",
-    "@babel/plugin-transform-destructuring": "^7.20.0",
-    "@babel/plugin-transform-react-constant-elements": "^7.18.12",
-    "@babel/plugin-transform-runtime": "^7.19.6",
-    "@babel/plugin-transform-typescript": "^7.18.12",
-    "@babel/preset-env": "^7.18.10",
-    "@babel/preset-react": "^7.18.6",
-    "@babel/preset-typescript": "^7.18.6",
```
joeyorlando, line 56: why the move to `dependencies`?

grafana-plugin/src/network/index.ts
```diff
@@ -34,14 +38,63 @@ export const makeRequest = async <RT = any>(path: string, config: RequestConfig)
...
+    let span = otel.trace.getActiveSpan();
+
+    if (!span) {
+      span = tracer.startSpan('http-request');
+      span.setAttribute('page_url', document.URL.split('//')[1]);
+      span.setAttribute(SemanticAttributes.HTTP_URL, url);
+      span.setAttribute(SemanticAttributes.HTTP_METHOD, method);
+    }
+
+    return new Promise<RT>((resolve, reject) => {
```
joeyorlando, line 54: would it be possible to de-duplicate this a bit? The two `try/catch` clauses are nearly identical and there's quite a lot of duplication. Also, any reason why to mix and match `async/await` and `Promise` in the same function?

joeyorlando:
so atm this will only work locally, correct?

joeyorlando (changes requested):
I think the PR may be missing changes to the Drone dev pipeline, specifically providing these three env vars at plugin build time. Is including `FARO_API_KEY` in the built plugin not a security concern?

grafana-plugin/src/utils/faro.ts
```diff
@@ -0,0 +1,71 @@
...
+import { initializeFaro, getWebInstrumentations } from '@grafana/faro-web-sdk';
+import { TracingInstrumentation } from '@grafana/faro-web-tracing';
+import { DocumentLoadInstrumentation } from '@opentelemetry/instrumentation-document-load';
+import { FetchInstrumentation } from '@opentelemetry/instrumentation-fetch';
+import { UserInteractionInstrumentation } from '@opentelemetry/instrumentation-user-interaction';
+import { XMLHttpRequestInstrumentation } from '@opentelemetry/instrumentation-xml-http-request';
+
+import plugin from '../../package.json'; // eslint-disable-line
+
+const IGNORE_URLS = [/^((?!\/{0,1}a\/grafana\-oncall\-app\\).)*$/];
```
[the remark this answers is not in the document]
teodosii: It's a bug on faro, please see faro channel where I poster this. It needs the regex

grafana-plugin/webpack.config.js
```diff
@@ -117,6 +120,13 @@ module.exports.getWebpackConfig = (config, options) => {
             'sass-loader',
           ],
         },
+
+        {
+          test: /\.m?js/,
```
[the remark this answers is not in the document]
teodosii: It was required by faro because they are using modules and not CommonJS pattern for modules. I believe this got fixed in beta5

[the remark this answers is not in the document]
teodosii: It can be removed now but it was required in version beta4 of faro

grafana-plugin/package.json
teodosii: Because they should be treated as dependencies and not dev dependencies. This is not related to thisPR but thought it will be good idea to do move them where they belong

teodosii: There's no difference in the way they are being used right now unless you pass special flags to yarn

grafana-plugin/src/network/index.ts
teodosii: I'll change it so that it's using just async/await, but I will leave it as 2 separate blocks for now because the first one relies on OTEL implementation and actually calls methods from OTEL, while the other one is just a try catch with log. It will also benefit to have them separate in case we want to add more span attributes on the open telemetry side.

grafana-plugin/package.json
teodosii: oh snap. I moved more than required - fixed now

joeyorlando (commented):
Overall LGTM. We discussed the env vars. Still need to discuss plan for building the plugin in different envs.

iskhakov:
we need to differentiate FARO_ENV for different envs, in the call with Maxim we decided to calculate env based on jsonData.oncallApiUrl
Example:
```
if oncallApiUrl == "https://a-prod-us-central-0.grafana.net"
    FARO_ENV = "prod"
elif oncallApiUrl == "https://a-ops-us-east-0.grafana.net"
    FARO_ENV = "ops"
if oncallApiUrl == "https://a-dev-us-central-0.grafana.net"
    FARO_ENV = "dev"
else:
   # This opensource, don't send traces
   FARO_ENABLED=false
```

joeyorlando:
> we need to differentiate FARO_ENV for different envs, in the call with Maxim we decided to calculate env based on jsonData.oncallApiUrl Example:
> 
> ```
> if oncallApiUrl == "https://oncall-prod-us-central-0.grafana.net"
>     FARO_ENV = "prod"
> elif oncallApiUrl == "https://oncall-ops-us-east-0.grafana.net"
>     FARO_ENV = "ops"
> if oncallApiUrl == "https://oncall-dev-us-central-0.grafana.net"
>     FARO_ENV = "dev"
> else:
>    # This opensource, don't send traces
>    FARO_ENABLED=false
> ```

what about for different regions? (ie. `oncallApiUrl == "https://oncall-prod-eu-west-0.grafana.net"`)

iskhakov:
right now we'll need to keep adding them manually, and add the correct label `environment` to jsonData in the future

maskin25:
@iskhakov @joeyorlando what about fuzzy search like:

```
if oncallApiUrl.includes('oncall-prod') 
      FARO_ENV = "prod"
```

joeyorlando:
@maskin25, @iskhakov and I discussed potentially building different plugin `.zip` files for the various environments, and including a different `process.env.FARO_ENV` variable into the build.

IMO I think environment variables make more sense in the long-term. If the OnCall API URL changes (probably unlikely) we'd have to remember to update these `if` statements.

maskin25:
right now we need  to merge this to start collection metrics, and you are absolutely right, we have to implement more reliable solution soon, it seems our deployment workflow is not ready for different zip's for different environments yet

iskhakov:
I think we should enable Faro integration if backend url is `<subdomain>.grafana.net` and set `environment=subdomain`, instead of dev, ops and prod

PR #1014, opened 2022-12-16
oncall_schedule with value None

gilbertobr:
When there is a migration of plans in pagerduty from **Business** to **Professional**, this type of error occurs.

Line https://github.com/grafana/oncall/blob/275aa94c71b18c6b14c9037247234ff76145ba39/tools/pagerduty-migrator/migrator/resources/escalation_policies.py#L95
image

### DIFF
--- a/tools/pagerduty-migrator/migrator/resources/escalation_policies.py
+++ b/tools/pagerduty-migrator/migrator/resources/escalation_policies.py
@@ -92,14 +92,15 @@ def transform_rule(
         if schedule is None:
             continue
 
-        oncall_schedule_id = schedule["oncall_schedule"]["id"]
+        if schedule["oncall_schedule"] is not None:
+            oncall_schedule_id = schedule["oncall_schedule"]["id"]
 
-        escalation_policy = {
-            "escalation_chain_id": escalation_chain_id,
-            "type": "notify_on_call_from_schedule",
-            "notify_on_call_from_schedule": oncall_schedule_id,
-        }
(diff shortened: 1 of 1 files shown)


tools/pagerduty-migrator/migrator/resources/escalation_policies.py
```diff
@@ -92,14 +92,15 @@ def transform_rule(
...
-        oncall_schedule_id = schedule["oncall_schedule"]["id"]
-
-        escalation_policy = {
-            "escalation_chain_id": escalation_chain_id,
-            "type": "notify_on_call_from_schedule",
-            "notify_on_call_from_schedule": oncall_schedule_id,
-        }
-        escalation_policies.append(escalation_policy)
+        if schedule["oncall_schedule"] is not None:
+            oncall_schedule_id = schedule["oncall_schedule"]["id"]
```
joeyorlando, line 96:
```suggestion
        if (oncall_schedule_id := schedule["oncall_schedule"]["id"]) is not None:
```

vadimkerr:
Hey @gilbertobr 👋 Thank you for the contribution!
I couldn't reproduce this, seems like this error only could happen if you apply the changes from https://github.com/grafana/oncall/pull/1007, could you please confirm if it's the case here?

vadimkerr:
I'm unable to reproduce this, so closing the PR. @gilbertobr feel free to reopen in case there's more info 🙂

PR #1015, opened 2022-12-16
[pagerduty-migrator]Added fallback if the integration type does not exist

gilbertobr:
image

As @vadimkerr commented in [PR](https://github.com/grafana/oncall/pull/401/files#r953611949), I added a fallback if the integration type does not exist.

If run without the fallback, it will get the default

Before:
```
docker run --rm -e PAGERDUTY_API_TOKEN="TOKEN" -e ONCALL_API_URL="http://localhost:8080" -e ONCALL_API_TOKEN="TOKEN_ONCALL" -e MODE="migrate" pd-oncall-migrator
```

Later:
```
docker run --rm -e PAGERDUTY_API_TOKEN="TOKEN" -e ONCALL_API_URL="http://localhost:8080" -e ONCALL_API_TOKEN="TOKEN_ONCALL" -e MODE="migrate" -e FALLBACK_TO_WEBHOOK_INTEGRATION_TYPE="webhook" pd-oncall-migrator
```


tools/pagerduty-migrator/migrator/resources/integrations.py
```diff
@@ -52,7 +53,14 @@ def migrate_integration(integration: dict, escalation_policies: list[dict]) -> N
...
 ) -> None:
-    payload = {"name": name, "type": integration_type, "team_id": None}
+
+    DEFAULT_INTEGRATION_TYPE = ["alertmanager", "grafana", "webhook", "kapacitor", "elastalert", "zabbix"]
+    fallback_to_integration_type = os.environ.get('FALLBACK_TO_WEBHOOK_INTEGRATION_TYPE')
+    
+    if integration_type in DEFAULT_INTEGRATION_TYPE:
+        payload = {"name": name, "type": integration_type, "team_id": None}
+    else:
+        payload = {"name": name, "type": fallback_to_integration_type, "team_id": None}
```
joeyorlando, line 63:
```suggestion
    payload = {
        "name": name,
        "type": integration_type if integration_type in DEFAULT_INTEGRATION_TYPE else fallback_to_integration_type,
        "team_id": None
    } 
```

tools/pagerduty-migrator/migrator/resources/integrations.py
```diff
@@ -52,7 +53,14 @@ def migrate_integration(integration: dict, escalation_policies: list[dict]) -> N
 def create_integration(
     name: str, integration_type: str, escalation_chain_id: str
 ) -> None:
-    payload = {"name": name, "type": integration_type, "team_id": None}
+
+    DEFAULT_INTEGRATION_TYPE = ["alertmanager", "grafana", "webhook", "kapacitor", "elastalert", "zabbix"]
+    fallback_to_integration_type = os.environ.get('FALLBACK_TO_WEBHOOK_INTEGRATION_TYPE')
```
joeyorlando, line 58:
```suggestion
    fallback_to_integration_type = os.environ.get('FALLBACK_INTEGRATION_TYPE')
```

joeyorlando:
hi there! thanks for your contribution.

do you mind adding a small note about the env var to `tools/pagerduty-migrator/README.md`?

tools/pagerduty-migrator/migrator/resources/integrations.py
```diff
@@ -52,7 +53,14 @@ def migrate_integration(integration: dict, escalation_policies: list[dict]) -> N
 def create_integration(
     name: str, integration_type: str, escalation_chain_id: str
 ) -> None:
-    payload = {"name": name, "type": integration_type, "team_id": None}
+
+    DEFAULT_INTEGRATION_TYPE = ["alertmanager", "grafana", "webhook", "kapacitor", "elastalert", "zabbix"]
```
vadimkerr, line 57: Please see [PAGERDUTY_TO_ONCALL_VENDOR_MAP](https://github.com/grafana/oncall/blob/dev/tools/pagerduty-migrator/migrator/config.py#L26), I think here it's better to use `PAGERDUTY_TO_ONCALL_VENDOR_MAP.values()` instead of introducing a new variable.

tools/pagerduty-migrator/migrator/resources/integrations.py
vadimkerr, line 58: I think this should be in [config.py](https://github.com/grafana/oncall/blob/dev/tools/pagerduty-migrator/migrator/config.py), so it's consistent with other constants.

vadimkerr (changes requested):
Thank you for the contribution!
The only place where the oncall type is set for an integration is here: https://github.com/grafana/oncall/blob/dev/tools/pagerduty-migrator/migrator/resources/integrations.py#L29

I think this should be implemented as following:
```python3
integration["oncall_type"] = PAGERDUTY_TO_ONCALL_VENDOR_MAP.get(vendor_name, default=FALLBACK_INTEGRATION_TYPE)
```

This seems to be more elegant and generate a smaller diff, also this wouldn't break other logic related to "oncall_type" of the integration (e.g. [this](https://github.com/grafana/oncall/blob/dev/tools/pagerduty-migrator/migrator/report.py#L52))

gilbertobr:
@vadimkerr , I made the changes as passed, but I get error 400.
image
[Line: 50](https://github.com/gilbertobr/oncall/blob/4dac49c08e306fa0723af1a5d0ae505f93011800/tools/pagerduty-migrator/migrator/resources/integrations.py#L50)

* Payload not working:

`payload = {'name': 'App - New Relic', 'type': 'newrelic', 'team_id': None}
`
* Payload that is working:

`payload = {'name': 'Example Webhook', 'type': 'webhook', 'team_id': None}`

gilbertobr:
image
The problem is that there is no newrelic integration in oncall

tools/pagerduty-migrator/README.md
```diff
@@ -80,6 +80,7 @@ docker run --rm \
 -e ONCALL_API_URL="<ONCALL_API_URL>" \
 -e ONCALL_API_TOKEN="<ONCALL_API_TOKEN>" \
 -e ONCALL_DEFAULT_CONTACT_METHOD="sms" \
+-e FALLBACK_INTEGRATION_TYPE="webhook" \
```
vadimkerr, line 83: We probably could rename this to `ONCALL_DEFAULT_INTEGRATION_TYPE` similar to `ONCALL_DEFAULT_CONTACT_METHOD` so it's more consistent

tools/pagerduty-migrator/migrator/resources/integrations.py
```diff
@@ -52,7 +55,14 @@ def migrate_integration(integration: dict, escalation_policies: list[dict]) -> N
 def create_integration(
     name: str, integration_type: str, escalation_chain_id: str
 ) -> None:
-    payload = {"name": name, "type": integration_type, "team_id": None}
+
```
vadimkerr, line 58: Seems like `create_integration` should be left unchaged since we already have set the `oncall_type` field earlier:
```
integration["oncall_type"] = PAGERDUTY_TO_ONCALL_VENDOR_MAP.get(vendor_name, default=FALLBACK_INTEGRATION_TYPE)
```

So I propose to remove this and only keep
```
payload = {"name": name, "type": integration_type, "team_id": None}
```
as it was before.

tools/pagerduty-migrator/migrator/resources/integrations.py
```diff
@@ -52,7 +55,14 @@ def migrate_integration(integration: dict, escalation_policies: list[dict]) -> N
...
+    DEFAULT_INTEGRATION_TYPE = list(PAGERDUTY_TO_ONCALL_VENDOR_MAP.values())
+    
+    payload = {
+        "name": name,
+        "type": integration_type if integration_type in DEFAULT_INTEGRATION_TYPE else FALLBACK_INTEGRATION_TYPE,
```
vadimkerr, line 63: I think this introduces the 400 error, so removing this as per previous comment should resolve the issue.

vadimkerr (changes requested):

vadimkerr:
@gilbertobr you're right, seems like newrelic is only available on cloud instances and not OSS. I'll take a closer look at this this week. I think it would make sense to solve this in a separate PR (e.g. generating `PAGERDUTY_TO_ONCALL_VENDOR_MAP` differently for OSS and cloud instances)

tools/pagerduty-migrator/README.md
gilbertobr: change applied: https://github.com/grafana/oncall/pull/1015/commits/6e26085464cd7b4d02b14a0186ac1456d1ca4520

PR #1017, opened 2022-12-17
Speedup search alertgroup to group alert

Konstantinov-Innokentii:


joeyorlando:
out of curiosity, what the background on why this sped things up?

PR #1028, opened 2022-12-20
RBAC: update RBAC permissions scopes for `plugins.app:access` action

IevaVasiljeva:
# What this PR does

Change RBAC permission scope to `plugins:id:grafana-oncall-app` from `plugins.app:id:grafana-oncall-app` for permissions with `plugins.app:access` action. This is needed because [Grafana is expecting](https://github.com/grafana/grafana/blob/main/pkg/services/accesscontrol/pluginutils/utils.go#L20-L21) `plugins:` not `plugins.app` prefix for permissions with `plugins.app:access` action.

## Checklist

- [ ] Tests updated
- [ ] Documentation added
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/grafana-plugin/src/plugin.json
+++ b/grafana-plugin/src/plugin.json
@@ -189,7 +189,7 @@
         "name": "Admin",
         "description": "Read/write access to everything in OnCall",
         "permissions": [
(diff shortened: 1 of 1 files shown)


joeyorlando (approved):
LGTM! Thanks for making this change 😄

PR #1052, opened 2022-12-29
Change wording from "incident" to "alert group" in the Telegram app

vadimkerr:
# What this PR does
Makes Telegram integration consistent with the rest of the system so it uses the word "alert group" instead of "incident" when referring to alert groups.

## Checklist

- [x] Tests updated
- [x] `CHANGELOG.md` updated


engine/apps/telegram/models/message.py
```diff
@@ -20,7 +20,7 @@ class TelegramMessage(models.Model):
         (ACTIONS_MESSAGE, "Actions message"),
         (LOG_MESSAGE, "Log message"),
         (FORMATTING_ERROR, "Alert can not be rendered"),
-        (PERSONAL_MESSAGE, "Alert group message with action buttons and incident log"),
+        (PERSONAL_MESSAGE, "Alert group message with action buttons and alert group log"),
```
vadimkerr, line 23: This line here generates a migration, but I think it won't really change anything at the DB level.

engine/apps/telegram/tasks.py
```diff
@@ -111,6 +111,19 @@ def send_link_to_channel_message_or_fallback_to_full_incident(
         )
 
 
+@shared_dedicated_queue_retry_task(bind=True, autoretry_for=(Exception,), retry_backoff=True, max_retries=None)
+def send_link_to_channel_message_or_fallback_to_full_incident(
```
vadimkerr, line 115: Can't just delete this straight away since it can cause celery issues. So introducing a new task with the same internals but a new name and "deprecating" this one.

vadimkerr (commented):
Most of the changes are just searching & replacing "incident" to "alert group", but there are couple places that are different.

iskhakov (approved):

PR #1068, opened 2023-01-03
Improve handling of template exceptions during group data creation

mderynck:
# What this PR does
With the addition of tighter controls on jinja templates handle exceptions while rendering group data as follows:
- Title will cache error message as title and display to user and the error will be logged
- Group distinction will be left as None and the error will be logged
- Is resolve signal will be treated as False and the error will be logged
- Is acknowledge signal will be treated as False and the error will be logged

https://github.com/grafana/oncall-private/issues/1542


matiasb (approved):
LGTM

PR #1073, opened 2023-01-04
Org soft-delete

Konstantinov-Innokentii:
# What this PR does
It introduces soft-delete of organization, since grafana stacks are soft-deleted too. Also, we had a problem with deleting orgs with large amounts of alerts, so soft-deletion will fix this problem. I think, that problem of cleaning alerts of deleted orgs should be solved as a part of alert retention  

## Checklist

- [x] Tests updated
- [ ] Documentation added
- [ ] `CHANGELOG.md` updated


engine/apps/user_management/models/organization.py
```diff
@@ -50,22 +53,37 @@ def create(self, **kwargs):
...
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.subscription_strategy = self._get_subscription_strategy()
 
+    def delete(self):
+        self.deleted_at = timezone.now()
+        self.save(update_fields=["deleted_at"])
+        id = self.public_primary_key
+        print(f"ID {id}")
```
joeyorlando, line 80:
```suggestion
```

joeyorlando, line 80:
```suggestion
```

engine/apps/user_management/models/organization.py
```diff
@@ -50,22 +53,37 @@ def create(self, **kwargs):
...
     def delete(self):
-        org_id = self.public_primary_key
-        super().delete(self)
-        if settings.FEATURE_MULTIREGION_ENABLED:
-            delete_oncall_connector_async.apply_async(
-                (org_id),
-            )
+        self.update(deleted_at=timezone.now())
+
+    def hard_delete(self):
```
joeyorlando, line 58: why is the `hard_delete` method needed on both the `OrganizationQuerySet` and `Organization` classes?

engine/apps/user_management/models/organization.py
```diff
@@ -317,3 +337,9 @@ def insight_logs_metadata(self):
     @property
     def is_moved(self):
         return self.migration_destination_id is not None
+
+
+@receiver(post_delete, sender=Organization)
+def test_post_delete(sender, instance, *args, **kwargs):
+    print(instance.public_primary_key)
+    print("org deleted")
```
joeyorlando, line 345: is this function still needed?

engine/apps/user_management/models/organization.py
Konstantinov-Innokentii: because it can be called both from queryset and object

mderynck (approved):

PR #1089, opened 2023-01-05
update mobile app push notification text + make telegram alert verbage consistent ("Firing" instead of "Alerting")

joeyorlando:
# What this PR does
Updates the mobile app push notification text to:

**iOS**
```
title: New Alert / New Critical Alert
subtitle:  #324134 KubePodCrashLooping via Amixr Prod AlertManager
body: Status: firing
```

**Android**
```
title: New Alert / New Critical Alert
body: #324134 KubePodCrashLooping via Amixr Prod AlertManager
        Status: firing
```

### DIFF
--- a/engine/apps/mobile_app/alert_rendering.py
+++ b/engine/apps/mobile_app/alert_rendering.py
@@ -10,4 +10,4 @@ class AlertMobileAppTemplater(AlertTemplater):
 def get_push_notification_message(alert_group):
     alert = alert_group.alerts.first()
     templated_alert = AlertMobileAppTemplater(alert).render()
-    return str_or_backup(templated_alert.title, "Alert Group")
+    return str_or_backup(f"#{alert_group.inside_organization_number} {templated_alert.title}", "Alert Group")
--- a/engine/apps/mobile_app/alert_rendering.py
+++ b/engine/apps/mobile_app/alert_rendering.py
@@ -10,4 +10,6 @@ class AlertMobileAppTemplater(AlertTemplater):
 def get_push_notification_message(alert_group):
     alert = alert_group.alerts.first()
     templated_alert = AlertMobileAppTemplater(alert).render()
-    return str_or_backup(f"#{alert_group.inside_organization_number} {templated_alert.title}", "Alert Group")
+    title = str_or_backup(templated_alert.title, "Alert Group")
+
+    return f"#{alert_group.inside_organization_number} {title} via {alert_group.channel.short_name}"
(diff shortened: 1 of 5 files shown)


engine/apps/mobile_app/tasks.py
```diff
@@ -51,11 +51,13 @@ def notify_user_async(user_pk, alert_group_pk, notification_policy_pk, critical)
...
         return
 
-    message = get_push_notification_message(alert_group)
     thread_id = f"{alert_group.channel.organization.public_primary_key}:{alert_group.public_primary_key}"
-    alert_title = f"Critical page: {message}" if critical else message
     number_of_alerts = alert_group.alerts.count()
 
+    alert_title = "New Critical Alert" if critical else "New Alert"
+    alert_subtitle = get_push_notification_message(alert_group)
+    alert_body = f"Status: Firing"  # TODO: is there a better spot to calculate this rather than hardcoding?
```
joeyorlando, line 59: not sure if there is a better way to generate the alert status here?

engine/apps/mobile_app/alert_rendering.py
```diff
@@ -10,4 +10,4 @@ def _render_for(self):
 def get_push_notification_message(alert_group):
     alert = alert_group.alerts.first()
     templated_alert = AlertMobileAppTemplater(alert).render()
-    return str_or_backup(templated_alert.title, "Alert Group")
+    return str_or_backup(f"#{alert_group.inside_organization_number} {templated_alert.title}", "Alert Group")
```
joeyorlando, line 13: for this @raphael-batte basically wants to mimic the format of Slack alerts ([example](https://raintank-corp.slack.com/archives/C025VMT6SPK/p1672821524988669)).

I think this may need to be modified to something like this:
```python3
return str_or_backup(f"#{alert_group.inside_organization_number} {templated_alert.title} via {alert_group.integration.name}", "Alert Group")
```
but how can I access the Integration name that fired this alert?

matiasb: I think that what you want is `{alert_group.channel.short_name}`

engine/apps/mobile_app/tasks.py
matiasb: I think it would be nice to have something we can reuse (I couldn't find something already there verbalizing the `alert_group.status`)

matiasb (approved):

engine/apps/mobile_app/alert_rendering.py
vadimkerr: But `short_name` would work too (wondering if we really need two different properties for the name)

vadimkerr (approved):
LGTM!

PR #1098, opened 2023-01-06
Draft initial logic for user/schedule paging

matiasb:


engine/apps/api/serializers/paging.py
```diff
@@ -0,0 +1,68 @@
...
+
+    title = serializers.CharField(required=False, default=None)
+    message = serializers.CharField(required=False, default=None)
+
+    def validate(self, attrs):
+        if len(attrs["users"]) == 0 and len(attrs["schedules"]) == 0:
+            raise serializers.ValidationError("Provide at least one user or schedule")
+
+        if not attrs["alert_group_id"] and not attrs["title"]:
+            raise serializers.ValidationError("Provide either alert_group_id or title")
```
matiasb, line 54: I think the idea was to fallback to a default title if none is provided? (the backend logic will set one if this is `None`).

matiasb (commented):
FYI, will leave the integration update (ie. not using the existing manual one) and the escalation support for a next PR to avoid adding more complexity here (I have some WIP). Added one minor comment, API changes look good to me.

engine/apps/alerts/escalation_snapshot/escalation_snapshot_mixin.py
```diff
@@ -253,6 +253,8 @@ def start_escalation_if_needed(self, countdown=ScenarioStep.CROSS_ACTION_DELAY,
 
         logger.debug(f"Start escalation for alert group with pk: {self.pk}")
 
+        # TODO: allow to start escalation at specific step?
```
vadimkerr, line 256: Do we plan to implement it as a part of this PR?

engine/apps/alerts/paging.py
```diff
@@ -0,0 +1,165 @@
...
+def _trigger_alert(organization: Organization, team: Team, title: str, message: str, from_user: User) -> AlertGroup:
+    """Trigger manual integration alert from params."""
+    alert_receive_channel = AlertReceiveChannel.get_or_create_manual_integration(
+        organization=organization,
+        team=team,
+        integration=AlertReceiveChannel.INTEGRATION_MANUAL,
+        deleted_at=None,
+        defaults={
+            "author": from_user,
+            "verbal_name": f"Manual incidents ({team.name if team else 'General'} team)",
```
vadimkerr, line 29: Probably change "incidents" to "alert groups"?

engine/apps/api/serializers/paging.py
vadimkerr: https://github.com/grafana/oncall/pull/1098/commits/55d248f9e612e1c47411f2ea2da27520d6acdee7

engine/apps/alerts/escalation_snapshot/escalation_snapshot_mixin.py
matiasb: No, will leave this for later, will remove the comment for now.

engine/apps/alerts/paging.py
matiasb: Yeah, sounds good, changing.

PR #1105, opened 2023-01-09
Simplify and speed up slack rendering

Konstantinov-Innokentii:
# What this PR does

This PR simplify and speed up slack rendering.

For alert group with large amount of alerts query `self.alert_group.alerts.filter(is_resolve_signal=False)` took a several seconds to execute, since **is_resolve_signal**not indexed. On the other hand, that query makes no sence, because anyway we using all alerts, when rendering their count in slack:
```
 def _get_text_alert_grouped(self):
        alert_count = self.alert_group.alerts.count()
        link = self.alert_group.web_link

        text = (
            f":package: Showing the last alert only out of {alert_count} total. "
            f"Visit <{link}|the plugin page> to see them all."
        )

        return text
```

So, I decided to remove filtering by **is_resolve_signal** in that case and leave only one` alert_group.alerts.count()` query.

Also, I checked that we have no any organizations which installed slack not via granular permissions and removed this check, which produced 4 SELECTs. For OSS slack we provide app manifest, so `self.alert_group.channel.organization.slack_team_identity.installed_via_granular_permissions `check can be safely removed.


vadimkerr (approved):
LGTM

PR #1113, opened 2023-01-09
update role description

IevaVasiljeva:
# What this PR does

Updates OnCall RBAC role description, as `Editor` role seems to include permissions to create schedules.


joeyorlando (approved):
good catch!

PR #1132, opened 2023-01-12
Enforce cloud connection to send push notifications on OSS

vadimkerr:
This PR modifies how OSS instances send mobile app push notifications. It also adds frontend warnings when user is trying to use the mobile app without connecting to cloud.

- [x] Add public API authentication to `FCMRelayView` and throttle the view to 300 push notifications per instance per minute. This is similar to how SMS and phone call notifications work on OSS instances.
- [x] Add frontend warnings based on cloud connectivity
- [x] Fix/add frontend tests
- [x] Add tests for FCMRelayView and mobile app backend

## Screenshots

When a user tries to connect the mobile app in his settings and cloud is not connected (clicking "Connect Cloud OnCall" redirects to the "Cloud" tab):

Screenshot 2023-01-12 at 18 48 58

When a user tries to use mobile push notifications as a personal notification step and cloud is not connected:

Screenshot 2023-01-12 at 19 01 10

Now on the "Cloud" tab there's some info about the mobile app (the last section at the bottom of the page):

Screenshot 2023-01-12 at 18 49 10

After connecting to the cloud instance, everything goes back to active and it's now possible to connect the mobile app:

Screenshot 2023-01-12 at 19 08 27

After connecting the app the warning is gone:

Screenshot 2023-01-12 at 19 07 00


grafana-plugin/src/containers/MobileAppConnection/MobileAppConnection.tsx
```diff
@@ -29,7 +32,28 @@ const INTERVAL_POLLING = 5000;
...
+      <VerticalGroup spacing="lg">
+        <Text>Please connect Cloud OnCall to use the mobile app</Text>
+        {isUserActionAllowed(UserActions.OtherSettingsWrite) ? (
+          <PluginLink query={{ page: 'cloud' }}>
+            <Button variant="secondary" icon="external-link-alt">
+              Connect Cloud OnCall
+            </Button>
+          </PluginLink>
+        ) : (
+          <Text>
```
teodosii, line 49: This should probably take a `type="secondary"` as otherwise it adds `undefined` in the snapshot on the class. Same for the other `<Text />` above.

engine/apps/mobile_app/urls.py
```diff
@@ -14,7 +12,6 @@
     optional_slash_path("auth_token", MobileAppAuthTokenAPIView.as_view(), name="auth_token"),
 ]
 
-if settings.FCM_RELAY_ENABLED:
-    urlpatterns += [
-        optional_slash_path("fcm_relay", FCMRelayView.as_view(), name="fcm_relay"),
-    ]
+urlpatterns += [
```
joeyorlando, line 15: shouldn't this endpoint should be disabled for OSS setups?

vadimkerr: This `FCM_RELAY_ENABLED` check is now here: https://github.com/grafana/oncall/pull/1132/files#diff-b163a51eb0fb7ccba294fc4366d56d9b3343cd04ec41d51cc6dad62a5dc0f1fcR37, it seems to me like it's easier to understand this way (also easier to enable/disable it when testing)

joeyorlando (approved):
LGTM! Good work on adding extensive tests for the mobile app backend logic. Only had one very minor question on the backend.

**Note**: I only reviewed the backend portion of the PR, relying on frontend team for the other half.

grafana-plugin/src/state/rootBaseStore/index.ts
```diff
@@ -98,9 +98,8 @@ export class RootBaseStore {
   // stores
 
   async updateBasicData() {
-    return Promise.all([
+    let storeUpdatePromises = [
```
maskin25, line 101: you can use `const` here

PR #1128, opened 2023-01-12
fake-data generation script + fixes for django-silk and django-debug-toolbar

joeyorlando:
# What this PR does

## Main stuff

- add Python script to populate local Grafana/OnCall setup w/ large amounts of fake data. Right now the data types that can be generated are:
  - teams and Admin users via the Grafana API (must be synced manually by going into the UI before going onto the next step)
  - Calendar Schedules which have three 8h oncall-shifts, via the OnCall public API
- fixes `django-debug-toolbar` when being run in `docker-compose` locally

## Other stuff
- documents how to easily modify the Grafana `docker-compose` container provisioning configuration
- document solutions for two backend setup related issues encountered when running the engine/celery workers locally, outside of `docker-compose`, on an Apple silicon Mac
- fixes small bug in `grafana_plugin.helpers.client.APIClient.call_api` where it would call `response.json()` for all requests, regardless of whether or not the response actually contained data or not
- in `engine/settings/dev.py`, properly setup `django-silk` and document the steps to use it locally
- make it possible to log out debug SQL queries by specifying `DEV_DEBUG_VIEW_SQL_QUERIES` env var, rather than having to uncomment out a section of `settings/dev.py`

## Which issue(s) this PR fixes

- Some local setup issues when trying to use `django-silk` and `django-debug-toolbar`
- Makes it much easier to populate your local setup with a lot of fake data
- Makes it possible to easily modify your local grafana's provisioning configuration

## Checklist

- [ ] Tests updated (N/A)
- [ ] Documentation added (N/A)
- [ ] `CHANGELOG.md` updated (N/A)

PR #1169, opened 2023-01-20
optimize GET /schedules internal API endpoint

joeyorlando:
# What this PR does

Fixes slow internal`GET /schedules` endpoints. Using the fake-data generation script in #1128, I generated 65 calendar schedules in my local setup. This resulted in the following endpoint performance:
Screenshot 2023-01-24 at 12 03 16

The responses which show ~76 queries were run on the latest `dev` branch. Responses w/ ~26 queries were run on this branch.

Additionally:
- add typing to a few methods in `apps/schedules/ical_utils.py`
- document `apps/api/permissions/__init__.py:user_is_authorized` function

https://github.com/grafana/oncall-private/issues/1552

## Checklist

- [ ] Tests updated
- [ ] Documentation added
- [ ] `CHANGELOG.md` updated


engine/apps/public_api/views/schedules.py
```diff
@@ -43,6 +43,8 @@ def get_queryset(self):
         if name is not None:
             queryset = queryset.filter(name=name)
 
+        queryset = self.serializer_class.setup_eager_loading(queryset)
```
iskhakov, line 46: 👍

iskhakov (approved):

joeyorlando:
Still TODO:
- [ ] investigate failing tests
- [ ] update internal API endpoint as well

engine/apps/slack/models/slack_usergroup.py
```diff
@@ -69,7 +69,7 @@ def can_be_updated(self) -> bool:
 
     @property
     def oncall_slack_user_identities(self):
-        users = self.oncall_schedules.get_oncall_users()
+        users = set(self.oncall_schedules.get_oncall_users().values())
```
joeyorlando: @vadimkerr  I think this won't work since `get_oncall_users` is returning a `Dict[Schedule, List[User]]`:
```python3
>>> schedules = {"foo": ["a", "b"], "bar": ["a", "c"]}
>>> set(schedules.values())
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: unhashable type: 'list'
```

we should probably do something like this:
```python3
>>> schedules = {"foo": ["a", "b"], "bar": ["a", "c"]}
>>> set(user for schedule in schedules.values() for user in schedule)
{'a', 'c', 'b'}
```

vadimkerr (approved):

PR #1180, opened 2023-01-20
Sync users and teams on tf requests

Konstantinov-Innokentii:
# What this PR does
This PR add sync with grafana on requests from terraform 

It's needed to fix case when customers want to create team via grafana terraform provider and use it in the oncall provider without having to log into Grafana Cloud.


engine/apps/user_management/sync.py
```diff
@@ -71,27 +70,44 @@ def _sync_instance_info(organization):
         organization.gcom_token_org_last_time_synced = timezone.now()
 
 
-def sync_users_and_teams(client, api_users, organization):
+def sync_users_and_teams(client: GrafanaAPIClient, organization):
+    sync_users(client, organization)
+    sync_teams(client, organization)
+    sync_team_members(client, organization)
+
+    organization.last_time_synced = timezone.now()
```
joeyorlando, line 78: nit; would suggest adding this on the line under the call to `sync_users_and_teams` rather than within this method

engine/apps/public_api/tf_sync.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def sync_teams_on_tf_request(organization):
+    cache_key = f"sync_teams_on_tf_request_{organization.id}"
+    if not cache.get(cache_key):
+        logger.info(f"Start sync_teams_on_tf_request organization_id={organization.id}")
+        cache.set(cache_key, True, SYNC_PERIOD)
+        client = GrafanaAPIClient(api_url=organization.grafana_url, api_token=organization.api_token)
+        sync_teams(client, organization, timeout=SYNC_REQUEST_TIMEOUT)
+        sync_team_members_for_organization_async.apply_async((organization.id,))
```
joeyorlando, line 37: did you mean to sync teams synchronously but team members asynchronously?

engine/apps/public_api/tf_sync.py
```diff
@@ -0,0 +1,37 @@
...
+    if not cache.get(cache_key):
+        logger.info(f"Start sync_users_on_tf_request organization_id={organization.id}")
+        client = GrafanaAPIClient(api_url=organization.grafana_url, api_token=organization.api_token)
+        cache.set(cache_key, True, SYNC_PERIOD)
+        sync_users(client, organization, timeout=SYNC_REQUEST_TIMEOUT)
+
+
+def sync_teams_on_tf_request(organization):
+    cache_key = f"sync_teams_on_tf_request_{organization.id}"
+    if not cache.get(cache_key):
```
joeyorlando, line 32: so lines 23 and 32 will only allow the sync to be triggered by a terraform API request once every `SYNC_PERIOD`? Aren't you concerned that this might lead to some strange behaviours?

joeyorlando (commented):
Overall LGTM. Had some minor questions/comments.

Don't forget to update the `CHANGELOG.md`. You may also want to add the "no public docs" label to your PR since I'm assuming it doesn't require changes to the public docs.

engine/apps/public_api/tf_sync.py
Konstantinov-Innokentii: Yep, I don't want to perform N requests for N team in the one request-responce cycle, because it might take more than 15 seconds and cortex-gw will drop such long request.

engine/apps/public_api/tf_sync.py
Konstantinov-Innokentii: @joeyorlando answered in DM :)

joeyorlando (approved):
LGTM 👍  consider adding some tests for the added terraform logic in the `GET /api/v1/<teams|users>` endpoints

Issue #1178, opened 2023-01-20
Allow outgoing webhooks to set custom headers


sd2k:
It would be great if outgoing webhooks could set additional headers as well as just `Authorization`. I'm imagining a similar UI/UX to adding custom headers when configuring HTTP-based Grafana datasources:

![image](https://user-images.githubusercontent.com/5464991/213693971-68d9291b-26e7-4810-9173-01ad48d81070.png)

mderynck:
Headers can now be set to anything including templates, might need to revisit the UI though

PR #1101, opened 2023-01-06
Add webhooks app and initial models

matiasb:

PR #1182, opened 2023-01-20
Add 'headers' field to outgoing_webhook models in the backend

sd2k:
This allows outgoing webhooks to specify custom headers as well as a
custom body, URL and authorization header.

The headers in the request must be a list of objects each representing a single header to be sent, containing `name` and `value` fields which must be strings. The headers will be encoded as JSON before being stored to the database. I chose to use a list of objects rather than a single object mapping header name -> header value because 1) it means we can preserve the order that the headers were originally added in the UI, which may be non-deterministic in an object, and 2) technically we should be able to pass the same header multiple times (although `requests` requires a dict to be passed to the `headers` arg so for now we can't handle that case; it's pretty rare anyway as far as I can tell).

This PR only contains the backend changes, I'm not really qualified to make the frontend changes unfortunately! I've added the `no changelog` and `no public docs` labels because of this.

Related to #1178.


mderynck:
This should be coordinated with:
https://github.com/grafana/oncall/pull/1101

engine/apps/alerts/models/custom_button.py
```diff
@@ -75,6 +75,7 @@ class CustomButton(models.Model):
     password = models.CharField(max_length=100, null=True, default=None)
     deleted_at = models.DateTimeField(blank=True, null=True)
     authorization_header = models.CharField(max_length=1000, null=True, default=None)
+    headers = models.TextField(null=True, default=None)
```
joeyorlando, line 78: did you consider using something like the following?
```python3
headers = models.JSONField(null=True, default=None)
```
this _may_ make the serialization/deserialization and validation simpler (would need to test)

engine/apps/api/tests/test_custom_button.py
```diff
@@ -208,11 +214,39 @@ def test_create_valid_data_after_render_use_all_data_button(custom_button_intern
...
+    url = reverse("api-internal:custom_button-list")
+
+    data = {
+        "name": "amixr_button",
+        "webhook": TEST_URL,
+        "team": None,
+        "headers": [{"name": "X-Grafana-API-Key", "value": "abcd1234"}],
+    }
+    response = client.post(url, data, format="json", **make_user_auth_headers(user, token))
+    print(response.data)
```
joeyorlando, line 236:
```suggestion
```

engine/apps/public_api/serializers/action.py
```diff
@@ -12,11 +12,41 @@
...
+
+        return json.dumps(data)
+
+
 class ActionCreateSerializer(serializers.ModelSerializer):
     id = serializers.CharField(read_only=True, source="public_primary_key")
     organization = serializers.HiddenField(default=CurrentOrganizationDefault())
     team_id = TeamPrimaryKeyRelatedField(required=False, allow_null=True, source="team")
     url = serializers.CharField(required=True, allow_null=False, allow_blank=False, source="webhook")
+    headers = HeadersField(allow_null=True, required=False)
```
joeyorlando, line 49: you may be able to avoid needing `HeadersField` by using a `serializers.JSONField` as such:
```python3
headers = serializers.JSONField(allow_null=True, required=False, validators=<some_validator_func_to_check_name_and_value>)
```
did not test this myself but some variation of this _should_ work

engine/apps/api/serializers/custom_button.py
```diff
@@ -11,11 +12,41 @@
...
+
+        return json.dumps(data)
+
+
 class CustomButtonSerializer(serializers.ModelSerializer):
     id = serializers.CharField(read_only=True, source="public_primary_key")
     organization = serializers.HiddenField(default=CurrentOrganizationDefault())
     team = TeamPrimaryKeyRelatedField(allow_null=True, default=CurrentTeamDefault())
     forward_whole_payload = serializers.BooleanField(allow_null=True, required=False)
+    headers = HeadersField(allow_null=True, required=False)
```
joeyorlando, line 49: similar comment to [here](https://github.com/grafana/oncall/pull/1182/files#r1084914184)

sd2k:
@joeyorlando thanks for the review, completely agreed on `JSONField`, I missed that in the Django docs.

@mderynck will #1101 supersede this work? If so I'd be happy to just close this and wait until that PR is complete instead.

mderynck:
@sd2k https://github.com/grafana/oncall/pull/1101 should cover the header functionality and also has the UI for it, we are moving to a new webhook model and are migrating the custom_buttons into it. It will likely be easier to not change the custom_buttons now and the new webhook model should be available soon™

sd2k:
Cool makes sense to me, let's close this and wait for #1101. Thanks!

PR #1185, opened 2023-01-21
[RBAC] - minor UI bug fixes

### LINKED ISSUE #971: When logged in as a viewer, the users page shows "Loading" indefinitely
Steps to reproduce:
- log in as a viewer
- go to the User's page
- "Loading" message is shown indefinitely

Screenshot 2022-12-09 at 11 53 56

joeyorlando:
# What this PR does

- remove hardcoded references to "Admin" and "Editor". Instead, use the `determineRequiredAuthString` and `generateMissingPermissionMessage` methods from `utils/authorization` which conditionally show the permission or role depending on whether RBAC is enabled or not
- fix bug on list users page that always showed "Loading...".
  Screenshot 2023-01-21 at 09 29 35
- only show users the user's table if they are allowed, otherwise show them this
  - RBAC enabled
    Screenshot 2023-01-26 at 09 09 57
  - RBAC disabled
    Screenshot 2023-01-26 at 09 05 30
- `Schedules Editor` role minor issue
  - Viewers are not allowed to list users by default, so granting schedule edit permission to them won’t allow them to see who rotations are assigned to or assign a rotation to a user. To make this a more straightforward user experience, instead of saying "No options found", we will tell the user that they are missing a particular role/permission to view these users:
    Before:
    Screenshot 2023-01-23 at 09 52 28
    After:
    Screenshot 2023-01-26 at 12 39 30

- closes #971 

## Checklist

- [x] Tests updated
- [x] `CHANGELOG.md` updated


grafana-plugin/src/pages/index.tsx
```diff
@@ -139,7 +138,6 @@ export const pages: { [id: string]: PageDefinition } = [
     icon: 'cloud',
     id: 'cloud',
     text: 'Cloud',
-    role: 'Admin',
```
joeyorlando, line 142: `role` is no longer used, this has been superseded by `action`. This should be safe to remove as the `PageDefinition` type does not contain a `role` attribute so any access to it would throw a type error during build time.

grafana-plugin/src/pages/users/Users.tsx
```diff
@@ -74,18 +76,20 @@ class Users extends React.Component<UsersProps, UsersState> {
...
+    await userStore.updateItems(usersFilters, page);
+
+    this.setState({ initialUsersLoaded: true });
   };
 
   componentDidUpdate(prevProps: UsersProps) {
-    if (!this.initialUsersLoaded && isUserActionAllowed(UserActions.UserSettingsWrite)) {
+    if (!this.state.initialUsersLoaded) {
       this.updateUsers();
-      this.initialUsersLoaded = true;
```
joeyorlando, line 88: this needs to be stored as state, not a class attribute. I believe the problem was that:
```tsx
this.initialUsersLoaded = true;
```
would not trigger a re-render.

grafana-plugin/src/containers/UserSettings/parts/index.tsx
```diff
@@ -133,7 +133,6 @@ export const TabsContent = observer(({ id, activeTab, onTabChange, isDesktopOrLa
         ) : (
           <PhoneVerification userPk={id} />
         ))}
-      {/* TODO: we should probably hide this tab when a user (ie. Admin) is viewing the user settings for another user. Would it make sense for an Admin to be able to link their mobile app to another user's profile */}
```
joeyorlando, line 136: this comment was outdated. This case is already being handled [here](https://github.com/grafana/oncall/blob/dev/grafana-plugin/src/containers/UserSettings/UserSettings.tsx#L56)

grafana-plugin/src/containers/RemoteSelect/RemoteSelect.tsx
```diff
@@ -62,17 +63,22 @@ const RemoteSelect = inject('store')(
...
-      });
+    const loadOptionsCallback = useCallback((query?: string): void => {
+      (async () => {
+        try {
+          const data = await makeRequest(href, { params: { search: query } });
+          const options = getOptions(data.results || data);
+          setOptions(options);
+        } catch (e) {
+          if (axios.isAxiosError(e) && e.response.status === 403) {
+            // TODO: disable dropdown/update placeholder
```
joeyorlando, line 74: this is still a WIP. we probably need to allow passing a `requiredUserAction` prop to `RemoteSelect`. `requiredUserAction` should be of type `src/utils/authorization/index.ts:UserAction`.

In the case that the `RemoteSelect` component gets a 403 when trying to fetch/search options, we should disable the dropdown and update the placeholder (consider using the `src/utils/authorization/index.ts:generateMissingPermissionMessage` method for this).

grafana-plugin/src/containers/ApiTokenSettings/ApiTokenSettings.tsx
```diff
@@ -67,6 +67,16 @@ class ApiTokens extends React.Component<ApiTokensProps, any> {
...
     ];
 
+    const requiredPermissionToView = UserActions.APIKeysWrite;
+    const authorizedToViewAPIKeys = isUserActionAllowed(requiredPermissionToView);
+
+    let emptyText = 'Loading...';
+    if (!authorizedToViewAPIKeys) {
+      emptyText = `${generateMissingPermissionMessage(requiredPermissionToView)} to be able to view API tokens.`;
+    } else if (apiTokens) {
+      emptyText = 'No tokens found';
```
teodosii, line 77: Not sure if there's any convention on oncall, but all phrases should end with `.`

grafana-plugin/src/containers/ApiTokenSettings/ApiTokenSettings.tsx
```diff
@@ -67,6 +67,16 @@ class ApiTokens extends React.Component<ApiTokensProps, any> {
       },
     ];
 
+    const requiredPermissionToView = UserActions.APIKeysWrite;
```
teodosii, line 70: Do we really need a separate const for this instead of just referencing `UserActions.APIKeysWrite` ?

grafana-plugin/src/containers/RemoteSelect/RemoteSelect.tsx
```diff
@@ -62,17 +68,22 @@ const RemoteSelect = inject('store')(
 
     const [options, setOptions] = useReducer(mergeOptions, []);
 
-    useEffect(() => {
-      makeRequest(href, {}).then((data) => {
-        setOptions(getOptions(data.results || data));
-      });
+    const loadOptionsCallback = useCallback((query?: string): void => {
+      (async () => {
```
teodosii, line 72: I understand the pattern you applied but why don't you just set the inner function of `useCallback` to be async instead of all the unnecessary lines? Unless I missed something :thinking:

grafana-plugin/src/containers/RemoteSelect/RemoteSelect.tsx
```diff
@@ -62,17 +68,22 @@ const RemoteSelect = inject('store')(
...
+        try {
+          const data = await makeRequest(href, { params: { search: query } });
+          const options = getOptions(data.results || data);
+          setOptions(options);
+        } catch (e) {
+          if (axios.isAxiosError(e) && e.response.status === 403 && requiredUserAction) {
```
teodosii, line 78: Is `axios.isAxisError(e)` really necessary here? Checking for status and user permission should be enough

grafana-plugin/src/containers/RemoteSelect/RemoteSelect.tsx
```diff
@@ -45,8 +48,11 @@ const RemoteSelect = inject('store')(
       openMenuOnFocus = true,
       showError,
       maxMenuHeight,
+      requiredUserAction,
     } = props;
 
+    const [noOptionsMessage, setNoOptionsMessage] = useState<string>('No options found');
```
teodosii, line 54: Add here an ending `.` too so at least we have some convention for phrases.

grafana-plugin/src/containers/RemoteSelect/RemoteSelect.tsx
joeyorlando: good point 👍 I initially tried this but it was complaining about the return type. Addressed this [here](https://github.com/grafana/oncall/pull/1185/commits/0f5f9ebfcfb0469aafbcfc470597f07fe5d32b33)

grafana-plugin/src/containers/RemoteSelect/RemoteSelect.tsx
joeyorlando: [`axios.isAxiosError`](https://github.com/axios/axios/blob/2a71f49bc6c68495fa419003a3107ed8bd703ad0/index.d.ts#L508) is a type guard, otherwise typescript is going to complain about me trying to access `e.response.status`

grafana-plugin/src/pages/users/Users.tsx
```diff
@@ -56,9 +57,10 @@ class Users extends React.Component<UsersProps, UsersState> {
     },
 
     errorData: initErrorDataState(),
+    initialUsersLoaded: false,
   };
 
-  initialUsersLoaded = false;
+  REQUIRED_PERMISSION_TO_VIEW_USERS = UserActions.UserSettingsWrite;
```
teodosii, line 63: Shouldn't this be `UserActions.UserSettingsRead` instead? Why does it need write permissions for viewing? :thinking: 

Also, the UserActions are pretty trivial and very readable, do we really need this const here? Such constant should not belong to the class anyway, either outside of it or in the constants file, but I don't think we need it - or do we?

joeyorlando:
@teodosii regarding the two punctuation related points. I didn't want to touch any of the pre-existing text (aside from the authorization related stuff).

I'm totally not a grammar expert. Did a quick search and [Grammarly suggests](https://www.grammarly.com/blog/bullet-points/#:~:text=If%20your%20points%20are%20not%20structured%20as%20proper%20sentences%2C%20you%20don%E2%80%99t%20need%20to%20end%20with%20punctuation): 

> If your points are not structured as proper sentences, you don’t need to end with punctuation

grafana-plugin/src/pages/users/Users.tsx
joeyorlando: on the backend, `GET /users` (list all users) currently requires `grafana-oncall-app:user-settings:write` ([or Editor basic role when RBAC is disabled](https://github.com/grafana/oncall/blob/dev/engine/apps/api/permissions/__init__.py#L175-L177))

I totally agree that the naming here is confusing, we should come back and address this specific point in a separate PR imo.

grafana-plugin/src/utils/authorization/index.ts
```diff
@@ -90,12 +90,30 @@ export const userHasMinimumRequiredRole = (minimumRoleRequired: OrgRole): boolea
...
+export const isUserActionAllowed = ({ permission, fallbackMinimumRoleRequired }: UserAction): boolean =>
+  config.featureToggles.accessControlOnCall
+    ? !!contextSrv.user.permissions?.[permission]
+    : userHasMinimumRequiredRole(fallbackMinimumRoleRequired);
+
+/**
+ * Given a `UserAction`, returns the permission or fallback-role, prefixed with "permission" or "role" respectively
+ * depending on whether or not RBAC is enabled/disabled
+ *
+ * @param UserAction
```
teodosii, line 102: We're not generating anything with JSDoc at the moment and we do not have any code documented with `@param` or `@returns` - maybe it'd make sense to remove these

teodosii (changes requested):
Looks good, just a few things

grafana-plugin/src/pages/users/Users.tsx
joeyorlando: 👍 addressed [here](https://github.com/grafana/oncall/pull/1185/commits/29c0b21803c27b8557d9f0ea9d24a62b7717276b)

teodosii:
> @teodosii regarding the two punctuation related points. I didn't want to touch any of the pre-existing text (aside from the authorization related stuff).
> 
> I'm totally not a grammar expert. Did a quick search and [Grammarly suggests](https://www.grammarly.com/blog/bullet-points/#:~:text=If%20your%20points%20are%20not%20structured%20as%20proper%20sentences%2C%20you%20don%E2%80%99t%20need%20to%20end%20with%20punctuation):
> 
> > If your points are not structured as proper sentences, you don’t need to end with punctuation

Yeah, maybe that was a nitpick, but my thinking was that we should start at some point to add a convention rather than stick to preferences.

grafana-plugin/src/containers/ApiTokenSettings/ApiTokenSettings.tsx
joeyorlando: addressed [here](https://github.com/grafana/oncall/pull/1185/commits/48cf7c3de02afd763f1fdf05cc0b9177c76eccaa) 👍

grafana-plugin/src/utils/authorization/index.ts
joeyorlando: addressed [here](https://github.com/grafana/oncall/pull/1185/commits/48cf7c3de02afd763f1fdf05cc0b9177c76eccaa) 👍

teodosii (approved):

grafana-plugin/src/containers/RemoteSelect/RemoteSelect.tsx
```diff
@@ -62,17 +68,23 @@ const RemoteSelect = inject('store')(
...
-      });
+    const loadOptionsCallback = useCallback(async (query?: string): Promise<SelectableValue[]> => {
+      try {
+        const data = await makeRequest(href, { params: { search: query } });
+        const options = getOptions(data.results || data);
+        setOptions(options);
+
+        return options;
+      } catch (e) {
+        if (axios.isAxiosError(e) && e.response.status === 403 && requiredUserAction) {
```
maskin25, line 79: can we omit checking `axios.isAxiosError(e)` here? because RemoteSelect shouldn't know which library we use for network requests

grafana-plugin/src/pages/users/Users.tsx
```diff
@@ -74,18 +75,20 @@ class Users extends React.Component<UsersProps, UsersState> {
     const { usersFilters, page } = this.state;
     const { userStore } = store;
 
-    if (!isUserActionAllowed(UserActions.UserSettingsWrite)) {
+    if (!isUserActionAllowed(REQUIRED_PERMISSION_TO_VIEW_USERS)) {
+      this.setState({ initialUsersLoaded: true });
```
maskin25, line 79: actually we don't load users if `!isUserActionAllowed(REQUIRED_PERMISSION_TO_VIEW_USERS)`, why do we need set `initialUsersLoaded` to`true`? looks contradictory

grafana-plugin/src/pages/users/Users.tsx
```diff
@@ -213,7 +220,7 @@ class Users extends React.Component<UsersProps, UsersState> {
                     </div>
 
                     <GTable
-                      emptyText={results ? 'No users found' : 'Loading...'}
+                      emptyText={initialUsersLoaded ? 'No users found' : 'Loading...'}
```
maskin25, line 223: if no REQUIRED_PERMISSION_TO_VIEW_USERS then let's display some appropriate message like 
`generateMissingPermissionMessage(REQUIRED_PERMISSION_TO_VIEW_USERS)`
instead of `'No users found'`

maskin25 (changes requested):

grafana-plugin/src/containers/RemoteSelect/RemoteSelect.tsx
teodosii: ^ I thought the same about this one, but Joey mentioned he encountered some type error there :thinking:

joeyorlando: I just double checked this morning:
```tsx
try {
  // makeRequest()
} catch (e) {
}
```

@teodosii you were right, `e` is typed as `any` here so technically accessing `e.response.status` is _okay_.

I like your point about the component not having to know about `axios`, but I also think a type-guard can be useful. I'd propose adding a custom type-guard to `src/network/index.ts` that wraps `axios.isAxiosError` (I'd also update the few other spots in the codebase where we use this function).

grafana-plugin/src/pages/users/Users.tsx
joeyorlando: `initialUsersLoaded` is used to conditionally show the `emptyText` on the `GTable` component.

But you're right, the fact that the `GTable` component is not rendered if `!authorizedToViewUsers`, means that technically I can remove this `setState` call 👍

grafana-plugin/src/pages/users/Users.tsx
joeyorlando: `GTable` won't get rendered if `!authorizedToViewUsers`, see [here](https://github.com/grafana/oncall/pull/1185/files#diff-9b30c8007fa630a2cafb31b17346113da5c47ded8b031c3e283e65c8cf7de99eR204) and the screenshots in the PR description

grafana-plugin/src/pages/users/Users.tsx
joeyorlando: this should be addressed in [ee9712f](https://github.com/grafana/oncall/pull/1185/commits/ee9712f92c9264037f077c7d30c327261773907f)

joeyorlando:
@teodosii & @maskin25 I refactored references to `axios.isAxiosError` in [this commit](https://github.com/grafana/oncall/pull/1185/commits/7b6e6140c4672a376f485fd896aa9f291b6d88b4), in favour of an `isNetworkError` function in `src/network/index.ts`. Let me know your thoughts.

maskin25 (approved):

PR #1190, opened 2023-01-23
Update message when user needs to connect their profile

DanCech:
# What this PR does

This just tweaks the message users get when they try to interact via slack but haven't connected their profile, it fixes a typo and streamlines the text.


joeyorlando (commented):
thanks for this! 😄

joeyorlando (approved):

Issue #1194, opened 2023-01-23
Container doesn't properly start when MySQL is not available


roock:
The oncall container doesn't properly start when MySQL is not (yet) available. I've only tested this issue with the docker-compose stack (mysql+rabbitmy). I think this could be a docker-compose specific issue and not relevant for kubernetes setups, since in kubernetes the mysql service would resolve even if the database is not (yet) ready.

Log output:
```
[log-encoder] registered format ${strftime:%Y-%m-%d %H:%M:%S} ${msgnl}
2023-01-23 15:28:38 *** Starting uWSGI 2.0.20 (64bit) on [Mon Jan 23 15:28:33 2023] ***
2023-01-23 15:28:38 compiled with version: 8.3.0 on 18 January 2023 20:13:01
2023-01-23 15:28:38 os: Linux-5.15.0-50-generic #56-Ubuntu SMP Tue Sep 20 13:23:50 UTC 2022
2023-01-23 15:28:38 nodename: 01d31400f874
2023-01-23 15:28:38 machine: aarch64
2023-01-23 15:28:38 clock source: unix
2023-01-23 15:28:38 detected number of CPU cores: 4
2023-01-23 15:28:38 current working directory: /etc/app
2023-01-23 15:28:38 writing pidfile to /tmp/project-master.pid
2023-01-23 15:28:38 detected binary path: /usr/local/bin/uwsgi
2023-01-23 15:28:38 !!! no internal routing support, rebuild with pcre support !!!
2023-01-23 15:28:38 uWSGI running as root, you can use --uid/--gid/--chroot options
2023-01-23 15:28:38 setgid() to 2000
2023-01-23 15:28:38 setuid() to 1000
2023-01-23 15:28:38 chdir() to /etc/app
2023-01-23 15:28:38 your memory page size is 4096 bytes
2023-01-23 15:28:38 detected max file descriptor number: 1048576
2023-01-23 15:28:38 lock engine: pthread robust mutexes
2023-01-23 15:28:38 thunder lock: disabled (you can enable it with --thunder-lock)
2023-01-23 15:28:38 uWSGI http bound on 0.0.0.0:8080 fd 7
2023-01-23 15:28:38 uwsgi socket 0 bound to TCP address 127.0.0.1:42421 (port auto-assigned) fd 6
2023-01-23 15:28:38 Python version: 3.9.16 (main, Jan 18 2023, 02:33:26)  [GCC 8.3.0]
[...]

PR #1195, opened 2023-01-23
Adding Container Entrypoint

roock:
# What this PR does

Adding Container Entrypoint
- entrypoint will wait for database to be available before starting the oncall service

see #1194

## Checklist

- [ ] Tests updated
- [ ] Documentation added
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/engine/Dockerfile
+++ b/engine/Dockerfile
@@ -36,4 +36,5 @@ RUN mkdir -p /tmp/prometheus_django_metrics;
 RUN chown -R 1000:2000 /tmp/prometheus_django_metrics
 ENV prometheus_multiproc_dir "/tmp/prometheus_django_metrics"
 
+ENTRYPOINT [ "/etc/app/docker-entrypoint.sh" ]
 CMD [ "uwsgi", "--ini", "uwsgi.ini" ]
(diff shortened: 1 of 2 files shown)


iskhakov:
Hi @roock, thank you fr the contribution. This script duplicates what oncall already has in the init containers and would be great to keep it in one place (orchestrators in this case). I read the connected issue and I think that adding `service_healthy` condition [as we have for redis](https://github.com/grafana/oncall/blob/dev/docker-compose.yml#L32) might help resolve the issue

roock:
> Hi @roock, thank you for the contribution. This script duplicates what oncall already has in the init containers and would be great to keep it in one place (orchestrators in this case). I read the connected issue and I think that adding `service_healthy` condition [as we have for redis](https://github.com/grafana/oncall/blob/dev/docker-compose.yml#L32) might help resolve the issue

@iskhakov Thank you for the quick response. You are absolutely right, a health check makes more sense.  After creating the PR I discovered how this is handled in the helm chart. I'm closing this one and will look into the service health check for the docker-compose stack.

roock:
@iskhakov FYI
Concerning the service_healthy condition: This is already set, but I think this is only evaluated by docker-compose and not by docker itself. When the docker service (re)starts, it will ignore the `service_healthy` description, since this is set by docker-compose as the container label `com.docker.compose.depends_on`.

Issue #1024, opened 2022-12-20
Improve startup time 


torkelo:
It feels like there is could be some easy quick wins to improve the loading time, especially when visiting OnCall again after the first time it takes almost the same amount of time (multiple seconds, more than Grafana itself). 

Seems to be multiple network requests (7x to /sync)  (and svg logo) that take up the booting time. 
![image](https://user-images.githubusercontent.com/10999/208599462-4d0f9776-4dbb-4065-ae86-648fcb11e5c0.png)
 
Also try to Use the PluginPage component when showing the loading state, so the user does not loose the left side section nav. 

![Screenshot from 2022-12-20 07-36-03](https://user-images.githubusercontent.com/10999/208600109-618c2fdf-419f-4dba-aa22-24e0ee42a39d.png)

Matvey-Kuk:
Duplicate #816

PR #1200, opened 2023-01-24
Refactor plugin sync

iskhakov:
# What this PR does

This PR adds a shortcut in the plugin synchronisation process, so the existing users will be able login without waiting for the sync task. Every request still starts the background synchronisation task, to be able to propagate the organisation changes faster than periodic task. It means that we don't necessarily need "force reload" button in the interface.
For all the other cases (user does not exist, organisation token "not ok", etc) process remains same - plugin will show "Initialising plugin..." until the background task in successfully completed

## Checklist

- [ ] Tests updated
- [ ] Documentation added
- [ ] `CHANGELOG.md` updated


engine/apps/auth_token/auth.py
```diff
@@ -96,6 +96,10 @@ def _get_user(request: Request, organization: Organization) -> User:
             logger.debug(f"Could not get user from grafana request. Context {context}")
             raise exceptions.AuthenticationFailed("Non-existent or anonymous user.")
 
+    @staticmethod
+    def get_user(cls, request: Request, organization: Organization) -> User:
```
Konstantinov-Innokentii, line 100: @staticmethod doesn't receive cls argument, probably you want to use @classmethod

engine/apps/grafana_plugin/views/sync.py
```diff
@@ -25,12 +27,13 @@ def post(self, request: Request) -> Response:
 
         try:
             organization = Organization.objects.get(stack_id=stack_id, org_id=org_id)
-
-            if organization.api_token_status == Organization.API_TOKEN_STATUS_OK:
-                is_installed = True
-            organization.api_token_status = Organization.API_TOKEN_STATUS_PENDING
-
-            organization.save(update_fields=["api_token_status"])
+            try:
```
Konstantinov-Innokentii, line 30: Could you please add comments here? I didn't catch the meaning of that check and why we setting API_TOKEN_STATUS_PENDING if we failed to get user from grafana context.

Konstantinov-Innokentii (changes requested):

engine/apps/grafana_plugin/views/sync.py
iskhakov: For all the other cases (user does not exist, organisation token "not ok", etc) process remains same - plugin will show "Initialising plugin..." until the background task in successfully completed

engine/apps/auth_token/auth.py
iskhakov: exactly, thanks!

engine/apps/grafana_plugin/views/sync.py
Konstantinov-Innokentii: Yep, but what is the case you are checking here and what is the expected behaviour? I mean, in which case we are unable to get user from grafana context?

iskhakov: btw, cleaned up a little and added pr description
you were too fast:)

iskhakov: fixed is_installed part, thanks

Konstantinov-Innokentii (approved):

engine/apps/grafana_plugin/views/sync.py
```diff
@@ -22,26 +23,27 @@ def post(self, request: Request) -> Response:
...
         try:
             organization = Organization.objects.get(stack_id=stack_id, org_id=org_id)
-
             if organization.api_token_status == Organization.API_TOKEN_STATUS_OK:
                 is_installed = True
-            organization.api_token_status = Organization.API_TOKEN_STATUS_PENDING
 
-            organization.save(update_fields=["api_token_status"])
+            is_user = PluginAuthentication.is_user_from_request_present_in_organization(request, organization)
+            if not is_user:
```
joeyorlando, line 33: _nit_;
```suggestion
            user_is_present_in_org = PluginAuthentication.is_user_from_request_present_in_organization(request, organization)
            if not user_is_present_in_org:
```

mderynck (approved):
LGTM
We can further optimize on UI side if we add the token_ok, license and version from the GET call to the data for the POST call. But that should probably be a different PR, this accomplishes most of the speedup while not needing to change UI.

iskhakov:
Fixes part of this [issue](https://github.com/grafana/oncall/issues/1024)

PR #1204, opened 2023-01-24
Remove DynamicSetting usage for mobile app backend on OSS

vadimkerr:
# What this PR does
Make so there's no need to populate `mobile_app_settings` DynamicSetting when using the OSS license to turn on the mobile app backend.

### DIFF
--- a/engine/apps/mobile_app/backend.py
+++ b/engine/apps/mobile_app/backend.py
@@ -53,6 +53,10 @@ class MobileAppBackend(BaseMessagingBackend):
 
     @staticmethod
     def is_enabled_for_organization(organization):
+        # Setting FEATURE_MOBILE_APP_INTEGRATION_ENABLED to True is enough to enable mobile app on OSS instances
+        if settings.LICENSE == settings.OPEN_SOURCE_LICENSE_NAME:
+            return True
+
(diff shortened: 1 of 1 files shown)


joeyorlando (approved):
LGTM!

I think with this change we can consider getting rid of the `enable_mobile_app` Django manage command ([here](https://github.com/grafana/oncall/blob/dev/engine/engine/management/commands/enable_mobile_app.py)) and the `enable-mobile-app-feature-flags` `make` command ([here](https://github.com/grafana/oncall/blob/dev/Makefile#L129-L133) + the references to this command in `./dev/README.md`)

vadimkerr:
> I think with this change we can consider getting rid of the enable_mobile_app Django manage command ([here](https://github.com/grafana/oncall/blob/dev/engine/engine/management/commands/enable_mobile_app.py)) and the enable-mobile-app-feature-flags make command ([here](https://github.com/grafana/oncall/blob/dev/Makefile#L129-L133) + the references to this command in ./dev/README.md)

It can be still useful when setting the license to `settings.CLOUD_LICENSE_NAME` locally, so I think it should stay until the dynamic setting usage is removed completely.

PR #1213, opened 2023-01-25
Cleanup ScenarioStep

Konstantinov-Innokentii:
# What this PR does
This PR cleanup ScenarioStep. It's needed to simplify moving Slack to the messaging backends in future.

1. Introduce AlertGroupSlackService to move logic from ScenarioStep. Also it allowed to get rid of importing ScenarioSteps in the code not related to processing of slack callbacks.
2. Remove tags from ScenarioSteps, they are unused.
3. Remove ScenarioStep.dispatch method. It just was calling ScenarioStep.process_scenario.
4. Remove "action" param from process_scenario, it was unused.
5. Remove creation of SlackActionRecord on handling SlackEvents. We are not using it, but it generates INSERT query on most of the user-slack interactions.
6. Remove "random_prefix_for_routing" from ScenarioStep, it was unused.
## Checklist

- [ ] Tests updated
- [ ] Documentation added
- [ ] `CHANGELOG.md` updated


engine/apps/slack/scenarios/notification_delivery.py
```diff
@@ -46,17 +46,17 @@ def process_signal(self, log_record):
...
                         )
                     elif log_record.notification_channel == UserNotificationPolicy.NotificationChannel.PHONE_CALL:
-                        self.post_message_to_channel(
+                        self.post_notification_failed_message_to_channel(
                             f"Failed to call to {user_verbal_with_mention}. Phone number is not verified",
                             alert_group.slack_message.channel_id,
                         )
 
-    def post_message_to_channel(self, text, channel):
+    def post_notification_failed_message_to_channel(self, text, channel):
```
Ferril, line 59: I would keep the old name because nothing here says about **failed** message

engine/apps/slack/views.py
```diff
@@ -507,9 +492,36 @@ def _get_organization_from_payload(self, payload, slack_team_identity):
...
+                        "text": {
+                            "type": "mrkdwn",
+                            "text": warning_text,
+                        },
+                    },
+                ],
+            }
             try:
-                step.open_warning_window(payload, warning_text)
+                slack_client.api_call(
```
Ferril, line 520: I wonder why not use `open_warning_window` here?

Ferril (approved):
LGTM, just a couple of minor comments

vadimkerr (approved):

engine/apps/alerts/incident_appearance/renderers/slack_renderer.py
```diff
@@ -362,3 +358,74 @@ def _get_invitation_attachment(self):
...
+    def _get_select_user_element(
+        self, action_id, multi_select=False, initial_user=None, initial_users_list=None, text=None
+    ):
+        MAX_STATIC_SELECT_OPTIONS = 100
+
+        if not text:
+            text = f"Select User{'' if not multi_select else 's'}"
+        element = {
+            "action_id": action_id,
+            "type": "static_select" if not multi_select else "multi_static_select",
```
joeyorlando, line 371:
```suggestion
            "type": "multi_static_select" if multi_select else "static_select",
```

engine/apps/slack/alert_group_slack_service.py
```diff
@@ -0,0 +1,136 @@
...
+    SlackAPIRateLimitException,
+    SlackAPITokenException,
+)
+
+logger = logging.getLogger(__name__)
+
+
+class AlertGroupSlackService:
+    def __init__(self, slack_team_identity, slack_client=None) -> None:
+        self.slack_team_identuty = slack_team_identity
```
joeyorlando, line 19:
```suggestion
        self.slack_team_identity = slack_team_identity
```

joeyorlando: actually, do we need to set `self.slack_team_identity` here? I don't see `slack_team_identuty` referenced anywhere else in this PR

engine/apps/slack/alert_group_slack_service.py
```diff
@@ -0,0 +1,136 @@
...
+        if slack_client is not None:
+            self.slack_client = slack_client
+        else:
+            self._slack_client = SlackClientWithErrorHandling(slack_team_identity.bot_access_token)
```
joeyorlando, line 23: do we need two slack client related instance attributes?
```suggestion
        self.slack_client = slack_client if slack_client else SlackClientWithErrorHandling(slack_team_identity.bot_access_token)
```

engine/apps/slack/alert_group_slack_service.py
```diff
@@ -0,0 +1,136 @@
...
+        logger.info(f"Started _update_slack_message for alert_group {alert_group.pk}")
+        SlackMessage = apps.get_model("slack", "SlackMessage")
+        AlertReceiveChannel = apps.get_model("alerts", "AlertReceiveChannel")
+
+        slack_message = alert_group.slack_message
+        attachments = alert_group.render_slack_attachments()
+        blocks = alert_group.render_slack_blocks()
+        logger.info(f"Update message for alert_group {alert_group.pk}")
+        try:
+            self._slack_client.api_call(
```
joeyorlando, line 35:
```suggestion
            self.slack_client.api_call(
```

engine/apps/slack/alert_group_slack_service.py
```diff
@@ -0,0 +1,136 @@
...
+                    logger.info(
+                        f"Message has not been updated for alert_group {alert_group.pk} due to slack rate limit."
+                    )
+            else:
+                raise e
+
+        except SlackAPIException as e:
+            if e.response["error"] == "message_not_found":
+                logger.info(f"message_not_found for alert_group {alert_group.pk}, trying to post new message")
+                result = self._slack_client.api_call(
```
joeyorlando, line 57:
```suggestion
                result = self.slack_client.api_call(
```

engine/apps/slack/alert_group_slack_service.py
```diff
@@ -0,0 +1,136 @@
...
+        # TODO: refactor checking the possibility of sending message to slack
+        # do not try to post message to slack if integration is rate limited
+        if alert_group.channel.is_rate_limited_in_slack:
+            return
+
+        SlackMessage = apps.get_model("slack", "SlackMessage")
+        slack_message = alert_group.get_slack_message()
+        channel_id = slack_message.channel_id
+        try:
+            result = self._slack_client.api_call(
```
joeyorlando, line 93:
```suggestion
            result = self.slack_client.api_call(
```

engine/apps/slack/views.py
joeyorlando: agreed here, this seems to duplicate a lot of the logic in `scenario_step.ScenarioStep.open_warning_window`

engine/apps/slack/models/slack_action_record.py
```diff
@@ -1,17 +1,21 @@
 from django.db import models
 
-from apps.slack.scenarios.scenario_step import ScenarioStep
+# from apps.slack.scenarios.scenario_step import ScenarioStep
```
joeyorlando, line 3: can we remove this model, and any remaining references, completely? if I understand correctly, you removed the last usage of it [here](https://github.com/grafana/oncall/pull/1213/files#diff-1784f1d0d65fa477f4562e73aa23fe1c757b171f36e03f12600bdb021f121307L292).

engine/apps/slack/alert_group_slack_service.py
```diff
@@ -0,0 +1,136 @@
...
+
+class AlertGroupSlackService:
+    def __init__(self, slack_team_identity, slack_client=None) -> None:
+        self.slack_team_identuty = slack_team_identity
+        if slack_client is not None:
+            self.slack_client = slack_client
+        else:
+            self._slack_client = SlackClientWithErrorHandling(slack_team_identity.bot_access_token)
+
+    def _update_alert_group_slack_message(self, alert_group):
```
joeyorlando, line 25: can we rename `_update_alert_group_slack_message` to `update_alert_group_slack_message`? The leading underscore by convention means that the method/variable is meant to be accessed only internally, but it's being used outside of this module.

engine/apps/slack/scenarios/notification_delivery.py
joeyorlando: +1 here, this method doesn't care about the contents of the message

joeyorlando (changes requested):
good work, this is a pretty big cleanup. I had a few questions/suggestions that I think are worth discussing.

engine/apps/slack/scenarios/notification_delivery.py
Konstantinov-Innokentii: This method is used only in NotificationDeliveryStep and only to send message about failed notification. One of the problem, which this PR is solving is that different parts of code used slack steps methods, which broke the abstraction. That is why I renamed this method. I want to avoid usage of that method in other places, that this step.

engine/apps/slack/views.py
Konstantinov-Innokentii: Reason is same, as in comment above.  I believe, that creating step only to call it's method is a bad practice.

engine/apps/slack/alert_group_slack_service.py
Konstantinov-Innokentii: Good catch, thanks

engine/apps/slack/models/slack_action_record.py
Konstantinov-Innokentii: Yes, we can remove SlackActionRecord, but I don't want to do it in this PR, it's too big already. I prefer to do it in separate one, since migration is involved.

engine/apps/slack/alert_group_slack_service.py
Konstantinov-Innokentii: Yep, you are right, it's legacy leftover.

engine/apps/slack/views.py
Konstantinov-Innokentii: On the other hand, this call is inside slack module, so it's kind of fair tradeoff

engine/apps/slack/scenarios/notification_delivery.py
joeyorlando: I would recommend naming it something like `_post_message_to_channel`. The method itself has nothing to do w/ a failure message, but if the intent is to keep it "private" to the `NotificationDeliveryStep` class, then a method named w/ a preceding underscore would be appropriate imo

engine/apps/slack/views.py
joeyorlando: would it make sense to refactor the `open_warning_window` from the `ScenarioStep` to some other Slack related class? imo it's not worth duplicating this logic

Konstantinov-Innokentii: Decided to return original variant, refactoring ScenarioSteps themselves, like checking which methods are used for slack interactions and so on are topic of different big PR.

engine/apps/slack/alert_group_slack_service.py
joeyorlando: bump

joeyorlando (approved):

joeyorlando:
I just re-triggered the build, but noticed a failing test [here](https://github.com/grafana/oncall/actions/runs/4231399697/jobs/7350174355#step:4:765):
```python3
>       if len(user_options) > scenario_step.MAX_STATIC_SELECT_OPTIONS:
E       AttributeError: module 'apps.slack.scenarios.scenario_step' has no attribute 'MAX_STATIC_SELECT_OPTIONS'
```

PR #1222, opened 2023-01-25
Add support for web overrides to Terraform schedules

matiasb:


engine/apps/schedules/models/on_call_schedule.py
```diff
@@ -401,6 +401,66 @@ def _merge_events(self, events):
             events = merged
         return events
 
+    def _generate_ical_file_from_shifts(self, qs, extra_shifts=None, allow_empty_users=False):
+        raise NotImplementedError
+
+    def preview_shift(self, custom_shift, user_tz, starting_date, days, updated_shift_pk=None):
```
matiasb, line 407: FYI, I'm just moving this to the upper class (from `OnCallScheduleWeb` to `OnCallSchedule`), without any changes.

grafana-plugin/src/containers/Rotations/ScheduleOverrides.tsx
```diff
@@ -173,13 +172,17 @@ class ScheduleOverrides extends Component<ScheduleOverridesProps, ScheduleOverri
...
 
-  onRotationClick = (shiftId: Shift['id'], moment: dayjs.Dayjs) => {
+  onRotationClick = (shiftId: Shift['id'], source: string, moment: dayjs.Dayjs) => {
     const { disabled } = this.props;
 
     if (disabled) {
       return;
     }
 
+    if (source !== 'web') {
```
matiasb, line 182: Wondering if there is a better way to do this?

Konstantinov-Innokentii:
@matiasb Are overrides included in the same shifts field in the api? If so, when you do override and then terraform plan, terraform will detect a difference in its config and data from api (since overrides added to the shifts)  and will try to remove them on terraform apply. It seems we need to return shifts and overrides separately in the api and add overrides field to the terraform provider.
And a second question - now it's just web ui, you can't make an override from terraform file, right? I'm asking because I don't want to give user "powers" to make override with same configurations as shifts ( rolling users, etc..)

matiasb:
Good questions! Thanks for checking.

> @matiasb Are overrides included in the same shifts field in the api? If so, when you do override and then terraform plan, terraform will detect a difference in its config and data from api (since overrides added to the shifts) and will try to remove them on terraform apply. It seems we need to return shifts and overrides separately in the api and add overrides field to the terraform provider ( or not add it, just return overrides separately to not allow users edit overrides from tf yet, see second question :) .

Overrides are not included with shifts. Terraform schedules are using the M2M relation for shifts (`custom_on_call_shifts`), while web shifts use the FK (from shift to schedule, `custom_shifts`; side note: I think we should eventually migrate everything to the FK implementation), so shifts defined via Terraform are still separated from web overrides (which is the only thing we are enabling atm).
About API, there are no changes there, so returned `shifts` will be coming from Terraform rotations, and for overrides we are only returning the `ical_url_overrides` field (so we may need to do something here to include web defined overrides somewhere). I won't add override support to TF for now.

> And a second question - now it's just web ui, you can't make an override from terraform file, right? I'm asking because I don't want to give user "powers" to make override with same configurations as shifts ( rolling users, etc..). I believe we should keep overrides as simple as possible.

Correct, overrides will be possible through the ical imported calendar and/or using the web UI. FWIW, I would make it explicit in the UI and in the docs that it is better to stick to one schema to avoid confusion (ie. use only ical, or only web).

Makes sense?

engine/apps/schedules/models/on_call_schedule.py
```diff
@@ -403,6 +402,83 @@ def _merge_events(self, events):
             events = merged
         return events
 
+    def _generate_ical_file_from_shifts(self, qs, extra_shifts=None, allow_empty_users=False):
```
matiasb, line 405: `_generate_ical_file_from_shifts` and `preview_shift` methods are moved, unchanged, from Web schedule class to the base class to be reused by Calendar (Terraform) schedules too.

matiasb:
> Correct, overrides will be possible through the ical imported calendar and/or using the web UI. FWIW, I would make it explicit in the UI and in the docs that it is better to stick to one schema to avoid confusion (ie. use only ical, or only web).

Changed plans, with the updated PR you should choose one schema at any time. Enabling web overrides will ignore any overrides coming from iCal URL.

grafana-plugin/src/pages/schedule/Schedule.tsx
```diff
@@ -116,11 +116,13 @@ class SchedulePage extends React.Component<SchedulePageProps, SchedulePageState>
     const users = store.userStore.getSearchResult().results;
     const schedule = scheduleStore.items[scheduleId];
 
-    const disabled =
+    const disabledRotationForm =
       !isUserActionAllowed(UserActions.SchedulesWrite) ||
       schedule?.type !== ScheduleType.API ||
-      shiftIdToShowRotationForm ||
-      shiftIdToShowOverridesForm;
+      shiftIdToShowRotationForm;
```
teodosii, line 122: Shouldn't this be converted to boolean instead if we're going to use it such?

```jsx
const disabledRotationForm =
    !isUserActionAllowed(UserActions.SchedulesWrite) ||
    schedule?.type !== ScheduleType.API ||
    !!shiftIdToShowRotationForm;
```

Because `shiftIdToShowRotationForm` doesn't appear to be a bolean, but `disabledRotationForm` is expected to be a boolean at the first glance.

grafana-plugin/src/containers/Rotations/ScheduleOverrides.tsx
```diff
@@ -73,8 +73,7 @@ class ScheduleOverrides extends Component<ScheduleOverridesProps, ScheduleOverri
 
     const schedule = store.scheduleStore.items[scheduleId];
 
-    const isTypeReadOnly =
-      schedule && (schedule?.type === ScheduleType.Ical || schedule?.type === ScheduleType.Calendar);
+    const isTypeReadOnly = schedule && !schedule?.enable_web_overrides;
```
teodosii, line 76: The `?` is not needed, you can also shorthand the syntax if you want.

```jsx
const isTypeReadOnly = !schedule?.enable_web_overrides;
```

teodosii (approved):
Approved, just 2 minor things.

PR #1226, opened 2023-01-26
Improve Healthchecks in Docker-Compose

roock:
Additional health checks for most Containers in the Docker-Compose setup.
Also health checks between the different compose files have been unified.
Health Checks used are those already used in the helm charts.

It is related to #1194

## Checklist

- [ ] Tests updated
- [ ] Documentation added
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/docker-compose-mysql-rabbitmq.yml
+++ b/docker-compose-mysql-rabbitmq.yml
@@ -31,6 +31,11 @@ services:
     command: >
       sh -c "uwsgi --ini uwsgi.ini"
     environment: *oncall-environment
+    healthcheck:
+      test: ["CMD", "bash", "-c" , 'curl --fail http://127.0.0.1:8080/health/ || exit 1']
+      timeout: 5s
+      retries: 5
+      start_period: 20s
     depends_on:
       mysql:
         condition: service_healthy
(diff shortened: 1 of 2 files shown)


docker-compose-mysql-rabbitmq.yml
```diff
@@ -63,6 +75,8 @@ services:
     depends_on:
       mysql:
         condition: service_healthy
+      oncall_db_migration:
+        condition: service_completed_successfully
```
iskhakov, line 79: `oncall_db_migration` depends on `oncall_db_migration`?

docker-compose-mysql-rabbitmq.yml
```diff
@@ -31,6 +31,11 @@ services:
     command: >
       sh -c "uwsgi --ini uwsgi.ini"
     environment: *oncall-environment
+    healthcheck:
+      test: ["CMD", "bash", "-c" , 'curl --fail http://127.0.0.1:8080/health/ || exit 1']
+      timeout: 5s
+      retries: 5
+      start_period: 20s
     depends_on:
```
iskhakov, line 39: let's add the same dependencies [here](https://github.com/grafana/oncall/pull/1226/files#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3R28) as well

docker-compose-mysql-rabbitmq.yml
roock: @iskhakov ah, right. Copy/paste error. Fixed now.

PR #1227, opened 2023-01-26
Use latest Grafana Container Images

roock:
additionally allow oeverride grafana version in .env file

The default docker-compose file should not use a grafan version several months old.
This uses the same logic as the development docker-compose file.

## Checklist

- [ ] Tests updated
- [ ] Documentation added
- [ ] `CHANGELOG.md` updated


joeyorlando (approved):
LGTM! thanks for the contribution @roock 🙏

PR #1234, opened 2023-01-27
Remove unused fields rendering from alert groups list serialiser

iskhakov:
# What this PR does
We render unnecessary fields which are not used by web plugin. This PR removes them
TODOS: 
- [] check if they are used on mobile app as it shares the same API
- [] render_for_web should return markdown, markdown should be converted to html on frontend side

## Checklist

- [ ] Tests updated
- [ ] Documentation added
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/engine/apps/alerts/incident_appearance/renderers/classic_markdown_renderer.py
+++ b/engine/apps/alerts/incident_appearance/renderers/classic_markdown_renderer.py
@@ -8,14 +8,17 @@ class AlertClassicMarkdownRenderer(AlertBaseRenderer):
     def templater_class(self):
         return AlertClassicMarkdownTemplater
 
-    def render(self):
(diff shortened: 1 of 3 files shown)


Konstantinov-Innokentii:
I think, that we still rendering that fields, since actual applying of Jinja templates happening in the alert templater

joeyorlando (commented):
_nit_ comment. Wondering if it'd be more maintainable to try using a `dict` or `class` to hold the possible `render_fields` values?

An `enum.StrEnum` class could be nice (like [this](https://realpython.com/python-enum/#:~:text=In%20the%20current,as%20parent%20classes.)), but this isn't available until Python 3.11.

iskhakov:
decided to go with caching

PR #1236, opened 2023-01-27
Cache render_for_web field for alertgroups list serializer

iskhakov:
# What this PR does
This PR caches the field `render_for_web` with lifetime 1 day and cache becomes invalid if it was created before 
* last alert received
* template changed

## Checklist

- [ ] Tests updated
- [ ] Documentation added
- [ ] `CHANGELOG.md` updated


Konstantinov-Innokentii:
Some time ago we had a discussion about storing first render result for alert-group and not to re-render on templates changes, what do you think?

iskhakov:
Let's try caching to fix draining SLO, let's do that iteration after

engine/apps/api/serializers/alert_group.py
```diff
@@ -92,10 +95,33 @@ def get_render_for_web(self, obj):
         if not obj.last_alert:
             return {}
 
-        return AlertGroupWebRenderer(obj, obj.last_alert).render()
+        web_templates_modified_at = obj.channel.web_templates_modified_at
```
vadimkerr, line 98: Shouldn't this be on `AlertGroupListSerializer.get_render_for_web` as well so the list view is cached? I believe `ShortAlertGroupSerializer` is only used to render the root alert group and dependant alert groups.
Probably this caching stuff can be moved to a separate function that's used both on `AlertGroupListSerializer` and `ShortAlertGroupSerializer`?

iskhakov: yep, thank you, I want to use this approach in all such serializer methods after I confirm it is working on dev cluster

engine/apps/api/serializers/alert_group.py
```diff
@@ -92,10 +95,33 @@ def get_render_for_web(self, obj):
...
+        CACHE_LIFEIME = 60 * 60 * 24
+        cached_render_for_web = cache.get(CACHE_KEY, None)
+
+        # use cache only if cache exists
+        # and cache was created after the last alert created
+        # and either web templates never modified
+        # or cache was created after templates were modified
+        if (
+            cached_render_for_web is not None
+            and cached_render_for_web.get("cache_created_at") > last_alert_created_at
```
Konstantinov-Innokentii, line 111: Is cache_created_at value from cache datetime so you can compare it with last_alert_created_at? I remember, that after putting to and retrieving from cache and became a string.

engine/apps/api/serializers/alert_group.py
vadimkerr: Ignore this comment, it's already on `AlertGroupListSerializer`, it's just GitHub merging diff lines 🙂

vadimkerr (approved):

Issue #1119, opened 2023-01-10
Filter alerts assigned by user


marctc:
I was oncall last weekend and I got multiple pages that I had to acknowledge.
The problem is that I had to scroll up on a slack channel to find all the alerts (more than 20)
and acknowledge one by one.

I tried to do that in the UI but I couldn't find any option to filter firing alerts assigned to
me. The only option that I found is filter by integration, but that's not very useful if someone
belongs to multiple rotations (i.e integrations). 

Would it make sense to add a filter for alerts by user (assigned)?

stevesg:
+1 for an "Ack All" feature for coping with pager storms.

iskhakov:
closed by https://github.com/grafana/oncall/pull/1240

PR #1240, opened 2023-01-27
Add a filter by involved users to alert groups page

matiasb:
Related to #1119 

It also adds a shortcut to filter current user's related alert groups (alert groups user was notified by, or in which user participated). Make the filter visible by default, with a false value.

### DIFF
--- a/engine/apps/api/views/alert_group.py
+++ b/engine/apps/api/views/alert_group.py
@@ -85,7 +85,11 @@ class AlertGroupFilter(DateRangeFilterMixin, ModelFieldFilterMixin, filters.Filt
     invitees_are = filters.ModelMultipleChoiceFilter(
         queryset=get_user_queryset, to_field_name="public_primary_key", method="filter_invitees_are"
     )
+    involved_users_are = filters.ModelMultipleChoiceFilter(
+        queryset=get_user_queryset, to_field_name="public_primary_key", method="filter_by_involved_users"
+    )
     with_resolution_note = filters.BooleanFilter(method="filter_with_resolution_note")
+    mine = filters.BooleanFilter(method="filter_mine")
 
     class Meta:
         model = AlertGroup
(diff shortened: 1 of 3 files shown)


Matvey-Kuk (approved):

engine/apps/api/views/alert_group.py
```diff
@@ -85,7 +85,11 @@ class AlertGroupFilter(DateRangeFilterMixin, ModelFieldFilterMixin, filters.Filt
     invitees_are = filters.ModelMultipleChoiceFilter(
         queryset=get_user_queryset, to_field_name="public_primary_key", method="filter_invitees_are"
     )
+    involved_users_are = filters.ModelMultipleChoiceFilter(
+        queryset=get_user_queryset, to_field_name="public_primary_key", method="filter_by_involved_users"
+    )
     with_resolution_note = filters.BooleanFilter(method="filter_with_resolution_note")
+    mine = filters.BooleanFilter(method="filter_mine")
```
vadimkerr, line 92: Alternative naming idea: "I'm involved" or something similar, seems to be easier to understand than "mine"

engine/apps/api/views/alert_group.py
```diff
@@ -85,7 +85,11 @@ class AlertGroupFilter(DateRangeFilterMixin, ModelFieldFilterMixin, filters.Filt
     invitees_are = filters.ModelMultipleChoiceFilter(
         queryset=get_user_queryset, to_field_name="public_primary_key", method="filter_invitees_are"
     )
+    involved_users_are = filters.ModelMultipleChoiceFilter(
```
vadimkerr, line 88: Probably we can name it just "Involved users" instead of "Involved users are"?

vadimkerr (approved):
LGTM! Added a couple minor comments related to naming.
Also probably a good idea to add some tests covering these filters.

engine/apps/api/views/alert_group.py
matiasb: Ack, makes sense. Will change in a next PR.

engine/apps/api/views/alert_group.py
matiasb: I like that better too, but we already had "Invitees are". We may change both in a later PR?

PR #1243, opened 2023-01-27
Correct init command for Postgres Database

scubbo:
It seems that the MySQL syntax command was mistakenly copied to the instructions for Postgres, which uses a different syntax - ref [here](https://www.postgresql.org/docs/current/sql-createdatabase.html) for creation syntax, and
[here](https://www.postgresql.org/docs/current/multibyte.html) for encoding information. I did read through
[here](https://www.postgresql.org/docs/current/collation.html) looking for an appropriate collation, but they appear to be installation-specific rather than having a standard set defined in documentation - I omitted this value (thus setting to the default), and haven't noted any immediate problems.


iskhakov:
thank you for the contribution, LGTM

iskhakov (approved):

Issue #1083, opened 2023-01-05
Integration settings drawer randomly opens back


maskin25:
more details https://raintank-corp.slack.com/archives/C0229FD3CE9/p1672903035392769

thomasmitchell:
This appears to happen when steps of an escalation chain for the current integration are still loading when you navigate away - when that network call returns, it will go back to the integrations screen. This is particularly common for us at this point because our Manual Incidents integration (the one that's loaded by default when going to the Integrations page) now has quite a lot of routes in order to facilitate the Slack slash command, and therefore, once you've gone to the Integrations page, you need to wait 5-10 seconds before navigating away or else you'll get snapped right back.

PR #900, opened 2022-11-24
WIP: first basic UI integration test

joeyorlando:
**What this PR does**:
Adds our first UI integration test using [Playwright](https://playwright.dev/) and runs on CI 🥳 (currently run the 3 default headless browsers, chromium, firefox, and webkit)

**Which issue(s) this PR fixes**:
Closes #873 

**Checklist**
- [ ] Tests updated
- [ ] Documentation added
- [ ] `CHANGELOG.md` updated

PR #1256, opened 2023-01-30
Rares/1083 fix integrations

teodosii:
# What this PR does

- Fix for #1083 
- Fix for #1257 

- [x] `CHANGELOG.md` updated

### DIFF
--- a/grafana-plugin/src/pages/integrations/Integrations.tsx
+++ b/grafana-plugin/src/pages/integrations/Integrations.tsx
@@ -53,24 +53,21 @@ class Integrations extends React.Component<IntegrationsProps, IntegrationsState>
 
   private alertReceiveChanneltoPoll: { [key: string]: number } = {};
   private alertReceiveChannelTimerId: ReturnType<typeof setTimeout>;
-  private isMounted: boolean;
 
   async componentDidMount() {
-    this.isMounted = true;
-    this.update().then(this.parseQueryParams);
+    this.update().then(() => this.parseQueryParams(true));
   }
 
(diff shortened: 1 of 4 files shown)


grafana-plugin/src/pages/integrations/Integrations.tsx
```diff
@@ -122,6 +128,7 @@ class Integrations extends React.Component<IntegrationsProps, IntegrationsState>
   }
 
   componentWillUnmount() {
+    this.isMounted = false;
     clearInterval(this.alertReceiveChannelTimerId);
```
maskin25, line 132: hmm, timer is cleared here, didn't get why should we check `this.isMounted`...

joeyorlando (changes requested):
Are you also able to still reproduce the issue?
https://www.loom.com/share/1e81d6b1e2aa4b8ab7be24d3b7d38c59

teodosii:
Thanks @joeyorlando - will further investigate - seems there's more around that

grafana-plugin/src/pages/integrations/Integrations.tsx
teodosii: That has nothing to do with timer, it's about when you move from integrations to a separate page. Once it loads the escalation chains it calls that method but you're on a different page already.

joeyorlando (approved):
good stuff! aside from the `markdownlint` failing, LGTM 🚢

joeyorlando:
@maskin25 are you okay with merging this in today? Will unblock #900

PR #1261, opened 2023-01-31
Cache web template rendered fields for alert and alertgroup endpoints

iskhakov:
# What this PR does
This PR adds same approach as introduced [here](https://github.com/grafana/oncall/pull/1236) to all alert and alertgroup endpoints

## Checklist

- [ ] Tests updated
- [ ] Documentation added
- [ ] `CHANGELOG.md` updated


joeyorlando (approved):
overall looks good. Wondering if it'd be possible to combine `AlertFieldsCacheSerializerMixin` and `AlertGroupFieldsCacheSerializerMixin` as there's quite a bit of overlap in their logic (or maybe extract out most of the logic into a parent class that they both inherit)?

iskhakov:
The only difference is the cache invalidation condition, but combining the rest into a single function just made the code less clean. Decided to optimise it next time if we start using this approach elsewhere.

PR #1262, opened 2023-01-31
add django-dbconn-retry library

joeyorlando:
# What this PR does

We occasionally see `django.db.utils.OperationalError` pop up. These seem to be sporadic and for whatever reason the engine/celery processes cannot connect to the database, but succeed on subsequent requests. This results in HTTP 500s for the original requests. The `django-dbconn-retry` library states in its documentation:

> This library monkeypatches django.db.backends.base.BaseDatabaseWrapper so that when a database operation fails because the underlying TCP connection was already closed, it first tried to reconnect, instead of immediately raising an OperationException.

I would like to give this a try and see if this patches these `django.db.utils.OperationalError` errors.

## Checklist

- [ ] Tests updated (N/A)
- [x] `CHANGELOG.md` updated


matiasb (approved):
Makes sense to me :+1:

PR #1263, opened 2023-01-31
Add default slack channel info to direct paging dialog

matiasb:


engine/apps/slack/scenarios/paging.py
```diff
@@ -356,20 +357,25 @@ def render_dialog(slack_user_identity, slack_team_identity, payload, initial=Fal
...
-def _get_form_view(routing_uid, blocks, private_metatada):
+def _get_form_view(routing_uid, blocks, private_metatada, organization):
+    try:
+        channel = organization.slack_team_identity.get_cached_channels().get(
+            slack_id=organization.general_log_channel_id
+        )
+        additional_info = f":information_source: The alert group will be posted to the #{channel.name} Slack channel"
+    except SlackChannel.DoesNotExist:
+        additional_info = (
+            ":information_source: The alert group will be posted to the default Slack channel if there is one setup"
```
joeyorlando, line 372: wouldn't a `SlackChannel.DoesNotExist` exception mean that there is no default slack channel configured for the organization? If that is the case, should the wording on this message be changed (or maybe even excluded)?

matiasb: There could be a channel setup for which we don't have the cached information yet (unlikely, but still, when the exception is raised, I think the displayed message is useful to know that you can setup a default channel and it will be used in that case)

joeyorlando: makes sense 👍

engine/apps/api/tests/test_schedules.py
```diff
@@ -1103,7 +1103,7 @@ def test_merging_same_shift_events(
     )
     now = timezone.now().replace(hour=0, minute=0, second=0, microsecond=0)
     start_date = now - timezone.timedelta(days=7)
-    request_date = start_date
+    request_date = start_date + timezone.timedelta(days=1)
```
joeyorlando, line 1106: why do we suddenly need to add 1 day to this?

joeyorlando (approved):
LGTM 👍  just one last small question.

engine/apps/api/tests/test_schedules.py
matiasb: This test has been failing a few times in CI, I'm guessing it could be related to the fact that the shift is setup for a couple of hours starting on `start_date`, and we were using that date for the request, potentially leaving the shift off during filtering (depending on timezones, etc). So moving the request date one day later I expect we can get a consistent output for this.

joeyorlando: oh nice, good idea! 🤞

PR #1264, opened 2023-02-01
Add utility function to get readonly db key if defined

joeyorlando:
# What this PR does

This is a minor refactor before implementing https://github.com/grafana/oncall-private/issues/1558.

Additionally, it cleans up a few spots where we do this:
```
# Re-take in case we are in the readonly db context.
```
We currently don't read anything from a read-only database, so this should be not necessary.

## Checklist

- [x] Tests updated
- [ ] `CHANGELOG.md` updated (N/A)

### DIFF
--- a/engine/apps/slack/models/slack_message.py
+++ b/engine/apps/slack/models/slack_message.py
@@ -68,12 +68,8 @@ class SlackMessage(models.Model):
                     f"It is strange!"
                 )
                 return None
-            # Re-take object to switch connection from readonly db to master.
             self._slack_team_identity = self.organization.slack_team_identity
-
-            _self = SlackMessage.objects.get(pk=self.pk)
-            _self._slack_team_identity = _self.organization.slack_team_identity
-            _self.save()
+            self.save()
         return self._slack_team_identity
 
     def get_alert_group(self):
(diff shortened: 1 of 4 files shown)


engine/apps/slack/models/slack_message.py
```diff
@@ -68,12 +68,8 @@ def slack_team_identity(self):
                     f"It is strange!"
                 )
                 return None
-            # Re-take object to switch connection from readonly db to master.
             self._slack_team_identity = self.organization.slack_team_identity
-
-            _self = SlackMessage.objects.get(pk=self.pk)
-            _self._slack_team_identity = _self.organization.slack_team_identity
-            _self.save()
```
joeyorlando, line 76: AFAIK, we don't make any usage of read-only databases atm, so there should be no need to refetch `self`

engine/apps/slack/models/slack_user_identity.py
```diff
@@ -207,13 +201,11 @@ def timezone(self):
     @property
     def im_channel_id(self):
         if self.cached_im_channel_id is None:
-            _self = SlackUserIdentity.objects.get(pk=self.pk)  # Re-take in case we are in the readonly db context.
             sc = SlackClientWithErrorHandling(self.slack_team_identity.bot_access_token)
             try:
                 result = sc.api_call("conversations.open", users=self.slack_id, return_im=True)
                 self.cached_im_channel_id = result["channel"]["id"]
-                _self.cached_im_channel_id = result["channel"]["id"]
-                _self.save()
```
joeyorlando, line 216: same thing in this file as the above comment

iskhakov (approved):

PR #1266, opened 2023-02-01
modify check_escalation_finished_task task

joeyorlando:
# What this PR does

This PR:
- modifies the `check_escalation_finished_task` celery task to:
  - do stricter escalation validation based on the alert group's
escalation snapshot (see the `audit_alert_group_escalation` method in `engine/apps/alerts/tasks/check_escalation_finished.py` for the validation logic)
  - use a read-only database for querying alert-groups if one is configured, otherwise use the "default" one
  - ping a configurable heartbeat (new env var `ALERT_GROUP_ESCALATION_AUDITOR_CELERY_TASK_HEARTBEAT_URL` added)
  - increase the task frequency from every 10 to every 13 minutes (this can be configured via an env variable)
  - adds public documentation on how to configure this auditor task
- modifies the local celery startup command to properly take into consideration all celery related env vars (similar to the ones we use in `engine/celery_with_exporter.sh`; this made it easier to enable `celery beat` locally for testing)
- removes the following code:
  - removes references to `AlertGroup.estimate_escalation_finish_time` and
marks the model field as deprecated using the [`django-deprecate-fields` library](https://pypi.org/project/django-deprecate-fields/). This field was only used for the previous version of this validation task
  - `EscalationSnapshotMixin.calculate_eta_for_finish_escalation` was only used to calculate the value for `AlertGroup.estimate_escalation_finish_time`
  - `calculate_escalation_finish_time` celery task
  

https://github.com/grafana/oncall-private/issues/1558

## Checklist

- [x] Tests updated
- [x] Documentation added
- [x] `CHANGELOG.md` updated


Matvey-Kuk (changes requested):
Now this task is failing in case of problems. We rely on it to blow our rabbit which should trigger monitoring. I wish this task to start explicitly sending alerts about problems to our OnCall instance. Also would be nice to make it check with heartbeat to make sure it's not failing silently. Also it's pretty heavy so retrying it is dangerous. Even 10 retrying tasks could damage DB.

Could you please check with @iskhakov about other probable improvements?

engine/apps/alerts/tasks/check_escalation_finished.py
```diff
@@ -32,6 +34,17 @@ def check_escalation_finished_task():
     )
```
iskhakov, line 35: ### Problem
I find calculating `estimate_escalation_finish_time` and use it in the query error prone.
For example if `estimate_escalation_finish_time` was never set, auditor will never evaluate this alert group and it will be missed.

### Possible Solution
I propose to use the following approach:
The desired state of the alert group is stored in the _escalation snapshot_ (see it attached below)
Escalation snapshot consists of:
* all escalation steps 
* current escalation step
* next escalation step ETA (this ETA is used by celery, which means we expect celery to run the next escalation step by the ETA)

Auditor retrieves all active alert groups and flags alert groups which have celery ETA < current time.

> ⚠️ We need to make sure that escalation snapshot was successfully set for the alert group, otherwise flag the alert group.

### Bonus task 😄 
We know all the escalation steps from snapshot which are expected to be completed and we can compare them to the alert group log records. Auditor can flag the alert group if it sees that some escalations steps were not completed.

### Example of the escalation snapshot
```json
{
  "channel_filter_snapshot": {
    "id": 1,
    "str_for_clients": "default",
    "notify_in_slack": true,
    "notify_in_telegram": false,
    "notification_backends": null
  },
  "escalation_chain_snapshot": {
    "id": 1,
    "name": "ildar"
  },
  "last_active_escalation_policy_order": 1,
  "escalation_policies_snapshots": [
    {
      "id": 1,
      "order": 0,
      "step": 13,
      "wait_delay": null,
      "notify_to_users_queue": [
        1
      ],
      "last_notified_user": null,
      "from_time": null,
      "to_time": null,
      "num_alerts_in_window": null,
      "num_minutes_in_window": null,
      "custom_button_trigger": null,
      "notify_schedule": null,
      "notify_to_group": null,
      "escalation_counter": 0,
      "passed_last_time": "2023-02-02T07:11:33.560022Z",
      "pause_escalation": false
    },
    {
      "id": 2,
      "order": 1,
      "step": 0,
      "wait_delay": "00:05:00",
      "notify_to_users_queue": [],
      "last_notified_user": null,
      "from_time": null,
      "to_time": null,
      "num_alerts_in_window": null,
      "num_minutes_in_window": null,
      "custom_button_trigger": null,
      "notify_schedule": null,
      "notify_to_group": null,
      "escalation_counter": 0,
      "passed_last_time": "2023-02-02T07:11:39.324953Z",
      "pause_escalation": false
    },
    {
      "id": 3,
      "order": 2,
      "step": 13,
      "wait_delay": null,
      "notify_to_users_queue": [
        1
      ],
      "last_notified_user": null,
      "from_time": null,
      "to_time": null,
      "num_alerts_in_window": null,
      "num_minutes_in_window": null,
      "custom_button_trigger": null,
      "notify_schedule": null,
      "notify_to_group": null,
      "escalation_counter": 0,
      "passed_last_time": null,
      "pause_escalation": false
    }
  ],
  "slack_channel_id": null,
  "pause_escalation": false,
  "next_step_eta": "2023-02-02T07:16:39.273824Z"
}
```

iskhakov (commented):
Can we consider

engine/apps/alerts/escalation_snapshot/snapshot_classes/escalation_snapshot.py
```diff
@@ -71,6 +79,20 @@ def next_active_escalation_policy_order(self) -> int:
...
+        as this field should be updated as the escalation policy is executed over time. If it is, this means that
+        an escalation policy step has been missed, or is substantially delayed
+
+        if `next_step_eta` is `None` then `None` is returned, otherwise a boolean is returned
+        representing the result of the time comparision
+        """
+        if self.next_step_eta is None:
+            # TODO: this is a valid case I assume?
+            return None
+        return self.next_step_eta > (timezone.now() - timezone.timedelta(minutes=5))
```
joeyorlando, line 94: debatable whether or not it makes more sense to have this method belong inside the celery task file, rather than here on `EscalationSnapshot`

engine/apps/alerts/escalation_snapshot/snapshot_classes/escalation_snapshot.py
```diff
@@ -28,34 +36,34 @@ class EscalationSnapshot:
...
-        pause_escalation,
-        next_step_eta,
+        alert_group: "AlertGroup",
+        channel_filter_snapshot: "ChannelFilterSnapshot",
+        escalation_chain_snapshot: "EscalationChainSnapshot",
+        last_active_escalation_policy_order: int,
+        escalation_policies_snapshots: typing.List["EscalationPolicySnapshot"],
+        slack_channel_id: str,
+        pause_escalation: bool,
+        next_step_eta: str,
```
joeyorlando, line 46: having the type hints made it easier to work with in my IDE 👍

engine/apps/alerts/escalation_snapshot/escalation_snapshot_mixin.py
```diff
@@ -87,8 +84,7 @@ def build_raw_escalation_snapshot(self) -> dict:
             'next_step_eta': '2021-10-18T10:28:28.890369Z
         }
         """
-
-        escalation_snapshot = None
+        data = {}
```
joeyorlando, line 87: this means we will no longer be setting `AlertGroup.raw_escalation_snapshot` to `None`. Instead:
```python3
>>> from apps.alerts.escalation_snapshot.snapshot_classes import EscalationSnapshot
>>> EscalationSnapshot.serializer({}).data
{'channel_filter_snapshot': None, 'escalation_chain_snapshot': None, 'last_active_escalation_policy_order': None, 'escalation_policies_snapshots': [], 'slack_channel_id': None, 'pause_escalation': False, 'next_step_eta': None}
```

engine/apps/alerts/escalation_snapshot/serializers/escalation_snapshot.py
```diff
@@ -8,11 +8,11 @@
...
 class EscalationSnapshotSerializer(serializers.Serializer):
-    channel_filter_snapshot = ChannelFilterSnapshotSerializer()
-    escalation_chain_snapshot = EscalationChainSnapshotSerializer()
+    channel_filter_snapshot = ChannelFilterSnapshotSerializer(allow_null=True, default=None)
+    escalation_chain_snapshot = EscalationChainSnapshotSerializer(allow_null=True, default=None)
     last_active_escalation_policy_order = serializers.IntegerField(allow_null=True, default=None)
-    escalation_policies_snapshots = EscalationPolicySnapshotSerializer(many=True)
-    slack_channel_id = serializers.CharField(allow_null=True)
+    escalation_policies_snapshots = EscalationPolicySnapshotSerializer(many=True, default=list)
+    slack_channel_id = serializers.CharField(allow_null=True, default=None)
```
joeyorlando, line 15: related to [this change](https://github.com/grafana/oncall/pull/1266/files#r1094955383)

engine/apps/alerts/tasks/check_escalation_finished.py
```diff
@@ -1,48 +1,115 @@
...
+        # TODO: uncomment this out
+        # is_escalation_finished=False,
         resolved=False,
         acknowledged=False,
         root_alert_group=None,
-        estimate_escalation_finish_time__lte=CHECKING_TIME,
+        # TODO: finish adding this filtering logic
+        # get all the alert groups older than 5 minutes
+        # but greater than February 1, 2023
+        # created_at__gte=<5 mins ago>,
```
joeyorlando, line 106: TODO: finish this filtering logic

engine/apps/alerts/tasks/check_escalation_finished.py
```diff
@@ -32,6 +34,17 @@ def check_escalation_finished_task():
     )
```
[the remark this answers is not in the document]
joeyorlando: > alert group was created, but the task failed during snapshot creation, therefore escalation snapshot field still has the default value None, and the periodic "checker" task should flag this alert group.

this case would be handled [here](https://github.com/grafana/oncall/pull/1266/files#diff-44a1662dc9824366dc3e35fd5a0ed44fb14e36bff91ee3d6eef724095c42b6e7R37-R40). I'm not currently setting a "flag" anywhere. What did you have in mind for how such a flag would be used?

engine/apps/alerts/tasks/check_escalation_finished.py
```diff
@@ -1,48 +1,115 @@
...
+        )
+        return
+    task_logger.info(f"{base_msg}'s escalation snapshot has a last_active_escalation_policy_snapshot")
+
+    # TODO: I'm assuming these are the proper alert group log record types to check for but want to verify
+    if not last_active_escalation_policy_snapshot.escalation_policy.log_records.filter(
+        type__in=[AlertGroupLogRecord.TYPE_ESCALATION_TRIGGERED, AlertGroupLogRecord.TYPE_ESCALATION_FINISHED]
+    ).exists():
+        # TODO: do we actually need to go through all of the escalation policy snapshots and validate log records
+        # for all of those, or is the last_active one sufficient?
```
joeyorlando, line 77: @iskhakov had a few minor questions for you within this block, when you have time

joeyorlando, line 77: @iskhakov had a few minor questions for you within this block when you have time

engine/apps/alerts/tasks/check_escalation_finished.py
joeyorlando: @iskhakov lets discuss this. I think all of these points should be addressed now 😄

engine/apps/alerts/tests/test_escalation_snapshot.py
```diff
@@ -11,54 +9,6 @@
...
-        from_time=from_time,
-        to_time=to_time,
-    )
-
-    alert_group = make_alert_group(alert_receive_channel, channel_filter=channel_filter)
-    alert_group.raw_escalation_snapshot = alert_group.build_raw_escalation_snapshot()
-    alert_group.save()
-    return alert_group, notify_to_multiple_users_step, wait_step, notify_if_time_step
-
-
```
joeyorlando, line 61: this was moved to `apps/alerts/tests/conftest.py`

engine/apps/alerts/tests/conftest.py
```diff
@@ -9,3 +12,51 @@ def mock_render_for(*args, **kwargs):
...
+        escalation_chain=channel_filter.escalation_chain,
+        escalation_policy_step=EscalationPolicy.STEP_NOTIFY_IF_TIME,
+        from_time=from_time,
+        to_time=to_time,
+    )
+
+    alert_group = make_alert_group(alert_receive_channel, channel_filter=channel_filter)
+    alert_group.raw_escalation_snapshot = alert_group.build_raw_escalation_snapshot()
+    alert_group.save()
+    return alert_group, notify_to_multiple_users_step, wait_step, notify_if_time_step
```
joeyorlando, line 62: this was moved from `engine/apps/alerts/tests/test_escalation_snapshot.py` so that it could be used by some other tests within that directoryso that it could be used by some other tests within that directory

engine/apps/alerts/tasks/check_escalation_finished.py
```diff
@@ -1,48 +1,130 @@
...
+    if not escalation_snapshot:
+        raise AlertGroupEscalationPolicyExecutionAuditException(
+            f"{base_msg} does not have an escalation snapshot associated with it, this should never occur"
+        )
+    task_logger.info(f"{base_msg} has an escalation snapshot associated with it, auditing if it executed properly")
+
+    escalation_policies_snapshots = escalation_snapshot.escalation_policies_snapshots
+
+    if not escalation_policies_snapshots:
+        # TODO: in what case is this valid?
```
joeyorlando, line 46: question here

engine/apps/alerts/tasks/check_escalation_finished.py
```diff
@@ -1,48 +1,130 @@
...
+        task_logger.info(
+            f"{base_msg}'s escalation snapshot has an empty escalation_policies_snapshots, skipping further validation"
+        )
+        return
+    task_logger.info(
+        f"{base_msg}'s escalation snapshot has a populated escalation_policies_snapshots, continuing validation"
+    )
+
+    if escalation_snapshot.next_step_eta_is_valid() is False:
+        # TODO: is a next_step_eta value of None actually valid here, or should we raise an error?
```
joeyorlando, line 56: question here

engine/apps/alerts/tasks/check_escalation_finished.py
```diff
@@ -1,48 +1,130 @@
...
+        )
+        return
+    task_logger.info(
+        f"{base_msg}'s escalation snapshot has {num_of_executed_escalation_policy_snapshots} executed escalation policies"
+    )
+
+    for executed_escalation_policy_snapshot in executed_escalation_policy_snapshots:
+        escalation_policy_id = executed_escalation_policy_snapshot.id
+
+        # TODO: is it valid to only check for the finished log record type here?
```
joeyorlando, line 77: question here

Ferril: We create this log if all escalation steps were executed or escalation chain didn't have any steps. As far as I know we don't create it only on step `Resolve incident automatically`, but in this case alert group would have status `resolved`, so it doesn't count.

joeyorlando: got it 👍

so it should be okay to just check for this log record type? I wasn't sure if I might also want to look for the "pending" type

engine/apps/alerts/escalation_snapshot/escalation_snapshot_mixin.py
Ferril: Please check if there will be no issues in places where we check `raw_escalation_snapshot` or `escalation_snapshot` since it also won't be `None`.
For example, we may encounter issues in escalate alert group and incident log builder (not sure, need to check)

engine/apps/alerts/tasks/check_escalation_finished.py
Ferril: It happens when escalation chain doesn't have any escalation steps. I think it is a valid case. But it may work different with [this](https://github.com/grafana/oncall/pull/1266/files#diff-3659a7ccf1ac8d0447f81d672f472b61355c2ac077b8188b13be982687263822R88) changes :thinking:

engine/apps/alerts/tasks/check_escalation_finished.py
Ferril: `next_step_eta` could be `None` if there are no escalation steps in the chain or escalation was not started. After executing the first escalation step, this field is set to some time value, depending on the step type. I'm not sure if there is any other way to make this happen.
But since you check above if escalation policies exist, I think, any other cases of `next_step_eta is None` should be invalid here 🤔

engine/apps/alerts/tasks/check_escalation_finished.py
Ferril: What do you mean by "pending type"?

I think for all successfully finished escalations this log should exist. 
But there can be a case, when escalation was finished, this log was created, and then user started escalation again (for example, resolve/unresolve). Seems like in this case it is possible to have the log record `TYPE_ESCALATION_FINISHED` and unfinished escalation at the same time 🤔

engine/apps/alerts/tasks/check_escalation_finished.py
```diff
@@ -1,48 +1,130 @@
...
-    exception_text = "\n".join(
-        exception_template.format(alert_group.pk, alert_group.estimate_escalation_finish_time, now)
-        for alert_group in alert_groups
-    )
+        task_logger.info("There are no alert groups to audit, everything is good :)")
 
-    ids = alert_groups.values_list("pk", flat=True)
-    task_logger.debug(ids)
+    for alert_group in alert_groups:
+        audit_alert_group_escalation(alert_group)
```
Ferril, line 128: As far as I understand this checking will stop after finding the first "stuck" alert group. So we won't collect the ids "stuck" of alert groups, right?

engine/apps/alerts/tasks/check_escalation_finished.py
joeyorlando: good point. I'll raise an exception here 👍

engine/apps/alerts/tasks/check_escalation_finished.py
joeyorlando: > What do you mean by "pending type"?

ah sorry, ignore this. I thought there was a `AlertGroupLogRecord.TYPE_ESCALATION_PENDING` log record type, doesn't look like there is.

> But there can be a case, when escalation was finished, this log was created, and then user started escalation again (for example, resolve/unresolve). Seems like in this case it is possible to have the log record TYPE_ESCALATION_FINISHED and unfinished escalation at the same time

do you have any suggestions for the query in `has_finished_log_record` [method](https://github.com/grafana/oncall/pull/1266/files#diff-167ec80d5717a06a85c122857926e50b393abfdcae1fe511ca83ac00f256e23fR108-R113) on how to handle this edge case?

engine/apps/alerts/tasks/check_escalation_finished.py
joeyorlando: good point. you're correct, currently this would raise on the first stuck alert group, and skip checking the rest. I'll modify this

joeyorlando: this should be handled in [this commit](https://github.com/grafana/oncall/pull/1266/commits/ef7fce9ba0c7e5e5a42eb77b0cc552033b2c8871). Added a test for this scenario.

engine/apps/alerts/escalation_snapshot/escalation_snapshot_mixin.py
joeyorlando: I've added a suite of tests for the `EscalationSnapshotMixin` class ([commit](https://github.com/grafana/oncall/pull/1266/commits/21424c15c7bab84d9dec99d276e4ca441d14cf1e)), so all should be good there.

But it seems there are a few spots that would need to be further refactored:
- [`apps/alerts/incident_log_builder/incident_log_builder.py`](https://github.com/grafana/oncall/blob/dev/engine/apps/alerts/incident_log_builder/incident_log_builder.py#L124-L240)
- [`apps/alerts/tasks/escalate_alert_group.py`](https://github.com/grafana/oncall/blob/dev/engine/apps/alerts/tasks/escalate_alert_group.py#L62-L82)
- [`apps/alerts/tasks/notify_all.py`](https://github.com/grafana/oncall/blob/dev/engine/apps/alerts/tasks/notify_all.py#L26-L81)
- [`apps/alerts/tasks/notify_group.py`](https://github.com/grafana/oncall/blob/dev/engine/apps/alerts/tasks/notify_group.py#L37-L74)

The main thing is how to properly refactor the checks like:
```python
if escalation_snapshot is not None:
   # do something
```

@Ferril would you mind giving some guidance on this (or hoping in to help refactor these few spots? 😄 )

Ferril: In the case `incident_log_builder.py`:
we can check if there are any escalation policies in escalation snapshot instead. By this checking we avoid unnecessary requests to db.

In `escalate_alert_group.py`:
if we remove this check, escalation should go like there are just no escalation policies. I'm not sure if we need to make any distinction between "no escalation policies" and "no escalation chain". If it is important, we can check if `escalation_chain` in the snapshot is not `None`. The difference is in additional log and `is_escalation_finished` flag.

In cases
`apps/alerts/tasks/notify_all.py`
`apps/alerts/tasks/notify_group.py`
it looks like this check doesn't make any sense now because we work with escalation snapshot like it is not `None` a few lines above in the same task 🤔

engine/apps/alerts/escalation_snapshot/snapshot_classes/escalation_policy_snapshot.py
```diff
@@ -105,6 +105,13 @@ def next_user_in_sorted_queue(self) -> User:
         next_user = users_queue[(last_user_index + 1) % len(users_queue)]
         return next_user
 
+    def has_finished_log_record(self, alert_group_id: int) -> bool:
+        return AlertGroupLogRecord.objects.filter(
+            escalation_policy_id=self.id,
+            alert_group_id=alert_group_id,
+            type=AlertGroupLogRecord.TYPE_ESCALATION_FINISHED,
```
Ferril, line 112: This type of logs doesn't have `escalation_policy_id` because we create it when there are no more escalation steps left

engine/apps/alerts/tasks/check_escalation_finished.py
Ferril: I don't have any suggestions how to do it in one query 😞  The main thing is that we should check if escalation wasn't triggered again after this log, like there are no `TYPE_ESCALATION_TRIGGERED` or `TYPE_ESCALATION_FAILED` logs after `TYPE_ESCALATION_FINISHED` log

engine/apps/alerts/escalation_snapshot/snapshot_classes/escalation_policy_snapshot.py
joeyorlando: good catch 👍 this should be addressed in [this commit](https://github.com/grafana/oncall/pull/1266/commits/c78e45f48f3a1f0099f969a7cce8b69b020b78e5)

engine/apps/alerts/escalation_snapshot/escalation_snapshot_mixin.py
joeyorlando: for the `incident_log_builder.py` case, this should be addressed in [this commit](https://github.com/grafana/oncall/pull/1266/commits/992200f2f0468761e44612aec0fbcd489716843f).

If I understand your comments correctly on the other three files, those _should_ be safe?

Ferril: Right 👍

Matvey-Kuk (approved):

docs/sources/open-source/_index.md
```diff
@@ -234,3 +234,23 @@ For Grafana OnCall OSS, the mobile app QR code includes an authentication token
 Your Grafana OnCall OSS instance should be reachable from the same network as your mobile device, preferably from the internet.
 
 For more information, see [Grafana OnCall mobile app]({{< relref "../mobile-app" >}})
+
+## Alert Group Escalation Auditor
```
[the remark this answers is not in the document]
joeyorlando: good suggestion! addressed in [this commit](https://github.com/grafana/oncall/pull/1266/commits/e0e273eff7a8bb2462252a58794aa8299b3d6b96)

PR #1267, opened 2023-02-01
Better FCM error handling / retries

vadimkerr:
# What this PR does
Raise `FirebaseError` in celery tasks contacting FCM instead of just logging it + add tests

## Checklist

- [x] Tests updated

### DIFF
--- a/engine/apps/mobile_app/fcm_relay.py
+++ b/engine/apps/mobile_app/fcm_relay.py
@@ -3,6 +3,7 @@ import logging
 from celery.utils.log import get_task_logger
 from django.conf import settings
 from fcm_django.models import FCMDevice
+from firebase_admin.exceptions import FirebaseError
 from firebase_admin.messaging import APNSConfig, APNSPayload, Aps, ApsAlert, CriticalSound, Message
 from rest_framework import status
 from rest_framework.permissions import IsAuthenticated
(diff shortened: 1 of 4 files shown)


engine/apps/mobile_app/fcm_relay.py
```diff
@@ -49,14 +50,17 @@ def post(self, request):
...
     message = Message(token=token, data=data, apns=deserialize_apns(apns))
 
-    task_logger.debug(f"Sending message to FCM: {message}")
-    result = FCMDevice(registration_id=token).send_message(message)
-    task_logger.debug(f"FCM response: {result.success}")
+    # https://firebase.google.com/docs/cloud-messaging/http-server-ref#interpret-downstream
+    response = FCMDevice(registration_id=token).send_message(message)
+    task_logger.debug(f"FCM response: {response}")
+
+    if isinstance(response, FirebaseError):
```
matiasb, line 62: Eventually would it be worth to check the status code and retry only if it makes sense?

matiasb (approved):
LGTM, one inline question

engine/apps/mobile_app/fcm_relay.py
vadimkerr: Yes, I think this makes sense not to retry for 4xx errors for example. But I'm not sure how this will behave on production just yet, so I propose to narrow this down later.

PR #1270, opened 2023-02-01
Fix local dev setup slowness

vadimkerr:
# What this PR does
Fixes an issue when a local dev setup becomes extremely slow.

- Set `DEBUG` and `SILK_PROFILER_ENABLED` to `False` by default + add utility make commands to toggle it
- Use `uwsgi` instead of Django's built-in `runserver` for local dev setup
- Limit Celery concurrency to 3 for local dev setup (previously was 20, used >1GB RAM on my machine)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -126,6 +126,17 @@ engine-manage:
 exec-engine:
 	docker exec -it oncall_engine bash
 
+_backend-debug-enable:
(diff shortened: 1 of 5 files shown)


iskhakov:
I like the idea to switch to uwsgi in local development as it is closer to what we have in production

iskhakov (approved):

joeyorlando (approved):
wow, this is way faster, good stuff!
🚢 🚢 🚢

PR #1272, opened 2023-02-02
Rares/coming soon

teodosii:
# What this PR does

Added Coming Soon label for `iOS`

- [x] `CHANGELOG.md` updated

### DIFF
--- a/grafana-plugin/src/pages/incident/Incident.helpers.tsx
+++ b/grafana-plugin/src/pages/incident/Incident.helpers.tsx
@@ -13,12 +13,13 @@ import { User } from 'models/user/user.types';
 import SilenceDropdown from 'pages/incidents/parts/SilenceDropdown';
 import { move } from 'state/helpers';
 import { UserActions } from 'utils/authorization';
+import { COLOR_DANGER, COLOR_PRIMARY, COLOR_SECONDARY, COLOR_WARNING } from 'utils/consts';
 
 export function getIncidentStatusTag(alert: Alert) {
   switch (alert.status) {
     case IncidentStatus.New:
       return (
-        <Tag color="#E02F44">
+        <Tag color={COLOR_DANGER}>
           <Text strong size="small">
             Firing
           </Text>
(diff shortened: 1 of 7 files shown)


grafana-plugin/src/utils/consts.ts
```diff
@@ -30,3 +30,8 @@ export const FARO_ENDPOINT_PROD =
 
 export const DOCS_SLACK_SETUP = 'https://grafana.com/docs/grafana-cloud/oncall/open-source/#slack-setup';
 export const DOCS_TELEGRAM_SETUP = 'https://grafana.com/docs/grafana-cloud/oncall/chat-options/configure-telegram/';
+
+export const COLOR_DANGER = '#E02F44';
+export const COLOR_WARNING = '#C69B06';
+export const COLOR_PRIMARY = '#299C46';
+export const COLOR_SECONDARY = '#464C54';
```
joeyorlando, line 37: these three hex-codes seem to be used in a few spots elsewhere in the codebase. I'd suggest updating those spots as well to use these.

joeyorlando (approved):
LGTM. may need to update one of the `jest` snapshots

grafana-plugin/src/pages/incident/Incident.helpers.tsx
```diff
@@ -13,36 +13,37 @@ import { User } from 'models/user/user.types';
...
 import { move } from 'state/helpers';
 import { UserActions } from 'utils/authorization';
+import { COLOR_DANGER, COLOR_PRIMARY, COLOR_SECONDARY, COLOR_WARNING } from 'utils/consts';
 
 export function getIncidentStatusTag(alert: Alert) {
   switch (alert.status) {
     case IncidentStatus.New:
       return (
-        <Tag color="#E02F44">
+        <Tag color={COLOR_DANGER}>
```
maskin25, line 22: is it possible to pass string like `color="var(--color-danger)"`? and add `--color-danger` to vars.css as all other color consts

teodosii: Let's do this in a separate PR

grafana-plugin/src/utils/consts.ts
teodosii: You are right :thumbsup:, I just pulled those from `Tag` because I knew they were used there but I wasn't aware of other usages. Will do this in a separate PR

PR #1277, opened 2023-02-02
Add sequence and last-modified info to custom shifts

matiasb:
There are some issues with exported ical for Terraform-based schedules (e.g. google calendar not syncing on changes). When terraform applies changes we update existing shifts without changing their UUID, which means there is no indication of the event having changed for clients reading the exported ical file.
To fix that, this PR introduces `SEQUENCE` and `LAST-MODIFIED` fields, keeping those values up to date for custom shifts, so that the generated ical file includes the information and indicates an event was updated [1]. This should help with Google calendar sync.

[1] https://stackoverflow.com/questions/60560457/google-doesnt-sync-my-subscribed-ics-feed


engine/apps/schedules/ical_utils.py
```diff
@@ -589,6 +589,8 @@ def create_base_icalendar(name: str) -> Calendar:
     cal.add("x-wr-calname", name)
     cal.add("x-wr-timezone", "UTC")
     cal.add("prodid", "//Grafana Labs//Grafana On-Call//")
+    # suggested minimum interval for polling for changes
+    cal.add("REFRESH-INTERVAL;VALUE=DURATION", "P1H")
```
vadimkerr, line 593: Does it mean that Google calendar will refresh an exported ICal file every hour **at most**?
If that's the case, is it possible to set a lower value to get it update faster?

vadimkerr (approved):
LGTM! 
Just one minor question and one nitpick suggestion.

engine/apps/schedules/ical_utils.py
matiasb: Not really, this is just a "suggestion" on how much (minimum) time wait before trying a refresh... Google is expected to still sync every 12-24hs. I didn't want to set a too lower threshold just in case some other clients keep hitting the endpoint (and 1h seemed frequent enough?).

engine/apps/schedules/models/custom_on_call_shift.py
```diff
@@ -212,9 +212,21 @@ class CustomOnCallShift(models.Model):
         related_name="parent_shift",
     )
 
+    last_modified = models.DateTimeField(default=None, null=True)
+    sequence = models.IntegerField(default=None, null=True)
```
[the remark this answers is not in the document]
matiasb: Let me check (in this case may be worth it given we don't have too many instances in the DB)

PR #1280, opened 2023-02-03
Include alert details in Grafana Incident alert-group endpoint

sd2k:
This will be used by Grafana Incident to infer details about the
incident if alert groups are present.

Depends on #1279

### DIFF
--- a/engine/apps/api_for_grafana_incident/serializers.py
+++ b/engine/apps/api_for_grafana_incident/serializers.py
@@ -3,12 +3,11 @@ import logging
 from rest_framework import serializers
 
 from apps.alerts.models import Alert, AlertGroup
-from common.api_helpers.mixins import EagerLoadingMixin
 
 logger = logging.getLogger(__name__)
 
 
-class AlertSerializer(EagerLoadingMixin, serializers.ModelSerializer):
+class AlertSerializer(serializers.ModelSerializer):
 
     id_oncall = serializers.CharField(read_only=True, source="public_primary_key")
     payload = serializers.JSONField(read_only=True, source="raw_request_data")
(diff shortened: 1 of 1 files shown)


engine/apps/api_for_grafana_incident/serializers.py
```diff
@@ -2,17 +2,31 @@
...
-from apps.alerts.models import AlertGroup
+from apps.alerts.models import Alert, AlertGroup
 from common.api_helpers.mixins import EagerLoadingMixin
 
 logger = logging.getLogger(__name__)
 
 
+class AlertSerializer(EagerLoadingMixin, serializers.ModelSerializer):
+
+    id_oncall = serializers.CharField(read_only=True, source="public_primary_key")
```
iskhakov, line 13: I think we can call it just `id` as it is part of oncall api already

engine/apps/api_for_grafana_incident/serializers.py
```diff
@@ -2,17 +2,31 @@
...
 from rest_framework import serializers
 
-from apps.alerts.models import AlertGroup
+from apps.alerts.models import Alert, AlertGroup
 from common.api_helpers.mixins import EagerLoadingMixin
 
 logger = logging.getLogger(__name__)
 
 
+class AlertSerializer(EagerLoadingMixin, serializers.ModelSerializer):
```
iskhakov, line 11: `EagerLoadingMixin` is not used here and can be removed.

It is not needed in AlertGroupSerializer below as well, as it doesn't work without PREFETCH_RELATED or SELECT_RELATED properties.

iskhakov (approved):
Overall looks good to me, thank you!
Approved, but left a few nitpicks

sd2k:
Thanks @iskhakov! I'll wait until #1279 is approved before merging.

PR #1286, opened 2023-02-03
Enable mobile app backend by default on OSS

vadimkerr:
# What this PR does
Enables mobile app backend by default on OSS.

## Checklist
- [x] `CHANGELOG.md` updated


joeyorlando:
how would this affect [this](https://github.com/grafana/oncall/blob/dev/engine/settings/base.py#L571)?

I think `initialize_app` _might_ complain. In `docker-compose-developer.yml` it's set [here](https://github.com/grafana/oncall/blob/dev/docker-compose-developer.yml#L28), but I'm wondering if that is a mistake. OSS users should not need to care about which GCP project this uses, they'll only be communicating with a cloud FCM relay, right?

vadimkerr:
@joeyorlando seems like `initialize_app` won't complain unless you don't use the `FIREBASE_APP` (which OSS instances don't use, OSS instances use the FCM relay by default, right).

joeyorlando (approved):

PR #1315, opened 2023-02-09
Fix route regex debugger to use only alerts from certain integration

iskhakov:
# What this PR does

Right now route regex debugger searches for every alertgroup in the team. We only need to compare alerts from the current integration when adding the route

## Checklist

- [ ] Tests updated
- [ ] Documentation added
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/engine/apps/api/views/route_regex_debugger.py
+++ b/engine/apps/api/views/route_regex_debugger.py
@@ -7,7 +7,7 @@ from rest_framework.response import Response
 from rest_framework.views import APIView
 
 from apps.alerts.incident_appearance.renderers.web_renderer import AlertWebRenderer
-from apps.alerts.models import Alert, AlertGroup
+from apps.alerts.models import Alert, AlertGroup, AlertReceiveChannel
 from apps.auth_token.auth import PluginAuthentication
 from common.api_helpers.exceptions import BadRequest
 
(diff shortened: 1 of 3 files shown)


grafana-plugin/src/containers/ChannelFilterForm/ChannelFilterForm.tsx
```diff
@@ -101,6 +101,7 @@ const ChannelFilterForm = observer((props: ChannelFilterFormProps) => {
       {!data?.is_default && (
         <IncidentMatcher
           regexp={filteringTerm}
+          alert_receive_channel_id={alertReceiveChannelId}
```
joeyorlando, line 104: _nitpick_ I believe React props tend to be camel cased vs snake case

engine/apps/api/views/route_regex_debugger.py
```diff
@@ -31,12 +32,28 @@ def get(self, request):
...
+                    team_id=team,
+                ).values_list("id", flat=True)
+            )
+        else:
+            alert_receive_channels_ids = list(
+                AlertReceiveChannel.objects.get(public_primary_key=alert_receive_channel_id).values_list(
+                    "id", flat=True
+                )
+            )
+
```
joeyorlando, line 50: what about if we made a v2 version of this api endpoint? it would avoid having to have backwards-compatible logic like this within the endpoint

iskhakov: I want to add some more thought on how we would start versioning internal api without duplicating the code. Let's merge this for now and discuss how to do that

grafana-plugin/src/containers/ChannelFilterForm/ChannelFilterForm.tsx
iskhakov: 🤦 thank you

Issue #1283, opened 2023-02-03
PagerDuty migrator: migrate on-call shifts


Matvey-Kuk:
Current implementation of PagerDuty migrator doesn't migrate on-call shifts, but only "imports" the ICal file from PagerDuty. This makes migrated schedules read-only. This is listed as one of the limitations of the tool: https://github.com/grafana/oncall/tree/dev/tools/pagerduty-migrator#limitations

Migrating on-call shifts instead of the ICal file will allow editing migrated schedules from the web UI directly and improve the migration process.

carlmagnusholmstrom-tomtom:
Actually, it does migrate the schedules, but they are read-only in the UI.

Matvey-Kuk:
Fixed & rolled to production.

PR #1317, opened 2023-02-09
PD migrator: migrate on-call shifts using public API

vadimkerr:
Allow PD migrator tool to migrate on-call shifts when migrating schedules (currently it migrates schedules using PD ICal file): https://github.com/grafana/oncall/issues/1283.

This PR will allow to select the mode of schedule migration via `SCHEDULE_MIGRATION_MODE_WEB` env variable (`ical` or `web`). Due to differences in the scheduling systems of PD and OnCall, it's not always possible to migrate shifts automatically (migration plan will show any schedules and layers that can't be migrated).

PD rotations that will be possible to migrate:
- Any rotation without restrictions ("restriction" is a PD term for describing active periods for rotation)
- Daily rotations with daily restrictions
- Weekly rotations with weekly restrictions
- Some weekly rotations with daily restrictions
- Some daily rotations with weekly restrictions

There will be a separate PR to update the [instruction](https://github.com/grafana/oncall/tree/dev/tools/pagerduty-migrator#readme) since this one is pretty huge already.


engine/apps/public_api/views/schedules.py
```diff
@@ -56,9 +56,6 @@ def get_object(self):
             raise NotFound
 
     def perform_create(self, serializer):
-        if serializer.validated_data["type"] == "web":
```
vadimkerr, line 59: Discussed this with @matiasb, seems like we're fine with allowing to create web shifts using public API

Konstantinov-Innokentii: If we don't want to allow to create web schedules from api, we can check User-Agent to allow to create them only from migrator. One reason to do that is because we have no any docs about creating web schedules from the public api.

Konstantinov-Innokentii (approved):

PR #1319, opened 2023-02-10
Jinja2 based routes

iskhakov:
# What this PR does

This PR adds the new way to set up routes using jinja2 templating language

Screenshot 2023-03-06 at 22 11 13
Screenshot 2023-03-06 at 22 11 34

## Checklist

- [ ] Tests updated
- [ ] Documentation added
- [ ] `CHANGELOG.md` updated


grafana-plugin/src/containers/ChannelFilterForm/ChannelFilterForm.tsx
```diff
@@ -33,7 +33,7 @@ const ChannelFilterForm = observer((props: ChannelFilterFormProps) => {
   const [filteringTermJinja2, setFilteringTermJinja2] = useState<string>(
     data ? data.filtering_term_jinja2 : '{payload.foo == "bar"}'
   );
-  const [filteringTermType, setFilteringTermType] = useState<number>(data ? data.filtering_term_type : 1);
+  const [filteringTermType, setFilteringTermType] = useState<FilteringTermType>(data ? data.filtering_term_type : 1);
```
iskhakov, line 36: @maskin25 can you please help

When I try using enum Value like here:

```
const [filteringTermType, setFilteringTermType] = useState<FilteringTermType>(data ? data.filtering_term_type : FilteringTermType.jinja2);
  
```

frontend fails with 


```
TypeError: undefined is not an object (evaluating 'models_channel_filter_channel_filter_types__WEBPACK_IMPORTED_MODULE_8__.FilteringTermType.jinja2')
```
and compliler shows:
```
Compiled with warnings!
  Attempted import error: 'FilteringTermType' is not exported from 'models/channel_filter/channel_filter.types' (imported as 'FilteringTermType').
  No errors found.

```
Screenshot 2023-02-10 at 21 48 15

grafana-plugin/src/models/channel_filter/channel_filter.types.ts
```diff
@@ -3,10 +3,10 @@ import { EscalationChain } from 'models/escalation_chain/escalation_chain.types'
 import { SlackChannel } from 'models/slack_channel/slack_channel.types';
 import { TelegramChannel } from 'models/telegram_channel/telegram_channel.types';
 
-// export const enum FilteringTermType {
-//   regex=0,
-//   jinja2=1,
-// }
+export const enum FilteringTermType {
```
maskin25, line 6: `export enum`

engine/apps/api/serializers/channel_filter.py
```diff
@@ -90,6 +91,10 @@ def validate_filtering_term(self, filtering_term):
                 raise serializers.ValidationError(["Filtering term is incorrect"])
         return filtering_term
 
+    def validate_filtering_term_jinja2(self, filtering_term_jinja2):
+        # TODO: check if valid
+        return filtering_term_jinja2
```
joeyorlando, line 96: `apply_jinja_template` function from `engine/common/jinja_templater` might be useful here

engine/apps/alerts/models/channel_filter.py
```diff
@@ -68,6 +70,16 @@ class ChannelFilter(OrderedModel):
 
     created_at = models.DateTimeField(auto_now_add=True)
     filtering_term = models.CharField(max_length=1024, null=True, default=None)
```
joeyorlando, line 72: could we get away with having just the `filtering_term` field?

Store `filtering_term_jinja2` inside of `filtering_term` and use the value of `filtering_term_type` to figure out whether or not it's regex or jinja

engine/apps/alerts/models/alert_group_log_record.py
```diff
@@ -262,6 +262,7 @@ def rendered_log_line_action(self, for_slack=False, html=False, substitute_autho
             escalation_chain = self.alert_group.escalation_chain_with_respect_to_escalation_snapshot
 
             if channel_filter is not None:
+                print(f"$$$$$$$$$$$$$$$$$$$$$$$$$ 1232 {channel_filter.str_for_clients}")
```
joeyorlando, line 265:
```suggestion
```

engine/apps/alerts/models/channel_filter.py
iskhakov: yep, thank you, I was not using `filtering_term_type` in the first prototype

I had to add a bit more complicated logic to default values on the frontend, but worth of it

engine/apps/api/serializers/channel_filter.py
```diff
@@ -108,7 +92,14 @@ def validate_notification_backends(self, notification_backends):
                 # update existing backend data
                 updated[backend_id] = updated.get(backend_id, {}) | updated_data
             notification_backends = updated
-        return notification_backends
+        return
```
vadimkerr, line 95: Seems like this change is not intended?

vadimkerr (approved):
LGTM, probably a good idea to add a couple tests on this, e.g. somewhere here: https://github.com/grafana/oncall/blob/19f1491e3ab4b15e8872ab8f5f4c7d4a74af416b/engine/apps/alerts/tests/test_channel_filter.py#L9

engine/apps/api/serializers/channel_filter.py
iskhakov: yep

engine/apps/alerts/migrations/0010_channelfilter_filtering_term_type.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('alerts', '0009_alertreceivechannel_web_templates_modified_at'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='channelfilter',
+            name='filtering_term_type',
+            field=models.IntegerField(choices=[(0, 'regex'), (1, 'jinja2')], default=None, null=True),
```
Konstantinov-Innokentii, line 16: Can we have this field non-nullable, but with default value 0?

engine/apps/api/serializers/channel_filter.py
```diff
@@ -110,6 +95,13 @@ def validate_notification_backends(self, notification_backends):
             notification_backends = updated
         return notification_backends
 
+    def to_representation(self, obj):
```
Konstantinov-Innokentii, line 98: See comment about migration, I think it's possible to get rid of that.

engine/apps/api/serializers/channel_filter.py
```diff
@@ -56,22 +57,6 @@ def get_slack_channel(self, obj):
             "id": obj.slack_channel_pk,
         }
 
-    def validate(self, attrs):
```
Konstantinov-Innokentii, line 59: Where jinja template is validated?

engine/apps/public_api/serializers/routes.py
```diff
@@ -176,20 +176,6 @@ def create(self, validated_data):
 
         return instance
 
-    def validate(self, attrs):
```
Konstantinov-Innokentii, line 179: Same question as in internal API - where jinja template is validated?

engine/apps/alerts/models/channel_filter.py
```diff
@@ -107,20 +118,29 @@ def select_filter(cls, alert_receive_channel, raw_request_data, title, message=N
...
+        if self.filtering_term_type == ChannelFilter.FILTERING_TERM_TYPE_JINJA2:
+            try:
+                is_matching = apply_jinja_template(self.filtering_term, payload=value)
+                return is_matching.strip().lower() in ["1", "true", "ok"]
+            except (JinjaTemplateError, JinjaTemplateWarning):
+                logger.error(f"channel_filter={self.id} failed to parse jinja2={self.filtering_term}")
+                return False
+        if self.filtering_term is not None and self.filtering_term_type == ChannelFilter.FILTERING_TERM_TYPE_REGEX:
+            try:
+                return re.search(self.filtering_term, json.dumps(value))
```
Konstantinov-Innokentii, line 139: json.dumps(value) is introduced here, could you please describe why?

Konstantinov-Innokentii:
Great idea, left comments. It will be great to support new filter term type in public api and in the terraform provider. Also, i think it is needed to provide some info in [docs](https://grafana.com/docs/oncall/latest/escalation-policies/configure-routes/#create-a-route) about jinja based routes.

engine/apps/alerts/models/channel_filter.py
iskhakov: it's moved from here https://github.com/grafana/oncall/pull/1319/files#diff-5b78eae626bcf20076ff2f5f25daa3f5703ff8dd9336e5d3e3dbbd7aa827d6b2L116

engine/apps/alerts/migrations/0010_channelfilter_filtering_term_type.py
Konstantinov-Innokentii: I think it's valid tradeoff to experience 500's to have more consistent database and codebase. With nullable we need a lot of code to show this field in private/public API correctly and to check for null to write to that field from API.

engine/apps/api/serializers/channel_filter.py
iskhakov: This was used to ensure that the routes are based on the unique regular expressions. I decided to drop uniqueness, if user wants to create exactly same route which will never get it's alert, let them do that, may be they'll want to modify it later.

engine/apps/alerts/migrations/0010_channelfilter_filtering_term_type.py
```diff
@@ -0,0 +1,26 @@
...
+
+    operations = [
+        migrations.AddField(
+            model_name='channelfilter',
+            name='filtering_term_type',
+            field=models.IntegerField(choices=[(0, 'regex'), (1, 'jinja2')], default=0),
+        ),
+        # Adding same default value on the database level
+        # for database backwards-compatibility with older versions of code
+        AddDefaultValue(
```
iskhakov, line 21: this is possible with RunSQL to insert default values, but we'll need to do that manually for each database. django_add_default_value does that for us

Konstantinov-Innokentii: I didn't get that idea.  Why we need AddDefaultValue, when we have default=0?

iskhakov: I left better comment
for more details I recommend to read this [article](../engine/apps/alerts/migrations/0010_channelfilter_filtering_term_type.py)

Konstantinov-Innokentii (approved):

PR #1327, opened 2023-02-15
PD migrator: handle resource names with extra spaces

vadimkerr:
# What this PR does
Currently PD migrator can fail with HTTP 400 error when dealing with PD resource names like " Schedule" (note the extra space). This PR makes sure that PD migrator script can handle such cases.

### DIFF
--- a/tools/pagerduty-migrator/migrator/report.py
+++ b/tools/pagerduty-migrator/migrator/report.py
@@ -47,7 +47,7 @@ def format_escalation_policy(policy: dict) -> str:
 
 
 def format_integration(integration: dict) -> str:
-    result = integration["service"]["name"] + " - " + integration["name"]
+    result = "{} - {}".format(integration["service"]["name"], integration["name"])
 
     if not integration["oncall_type"]:
         result = (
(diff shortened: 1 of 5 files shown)


tools/pagerduty-migrator/migrator/report.py
```diff
@@ -47,7 +47,7 @@ def format_escalation_policy(policy: dict) -> str:
 
 
 def format_integration(integration: dict) -> str:
-    result = integration["service"]["name"] + " - " + integration["name"]
+    result = "{} - {}".format(integration["service"]["name"], integration["name"])
```
joeyorlando, line 50: _total nit_
```suggestion
    result = f"{integration['service']['name']} - {integration['name']}"
```

tools/pagerduty-migrator/migrator/resources/integrations.py
```diff
@@ -6,10 +6,12 @@
 def match_integration(integration: dict, oncall_integrations: list[dict]) -> None:
     oncall_integration = None
     for candidate in oncall_integrations:
-        name = "{} - {}".format(
-            integration["service"]["name"], integration["name"]
-        ).lower()
-        if candidate["name"].lower() == name:
+        name = (
+            "{} - {}".format(integration["service"]["name"], integration["name"])
```
joeyorlando, line 10: up to you on the f-string here, set nit comment above

joeyorlando (approved):

tools/pagerduty-migrator/migrator/report.py
vadimkerr: I kinda prefer `format` here, seems to be easier to understand what the result would be from a quick look. Probably a taste thing 😁

PR #1340, opened 2023-02-19
Refactor maintenance

Konstantinov-Innokentii:
# What this PR does
This PR simplifies code of maintenance mode.
1. Perform distribution/escalation maintenance checks in send_signal... tasks.
2. Use usual alert distribution flow for the maintenance incident.
3. Decouple maintenance mode from slack (all, except **notify_about_maintenance_action** methods, I don't want to make this PR too big)

As a bonus from these changes, maintenance mode now mute alert group delivery in all chatops integrations, not only in slack. (Before, incidents happened while maintenance were posted to telegram and msteams anyway)

## Checklist

- [ ] Tests updated
- [ ] Documentation added
- [ ] `CHANGELOG.md` updated


engine/apps/alerts/escalation_snapshot/escalation_snapshot_mixin.py
```diff
@@ -245,6 +245,11 @@ def start_escalation_if_needed(self, countdown=ScenarioStep.CROSS_ACTION_DELAY,
         """
         AlertGroup = apps.get_model("alerts", "AlertGroup")
 
+        is_on_maintenace_or_debug_mode = (
```
vadimkerr, line 248: This is used ~3 times in this PR, so probably makes sense to move it to `AlertReceiveChannel` as a property

engine/apps/alerts/incident_appearance/renderers/slack_renderer.py
```diff
@@ -169,7 +169,7 @@ def _get_buttons_attachments(self):
     def _get_buttons_blocks(self):
         AlertGroup = apps.get_model("alerts", "AlertGroup")
         buttons = []
-        if self.alert_group.maintenance_uuid is None:
+        if not self.alert_group.is_maintenance_incident:
```
vadimkerr, line 172: Consider renaming `is_maintenance_incident` to `is_maintenance_alert_group` for consistency

engine/apps/alerts/models/alert.py
```diff
@@ -127,29 +127,30 @@ def create(
...
-                )
-                logger.debug(
-                    f"call send_alert_group_signal for alert_group {maintenance_incident.pk} (maintenance), "
-                    f"log record {log_record_for_root_incident.pk} with type "
-                    f"'{log_record_for_root_incident.get_type_display()}'"
-                )
-                send_alert_group_signal.apply_async((log_record_for_root_incident.pk,))
-            except AlertGroup.DoesNotExist:
-                pass
+        if group_created:
```
vadimkerr, line 130: nit
```
if not group_created:
    return alert

maintenance_uuid = None
...
```

also probably a comment would be useful (e.g. `# all the code below is related to maintenance`)

vadimkerr (approved):
LGTM 👍 
Couple of nitpick comments

engine/apps/alerts/models/alert.py
Konstantinov-Innokentii: I'll add the comment, thanks

engine/apps/alerts/escalation_snapshot/escalation_snapshot_mixin.py
Konstantinov-Innokentii: The problem is that I don't want to move this check to alert_recevie_channel, since it checks organization too. Also, sometimes it's needed to check only debug/maintenance mode, or check it only for organization. You are right, that it's duplication of code, but I don't want to sacrifice SRP principle in favour of DRY. I have a concept of Maintenace v2, which should simplify all concept and code even further  and will require no such checks at all.

PR #1373, opened 2023-02-21
Add Google reCAPTCHA for mobile app phone verification

joeyorlando:
# What this PR does

Adds reCAPTCHA validation to the get mobile verification code endpoint

## Checklist

- [x] Tests updated
- [x] `CHANGELOG.md` updated

### DIFF
--- a/engine/apps/api/views/user.py
+++ b/engine/apps/api/views/user.py
@@ -298,6 +298,8 @@ class UserView(
     )
     def get_verification_code(self, request, pk):
         if not settings.OSS_INSTALLATION:
+            logger.info("Validating reCAPTCHA code")
+
             serializer = MobileVerificationCodeRecaptchaSerializer(
                 data={"recaptcha": request.headers["X-OnCall-Recaptcha"]},
                 context={"request": request},
@@ -306,6 +308,7 @@ class UserView(
             if not serializer.is_valid():
                 logger.warning(f"Invalid reCAPTCHA validation: {serializer.fields['recaptcha']}")
                 return Response(status=status.HTTP_400_BAD_REQUEST)
+            logger.info("reCAPTCHA code is valid")
 
         user = self.get_object()
         phone_manager = PhoneManager(user)
--- a/engine/apps/api/views/user.py
+++ b/engine/apps/api/views/user.py
@@ -297,18 +297,22 @@ class UserView(
         throttle_classes=[GetPhoneVerificationCodeThrottlerPerUser, GetPhoneVerificationCodeThrottlerPerOrg],
     )
     def get_verification_code(self, request, pk):
-        if not settings.DRF_RECAPTCHA_TESTING:
-            logger.info("Validating reCAPTCHA code")
+        """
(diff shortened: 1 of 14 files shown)


grafana-plugin/src/containers/UserSettings/parts/tabs/PhoneVerification/PhoneVerification.tsx
```diff
@@ -99,26 +100,32 @@ const PhoneVerification = observer((props: PhoneVerificationProps) => {
...
+                if (codeInputRef.current) {
+                  codeInputRef.current.focus();
+                }
+              })
+              .catch(() => {
+                openErrorNotification(
+                  'Grafana OnCall is unable to verify your phone number due to incorrect number or verification service being unavailable.'
+                );
+              });
+          });
```
joeyorlando, line 127: on OSS installations we don't need to make a call to Google to fetch a client-side reCAPTCHA token. It's not required on the backend in that scenario.

grafana-plugin/src/utils/consts.ts
```diff
@@ -15,6 +15,9 @@ export const DEFAULT_PAGE = 'incidents';
 
 export const PLUGIN_ROOT = '/a/grafana-oncall-app';
 
+// https://developers.google.com/recaptcha/docs/v3
+export const reCAPTCHA_site_key = '6LeIPJ8kAAAAAJdUfjO3uUtQtVxsYf93y46mTec1';
```
joeyorlando, line 19: is there a way we can add constants like this into `process.env` at build time?

I'm a bit weary of adding constants like this because if we decide to add different recaptcha environments (ex. one for dev/staging/prod), how would we handle that here?

iskhakov (approved):

engine/apps/api/views/user.py
```diff
@@ -292,11 +297,29 @@ def timezone_options(self, request):
...
+        when the recaptcha checks are actually made
+        """
+        logger.info("Validating reCAPTCHA code")
+
+        serializer = MobileVerificationCodeRecaptchaSerializer(
+            data={"recaptcha": request.headers.get("X-OnCall-Recaptcha", "some-non-null-value")},
+            context={"request": request},
+        )
+
+        if not serializer.is_valid():
```
joeyorlando, line 312: since `DRF_RECAPTCHA_TESTING` is set to `True` in `settings/base.py` this will always return true (without the default of `some-non-null-value` this would complain about the serializer field being blank).

When `DRF_RECAPTCHA_TESTING` is set to `False` (ex. for cloud stacks), `drf_django` will make actual calls to the Google reCAPTCHA server with the value of `X-OnCall-Recaptcha` (and thus `some-non-null-value` would fail).

grafana-plugin/src/utils/consts.ts
maskin25: you are right, let's discuss https://raintank-corp.slack.com/archives/C04JCU51NF8/p1677059171447329

PR #1376, opened 2023-02-21
Fix the cause of retry of notify_all and notify_group tasks

Ferril:
Fix the cause of retry of notify_all and notify_group tasks that was related to an incorrect step order.

### DIFF
--- a/engine/apps/alerts/tasks/notify_all.py
+++ b/engine/apps/alerts/tasks/notify_all.py
@@ -18,8 +18,24 @@ def notify_all_task(alert_group_pk, escalation_policy_snapshot_order=None):
 
     alert_group = AlertGroup.all_objects.get(pk=alert_group_pk)
 
+    # check alert group state before notifying all users in the channel
+    if alert_group.resolved or alert_group.acknowledged or alert_group.silenced:
+        task_logger.info(f"alert_group {alert_group.pk} was resolved, acked or silenced forever. No need to notify all")
+        return
+
     escalation_snapshot = alert_group.escalation_snapshot
-    escalation_policy_snapshot = escalation_snapshot.escalation_policies_snapshots[escalation_policy_snapshot_order]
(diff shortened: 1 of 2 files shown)


engine/apps/alerts/tasks/notify_all.py
```diff
@@ -18,8 +18,24 @@ def notify_all_task(alert_group_pk, escalation_policy_snapshot_order=None):
...
     alert_group = AlertGroup.all_objects.get(pk=alert_group_pk)
 
+    # check alert group state before notifying all users in the channel
+    if alert_group.resolved or alert_group.acknowledged or alert_group.silenced:
+        task_logger.info(f"alert_group {alert_group.pk} was resolved, acked or silenced forever. No need to notify all")
+        return
+
     escalation_snapshot = alert_group.escalation_snapshot
-    escalation_policy_snapshot = escalation_snapshot.escalation_policies_snapshots[escalation_policy_snapshot_order]
+    escalation_policy_snapshot = escalation_snapshot.escalation_policies_snapshots.get(escalation_policy_snapshot_order)
```
vadimkerr, line 27: Isn't `escalation_policies_snapshots` a list? Seems like using `get` here won't work

vadimkerr (changes requested):
LGTM overall, requesting changes to address the comment.
This is good to stop tasks from retrying, ideally we need to figure out why it was scheduled incorrectly.

engine/apps/alerts/tasks/notify_all.py
Ferril: You are right, thank you, I'll fix it

Ferril:
> This is good to stop tasks from retrying, ideally we need to figure out why it was scheduled incorrectly.

The only way I found to reproduce it is to change step order manually using terraform.

vadimkerr (approved):

PR #1377, opened 2023-02-21
Revert "Rework schedules cached ical file values"

matiasb:
Reverts grafana/oncall#1312

This change seems to have introduced some unexpected behavior with slack user groups. Reverting to reproduce locally and push an improved update.


joeyorlando (approved):

matiasb:
After a closer look and discussion with @Konstantinov-Innokentii, this doesn't seem to be the real cause behind the issues described in https://github.com/grafana/oncall-private/issues/1641, so closing for now. Will need some more work to reproduce locally and find the problem.

PR #1401, opened 2023-02-23
Matiasb/fix task refresh ical when empty value

matiasb:
This should fix task error as seen in logs, trying to parse an empty string as ical value:
```
Task apps.schedules.tasks.refresh_ical_files.refresh_ical_file[] raised unexpected: ValueError("Found no components where exactly one is required: ''")
```

### DIFF
--- a/engine/apps/schedules/tasks/refresh_ical_files.py
+++ b/engine/apps/schedules/tasks/refresh_ical_files.py
@@ -41,9 +41,9 @@ def refresh_ical_file(schedule_pk):
         notify_ical_schedule_shift.apply_async((schedule.pk,))
 
     run_task_primary = False
-    if bool(schedule.cached_ical_file_primary):
+    if schedule.cached_ical_file_primary:
         # ie. primary schedule is not empty (None -> no ical, "" -> empty cached value)
-        if not bool(schedule.prev_ical_file_primary):
+        if not schedule.prev_ical_file_primary:
             # prev value is empty
             run_task_primary = True
             task_logger.info(f"run_task_primary {schedule_pk} {run_task_primary} prev_ical_file_primary is None")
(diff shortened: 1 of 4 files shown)


engine/apps/schedules/tasks/refresh_ical_files.py
```diff
@@ -41,22 +41,29 @@ def refresh_ical_file(schedule_pk):
         notify_ical_schedule_shift.apply_async((schedule.pk,))
 
     run_task_primary = False
-    if schedule.cached_ical_file_primary is not None:
-        if schedule.prev_ical_file_primary is None:
+    if bool(schedule.cached_ical_file_primary):
```
Konstantinov-Innokentii, line 44: Why to convert to bool? Can't we check if schedule.cached_ical_file_primary, which will handle None end ""?

Konstantinov-Innokentii:
@matiasb could you also check notify_ical_schedule_shift task for places with comparing ical to none? I'm sure there are some, there is one:

```
    for prev_ical_file, current_ical_file in prev_and_current_ical_files:
        if prev_ical_file is not None and (
            current_ical_file is None or not is_icals_equal(current_ical_file, prev_ical_file)
        ):
            # If icals are not equal then compare current_events from them
            is_prev_ical_diff = True
            prev_calendar = icalendar.Calendar.from_ical(prev_ical_file)

            prev_shifts_result, prev_users_result = get_current_shifts_from_ical(
                prev_calendar,
                schedule,
                prev_overrides_priority,
            )
            if prev_overrides_priority == 0 and prev_shifts_result:
                prev_overrides_priority = max([prev_shifts_result[uid]["priority"] for uid in prev_shifts_result]) + 1

            prev_shifts.update(prev_shifts_result)
            prev_users.update(prev_users_result)
```

engine/apps/schedules/tasks/refresh_ical_files.py
matiasb: I think it was more explicit, but switched back to just check for cached_ical_file_primary

matiasb:
> @matiasb could you also check notify_ical_schedule_shift task for places with comparing ical to none? I'm sure there are some, there is one:

Checked and updated. Also did a search for any other occurrences of similar scenarios (nothing to add)

mderynck (approved):

PR #1406, opened 2023-02-24
Schedule Filters tweaks

### LINKED ISSUE #941: OnCall Schedule Filters Ignored
Visiting the `Schedules` page, if filters are selected before the initial load of schedule is completed, the filters will be ignored. As an example:

* Click on Schedules
* Immediately select filters (Ex: `Unused` & `Web`)
* Expected Result: Only unused web schedules will show in th e list
* Actual Result: All schedules are listed

teodosii:
# What this PR does

Improvements over schedule filters

#941 

## Checklist

- [ ] Tests updated
- [ ] Documentation added
- [x] `CHANGELOG.md` updated


grafana-plugin/src/pages/schedules/Schedules.tsx
```diff
@@ -395,17 +412,21 @@ class SchedulesPage extends React.Component<SchedulesPageProps, SchedulesPageSta
...
-    const { filters } = this.state;
-    const { store } = this.props;
-    const { scheduleStore } = store;
-    scheduleStore.updateItems(filters);
+  applyFilters = (filters: SchedulesFiltersType) => {
+    const { scheduleStore } = this.props.store;
+
+    scheduleStore.updateItems(filters).finally(() => {
+      // check for any change in filters in the meanwhile
+      if (filters === this.state.filters) {
```
maskin25, line 423: but filters object is not a primitive, how do you compare them with just `===`

maskin25:
in my mind you just need to remove  line `store.scheduleStore.updateItems();` from `pages/schedules/Schedules.tsx -> componentDidMount` method, and add 
```
useEffect(() => {
    onChange(value);
  }, []);
```

in SchedulesFilters.tsx to immediately apply initial filters after `SchedulesFilters` mounts

grafana-plugin/src/pages/schedules/Schedules.tsx
```diff
@@ -155,23 +163,23 @@ class SchedulesPage extends React.Component<SchedulesPageProps, SchedulesPageSta
...
-              emptyText={
-                <div className={cx('loader')}>
-                  {data ? <Text type="secondary">Not found</Text> : <Text type="secondary">Loading schedules...</Text>}
-                </div>
-              }
-            />
+            {isLoading ? (
+              this.renderLoading()
+            ) : (
+              <Table
```
maskin25, line 169: Table itself has prop `loading`

maskin25: `<Table loading={!data} ...`

maskin25 (changes requested):

grafana-plugin/src/pages/schedules/Schedules.tsx
teodosii: This checks for reference as that's enough to know a change happened. It's overhead to check for actual change and doesn't bring any benefit I'd say.

This check is only so that we do not reset `isLoading` to false when there's like 2 requests ongoing.

grafana-plugin/src/pages/schedules/Schedules.tsx
```diff
@@ -196,6 +197,22 @@ class SchedulesPage extends React.Component<SchedulesPageProps, SchedulesPageSta
     );
   }
 
+  renderLoading() {
```
maskin25, line 200: seems it can be removed

grafana-plugin/src/pages/schedules/Schedules.tsx
```diff
@@ -394,18 +404,37 @@ class SchedulesPage extends React.Component<SchedulesPageProps, SchedulesPageSta
...
 
-  applyFilters = () => {
-    const { filters } = this.state;
-    const { store } = this.props;
-    const { scheduleStore } = store;
-    scheduleStore.updateItems(filters);
+  applyFilters = (filters: SchedulesFiltersType) => {
+    const { scheduleStore } = this.props.store;
+
+    scheduleStore.updateItems(filters).finally(() => {
```
maskin25, line 428: I think it's better to use  `then` because `finally` works for both success and failed requests

maskin25 (approved):

Issue #1423, opened 2023-02-28
Schedules: Status filter no longer working with pagination


teodosii:
We recently added pagination to the schedules endpoint and the client side filtering on status no longer works accordingly. 

We should change it so that the backend endpoint takes a parameter to filter out schedules based on whether the schedules are used in escalation chains or not (_All_, _Used in Escalations_ or _Not Used_) - the same we're doing it now except it's on client side.

Involves backend for adding the endpoint change and frontend for passing the parameter.

teodosii:
Frontend work done in #941 - Merged to `dev`

PR #1425, opened 2023-02-28
Add "used in escalation" filter for schedules internal API

vadimkerr:
# What this PR does
Adds a `used_in_escalations` filter on schedules endpoint for internal API.

https://github.com/grafana/oncall/issues/1423

## Checklist

- [ ] Tests updated
- [ ] Documentation added
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/engine/apps/api/views/schedule.py
+++ b/engine/apps/api/views/schedule.py
@@ -7,6 +7,7 @@ from django.utils.functional import cached_property
 from rest_framework import mixins, status
 from rest_framework.decorators import action
 from rest_framework.exceptions import NotFound
+from rest_framework.fields import BooleanField
 from rest_framework.filters import SearchFilter
 from rest_framework.pagination import PageNumberPagination
 from rest_framework.permissions import IsAuthenticated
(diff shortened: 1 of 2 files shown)


matiasb (approved):

engine/apps/api/views/schedule.py
```diff
@@ -154,6 +155,7 @@ def _annotate_queryset(self, queryset):
     def get_queryset(self):
         is_short_request = self.request.query_params.get("short", "false") == "true"
         filter_by_type = self.request.query_params.get("type")
+        used = BooleanField(allow_null=True).to_internal_value(data=self.request.query_params.get("used"))
```
Konstantinov-Innokentii, line 158: Maybe we can use something easier, like
` is_short_request = self.request.query_params.get("short", "false") == "true"`
( but with check is filter is present)

engine/apps/api/views/schedule.py
```diff
@@ -165,6 +167,8 @@ def get_queryset(self):
             queryset = self.serializer_class.setup_eager_loading(queryset)
         if filter_by_type is not None and filter_by_type in SCHEDULE_TYPE_TO_CLASS:
             queryset = queryset.filter().instance_of(SCHEDULE_TYPE_TO_CLASS[filter_by_type])
+        if used is not None:
+            queryset = queryset.filter(escalation_policies__isnull=not used).distinct()
```
Konstantinov-Innokentii, line 171: Can we check num_escalation_chains annotation instead of escalation_policies, I think it should be faster

engine/apps/api/views/schedule.py
vadimkerr: I like `BooleanField` here because it automatically handles null values as well as the case when the param is not provided. It also returns 400 in case non-boolean value is provided (e.g. 400 will be returned on `?used=random`)

engine/apps/api/views/schedule.py
vadimkerr: Thanks, I'll check this out!
In this particular case there are no annotations when `is_short_request` is set to `True` though. I'm not sure if `is_short_request` is used at all, I'll check this & update in a separate PR if it's not used.

PR #1449, opened 2023-03-02
Prevent redirecting to Integrations if you navigated away to a different screen

teodosii:
# What this PR does

A while ago we fixed a bug where each time you navigated to Integrations, moving to a different a screen would have redirected you back to the Integrations screen if the escalation chain didn't complete loading before you changed the screens. Added a check for `isMounted` which should prevent this bug to appear again in the future.


grafana-plugin/src/pages/integrations/Integrations.tsx
```diff
@@ -51,26 +51,31 @@ class Integrations extends React.Component<IntegrationsProps, IntegrationsState>
...
+    this.isMounted = false;
+    clearInterval(this.alertReceiveChannelTimerId);
+  }
+
   componentDidUpdate(prevProps: Readonly<IntegrationsProps>): void {
     if (prevProps.match.params.id && !this.props.match.params.id) {
-      this.setState({ errorData: initErrorDataState() }, () => {
-        this.parseQueryParams();
-      });
+      this.parseQueryParams();
```
teodosii, line 70: Removed the initial `setState` as it's doing the same thing inside `parseQueryParams`

iskhakov (approved):

maskin25 (approved):

PR #1461, opened 2023-03-06
Schedule quality backend improvements

vadimkerr:
# What this PR does

Changes the schedule quality API so it also returns types of comments (this is needed to address https://github.com/grafana/oncall/issues/118#issuecomment-1436954708).

Related to https://github.com/grafana/oncall/issues/118

## Checklist

- [x] Tests updated


engine/apps/schedules/quality_score.py
```diff
@@ -22,17 +22,27 @@ def get_schedule_quality_score(events: list[dict], days: int) -> dict:
...
-        balance_comment = "Schedule has balance issues"
+        warning_comments.append("Schedule has balance issues")
     elif 0.8 <= balance_score < 1:
-        balance_comment = "Schedule is well-balanced, but still can be improved"
+        info_comments.append("Schedule is well-balanced, but still can be improved")
     else:
-        balance_comment = "Schedule is perfectly balanced"
+        info_comments.append("Schedule is perfectly balanced")
+
+    comments = [{"type": "warning", "text": comment} for comment in warning_comments]
```
matiasb, line 40: maybe define some constant values for "warning" and "info" strings? also wondering if directly building the `comments` value would look too complex? (since it seems we don't need to have the intermediate lists?)

matiasb (approved):
LGTM, with a minor suggestion

engine/apps/schedules/quality_score.py
vadimkerr: Good call! Updated in https://github.com/grafana/oncall/pull/1461/commits/1e2faef4fba853fd05a01a77740261cc0297080f

PR #1537, opened 2023-03-14
Prevent viewer from opening link to mobile app

teodosii:
# What this PR does

Prevent a user with role = Viewer to access mobile app from the link on User Profile Page


joeyorlando (commented):
I would suggest using [this](https://github.com/grafana/oncall/blob/dev/grafana-plugin/src/utils/authorization/index.ts#L109) function to generate a more actionable "missing permission" message, as opposed to:
```
You do not have permission to perform this action. Ask an admin to upgrade your permissions.
```

teodosii:
@joeyorlando 
There's a few places that already use this phrase. I was thinking to put this into a constant to be reused by all usages, most likely will do it in a separate PR. I will keep in mind your remark when doing the refactoring.

maskin25 (approved):

joeyorlando (commented):
would it make sense to add a small integration test for this case?

teodosii:
Will plan to maybe add an e2e (or an integration test as we're also lacking in that area) test in a separate PR

PR #1538, opened 2023-03-14
UI Tweaks on schedules

teodosii:
# What this PR does

Adds a few display tweaks to schedules score quality display


grafana-plugin/src/style/utils.css
```diff
@@ -1,3 +1,7 @@
+.link {
+  text-decoration: none !important;
```
maskin25, line 2: plugin links don't have underlines, it comes from grafana styles 'a { text-decoration: none;}'

maskin25 (approved):

grafana-plugin/src/style/utils.css
teodosii: hmm, but they do. It's a different style affecting it 
image

PR #1546, opened 2023-03-14
add IS_OPEN_SOURCE env var to UI integration tests

joeyorlando:


mderynck:
Something I find kind of confusing here is the backend is set to use this:
`IS_OPEN_SOURCE = LICENSE == OPEN_SOURCE_LICENSE_NAME`
But does not read it from directly env, seems like it could cause some confusing scenarios in some environments.

matiasb (approved):

PR #1568, opened 2023-03-17
generate openapi spec from django-rest-framework schema

joeyorlando:
# What this PR does

## Checklist

- [ ] Tests updated
- [ ] Documentation added
- [ ] `CHANGELOG.md` updated


joeyorlando:
this is currently failing.. it seems like the `drf-spectacular` library doesn't like something about our current schema.. need to poke around a bit more.

Issue #1418, opened 2023-02-27
Refactor isInWorkingHours to use dayjs methods


Ukochka:
isInWorkingHours method uses simple string comparing and it is not that stable. we should use dayjs methods isSameOrBefore, isSameOrAfter

PR #1574, opened 2023-03-20
Refactoring of isInWorkingHours method

Ukochka:
# What this PR does
Instead of using string comparing now isBetween method from dayjs is used

https://github.com/grafana/oncall/issues/1418


grafana-plugin/src/components/WorkingHours/WorkingHours.helpers.ts
```diff
@@ -99,17 +99,30 @@ export const getNonWorkingMoments = (startMoment, endMoment, workingMoments) =>
...
+      .set('date', currentMoment.date())
+      .set('hour', Number(start_HH))
+      .set('minute', Number(start_mm))
+      .set('second', Number(start_ss));
+
+    const endTime = dayjs(momentToStart)
+      .set('date', currentMoment.date())
+      .set('hour', Number(end_HH))
+      .set('minute', Number(end_mm))
+      .set('second', Number(end_ss));
```
joeyorlando, line 123: can this logic which builds the `dayjs` object, be extracted into a separate function? Something like:

```ts
const myFunction = (currentMoment: dayjs.Dayjs, timezone: string, time: string): dayjs.Dayjs => {
  const momentToStart = dayjs().tz(timezone).utcOffset() === 0 ? currentMoment : currentMoment.tz(timezone);

  const [hour, minutes, seconds] = time.split(':');
  return dayjs(momentToStart)
      .set('date', currentMoment.date())
      .set('hour', Number(hour))
      .set('minute', Number(minutes))
      .set('second', Number(seconds));
}
```

then this function can be reused on lines 113 and 119.

joeyorlando:
also, would it make sense to add some unit tests for this function? this way we can be sure it handles all possible cases/edge-cases 🤓

grafana-plugin/src/components/WorkingHours/WorkingHours.helpers.ts
Ukochka: Done

maskin25 (changes requested):
great that you start using `isBetween` method from dayjs but
working hours can be a list of ranges, in your PR you take into account first range only, you should iterate on full list of ranges

PR #1578, opened 2023-03-20
change escalation chains searching to allow for partial searching

joeyorlando:

Previously if you had an Escalation Chain named "Something Critical" and tried searching for "Critical", it would return no results. This was because the backend was using a "starts-with" search on the `name` attribute.

This PR changes that to use "partial searching" + adds a few e2e test cases.

## Checklist

- [x] Tests updated
- [x] `CHANGELOG.md` updated


engine/apps/api/views/escalation_chain.py
```diff
@@ -46,7 +46,7 @@ class EscalationChainViewSet(
     }
 
     filter_backends = [SearchFilter]
-    search_fields = ("^name",)
+    search_fields = ("name",)
```
joeyorlando, line 49: main change ([`django-rest-framework` docs](https://www.django-rest-framework.org/api-guide/filtering/#searchfilter:~:text=By%20default%2C%20searches%20will%20use%20case%2Dinsensitive%20partial%20matches))

PR #1580, opened 2023-03-20
Keep webhook responses data

matiasb:
Track all webhook responses data, and allow using this between alertgroup-related webhooks (e.g. use firing webhook response data when templating the acknowledge webhook request data).

NOTE: dropping the table is not backwards compatible but the feature is not enabled (and in any case it would drop log entries only used for status display)


engine/apps/api/serializers/webhook.py
```diff
@@ -106,16 +104,13 @@ def validate_forward_all(self, data):
 
     def get_last_run(self, obj):
         last_run = ""
-        last_log = obj.logs.all().last()
+        last_log = obj.responses.all().last()
         if last_log:
```
joeyorlando, line 108:
```suggestion
        if last_log := obj.responses.all().last():
```

engine/apps/webhooks/tasks/trigger_webhook.py
```diff
@@ -91,54 +92,61 @@ def execute_webhook(webhook_pk, alert_group_id, user_id):
...
+    try:
+        alert_group = AlertGroup.unarchived_objects.get(pk=alert_group_id)
+    except AlertGroup.DoesNotExist:
+        return
+
+    data = _build_payload(webhook.trigger_type, alert_group, user_id)
     status = {
-        "last_run_at": timezone.now(),
-        "input_data": data,
+        "timestamp": timezone.now(),
```
joeyorlando, line 102: what about setting `timezone.now` as the default on the model's `timestamp` attribute?

joeyorlando (approved):
LGTM 👍 two very minor nit comments

vadimkerr (approved):

PR #1582, opened 2023-03-21
Add SLACK_INTEGRATION_MAINTENANCE env var

Konstantinov-Innokentii:
# What this PR does
Add SLACK_INTEGRATION_MAINTENANCE env var to be able to disable slack install/uninstall


engine/apps/api/views/auth.py
```diff
@@ -25,6 +26,8 @@ def overridden_login_slack_auth(request, backend):
     # We can't just redirect frontend here because we need to make a API call and pass tokens to this view from JS.
     # So frontend can't follow our redirect.
     # So wrapping and returning URL to redirect as a string.
+    if settings.SLACK_INTEGRATION_MAINTENANCE:
```
iskhakov, line 29: May be rename it to
```suggestion
    if settings.MAINTENANCE_FOR_SLACK_MESSAGING_BACKEND_ENABLED:
```

engine/apps/api/views/auth.py
```diff
@@ -25,6 +26,8 @@ def overridden_login_slack_auth(request, backend):
...
+        return Response("Temporary maintenance is being performed on Slack integration management", status=400)
```
iskhakov, line 30:
```suggestion
        return Response("Grafana OnCall is temporary unable to connect your slack account or install OnCall to your slack workspace", status=400)
```

engine/apps/slack/views.py
```diff
@@ -540,20 +540,22 @@ class ResetSlackView(APIView):
...
-                delete_slack_connector_async.apply_async((slack_team_identity.slack_id,))
-            write_chatops_insight_log(
-                author=request.user,
-                event_name=ChatOpsEvent.WORKSPACE_DISCONNECTED,
-                chatops_type=ChatOpsType.SLACK,
-            )
-            unpopulate_slack_user_identities(organization.pk, True)
-            response = Response(status=200)
+        if settings.SLACK_INTEGRATION_MAINTENANCE:
+            response = Response("Temporary maintenance is being performed on Slack integration management", status=400)
```
iskhakov, line 544:
```suggestion
            response = Response("Grafana OnCall is temporary unable to connect your slack account or install OnCall to your slack workspace", status=400)
```

engine/apps/api/views/auth.py
Konstantinov-Innokentii: I started name from SLACK since all slack related envs are named with that convention. It's called INTEGRATION, because we have FEATURE_SLACK_INTEGRATION_ENABLED, so I wanted to be consistent here too. I'm agree, that to be consistent it's better to add ENABLED to the end, as we do in other places.

iskhakov (approved):

engine/apps/api/views/auth.py
iskhakov: I had two points
* booleans end with _ENABLED
* MAINTENANCE be confused with maintenance mode in OnCall

Ukochka (approved):

PR #1594, opened 2023-03-22
Do not show team name if it doesn't exist in store

iskhakov:
# What this PR does

## Checklist

- [ ] Tests updated
- [ ] Documentation added
- [ ] `CHANGELOG.md` updated


grafana-plugin/src/containers/TeamName/TeamName.tsx
```diff
@@ -14,6 +14,9 @@ interface TeamNameProps {
 
 const TeamName = observer((props: TeamNameProps) => {
   const { team, size } = props;
+  if (!team) {
+    return '';
```
maskin25, line 18: it's ok, but `return null` is better

Issue #1490, opened 2023-03-08
Remove incident word from frontend


Ukochka:
1. from URL: instead of `incidents/<id>` should be `alertgroups/<id>` or `alert-groups/<id>`
2. could you remind me another place @raphael-batte  ?

Matvey-Kuk:
It was in routes, but removed with latest Ildar's update.

jtheory:
The routes like https://ops.grafana-ops.net/a/grafana-oncall-app/incidents are still there in Ops
 (but I'm guessing Ildar's changes aren't deployed here yet?)

I also see "incident" mentioned in the "Mode" dropdown when creating a new Maintenance.

Issue #1489, opened 2023-03-08
Modal of Editing rotation is sticking to the top of the screen 


Ukochka:
1. Scroll to the bottom
2. Click Add rotation -> Layer 1
<img width="1840" alt="image" src="https://user-images.githubusercontent.com/20494436/223697391-d45212ff-f4a5-4587-91d8-83c757c61766.png">

matiasb:
This should have been fixed by https://github.com/grafana/oncall/pull/1678 (checked).

Issue #1486, opened 2023-03-08
Linked integrations list appears only when you click on 'linked' icon in Integrations list


Ukochka:
List should be displayed when you select Integration from the list

https://user-images.githubusercontent.com/20494436/223695004-122c8b6f-207d-494d-898f-84b17668dbc5.mov

greatvovan:
Cannot reproduce.

PR #1600, opened 2023-03-22
Collection of bugs

Ukochka:
# What this PR does
List of fixes:

- Incident word removed from the url to `/alert-groups` page and `/alert-group/{id}`
- Incident word removed from other parts of UI (Maintenance page, Linked alert groups, Templates)
- Sticky rotation form was fixed at Schedules (when you scroll down and open Layer 1)
- Integrations linked to Escalations chains are displayed right after loading the page
- Autoresolve condition link to Alert behaviour template

https://github.com/grafana/oncall/issues/1490
https://github.com/grafana/oncall/issues/1489
https://github.com/grafana/oncall/issues/1486


helm/oncall/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: oncall
-description: Developer-friendly incident response with brilliant Slack integration
+description: Developer-friendly alert groups response with brilliant Slack integration
```
joeyorlando, line 3: see comment above in `grafana-plugin/src/utils/consts.ts`

grafana-plugin/src/pages/incident/Incident.tsx
joeyorlando, line 1: should the link to the individual alert group page be `{ page: 'alert-groups', id: <alert-group-id> }` or `{ page: 'alert-group', id: <alert-group-id> }` (without the "s")? it looks like there are both styles in this file.

grafana-plugin/src/utils/consts.ts
```diff
@@ -2,7 +2,7 @@ import plugin from '../../package.json'; // eslint-disable-line
 
 // Navbar
 export const APP_TITLE = 'Grafana OnCall';
-export const APP_SUBTITLE = `Developer-friendly incident response (${plugin?.version})`;
+export const APP_SUBTITLE = `Developer-friendly alert groups response (${plugin?.version})`;
```
joeyorlando, line 5: hmm, imo "alert groups" response doesn't fully summarize what our app does. From the perspective of someone who is new to the app, it might not be entirely clear what an "alert group" is. I would vote for keeping the word `incident` here.

whatever terminology we decide on, we should make sure we're consistent throughout the codebase. We have this same statement in a few places, here, in the helm chart, and the main README (maybe some other spots too?). 

If we do decide to go w/ "alert groups" I would suggest changing it to the singular version:
```suggestion
export const APP_SUBTITLE = `Developer-friendly alert group response (${plugin?.version})`;
```

joeyorlando (commented):
Just some questions/thoughts I had on things you may want to consider:
- `/incidents` will be redirected to `/alert-groups`, correct?
- is there any URL generation, for this route, happening on the backend that may need to be updated? (open question for the whole team)
- [`plugin.json`](https://github.com/grafana/oncall/blob/dev/grafana-plugin/src/plugin.json#L41-L47) still references the `/incidents` route. does that need to be updated as well?
- might be worth adding a very small e2e test that asserts that going to `/incidents` will still redirect you to the proper page

grafana-plugin/src/utils/consts.ts
Ukochka: Yes, I agree, here we should leave 'incident' word

grafana-plugin/src/pages/index.tsx
```diff
@@ -27,24 +27,24 @@ function getPath(name = '') {
...
     hideFromTabs: true,
     hideFromBreadcrumbs: true,
     parentItem: {
-      text: 'Incident',
-      url: `${PLUGIN_ROOT}/incidents`,
+      text: 'Alert Group',
+      url: `${PLUGIN_ROOT}/alert-groups`,
     },
-    path: getPath('incident'),
+    path: getPath('alert-groups'),
```
teodosii, line 47: This one is plural, the other one was singular. Is this on purpose?

grafana-plugin/src/pages/incident/Incident.tsx
```diff
@@ -256,12 +256,12 @@ class IncidentPage extends React.Component<IncidentPageProps, IncidentPageState>
                 {incident.root_alert_group && (
                   <Text type="secondary">
                     Attached to{' '}
-                    <PluginLink query={{ page: 'incident', id: incident.root_alert_group.pk }}>
+                    <PluginLink query={{ page: 'alert-groups', id: incident.root_alert_group.pk }}>
```
teodosii, line 259: Here too

grafana-plugin/src/pages/index.tsx
Ukochka: Yes it is. We had inconsistency before, and I fixed couple of places to make it all consistent. We have page name like 'alert-groups' in plural and after it will be parameter {id}

teodosii (approved):
Approved, there's only 1 bug remaining to be fixed.

PR #1605, opened 2023-03-23
OnCall prometheus metrics exporter

Ferril:
# What this PR does
Add OnCall prometheus metrics exporter

## Checklist

- [x] Tests updated
- [ ] Documentation added
- [ ] `CHANGELOG.md` updated


engine/settings/base.py
```diff
@@ -216,6 +216,7 @@ class DatabaseTypes:
     "apps.public_api",
     "apps.grafana_plugin",
     "apps.webhooks",
+    "apps.metrics_exporter",
```
iskhakov, line 219: I know this is still in draft, adding this comment just to doublecheck that un-authenticated /metrics/ endpoint won't be exposed outside the cluster before merging

iskhakov (dismissed):

engine/apps/metrics_exporter/helpers.py
```diff
@@ -0,0 +1,148 @@
...
+    for team_id, team_data in teams_updated_data.items():
+        for integration_id in alert_groups_total_metrics:
+            if alert_groups_total_metrics[integration_id]["team_id"] == team_id:
+                integration_response_time_metrics = response_time_metrics.get(integration_id)
+                if team_data["deleted"]:
+                    alert_groups_total_metrics[integration_id]["team_id"] = None
+                    alert_groups_total_metrics[integration_id]["team_name"] = "General"
+                    if integration_response_time_metrics:
+                        integration_response_time_metrics["team_id"] = None
+                        integration_response_time_metrics["team_name"] = "General"
```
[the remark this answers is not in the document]
Ferril: Yes, you are right, it should be "No team" now 👍

engine/apps/metrics_exporter/helpers.py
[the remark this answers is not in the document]
Ferril: Yes, each metric will be cached per organization

engine/apps/alerts/models/alert_group.py
```diff
@@ -1064,32 +1107,71 @@ def _bulk_acknowledge(user: User, alert_groups_to_acknowledge: "QuerySet[AlertGr
         AlertGroupLogRecord = apps.get_model("alerts", "AlertGroupLogRecord")
 
         # it is needed to unserolve those alert_groups which were resolved to build proper log.
-        alert_groups_to_unresolve_before_acknowledge = alert_groups_to_acknowledge.filter(resolved=True)
+        alert_groups_to_unresolve_before_acknowledge = alert_groups_to_acknowledge.filter(resolved=models.Value("1"))
 
         # it is needed to unsilence those alert_groups which were silenced to build proper log.
-        alert_groups_to_unsilence_before_acknowledge = alert_groups_to_acknowledge.filter(silenced=True)
+        alert_groups_to_unsilence_before_acknowledge = alert_groups_to_acknowledge.filter(silenced=models.Value("1"))
```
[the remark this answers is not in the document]
Ferril: It just works faster for the same result in mysql

Ferril:
> Have you considered calculating this on-the-fly during the background task which calculates the metrics? Something like this:
> 
> ```python
> response_time = alert_group.get_first_activity_log_timestamp() - alert_group.started_at
> ```

It's a good idea if we were talking only about the calculation task. But we also need to calculate `response_time` on every related action with alert group (ack, resolve, silence). It means that, if we want to calculate in "on-the-fly" we have to do an additional request to db on every ack, silence or resolve action to get alert group logs. 

> Testing
> Prod performance

Not yet, I'll do it later

engine/apps/metrics_exporter/tasks.py
```diff
@@ -0,0 +1,99 @@
...
+    autoretry_for=(Exception,), retry_backoff=True, max_retries=1 if settings.DEBUG else None
+)
+def calculate_and_cache_metrics():  # todo:metrics org
+    """todo:metrics: description"""
+
+    # check recalculation metric flag to avoid parallel or too frequent launch
+    if cache.get(METRICS_CACHE_TIMER):
+        return
+    recalculate_timeout = get_metrics_recalculate_timeout()
+    cache.set(METRICS_CACHE_TIMER, recalculate_timeout, timeout=recalculate_timeout)  # todo:metrics org
```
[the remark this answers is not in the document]
Ferril: As far as I understand, celery rate_limit works for all tasks with the same name per worker. I want to rate limit this task for all workers per organization

engine/apps/alerts/migrations/0015_alertgroup_response_time.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('alerts', '0014_alertreceivechannel_restricted_at'),
+    ]
+
+    operations = [
+        migrations.AddField(
```
mderynck, line 13: Currently I don't think we can't handle adding this to the table, migration takes too long in production.  Is there another way we can store it?

engine/settings/base.py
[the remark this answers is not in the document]
mderynck: Gateway will not forward traffic to that path (/metrics), see [routes](https://github.com/grafana/backend-enterprise/blob/20c7ca3a215ed54e35b4025bd38145a7df357677/pkg/gateway/routes.go#L727), but we should be sure to test it.

engine/apps/metrics_exporter/metrics_cache_manager.py
```diff
@@ -0,0 +1,108 @@
...
+    def metrics_update_diff_for_alert_group(
+        integration_id,
+        response_time_diff,
+        states_diff,
+        old_state=None,
+        new_state=None,
+        response_time=None,
+        started_at=None,
+    ):
+        # todo:metrics: add comment
```
joeyorlando, line 97: todo comment

engine/apps/metrics_exporter/metrics_collectors.py
```diff
@@ -0,0 +1,129 @@
...
+
+        metrics_to_recalculate = []
+        org_ids = get_organization_ids()
+
+        for organization_id in org_ids:
+            start_calculation_task = False
+            # get alert_groups_total metric
+            alert_groups_total_key = get_metric_alert_groups_total_key(organization_id)
+            ag_states: typing.Dict[int, AlertGroupsTotalMetricsDict] = cache.get(alert_groups_total_key)
+            print(f"STATES: {ag_states}")  # todo:metrics remove print
```
joeyorlando, line 54: todo comment + print

engine/apps/metrics_exporter/metrics_collectors.py
```diff
@@ -0,0 +1,129 @@
...
+                    labels_values = list(map(str, labels_values))
+                    for state in ALERTGROUP_STATES:
+                        alert_groups_total.add_metric(labels_values + [state], integration_data[state])
+
+            # get alert_groups_response_time metric
+            alert_groups_response_time_key = get_metric_alert_groups_response_time_key(organization_id)
+            ag_response_time: typing.Dict[int, AlertGroupsResponseTimeMetricsDict] = cache.get(
+                alert_groups_response_time_key
+            )
+            print(f"RESP TIME: {ag_response_time}")  # todo:metrics remove print
```
joeyorlando, line 79: todo/print

engine/apps/metrics_exporter/metrics_collectors.py
```diff
@@ -0,0 +1,129 @@
...
+                        integration_data["id"],  # id
+                    ]
+                    labels_values = list(map(str, labels_values))
+
+                    response_time_values = integration_data["response_time"]
+                    if not response_time_values:
+                        continue
+                    buckets, sum_value = self.get_buckets_with_sum(response_time_values)
+                    buckets = sorted(list(buckets.items()), key=lambda x: float(x[0]))
+                    print(f"BUCKETS: {buckets}")  # todo:metrics remove print
```
joeyorlando, line 100: todo/print

engine/apps/metrics_exporter/metrics_collectors.py
```diff
@@ -0,0 +1,129 @@
...
+
+            if start_calculation_task or not get_metrics_cache_timer_for_organization(organization_id):
+                org_to_recalculate: RecalculateOrgMetricsDict = {
+                    "organization_id": organization_id,
+                    "force": start_calculation_task,
+                }
+                metrics_to_recalculate.append(org_to_recalculate)
+
+        if metrics_to_recalculate:
+            print(f"RECALCULATE: {metrics_to_recalculate}")  # todo:metrics: remove print
```
joeyorlando, line 111: todo/print

engine/apps/metrics_exporter/helpers.py
```diff
@@ -0,0 +1,191 @@
...
+    metric_alert_groups_response_time_key = get_metric_alert_groups_response_time_key(integration.organization_id)
+
+    for metric_key in [metric_alert_groups_total_key, metric_alert_groups_response_time_key]:
+        metric_cache = cache.get(metric_key)
+        metric_cache.pop(integration.id, None)
+        cache.set(metric_key, metric_cache, timeout=metrics_cache_timeout)
+
+
+def metrics_add_integration_to_cache(integration):
+    # todo:metrics: description
```
joeyorlando, line 117: todo

engine/apps/metrics_exporter/helpers.py
```diff
@@ -0,0 +1,191 @@
...
+                cache_updated = True
+            if integration_metric_cache["integration_name"] != integration.emojized_verbal_name:
+                integration_metric_cache["integration_name"] = integration.emojized_verbal_name
+                cache_updated = True
+            if cache_updated:
+                cache.set(metric_key, metric_cache, timeout=metrics_cache_timeout)
+
+
+def metrics_remove_deleted_integration_from_cache(integration):
+    # todo:metrics: description
```
joeyorlando, line 105: todo

engine/apps/metrics_exporter/helpers.py
```diff
@@ -0,0 +1,191 @@
...
+def get_metric_alert_groups_total_key(organization_id):
+    return f"{ALERT_GROUPS_TOTAL}_{organization_id}"
+
+
+def get_metric_alert_groups_response_time_key(organization_id):
+    return f"{ALERT_GROUPS_RESPONSE_TIME}_{organization_id}"
+
+
+def metrics_update_integration_cache(integration):
+    # todo:metrics: description
```
joeyorlando, line 83: todo

engine/apps/alerts/models/alert_group.py
joeyorlando, line 1: few remaining todo comments in this file

joeyorlando (commented):
it's looking really good! few notes just about remaining todo comments. aside from that some general thoughts to consider:
- unit and/or integration tests. especially for the bulk of your new logic in the celery task, `MetricsCacheManager` and `ApplicationMetricsCollector` classes, added `AlertGroup` model logic, and helper methods
-  logging. consider adding some log statements where you think is appropriate. this will help a lot should we need to debug things later on

engine/apps/alerts/models/alert_group.py
```diff
@@ -1363,38 +1630,64 @@ def _bulk_silence(user: User, alert_groups_to_silence: "QuerySet[AlertGroup]", s
...
+            "resolved_by",
+            "silenced",
+            "silenced_at",
+            "silenced_until",
+            "silenced_by_user",
+            "is_escalation_finished",
+            "response_time",
+        ]
+
+        AlertGroup.all_objects.bulk_update(alert_groups_to_silence_list, fields_to_update, batch_size=1000)
```
joeyorlando, line 1685: there appears to be 5 other spots in this file's PR diff where you're modifying the object's fields like this. is it possible to deduplicate this into a method on the `AlertGroup` model class?

engine/apps/metrics_exporter/metrics_cache_manager.py
joeyorlando, line 1: few todo comments in this file

engine/apps/alerts/models/alert_group.py
Ferril: I'm not sure how to do it better, since these methods update fields to different values, `fields_to_update` is also different in some places 🤔

engine/apps/alerts/models/alert_group.py
```diff
@@ -1076,32 +1179,55 @@ def _bulk_acknowledge(user: User, alert_groups_to_acknowledge: "QuerySet[AlertGr
...
+            "silenced_until",
+            "silenced_by_user",
+            "silenced_at",
+            "silenced",
+            "acknowledged_at",
+            "acknowledged_by_user",
+            "acknowledged_by",
+            "is_escalation_finished",
+        ]
+        AlertGroup.all_objects.bulk_update(alert_groups_to_acknowledge_list, fields_to_update, batch_size=1000)
```
vadimkerr, line 1228: Here it's better to use `alert_groups_to_acknowledge.update` instead of `AlertGroup.all_objects.bulk_update` because regular queryset update is more efficient.

Regular update SQL will look similar to this:

```sql
UPDATE table
SET table.field = 1
WHERE table.id IN (1, 2, 3, 4, 5, 6, 7, 8, 9, 10);
```

Batch update SQL will look similar to this:

```sql
UPDATE table
SET table.field = CASE
                      WHEN (table.id = 1) THEN 1
                      WHEN (table.id = 2) THEN 1
                      WHEN (table.id = 3) THEN 1
                      WHEN (table.id = 4) THEN 1
                      WHEN (table.id = 5) THEN 1
                      WHEN (table.id = 6) THEN 1
                      WHEN (table.id = 7) THEN 1
                      WHEN (table.id = 8) THEN 1
                      WHEN (table.id = 9) THEN 1
                      WHEN (table.id = 10) THEN 1
                      ELSE NULL END
WHERE table.id IN (1, 2, 3, 4, 5, 6, 7, 8, 9, 10);
```

engine/apps/alerts/models/alert_group.py
```diff
@@ -1147,31 +1275,54 @@ def _bulk_resolve(user: User, alert_groups_to_resolve: "QuerySet[AlertGroup]") -
...
+            "resolved_by_user",
+            "resolved_by",
+            "is_escalation_finished",
+            "silenced_until",
+            "silenced_by_user",
+            "silenced_at",
+            "silenced",
+        ]
+
+        AlertGroup.all_objects.bulk_update(alert_groups_to_resolve_list, fields_to_update, batch_size=1000)
```
vadimkerr, line 1315: Use `alert_groups_to_resolve.update`

engine/apps/alerts/models/alert_group.py
```diff
@@ -1245,24 +1420,48 @@ def _bulk_restart_unresolve(user: User, alert_groups_to_restart_unresolve: "Quer
...
+            "is_open_for_grouping",
+            "resolved_by_user",
+            "resolved_by",
+            "silenced_until",
+            "silenced_by_user",
+            "silenced_at",
+            "silenced",
+        ]
+
+        AlertGroup.all_objects.bulk_update(alert_groups_to_restart_unresolve_list, fields_to_update, batch_size=1000)
```
vadimkerr, line 1458: Use `alert_groups_to_restart_unresolve.update`

engine/apps/alerts/models/alert_group.py
```diff
@@ -1281,24 +1480,48 @@ def _bulk_restart_unsilence(user: User, alert_groups_to_restart_unsilence: "Quer
...
+            "is_open_for_grouping",
+            "resolved_by_user",
+            "resolved_by",
+            "silenced_until",
+            "silenced_by_user",
+            "silenced_at",
+            "silenced",
+        ]
+
+        AlertGroup.all_objects.bulk_update(alert_groups_to_restart_unsilence_list, fields_to_update, batch_size=1000)
```
vadimkerr, line 1518: Use `alert_groups_to_restart_unsilence.update`

engine/apps/alerts/models/alert_group.py
```diff
@@ -1507,7 +1741,8 @@ def unacknowledge(self):
             self.acknowledged_at = None
             self.acknowledged_by_user = None
             self.acknowledged_by = AlertGroup.NOT_YET
-            self.save(update_fields=["acknowledged", "acknowledged_at", "acknowledged_by_user", "acknowledged_by"])
+            update_fields = ["acknowledged", "acknowledged_at", "acknowledged_by_user", "acknowledged_by"]
+            self.save(update_fields=update_fields)
```
vadimkerr, line 1745:
```suggestion
self.save(update_fields=["acknowledged", "acknowledged_at", "acknowledged_by_user", "acknowledged_by"])
```

engine/apps/alerts/models/alert_group.py
```diff
@@ -1527,7 +1763,8 @@ def unresolve(self):
             self.resolved_at = None
             self.resolved_by = AlertGroup.NOT_YET
             self.resolved_by_user = None
-            self.save(update_fields=["resolved", "resolved_at", "resolved_by", "resolved_by_user"])
+            update_fields = ["resolved", "resolved_at", "resolved_by", "resolved_by_user"]
+            self.save(update_fields=update_fields)
```
vadimkerr, line 1767:
```suggestion
self.save(update_fields=["resolved", "resolved_at", "resolved_by", "resolved_by_user"])
```

engine/apps/alerts/models/alert_group.py
```diff
@@ -1538,17 +1775,17 @@ def silence(self, **kwargs):
...
         self.silenced_until = None
         self.silenced_by_user = None
         self.silenced_at = None
         self.silenced = False
         self.unsilence_task_uuid = None
-        self.save(
-            update_fields=["silenced_until", "silenced", "silenced_by_user", "silenced_at", "unsilence_task_uuid"]
-        )
+        update_fields = ["silenced_until", "silenced", "silenced_by_user", "silenced_at", "unsilence_task_uuid"]
+        self.save(update_fields=update_fields)
```
vadimkerr, line 1788:
```suggestion
        self.save(
            update_fields=["silenced_until", "silenced", "silenced_by_user", "silenced_at", "unsilence_task_uuid"]
        )
```

engine/apps/alerts/tasks/unsilence.py
```diff
@@ -30,8 +30,14 @@ def unsilence_task(alert_group_pk):
             )
             return
         if alert_group.status == AlertGroup.SILENCED and alert_group.is_root_alert_group:
+            initial_state = alert_group.state
             task_logger.info(f"unsilence alert_group {alert_group_pk} and start escalation if needed")
+
             alert_group.un_silence()
+            # update metrics
```
vadimkerr, line 37: Maybe it's better to update metrics in `AlertGroup.un_silence`?

engine/apps/metrics_exporter/helpers.py
```diff
@@ -0,0 +1,227 @@
...
+    METRICS_RECALCULATION_CACHE_TIMEOUT_DISPERSE,
+    METRICS_RESPONSE_TIME_CALCULATION_PERIOD,
+    AlertGroupsResponseTimeMetricsDict,
+    AlertGroupsTotalMetricsDict,
+)
+
+
+def get_organization_ids():
+    """Try getting organizations ids from cache, otherwise get from db and save values in cache"""
+    Organization = apps.get_model("user_management", "Organization")
```
vadimkerr, line 26: Q: Is this here because of circular import?

engine/apps/metrics_exporter/helpers.py
```diff
@@ -0,0 +1,227 @@
...
+    metric_alert_groups_total = cache.get(metric_alert_groups_total_key, {})
+    metric_alert_groups_response_time = cache.get(metric_alert_groups_response_time_key, {})
+    for team_id, team_data in teams_updated_data.items():
+        for integration_id in metric_alert_groups_total:
+            if metric_alert_groups_total[integration_id]["team_id"] == team_id:
+                integration_response_time_metrics = metric_alert_groups_response_time.get(integration_id)
+                if team_data["deleted"]:
+                    metric_alert_groups_total[integration_id]["team_id"] = "no_team"
+                    metric_alert_groups_total[integration_id]["team_name"] = "No team"
+                    if integration_response_time_metrics:
```
vadimkerr, line 175: nit: Maybe it's possible to simplify this a bit? It's hard to keep context when there are 4 levels of indentation.

vadimkerr (changes requested):

engine/apps/alerts/models/alert_group.py
```diff
@@ -502,9 +516,58 @@ def declare_incident_link(self) -> str:
     def happened_while_maintenance(self):
         return self.root_alert_group is not None and self.root_alert_group.maintenance_uuid is not None
 
+    @property
+    def response_time(self):
```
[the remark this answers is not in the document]
matiasb: The idea is to reconsider adding response_time as an `AlertGroup` field (when we unblock migrations for that), so this is somehow keeping a similar behavior to what we would have if the column is here (we are assuming we would want this in the long-term). If after revisiting this in the short-term we think we want to keep a separated model, it makes sense to move things there.

engine/apps/alerts/models/alert_group.py
```diff
@@ -502,9 +516,58 @@ def declare_incident_link(self) -> str:
...
+        """Return time between alert group firing and the first action response, if any."""
+        result = None
+        try:
+            metrics = self.metrics
+        except AlertGroupMetrics.DoesNotExist:
+            pass
+        else:
+            result = metrics.response_time
+        return result
```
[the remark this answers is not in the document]
matiasb: Ack, fwiw I like the explicitness of knowing what's triggering the exception (if there is one), but no objections to change it.

engine/apps/alerts/models/alert_group.py
```diff
@@ -502,9 +516,58 @@ def declare_incident_link(self) -> str:
...
+            organization_id=organization_id,
+            old_state=previous_state,
+            new_state=self.state if not deleted else None,
+            response_time=updated_response_time,
+            started_at=self.started_at,
+        )
+
     def acknowledge_by_user(self, user: User, action_source: Optional[str] = None) -> None:
         AlertGroupLogRecord = apps.get_model("alerts", "AlertGroupLogRecord")
+        initial_state = self.state
```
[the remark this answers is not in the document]
matiasb: Right, I was thinking something more like a context manager, but was leaving for a next iteration (because I was also thinking we may want to do something around the state transitions related to this).

engine/apps/alerts/constants.py
```diff
@@ -10,3 +10,11 @@ class ActionSource:
 TASK_DELAY_SECONDS = 1
 
 NEXT_ESCALATION_DELAY = 5
+
+# AlertGroup states verbal
+STATE_FIRING = "firing"
```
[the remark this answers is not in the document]
matiasb: Makes sense

engine/apps/alerts/tasks/unsilence.py
matiasb: Updating metrics inside the state transition methods is tricky because you may lose track of the original state in some cases (e.g. we call un_silence in multiple places, for example we may call this during acknowledge), and also because bulk updates don't go through these methods.

engine/apps/metrics_exporter/helpers.py
matiasb: I don't think so, will change

engine/apps/metrics_exporter/helpers.py
matiasb: Checking, not seeing an easy refactoring atm, but for sure maybe something to revisit later when we deal with more metrics.

engine/apps/alerts/migrations/0016_alertgroupmetrics.py
```diff
@@ -0,0 +1,22 @@
...
+    dependencies = [
+        ('alerts', '0015_auto_20230508_1641'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='AlertGroupMetrics',
+            fields=[
+                ('id', models.BigAutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('response_time', models.DurationField(default=None, null=True)),
```
joeyorlando, line 18: what about storing this on the `alert_group` table itself?

engine/apps/metrics_exporter/helpers.py
matiasb: Double-checked, there is a circular import indeed (through alerts/alert group), keeping for now.

engine/apps/alerts/migrations/0016_alertgroupmetrics.py
matiasb: :+1:  Switched back to the `alert_group` column implementation.

engine/apps/alerts/constants.py
```diff
@@ -10,3 +13,11 @@ class ActionSource:
 TASK_DELAY_SECONDS = 1
 
 NEXT_ESCALATION_DELAY = 5
+
+
+# AlertGroup states verbal
+class AlertGroupState(str, Enum):
```
joeyorlando, line 19: I've yet to try this out, but with Python 3.11 there is a new class, [`StrEnum`](https://docs.python.org/3/library/enum.html#enum.StrEnum), might be worth giving a try

mderynck (approved):

joeyorlando (approved):

Issue #1611, opened 2023-03-23
Save selected teams in local storage


maskin25:
something like this:

```
if(localStorage['team-filters]){
filters.teams=localStorage['team-filters]
}else{
don't set filter default value
}


```

PR #1613, opened 2023-03-23
Save selected teams filter in local storage

maskin25:
# What this PR does

Selected teams filter is now saved in local storage

https://github.com/grafana/oncall/issues/1611

## Checklist


grafana-plugin/src/models/filters/filters.ts
```diff
@@ -3,21 +3,39 @@ import { action, observable } from 'mobx';
...
-  public globalValues: FiltersValues = {};
+  private _globalValues: FiltersValues = {};
 
   constructor(rootStore: RootStore) {
     super(rootStore);
+
+    const savedFilters = getItem(LOCAL_STORAGE_FILTERS_KEY);
+    if (savedFilters) {
+      this.globalValues = { ...savedFilters };
+    }
```
joeyorlando, line 28: should we add an `else` clause here which sets to the user's default configured team if nothing is stored in local storage?

joeyorlando (approved):

joeyorlando (approved):
could it be worthwhile adding a small e2e test that the filter values are persisted (for example on refresh)?

grafana-plugin/src/models/filters/filters.ts
maskin25: good idea, but default team is unknown at the moment of this code running...

PR #1615, opened 2023-03-24
speed up ci builds from 15 to <7 minutes

joeyorlando:
This PR cuts GitHub Action build times from 14-15 minutes, down to just under 7 minutes. It does this by:
- caching `grafana-plugins/node_modules` and `pip` dependencies based on their respective dependency files (eg. `requirements.txt` & `yarn.lock`). This step alone saves ~3 minutes.
- get rid of the "build-engine-docker-image" and "backend-integration-tests" jobs in the old "Integration Tests" workflow. This was split out this way so that we could build the backend docker image once, upload the artifact, and then reuse it across the backend and e2e tests. We no longer need these backend integration tests because we are testing the same thing in the e2e tests. This saves ~45 seconds of having to upload the image artifact.
- few improvements within the integration tests themselves:
  - move plugin configuration to the `globalSetup.ts`. This means that every test does not need to check if the plugin has been configured because it is done once before all the tests are run.
  - cache the plugin frontend build. If your commit doesn't change anything to `grafana-plugin/src` or `grafana-plugin/yarn.lock` it should be safe to reuse a previously built/cached version of the plugin frontend. This saves ~3 minutes
  - cache playwright binaries/dependencies. Only re-install them if the version of `@playwright/test` in `grafana-plugin/yarn.lock` changes. This saves ~3 minutes.
  
**Other things to mention**
Once we refactor the `GSelect` component to not call the `onChange` callback on every keyDown event (#1628), this should allow us to parallelize the integration tests, and cut the time required to execute the tests themselves in half


grafana-plugin/src/containers/GSelect/GSelect.tsx
joeyorlando, line 1: this is effecting the ability to parallelize the integration tests. Currently we make an API call on every keyDown event. This hammers the backend and makes it crawl to a halt on CI (mostly due to the limited CPU resources GitHub provides you on CI builds).

I tried refactoring the `loadOptions` callback but couldn't get it working 100%, could use some assistance here (cc: @grafana/grafana-oncall-frontend)

grafana-plugin/src/containers/GSelect/GSelect.tsx
```diff
@@ -61,7 +61,6 @@ const GSelect = observer((props: GSelectProps) => {
     showWarningIfEmptyValue = false,
     getDescription,
     filterOptions,
-    // fromOrganization,
```
joeyorlando, line 64: this is commented out, ie. never used, removed the few leftover references to it.

.github/workflows/linting-and-tests.yml
```diff
@@ -0,0 +1,435 @@
...
+  lint-entire-project:
+    name: "Lint entire project"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: "3.9.12"
+          cache: "pip"
+          cache-dependency-path: engine/requirements.txt
```
joeyorlando, line 36: this speeds things up - see PR description

.github/workflows/linting-and-tests.yml
```diff
@@ -0,0 +1,435 @@
...
+          cache: "yarn"
+          cache-dependency-path: grafana-plugin/yarn.lock
+      - name: Use cached frontend dependencies
+        id: cache-frontend-dependencies
+        uses: actions/cache@v3
+        with:
+          path: grafana-plugin/node_modules
+          key: ${{ runner.os }}-frontend-node-modules-${{ hashFiles('grafana-plugin/yarn.lock') }}
+      - name: Install frontend dependencies
+        if: steps.cache-frontend-dependencies.outputs.cache-hit != 'true'
```
joeyorlando, line 50: this speeds things up - see PR description

.github/workflows/linting-and-tests.yml
```diff
@@ -0,0 +1,435 @@
...
+        with:
+          path: "~/.cache/ms-playwright"
+          key: ${{ runner.os }}-playwright-${{ env.PLAYWRIGHT_VERSION }}
+
+      - name: Install Playwright binaries/dependencies
+        if: steps.playwright-cache.outputs.cache-hit != 'true'
+        # if more browsers are added, need to modify
+        run: |
+          npx playwright install --with-deps chromium firefox
+          npx playwright install-deps
```
joeyorlando, line 393: this speeds things up - see PR description

engine/settings/ci-test.py
joeyorlando, line 1: I changed this because if we want to leverage `pip` dependency caching we _should_ run the tests on the GitHub runner host, rather than inside a Docker container, on that host. 

Without doing this, the container doesn't have proper access to the cached files (maybe with some volume mounting it could, but I didn't mess around with this)

engine/wait_for_test_mysql_start.sh
joeyorlando, line 1: see comment above in `engine/settings/ci-test.py`

grafana-plugin/integration-tests/globalSetup.ts
```diff
@@ -1,6 +1,39 @@
...
+  const pluginIsAutoConfigured = (await onCallApiUrlInput.count()) === 0;
+
+  if (!pluginIsAutoConfigured) {
+    await onCallApiUrlInput.fill(ONCALL_API_URL);
+    await clickButton({ page, buttonText: 'Connect' });
+  }
+
+  // wait for the "Connected to OnCall" message to know that everything is properly configured
+  await expect(page.getByTestId('status-message-block')).toHaveText(/Connected to OnCall.*/);
+};
```
joeyorlando, line 36: this function already existed as-is. Just relocating it from `grafana-plugin/integration-tests/utils/configurePlugin.ts`, which has been removed, because we don't need to access it for every test.

.github/workflows/snyk.yml
```diff
@@ -17,16 +17,21 @@ jobs:
...
-        run: |
-          pip install -r engine/requirements.txt
-          cd grafana-plugin/
-          yarn --network-timeout 500000
+      - name: Install backend dependencies
+        working-directory: engine
+        run: pip install -r requirements.txt
+      - name: Install frontend dependencies
+        working-directory: grafana-plugin
+        run: yarn install --frozen-lockfile --prefer-offline --network-timeout 500000
```
teodosii, line 34: `--prefer-offline`

Interesting, does this work on CI or it will still attempt to download?

grafana-plugin/src/containers/GSelect/GSelect.tsx
teodosii: Maybe this should be debounced? I can definitely have a look. The problem is not really in `GSelect` but in the `EscalationPolicy` itself due to calling the backend.

teodosii (approved):

engine/wait_for_test_mysql_start.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
-until nc -z -v -w30 mysql_test 3306
+until nc -z -v -w30 "${DATABASE_HOST:mysql_test}" 3306
```
iskhakov, line 3: this won't work on bash, see [stackoverflow](https://stackoverflow.com/questions/2013547/assigning-default-values-to-shell-variables-with-a-single-command-in-bash) or experiment:
```
echo ${NON_EXISTANT_VAR:default}
zsh: unrecognized modifier `d'
echo ${NON_EXISTANT_VAR:-default}
default
echo ${NON_EXISTANT_VAR:=default}
default
```

PR #1646, opened 2023-03-28
Update apns payload with corrected `interruption-level` field

imtoori:
I just noticed that ios notifications are always critical, even when the switch ("override DND") is set to `false`. I believe this is because `critical` should be either 1 or 0 
https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/generating_a_remote_notification#2990112

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+
+- Addressed bug with iOS mobile push notifications always being set to critical by @imtoori and @joeyorlando ([#1646](https://github.com/grafana/oncall/pull/1646))
+
 ## v1.2.2 (2023-03-27)
 
 ### Changed
--- a/engine/apps/mobile_app/tasks.py
+++ b/engine/apps/mobile_app/tasks.py
@@ -208,7 +208,9 @@ def _get_fcm_message(alert_group, user, registration_id, critical):
                     alert=ApsAlert(title=alert_title, subtitle=alert_subtitle, body=alert_body),
                     sound=CriticalSound(
                         # The notification shouldn't be critical if the user has disabled "override DND" setting
-                        critical=1 if (critical and mobile_app_user_settings.important_notification_override_dnd) else 0,
+                        critical=1
+                        if (critical and mobile_app_user_settings.important_notification_override_dnd)
+                        else 0,
                         name=apns_sound_name,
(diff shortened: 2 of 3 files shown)


joeyorlando (approved):

engine/apps/mobile_app/tasks.py
```diff
@@ -208,7 +208,9 @@ def _get_fcm_message(alert_group, user, registration_id, critical):
                     alert=ApsAlert(title=alert_title, subtitle=alert_subtitle, body=alert_body),
                     sound=CriticalSound(
                         # The notification shouldn't be critical if the user has disabled "override DND" setting
-                        critical=(critical and mobile_app_user_settings.important_notification_override_dnd),
```
joeyorlando, line 211: just leaving a paper trail here in case others come across this. The in-code documentation for the `CriticalSound` class mentions that `critical` should be a boolean. However, the iOS [docs](https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/generating_a_remote_notification#2990112) say `True/False`.

The commit prior to this, [ed5b5e153d289d355fbd529b39bf5cb66cafdebd](https://github.com/grafana/oncall/commit/ed5b5e153d289d355fbd529b39bf5cb66cafdebd), had this set to 0/1 and it appeared to be working

vadimkerr: Yeah, https://firebase.google.com/docs/reference/admin/python/firebase_admin.messaging#firebase_admin.messaging.CriticalSound mentions it should be boolean and there's how the lib handles the flag: https://github.com/firebase/firebase-admin-python/blob/master/firebase_admin/_messaging_encoder.py#L584-L585.
@imtoori do you mean that notifications are critical even when `important_notification_override_dnd` is set to `False`?

vadimkerr:
I disabled auto-merge to make sure we track the root cause, I think this is not related to `critical` flag not being 0/1. It currently works for non-important push notifications, so I think there's some other issue.

vadimkerr (approved):

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+
+- Addressed bug with iOS mobile push notifications always being set to critical by @imtoori and @joeyorlando ([#1646](https://github.com/grafana/oncall/pull/1646))
```
vadimkerr, line 12: Not sure if we should include this to changelog, user settings are not released yet so this is not user-facing change, right?

vadimkerr: Nvm, this does actually make sense even if user settings are not released yet in the mobile app (user would need to update the backend to get this working correctly)

PR #1649, opened 2023-03-28
e2e - Add viewer/admin auth, added trivial test for viewer

teodosii:
# What this PR does

This PR is concering only the e2e tests, plus a few added new `data-testid` attributes.

- Added `auth.setup.ts` - this handles `admin` \ `viewer` authentication
- Removed most of the contents of `globalSetup.ts` as it was conflicting with the new authentication rules.
- Old `globalSetup.ts` was renamed to `configurePlugin.setup.ts` and it preservers the existing `configurePlugin` helper
- Added a new user (viewer) which is appended to the database on test load

```jsx
export const GRAFANA_VIEWER_USERNAME = 'oncall-viewer';
export const GRAFANA_VIEWER_PASSWORD = 'oncall';
export const GRAFANA_VIEWER_EMAIL = 'oncall@grafana.com';
```

Open to suggestions for the email, might not be best to name it `oncall@grafana.com`

- Minor refactoring (such as `OnCallPage` to be more friendly in terms of typing)
- Added the first test on Viewer permissions (more to come) - pretty trivial at the moment. Most important is that it asserts the logged in user is the actual viewer user we created.


grafana-plugin/integration-tests/auth.setup.ts
```diff
@@ -0,0 +1,69 @@
...
+  const { headless } = config.projects[0]!.use;
+  const browser = await chromium.launch({ headless, slowMo: headless ? 0 : 100 });
+  const browserContext = await browser.newContext();
+
+  try {
+    // Create a Viewer user if none exists
+    const createUserRes = await browserContext.request.post(`${BASE_URL}/api/admin/users`, {
+      data: {
+        name: GRAFANA_VIEWER_USERNAME,
+        email: GRAFANA_VIEWER_EMAIL,
```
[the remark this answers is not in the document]
joeyorlando: we may want to extract this API call to a separate function that also updates the role. This way we can easily reuse this to create an Editor if we want.

Something like this:
```ts
import { OrgRole } from '@grafana/data';
import { BrowserContext } from '@playwright/test';

const ORG_ID = process.env.ORG_ID || 1; // I believe for OSS this should always be 1 but could be useful for cloud to be able to dynamically pass this in

const createGrafanaUserWithRole = async (browserContext: BrowserContext, login: string, password: string, role: OrgRole): Promise<void> => {
  const createUserRes = await browserContext.request.post(`${BASE_URL}/api/admin/users`, {
      data: {
        name: login,
        login,
        password,
      },
    });

   expect(createUserRes.ok()).toBeTruthy();

  const updateUserRoleRes = await browserContext.request.patch(`${BASE_URL}/api/orgs/${ORG_ID}/users/${createUserRes.data.id}`, {
      data: {
        role,
      },
    });

   expect(updateUserRoleRes.ok()).toBeTruthy();
}
```
I haven't tested this out, you might need to slightly modify it to handle the case where the user already exists.

grafana-plugin/integration-tests/auth.setup.ts
```diff
@@ -0,0 +1,69 @@
...
+  const { headless } = config.projects[0]!.use;
+  const browser = await chromium.launch({ headless, slowMo: headless ? 0 : 100 });
+  const browserContext = await browser.newContext();
+
+  const res = await login(browserContext, GRAFANA_USERNAME, GRAFANA_PASSWORD);
+
+  expect(res.ok()).toBeTruthy();
+
+  await page.context().storageState({ path: ADMIN_FILE });
+  await browserContext.storageState({ path: ADMIN_FILE });
```
joeyorlando, line 26: do we need both `storageState` calls here? see [here](https://github.com/grafana/oncall/pull/1649/files#diff-698159ea62969664ec910a06dd9c79739dcc39baacd386b29b299888a7e20569L55) in the previous `globalSetup.ts`.

grafana-plugin/integration-tests/utils/userSettings.ts
```diff
@@ -2,19 +2,23 @@ import { Locator, Page } from '@playwright/test';
...
+  await goToOnCallPage(page, OnCallPage.USERS);
   await clickButton({ page, buttonText: 'View my profile' });
   await page.locator('text=To edit user details such as Username, email, and role').waitFor({ state: 'visible' });
 };
 
 const getForgetPhoneNumberButton = (page: Page): Locator => page.locator('button >> text=Forget Phone Number');
 
+export const openViewMyProfile = async (page: Page): Promise<void> => {
+  await openUserSettingsModal(page);
+};
```
joeyorlando, line 20: do we need this function? Could we just call `openUserSettingsModal(page)` directly instead?

grafana-plugin/integration-tests/utils/constants.ts
```diff
@@ -4,3 +4,7 @@ export const GRAFANA_USERNAME = process.env.GRAFANA_USERNAME || 'oncall';
 export const GRAFANA_PASSWORD = process.env.GRAFANA_PASSWORD || 'oncall';
 export const MAILSLURP_API_KEY = process.env.MAILSLURP_API_KEY;
 export const IS_OPEN_SOURCE = (process.env.IS_OPEN_SOURCE || 'true').toLowerCase() === 'true';
+
+export const GRAFANA_VIEWER_USERNAME = 'oncall-viewer';
+export const GRAFANA_VIEWER_PASSWORD = 'oncall';
+export const GRAFANA_VIEWER_EMAIL = 'oncall@grafana.com';
```
joeyorlando, line 10:
```suggestion
export const GRAFANA_VIEWER_USERNAME = process.env.GRAFANA_VIEWER_USERNAME || 'oncall-viewer';
export const GRAFANA_VIEWER_PASSWORD = process.env.GRAFANA_VIEWER_PASSWORD || 'oncall';
```

joeyorlando (commented):
this looks awesome! this will be really helpful for testing role specific stuff.

Mostly LGTM, just some questions/general suggestions.

grafana-plugin/integration-tests/utils/constants.ts
joeyorlando: one other small note, be sure to update the env vars in `grafana-plugin/.env.example`

grafana-plugin/playwright.config.ts
```diff
@@ -12,7 +12,7 @@ require('dotenv').config();
  */
 const config: PlaywrightTestConfig = {
   testDir: './integration-tests',
-  globalSetup: './integration-tests/globalSetup.ts',
+  // globalSetup: './integration-tests/globalSetup.ts',
```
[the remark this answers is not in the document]
teodosii: Dark magic 🕶️

grafana-plugin/integration-tests/auth.setup.ts
teodosii: Yea, I was wondering the same, but in a few tests I tried skipping the `browserContext.storageState` call and it didn't work 🤔 I'll check for the other one to see if it is required, the first line comes from playwright documentation

teodosii: Oh good catch. I bet we won't be needing `storageState.json` any more, cause right now we should use auth instead.

grafana-plugin/integration-tests/utils/userSettings.ts
teodosii: The plan was to add more stuff to it, todo for next time. I changed it as you suggested in the meanwhile.

grafana-plugin/adminState.json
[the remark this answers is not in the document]
teodosii: Hmm, actually I think this is needed because I ran into some cases where if that path is missing, then it won't allow running the tests and complain on that. It complains only for `adminState.json` because that's inside playwright config, I think we should add these files and mark them to .gitignore so we do not track changes for it.

grafana-plugin/viewerState.json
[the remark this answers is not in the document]
teodosii: See comment above.

grafana-plugin/integration-tests/.auth/viewerState.json
joeyorlando, line 1: for this file, and `grafana-plugin/integration-tests/.auth/adminState.json`, you likely need to run:
```bash
git rm --cached grafana-plugin/integration-tests/.auth  # or some variation of this
```

joeyorlando:
@teodosii I believe we can close this. This was done in #2417

PR #1652, opened 2023-03-28
Fix Grafana teams sync

vadimkerr:
# What this PR does
Sometimes plugin sync fails with the following exception:
```
Cannot delete or update a parent row: a foreign key constraint fails (`schedules_oncallschedule`, CONSTRAINT `alerts_oncallschedul_team_id_4e633f4b_fk_user_mana` FOREIGN KEY (`team_id`) REFERENCES `user_management_team` (`id`))'
```

How to reproduce:
1. Create a new Grafana team
2. Create two schedules with different types (e.g. ICal and Web) and assign both schedules to the new team
3. Delete the team in Grafana
4. Trigger plugin sync, the sync will fail with the exception above

This happens because the `OnCallSchedule` Django model is a polymorphic model and there's a [known bug](https://github.com/django-polymorphic/django-polymorphic/issues/229) in `django-polymorphic` with deleting related objects when using `SET_NULL` and `CASCADE`. This PR adds non-polymorphic versions of `SET_NULL` and `CASCADE` to use in schedule FKs as per this [comment](https://github.com/django-polymorphic/django-polymorphic/issues/229#issuecomment-398434412).

This also applies to two other schedule FKs: `organization` and `user_group`, which are not working properly as well.

## Checklist


engine/apps/schedules/models/on_call_schedule.py
```diff
@@ -96,12 +97,12 @@ class OnCallSchedule(PolymorphicModel):
...
 
     organization = models.ForeignKey(
-        "user_management.Organization", on_delete=models.CASCADE, related_name="oncall_schedules"
+        "user_management.Organization", on_delete=NON_POLYMORPHIC_CASCADE, related_name="oncall_schedules"
     )
 
     team = models.ForeignKey(
         "user_management.Team",
-        on_delete=models.SET_NULL,
+        on_delete=NON_POLYMORPHIC_SET_NULL,
```
vadimkerr, line 105: main change of the PR

engine/common/database.py
```diff
@@ -15,3 +16,23 @@ def get_random_readonly_database_key_if_present_otherwise_default() -> str:
     if hasattr(settings, "READONLY_DATABASES") and len(settings.READONLY_DATABASES) > 0:
         using_db = random.choice(list(settings.READONLY_DATABASES.keys()))
     return using_db
+
+
+def NON_POLYMORPHIC_SET_NULL(collector, field, sub_objs, using):
```
vadimkerr, line 21: main change of the PR

PR #1707, opened 2023-04-06
Use grafana theme colors in tags

iskhakov:
# What this PR does

Some colours are not unified with grafana theme. This PR is fixing colours in tags.

## Checklist


teodosii (approved):

grafana-plugin/src/style/vars.css
```diff
@@ -56,6 +46,16 @@
   --shadows-z1: 0 1px 2px rgba(24, 26, 27, 0.2);
   --shadows-z2: 0 4px 8px rgba(24, 26, 27, 0.2);
   --shadows-z3: 0 13px 20px 1px rgba(24, 26, 27, 0.18);
+  --tag-danger: #e0226e;
+  --tag-warning: #fad34a;
+  --tag-primary: #1b855e;
+  --tag-secondary: rgba(36, 41, 36, 0.16);
+  --tag-border-danger: #cf0e5b;
```
torkelo, line 53: How are these generated? What if the theme changes?


You can use Emotion la GlobalStyles to generate these at runtime based on the current theme. Grafana themes are dynamic and can change (ie more than the built in ones) in the future , the theme system in Grafana supports multiple themes so having colors hard coded in sass files could be problematic

iskhakov: oncall might not support that yet, @grafana/grafana-oncall-frontend can you please answer here? How do you think if we add it to the upcoming quarter as frontend refactoring?

iskhakov: > How are these generated? What if the theme changes?
I took the colors from [storybook](https://developers.grafana.com/ui/latest/index.html?path=/story/docs-overview-theme--theme-demo)

as I see oncall uses hard-coded colors everywhere, https://github.com/orgs/grafana/teams/grafana-oncall-frontend can you please answer here? How do you think if we add it to the upcoming quarter as frontend refactoring?

teodosii: I'd say let's merge it as it is because right now we're only hardcoding them. TODO for later time to add this

PR #1713, opened 2023-04-09
Phone notificator

Konstantinov-Innokentii:
# What this PR does

## Checklist


engine/apps/alerts/incident_appearance/templaters/phone_call_templater.py
```diff
@@ -14,14 +14,11 @@ def _postformat(self, templated_alert):
...
 
     def _slack_format_for_phone_call(self, data):
         sf = self.slack_formatter
         sf.user_mention_format = "{}"
         sf.channel_mention_format = "#{}"
         sf.hyperlink_mention_format = "{title}"
         return sf.format(data)
-
-    def _escape(self, data):
-        return escape_for_twilio_phone_call(data)
```
iskhakov, line 27: Is that expected that we don't escape `< > &` anymore

https://github.com/grafana/oncall/pull/1713/files#diff-007b962c05ebca7bac34170b3d6fa519225c2bc6d652371dfc27ae018f942b73L193-L195

engine/apps/phone_notifications/models/phone_call.py
```diff
@@ -0,0 +1,55 @@
+from django.db import models
+
+
+# Duplicate to avoid circular import
+class TwilioCallStatuses:
```
iskhakov, line 5: This will be removed after migration, right?

engine/apps/alerts/incident_appearance/templaters/phone_call_templater.py
Konstantinov-Innokentii: Yes, this escape was TwilioRelated, so it's moved to TwilioPhoneProvider https://github.com/grafana/oncall/pull/1713/files#diff-764f1941b04b97b1517990791c24bc6446e623f0326b95c44060dd4ec42afd75R26

engine/apps/twilioapp/gather.py
```diff
@@ -0,0 +1,79 @@
...
+            if digit == "1":
+                alert_group.acknowledge_by_user(user, action_source=ActionSource.PHONE)
+            elif digit == "2":
+                alert_group.resolve_by_user(user, action_source=ActionSource.PHONE)
+            elif digit == "3":
+                alert_group.silence_by_user(user, silence_delay=1800, action_source=ActionSource.PHONE)
+
+
+def get_gather_url():
+    return create_engine_url(reverse("twilioapp:gather"), "https://pretty-mosh-97.loca.lt")
```
iskhakov, line 75: hardcoded url

engine/apps/twilioapp/migrations/0003_auto_20230408_0711.py
```diff
@@ -0,0 +1,27 @@
...
+        migrations.DeleteModel('PhoneCall'),
+        migrations.DeleteModel('SMSMessage')
+    ]
+
+    operations = [
+        migrations.SeparateDatabaseAndState(
+            database_operations=database_operations,
+            state_operations=state_operations
+        )
+    ]
```
iskhakov, line 27: Consider using https://docs.djangoproject.com/en/3.2/ref/models/options/#db-table instead of renaming tables to avoid data migrations

iskhakov (approved):
looks good

engine/apps/twilioapp/gather.py
Konstantinov-Innokentii: Thanks

engine/apps/phone_notifications/models/phone_call.py
Konstantinov-Innokentii: Yes, I'm keeping it to be able to receive Gather (digits from phone call) for calls which were made shortly before migration.

engine/apps/phone_notifications/models/sms.py
```diff
@@ -0,0 +1,62 @@
...
+        (FAILED, "failed"),
+        (DELIVERED, "delivered"),
+        (UNDELIVERED, "undelivered"),
+        (RECEIVING, "receiving"),
+        (RECEIVED, "received"),
+        (READ, "read"),
+    )
+
+
+class OnCallSMS(models.Model):
```
vadimkerr, line 36: nit: Maybe just `SMS`?

vadimkerr (approved):
Great refactoring, looks like it will make adding new providers so much easier 👍

engine/engine/management/commands/verify_phone.py
joeyorlando, line 1: could this still be useful?

engine/apps/twilioapp/gather.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+def get_gather_message():
+    return "Press 1 to acknowledge, 2 to resolve, 3 to silence to 30 minutes"
```
joeyorlando, line 79: _nit_ move "1", "2", "3" into variables and reference those in the 3 places in this file. ex:
```python
ACK_DIGIT = "1"
RESOLVE_DIGIT = "2"
SILENCE_DIGIT = "3"

def get_gather_message():
    return f"Press {ACK_DIGIT} to acknowledge, {RESOLVE_DIGIT} to resolve, {SILENCE_DIGIT} to silence to 30 minutes"
```
same thing for the `process_digit` and `process_gather_data` functions

engine/apps/twilioapp/status_callback.py
```diff
@@ -0,0 +1,142 @@
...
+    }
+    return TWILIO_ERRORS_TO_ERROR_CODES_MAP.get(status, None)
+
+
+def get_call_status_callback_url():
+    return create_engine_url(reverse("twilioapp:call_status_events"), "https://pretty-mosh-97.loca.lt")
+
+
+def get_sms_status_callback_url():
+    return create_engine_url(reverse("twilioapp:sms_status_events"), "https://pretty-mosh-97.loca.lt")
```
joeyorlando, line 142: hardcoded URL

joeyorlando (approved):
this is great! LGTM.

few minor questions/comments:
- do we need the new `get_verification_call` endpoint?
- I see you have "refactor tests" listed as a todo, +1 on this

engine/apps/api/views/user.py
```diff
@@ -294,22 +300,44 @@ def timezone_options(self, request):
...
+        logger.info('get_verification_code_via_call: pass reCAPTCHA validation"')
+
+        user = self.get_object()
+        phone_backend = PhoneBackend()
+        try:
+            phone_backend.make_verification_call(user)
+        except NumberAlreadyVerified:
+            return Response("Phone number already verified", status=status.HTTP_400_BAD_REQUEST)
+        except FailedToStartVerification:
+            return Response("Something went wrong while calling", status=status.HTTP_500_INTERNAL_SERVER_ERROR)
```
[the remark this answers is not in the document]
Konstantinov-Innokentii: Yep, it is introduced keeping in mind that some providers can't sent sms, like asterisk. Twilio is also capable of verifying via call, but I didn't supported it to not overcomplicate refactoring.

engine/apps/twilioapp/twilio_client.py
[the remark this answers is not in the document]
Konstantinov-Innokentii: Yep, I'm still using it as a reference while working on PR, will delete it before merge, thanks!

engine/apps/phone_notifications/phone_backend.py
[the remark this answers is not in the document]
Konstantinov-Innokentii: Yep

engine/apps/phone_notifications/models/phone_call.py
```diff
@@ -0,0 +1,55 @@
...
+        (IN_PROGRESS, "in-progress"),
+        (COMPLETED, "completed"),
+        (BUSY, "busy"),
+        (FAILED, "failed"),
+        (NO_ANSWER, "no-answer"),
+        (CANCELED, "canceled"),
+    )
+
+
+class OnCallPhoneCall(models.Model):
```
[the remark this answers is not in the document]
Konstantinov-Innokentii: I tried just PhoneCall name, but it doesn't works
For example we have this method in ProveProvider interface. If I rename OnCallPhoneCall to PhoneCall, then I should rename oncall_phone_call parameter name to phone_call which makes no sense in function singanure. On the other side I don't like PhoneCall model name, but oncall_phone_call parameter, since it is confusing.
```
def make_notification_call(self, number, message, oncall_phone_call):
        ....
```
Same issue we have with TwilioPhoneCall model. Is has field oncall_phone_call. Renaming it to just phone call leads to lines like ```twilio_phone_call.phone_call```. 
I'm not 100% happy with OnCallPhoneCall name, but having just PhoneCall introduces more confusion.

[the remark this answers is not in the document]
Konstantinov-Innokentii: @vadimkerr what do you think about PhoneCallRecord instead of OnCallPhoneCall. Also it sounds well: twilio_phone_call.phone_call_record ( at least better that twilio_phone_call.oncall_phone_call)

engine/engine/management/commands/verify_phone.py
Konstantinov-Innokentii: Actually it is useful only for slack now. We process events triggered from slack synchronously, instead of running celery tasks. It is needed to make re-rendering of slack message faster, when button is pressed. We don't use this system for other chatops integrations, but ideally we should. Just keeping it consistent, waiting when we will refactor messaging backends.

Ukochka (approved):

PR #1726, opened 2023-04-11
Fix GForm Select dropdowns

mderynck:
Return behavior so GForm select drop-down accesses value field instead of using the KeyValuePair.


matiasb (approved):
LGTM

teodosii (approved):

PR #1728, opened 2023-04-12
New webhook improvements

mderynck:
- Enable/Disable Toggle
- Add Integrations Filter
- Enable log for non-triggered webhook for troubleshooting
- Improve UI to better show enabled, trigger status, request status
- Standardize terms New, Firing, Triggered to Firing
- Mask password field
- Fix bug where UI did not update when webhook is deleted


engine/apps/webhooks/tasks/trigger_webhook.py
```diff
@@ -123,6 +94,10 @@ def execute_webhook(webhook_pk, alert_group_id, user_id, escalation_policy_id):
 
     exception = error = None
     try:
+        if not webhook.check_integration_filter(alert_group):
+            status["request_trigger"] = f"Alert group was not from a selected integration"
+            return status, None, None
```
matiasb, line 99: Something I'm worried about here is that we will be creating a WebhookResponse entry when no request was triggered (same below for trigger template evaluation), and then if we want to use responses data, we won't have anything there, even not being called was the expected behavior. Wondering if we have any other way to let the user know, or maybe suggest that a non-triggered webhook could be related to this cases (integration not selected or trigger template evaluating to False).

matiasb (commented):
Looking good, just one question.

PR #1749, opened 2023-04-14
fix small bug in IRMBanner rendering logic

joeyorlando:


grafana-plugin/src/containers/IRMBanner/IRMBanner.tsx
```diff
@@ -15,9 +15,7 @@ const IRMBanner: React.FC = observer(() => {
   } = store;
 
   useEffect(() => {
-    if (store.isOpenSource()) {
-      alertGroupStore.fetchIRMPlan();
-    }
+    alertGroupStore.fetchIRMPlan();
```
joeyorlando, line 18: @teodosii I believe this should be `if (!store.isOpenSource()) {`, correct? I've moved that logic to the `alertGroupStore.fetchIRMPlan` method.

PR #1753, opened 2023-04-14
Webhooks 2 - Add copy ID to clipboard buttons

mderynck:
Add copy to clipboard buttons for integration, route, webhook IDs if webhooks 2 is enabled.

Screenshot from 2023-04-14 11-31-13

Screenshot from 2023-04-14 11-20-13

Updated tooltip:

Screenshot from 2023-04-14 12-00-08


engine/apps/api/views/features.py
```diff
@@ -58,4 +59,15 @@ def _get_enabled_features(self, request):
             if request.auth.organization.pk in enabled_web_schedules_orgs.json_value["org_ids"]:
                 enabled_features.append(FEATURE_WEB_SCHEDULES)
 
+        enabled_webhooks2_orgs = DynamicSetting.objects.get_or_create(
```
matiasb, line 62: FYI, there is a `is_webhooks_enabled_for_organization` helper in webhooks/utils.py that I think you could use here.

matiasb (approved):
Makes sense to me.

Not really sure if the copy icon is what we want though, can it be confusing what you will be copying at first sight, before reading the tooltip? Maybe we should somehow render the ID in the tooltip too?

Also, I wonder if we should add the same feature for webhooks in the webhooks2 listing?

grafana-plugin/src/containers/AlertReceiveChannelCard/AlertReceiveChannelCard.tsx
```diff
@@ -63,6 +65,22 @@ const AlertReceiveChannelCard = observer((props: AlertReceiveChannelCardProps) =
...
+                      ID {alertReceiveChannel.id}
+                      <br />
+                      (click to copy ID to clipboard)
+                    </div>
+                  }
+                  tooltipPlacement="top"
+                  name="info-circle"
+                />
+              </CopyToClipboard>
+            )}
```
joeyorlando, line 83: _nit_; might make sense to move this series of components to a shared component that is reused in these 3 cases in the PR

PR #1756, opened 2023-04-17
Allow use of dynamic payloads in alert receive channels preview templ…

iskhakov:
…ate in private api

# What this PR does

## Checklist

### DIFF
--- a/engine/common/api_helpers/mixins.py
+++ b/engine/common/api_helpers/mixins.py
@@ -17,7 +17,6 @@ from apps.alerts.incident_appearance.templaters import (
     AlertWebTemplater,
     TemplateLoader,
 )
-from apps.alerts.models import Alert, AlertGroup, AlertReceiveChannel
 from apps.api.permissions import LegacyAccessControlRole
 from apps.base.messaging import get_messaging_backends
 from common.api_helpers.exceptions import BadRequest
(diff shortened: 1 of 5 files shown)


engine/common/api_helpers/mixins.py
```diff
@@ -288,25 +289,38 @@ class PreviewTemplateMixin:
...
         if attr_name in APPEARANCE_TEMPLATE_NAMES:
             if notification_channel is None:
                 raise BadRequest(detail={"notification_channel": "notification_channel is required"})
             if notification_channel not in NOTIFICATION_CHANNEL_OPTIONS:
                 raise BadRequest(detail={"notification_channel": "Unknown notification_channel"})
 
-        alert_to_template = self.get_alert_to_template()
-        if alert_to_template is None:
-            raise BadRequest(detail="Alert to preview does not exist")
+        if isinstance(self.get_object(), AlertReceiveChannel):
```
Konstantinov-Innokentii, line 309: Original idea was to use get_alert_to_template method from AlertReceiveChannelView and AlertGroupView to not to check instance types here. I propose to pass payload argument to this method and keep logic there.

Konstantinov-Innokentii, line 309: Original idea was to use get_alert_to_template method from AlertReceiveChannelView and AlertGroupView to not to check instance types here. I propose to pass payload argument to this method and keep logic there.

Konstantinov-Innokentii (changes requested):

Konstantinov-Innokentii (approved):

engine/common/api_helpers/mixins.py
iskhakov: nice idea! Implemented it

PR #1789, opened 2023-04-19
Fix helm hook for migration job

Mattes83:
# What this PR does
This PR fixes the migration job when using helm hooks.
Currently when enabling the useHooks option in the helm chart there is a deadlock as the hook is defined as "post-upgrade, post-install". In this case the database migration will only be executed when all other pods are ready. This will not happen as the pods are waiting for the database migration. So we need to use "pre-install, preupgrade" instead of "post-install,post-upgrade"

## Checklist

- [ x] `CHANGELOG.md` updated (or `pr:no changelog` PR label added if not required)


Mattes83:
Can someone please add the pr:no public docs label?

iskhakov (approved):

PR #1798, opened 2023-04-19
Add users_to_be_notified to new webhooks payload

mderynck:
- Change FIRING trigger for webhooks to be sent after escalation snapshot has been computed
- Extract users from `notify_to_users_queue` and `notify_schedule` from escalation snapshot to populate `users_to_be_notified` in webhook payload

### DIFF
--- a/engine/apps/alerts/tasks/distribute_alert.py
+++ b/engine/apps/alerts/tasks/distribute_alert.py
@@ -26,7 +26,7 @@ def distribute_alert(alert_id):
     if alert.is_the_first_alert_in_group:
         alert_group = AlertGroup.all_objects.filter(pk=alert.group_id).get()
         alert_group.start_escalation_if_needed(countdown=TASK_DELAY_SECONDS)
-        alert_group_escalation_snapshot_built.send(sender=distribute_alert, alert_group=alert_group.id)
+        alert_group_escalation_snapshot_built.send(sender=distribute_alert, alert_group=alert_group)
 
     updated_rows = Alert.objects.filter(pk=alert_id, delivered=True).update(delivered=True)
     if updated_rows != 1:
--- a/engine/apps/alerts/tasks/distribute_alert.py
+++ b/engine/apps/alerts/tasks/distribute_alert.py
@@ -2,7 +2,7 @@ from django.apps import apps
 from django.conf import settings
 
 from apps.alerts.constants import TASK_DELAY_SECONDS
-from apps.alerts.signals import alert_create_signal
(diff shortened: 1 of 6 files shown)


engine/apps/alerts/tasks/distribute_alert.py
```diff
@@ -26,6 +26,7 @@ def distribute_alert(alert_id):
     if alert.is_the_first_alert_in_group:
         alert_group = AlertGroup.all_objects.filter(pk=alert.group_id).get()
         alert_group.start_escalation_if_needed(countdown=TASK_DELAY_SECONDS)
+        alert_group_escalation_snapshot_built.send(sender=distribute_alert, alert_group=alert_group.id)
```
matiasb, line 29: I think you need to pass the `alert_group` instance here, not the id (unless you change this to be a task)

engine/apps/webhooks/utils.py
```diff
@@ -124,6 +125,18 @@ def _serialize_event_user(user):
...
+def _extract_users_from_escalation_snapshot(escalation_snapshot):
+    users = []
+    if escalation_snapshot:
+        for policy_snapshot in escalation_snapshot.escalation_policies_snapshots:
+            for user in policy_snapshot.notify_to_users_queue:
+                users.append(_serialize_event_user(user))
+            if policy_snapshot.notify_schedule:
+                for user in list_users_to_notify_from_ical(policy_snapshot.notify_schedule):
+                    users.append(_serialize_event_user(user))
+    return list({u["id"]: u for u in users}.values())
```
matiasb, line 137: Maybe build a set from the beginning, adding the serialized user to it, and then return the list of the set?

mderynck: Not sure if it is correct but I was trying to maintain insertion order as well as eliminating duplicates, I don't think I can do that with a set?  Reason behind it was if the escalation snapshot has multiple steps that it can get users from we would have a more accurate order.  My assumption is generally it will be more useful to get the first element for our best guess at "next" rather than searching through the array.

matiasb: Oh, I see. In that case, are we interested in knowing if a person will be notified multiple times? (and then just having a list where we append users?)

matiasb (approved):

engine/apps/webhooks/utils.py
mderynck: Good catch, I'll fix it in the current webhooks PR, something like
`list({u["foo"]: u for u in users if u}.values()' shdould cover it

PR #1800, opened 2023-04-20
fix for topnav changes with old version of grafana

Ukochka:
# What this PR does
Fix frontend while using topnav flag with older version of grafana
https://raintank-corp.slack.com/archives/C04JCU51NF8/p1681273331570749

### DIFF
--- a/grafana-plugin/src/plugin/GrafanaPluginRootPage.helpers.tsx
+++ b/grafana-plugin/src/plugin/GrafanaPluginRootPage.helpers.tsx
@@ -1,7 +1,7 @@
 import { config } from '@grafana/runtime';
 
 export function isTopNavbar(): boolean {
-  return !!config.featureToggles.topnav;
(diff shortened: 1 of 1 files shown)


joeyorlando (commented):
I would suggest updating `dependencies.grafanaDependency` and/or `dependencies.grafanaVersion` in the `plugin.json` [here](https://github.com/grafana/oncall/blob/dev/grafana-plugin/src/plugin.json#L610-L612) instead. These properties seem better suited for what is trying to be accomplished here.

joeyorlando:
@Ukochka I believe this PR is no longer relevant. In [`plugin.json`](https://github.com/grafana/oncall/blob/dev/grafana-plugin/src/plugin.json#L619C3-L622C4) we specify that OnCall can only be installed on Grafana >= 9.2.0:
```json
"dependencies": {
    "grafanaDependency": ">=9.2.0",
    "plugins": []
  }
```

PR #1811, opened 2023-04-24
Edit template modal

Ukochka:
- Drawer for editing templates
- Saving
- Checking the result using latest endpoints with payload 
- cheatsheets
- Additional parameters with conditional templates
- Use Custom payload


joeyorlando (commented):
can we add some e2e tests here for the basic user flows that this PR modifies/adds?

PR #1814, opened 2023-04-24
"You're Going OnCall" mobile app push notification

joeyorlando:
# What this PR does

https://www.loom.com/share/c5deb35309604cfdab6176c44de7b15e

## Checklist


Dieterbe (approved):
i can't review the python code, but in principle, LGTM

engine/apps/mobile_app/tasks.py
joeyorlando, line 1: this looks like a bigger diff than it actually is. I extracted the logic that is now in the following methods to avoid having to duplicate w/ the addition of this new push notification:
- `send_push_notification_to_fcm_relay`
- `_send_push_notification`
- `_construct_fcm_message`

engine/apps/mobile_app/tests/test_your_going_oncall_notification.py
joeyorlando, line 1: still WIP, need to finish these up

engine/apps/mobile_app/tasks.py
```diff
@@ -153,78 +184,189 @@ def _get_fcm_message(alert_group, user, registration_id, critical):
...
+
+    for schedule_event in schedule.final_events("UTC", now, days=5):
+        users = schedule_event["users"]
+
+        for user in users:
+            user_pk = user["pk"]
+            logger.info(f"Evaluating if we should send push notification for schedule {schedule_pk} for user {user_pk}")
+
+            try:
+                user = User.objects.get(public_primary_key=user_pk)
```
matiasb, line 349: You will be iterating over multiple events, and potentially the same user will appear several times. Would it be worth it to keep a dict/cache with already fetched users (and device? and maybe already notified check?) instead of going to the DB each time?

engine/apps/mobile_app/tasks.py
```diff
@@ -153,78 +184,189 @@ def _get_fcm_message(alert_group, user, registration_id, critical):
...
+        return
+
+    message = _get_alert_group_escalation_fcm_message(alert_group, user, device_to_notify, critical)
+    _send_push_notification(device_to_notify, message, _create_error_log_record)
+
+
+def should_we_send_going_oncall_push_notification(
+    now: timezone.datetime, user_settings: "MobileAppUserSettings", schedule_event: ScheduleEvent
+) -> bool:
+    NOTIFICATION_TIMING_BUFFER = 15 * 60  # 15 minutes in seconds
```
matiasb, line 298: Since the task is scheduled to run every 10 mins, is it possible we may send 2 notifications for the same user/shift combination?

matiasb (commented):
Makes sense, minor comments/questions added :+1:

engine/apps/mobile_app/tasks.py
joeyorlando: good idea 👍  updated in [3335209](https://github.com/grafana/oncall/pull/1814/commits/3335209caeee989c25bb7c59a6c53e1bf5ed7684)

engine/apps/mobile_app/tasks.py
joeyorlando: good point 👍 I ended up storing a key in redis for ~80mins to keep track of these notifications ([change](https://github.com/grafana/oncall/pull/1814/commits/c323eac5d1af995f7ac467b1d488bf869d100d06#diff-b967f09465191f5c7bf2d5cca2356b32aeec88ef5d99946128b8edec0a084cb4R377-R429)) and avoid duplicates

joeyorlando:
discussed w/ @grafana/grafana-oncall-mobile, we will document this feature in a forthcoming PR

engine/apps/mobile_app/tasks.py
```diff
@@ -153,78 +185,256 @@ def _get_fcm_message(alert_group, user, registration_id, critical):
...
+    except OnCallSchedule.DoesNotExist:
+        logger.info(f"Tried to notify user about going on-call for non-existing schedule {schedule_pk}")
+        return
+
+    now = timezone.now()
+    current_date_with_hour = now.replace(minute=0, second=0, microsecond=0).strftime("%Y-%m-%d-%Hh")
+    schedule_final_events = schedule.final_events("UTC", now, days=7)
+
+    relevant_cache_keys = [
+        _generate_going_oncall_push_notification_cache_key(user["pk"], schedule_event, current_date_with_hour)
```
matiasb, line 381: edge case, but using `current_date_with_hour` couldn't still lead to duplicates? (e.g. my shift starts at 16:02?) wouldn't be enough to use user and event keys? (and caching the key for 1h)

matiasb (approved):
LGTM, 2 last questions added.

PR #1829, opened 2023-04-26
Webhook 2 improvements and fixes

mderynck:
- Rename Firing to Alert Group Created to reduce confusion as to why the event only first once and not when unresolve or unacknowledge returns the alert group to the firing state.
- Increase password field length
- Do not filter webhook execution by team, team is just for filtering ownership now
- Do not log webhook triggers in alert group escalation log if the webhook does not trigger (Status/response will still be stored)
- Fix formatting for response content and data fields on the Status page
- Add a content length limit for responses being stored (50000 characters)


engine/apps/webhooks/tasks/trigger_webhook.py
```diff
@@ -119,12 +121,16 @@ def make_request(webhook, alert_group, data):
...
-                status["content"] = response.content.decode("utf-8")
+            content_length = response.headers.get("Content-Length")
+            if content_length and int(content_length) < WEBHOOK_RESPONSE_LIMIT:
+                try:
+                    status["content"] = json.dumps(response.json())
+                except JSONDecodeError:
+                    status["content"] = response.content.decode("utf-8")
+            else:
+                status["content"] = f"Response content exceeds {WEBHOOK_RESPONSE_LIMIT} character limit"
         else:
```
matiasb, line 132: Wondering if we could remove this else block, and update the closing return to be:
`return triggered, status, error, exception`

grafana-plugin/src/containers/OutgoingWebhook2Status/OutgoingWebhook2Status.tsx
```diff
@@ -100,9 +113,7 @@ const OutgoingWebhook2Status = observer((props: OutgoingWebhook2StatusProps) =>
               {data.last_response_log.content && (
                 <VerticalGroup>
                   <Label>Response Body</Label>
-                  <SourceCode showClipboardIconOnly>
-                    {JSON.stringify(data.last_response_log.content, null, 4)}
-                  </SourceCode>
+                  <SourceCode showClipboardIconOnly>{format_response_field(data.last_response_log.content)}</SourceCode>
```
matiasb, line 116: FYI, Raphael suggested having response status code and response content at the top of the status log. I guess we can revisit some (easy) UI/UX improvements in a later PR.

matiasb (approved):
LGTM

PR #1830, opened 2023-04-26
Webhooks 2 Release

mderynck:
- Enables new webhooks functionality.
- Database migration will automatically convert existing webhooks to new ones.  Note: Converted webhooks are considered "legacy" they will continue to work as part of your escalation chain but will no longer be editable.  To make changes use the `Make a copy` action and edit that one, after you can delete your legacy webhook.  Remember to connect your escalation chain with your newly copied webhook!


joeyorlando (approved):
LGTM 👍 frontend changes seem fine as well but may be worthwhile getting a 2nd 👍 from someone on @grafana/grafana-oncall-frontend.

You might want to checkout [this frontend function](https://github.com/grafana/oncall/blob/dev/grafana-plugin/src/models/filters/filters.helpers.ts#L4-L5), it makes a reference to `outgoing_webhooks` vs `outgoing_webhooks_2`. Seems related w/ determining which filters to fetch from the backend to render.

grafana-plugin/src/models/filters/filters.helpers.ts
```diff
@@ -2,7 +2,6 @@ export const getApiPathByPage = (page: string) => {
   return (
     {
       outgoing_webhooks: 'custom_buttons',
-      outgoing_webhooks_2: 'webhooks',
```
mderynck, line 5: Turns out we don't pass `outgoing_webhooks_2` into this and rely on the fallback, webhooks is being passed in and returned.

PR #1833, opened 2023-04-27
Unify "name" and "title" field for oncall shift

Konstantinov-Innokentii:
# What this PR does
It drops "title" field from OnCall shift and makes "name" field not-required and nullable.  It is needed to provide ability to change shift name in web ui.

## Checklist


engine/apps/schedules/models/custom_on_call_shift.py
```diff
@@ -172,8 +170,9 @@ class CustomOnCallShift(models.Model):
         null=True,
         default=None,
     )
-    name = models.CharField(max_length=200)
-    title = models.CharField(max_length=200, null=True, default=None)
+    name = models.CharField(max_length=200, null=True, default=None)
+    # name = models.CharField(max_length=200)
+    # title = models.CharField(max_length=200, null=True, default=None)
```
Ferril, line 175: Probably forgot to remove commented code

engine/apps/schedules/models/custom_on_call_shift.py
```diff
@@ -705,10 +704,6 @@ def create_or_update_last_shift(self, data):
         instance_data["start_rotation_from_user_index"] = start_rotation_from_user_index
 
         if self.last_updated_shift is None or self.last_updated_shift.event_is_started:
-            # create new shift
```
Ferril, line 708: There is no need to remove this comment as it is related to the code below.
Also I think, we should set the current's shift name here or exclude the "name" from the `fields_to_remove` above, if we want shifts belonging to the same rotation to have the same name

engine/apps/schedules/migrations/0012_auto_20230427_0549.py
```diff
@@ -0,0 +1,36 @@
...
+            model_name='customoncallshift',
+            name='title',
+        ),
+        migrations.AlterField(
+            model_name='customoncallshift',
+            name='name',
+            field=models.CharField(default=None, max_length=200, null=True),
+        ),
+        migrations.RunPython(drop_autogenerated_web_shift_name, autogenerate_web_shift_name),
+    ]
```
Ferril, line 36: It's also probably needed to remove or change the `unique_together` constraint for the field `name`

Ferril (approved):

Konstantinov-Innokentii:
merged in maxim/1502-visualize-schedule-weekends branch

PR #1839, opened 2023-04-27
Docs: Update user notification setting

alfredo-d:
# What this PR does
Add more information on:
- configure ms teams users
- update slack user setting instruction with new UI


Matvey-Kuk:
I removed section about MS Teams because it's not actual anymore but thank you for the contribution!

Issue #1452, opened 2023-03-03
Can't create schedules with name used by the other team


orz80i1345:
I have some problems
show ststus=400  when I am create schedules 
How can I do?

log:
msg="Request Completed" method=POST path=/api/plugin-proxy/grafana-oncall-app/api/internal/v1/schedules/ status=400

![image](https://user-images.githubusercontent.com/126868261/222691553-e43d8c39-51b2-4d15-bf16-390967efbc5d.png)

orz80i1345:
![image](https://user-images.githubusercontent.com/126868261/222691414-fd6416cb-f9cb-4106-a212-03f36b1e65ab.png)

Konstantinov-Innokentii:
@orz80i1345 Oh, it seems that schedule with same name exists in your organization, but  in another team. It looks like a validation bug on oncall side. You can try different name for your schedule.

Thanks for reporting!

PR #1859, opened 2023-05-02
Remove schedule name uniqueness restriction

matiasb:
Related to https://github.com/grafana/oncall/issues/1452


mderynck (approved):

engine/apps/schedules/migrations/0012_auto_20230502_1259.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('schedules', '0011_oncallschedule_cached_ical_final_schedule'),
+    ]
+
+    operations = [
+        migrations.AlterModelOptions(
```
vadimkerr, line 13: Looks like nothing changed about the model manager, is this expected?

vadimkerr (approved):
LGTM 👍 
nit: maybe add a simple unit test creating two schedules with the same name?

engine/apps/schedules/migrations/0012_auto_20230502_1259.py
matiasb: This was added by the automatic migration generation, will double-check.

matiasb: This seems related to a previous change in the manager, for which there wasn't a migration created. I think the change was adding a method to the custom base queryset, which is used as the default `objects` manager via `PolymorphicManager` (so, not a real model/db change, but it seems this is detected as a modification for which a migration step is added).

PR #1865, opened 2023-05-03
Fix and tidy alertmanager heartbeat template

colega:
# What this PR does

There was an unnecessary indentation in the `rules:` key which made it invalid YAML.

Also replaced the mentions to Amixr with Grafana OnCall, used some `<code>` tags and reworded some sentences.

Also removed the anchor tag from the webhook link: we don't want people to follow that in their browser, we want them to copy it

## Result screenshot

image

## Checklist

### DIFF
--- a/engine/apps/integrations/templates/heartbeat_instructions/alertmanager.html
+++ b/engine/apps/integrations/templates/heartbeat_instructions/alertmanager.html
@@ -1,25 +1,25 @@
-<p>This configuration will send an alert once in a minute and if alertmanager stops working amixr will detect that it
-    doesn't work and notify you about that.</p>
+<p>This configuration will send an alert once in a minute and if alertmanager stops working OnCall will detect
+    it and notify you about that.</p>
 <ol>
-    <li>Add the alert generating script to prometheus.yaml file
-        Within prometheus it is trivial to create an expression that we can build a heartbeat with Amixr
-        expr: vector(1)
-        <p>That expression will always return true. Here is an alert that leverages the previous expression to create a
-            heartbeat alert</p>
+    <li>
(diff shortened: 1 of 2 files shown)


engine/apps/integrations/templates/heartbeat_instructions/alertmanager.html
```diff
@@ -1,38 +1,38 @@
-<p>This configuration will send an alert once in a minute and if alertmanager stops working amixr will detect that it
-    doesn't work and notify you about that.</p>
+<p>This configuration will send an alert once in a minute and if alertmanager stops working OnCall will detect
```
joeyorlando, line 1:
```suggestion
<p>This configuration will send an alert once a minute, and if alertmanager stops working, OnCall will detect
```

grafana-plugin/src/containers/HeartbeatModal/HeartbeatForm.tsx
joeyorlando, line 1: do you mind including a screenshot of this change? I vaguely recall this URL linking to an HTML page which had some relevant information for users (but I may be thinking of something else here).

joeyorlando (approved):

grafana-plugin/src/containers/HeartbeatModal/HeartbeatForm.tsx
colega: This URL links to the API endpoint where heartbeat alerts should be received. All it returns is a plain text:

```
":)"
```

colega: I've updated the PR description with a screenshot of the result.

colega:
Does this need a changelog entry?

Issue #1553, opened 2023-03-15
Slack integrations can rapidly repeat notices for unassigned time periods and shift changes


tleecsm:
When using an iCal link for attaching an Outlook calendar to Grafana OnCall, it is possible to trigger a very noisy bug that fills a target slack channel with notifications from the integrated Slack application.  We noticed after updating one of our Outlook calendars that is attached to Grafana OnCall, that the slack application started spamming "There are time periods that are unassigned..." and "New on-call shift:..." messages to the team's targeted Slack channel.  These messages seemed to be sending once every 10 minutes and quickly made the Slack channel unusable due to the spam.  

A potential workaround for this is disabling the meta notifications from Grafana OnCall, but we found in our case, deleting the Outlook calendar and attaching a new one resolved the issue.  Matvey mentioned a similar issue for Google calendars that was fixed in a recent update to OnCall.  

[Additional context and screenshots can be found on the public Grafana slack.](https://grafana.slack.com/archives/C02LSUUSE2G/p1678743172095379)

PR #1870, opened 2023-05-03
Update ical comparison to check only for event components

matiasb:
May be related to https://github.com/grafana/oncall/issues/1553.

We got feedback about that happening for Google Calendar imported icals.
Google Calendar exported ics URL was returning different VTIMEZONE components on different requests, triggering differences in the imported ical. Updated the comparison to only consider events (while keep assuming the sequence will reflect if there are any particular event change).


mderynck (approved):

engine/apps/schedules/ical_utils.py
```diff
@@ -509,21 +509,15 @@ def is_icals_equal(first, second):
...
-            if first_seq != second_seq:
-                return False
-        return True
+        # only consider VEVENT components
+        first_cal_events = {
+            cmp.get("UID", None): cmp.get("SEQUENCE", None) for cmp in first_subcomponents if cmp.name == "VEVENT"
+        }
+        second_cal_events = {
+            cmp.get("UID", None): cmp.get("SEQUENCE", None) for cmp in second_subcomponents if cmp.name == "VEVENT"
+        }
```
joeyorlando, line 518: _nit_ consider deduplicating the `dict` construction logic into a separate function

joeyorlando (approved):

PR #1885, opened 2023-05-05
Fix teams filter related issue setting maintenance mode

matiasb:
Related to https://github.com/grafana/support-escalations/issues/5862


mderynck (approved):
Thank you!
Was also mentioned in this one:
https://github.com/grafana/support-escalations/issues/5529

PR #1917, opened 2023-05-10
add the ability to set/display "currently undergoing maintenance message" in the UI

joeyorlando:
# What this PR does

add a new endpoint, `GET /maintenance-mode/`, which returns either a string message pulled from the `CURRENTLY_UNDERGOING_MAINTENANCE_MESSAGE` env var, or `None`  + update the UI to conditionally show this message if it is set

Screenshot 2023-05-10 at 11 28 16

## Checklist

- [x] Documentation added (or `pr:no public docs` PR label added if not required) (N/A)


mderynck:
I guess we would need to fallback to this for the DB being down

engine/apps/grafana_plugin/views/status.py
```diff
@@ -38,5 +38,6 @@ def get(self, _request: Request) -> Response:
                 "is_user_anonymous": self.grafana_context["IsAnonymous"],
                 "license": settings.LICENSE,
                 "version": settings.VERSION,
+                "currently_undergoing_maintenance_message": settings.CURRENTLY_UNDERGOING_MAINTENANCE_MESSAGE,
```
vadimkerr, line 41: Wouldn't this result in a 500 HTTP error in case DB is unreachable? There are some DB queries running before returning the response, maybe it's better to check for this setting before accessing the DB (including DB access in authentication classes)?

vadimkerr (approved):

engine/apps/grafana_plugin/views/status.py
mderynck: This has me thinking about whether maintenance mode should be more explicit as in a different settings file or possibly introducing middleware instead of handling different API calls case by case.

vadimkerr (approved):
LGTM, minor question/suggestion on HTTP codes

mderynck:
Take a look at the [gateway routes](https://github.com/grafana/backend-enterprise/blob/f5c4e2713c6832c179541c81f673d200cbe864c6/pkg/gateway/routes.go#L728) currently we don't have a route through to maintenance mode it will need to be added or moved to be under existing API paths. I would prefer it to go under existing API path since routes are classified for [alerting purposes + SLO calculation](https://github.com/grafana/deployment_tools/blob/0b87e6bf869b0e62366a86557470ce5df6222ac2/ksonnet/lib/slo/oncall.libsonnet#L31).  Also it would be nice not to have special handling for the URL in the plugin (includeApiPathPrefix) for one URL.

engine/engine/urls.py
```diff
@@ -68,3 +73,7 @@
     urlpatterns += [path(settings.SILK_PATH, include("silk.urls", namespace="silk"))]
 
 admin.site.site_header = settings.ADMIN_SITE_HEADER
+
+# NOTE: keep this at the end of the file to properly override urlpatterns
+if settings.IS_IN_MAINTENANCE_MODE:
+    urlpatterns = paths_to_work_even_when_maintenance_mode_is_active
```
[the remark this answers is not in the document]
joeyorlando: @mderynck raised a good point here about this that I agree with. Allowing traffic to these routes is less deterministic when the database is down. Some routes might work, some might complete a small amount of work before returning a 500.

engine/engine/celery.py
```diff
@@ -16,12 +16,9 @@
 
 os.environ.setdefault("DJANGO_SETTINGS_MODULE", "settings.prod")
 
-from django.db import connection  # noqa: E402
-
 logger = get_task_logger(__name__)
 logger.setLevel(logging.DEBUG)
 
-connection.cursor()
```
joeyorlando, line 24: had to remove this because the API container's startup command imports this file somewhere along the way, and `connection.cursor()` throws an exception if the database is unavailable

engine/engine/urls.py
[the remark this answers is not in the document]
mderynck: Maybe we need to check the setting when adding middleware too, add a custom middleware to return 503 for endpoints that would resolve normally. Returning 404 is bad because people are more likely to fiddle with their config if they didn't know we are doing maintenance (Did not pay attention to the warning).  503 at least lets them know its on our side.

grafana-plugin/src/plugin/PluginSetup/PluginSetup.skipped.tsx
joeyorlando, line 1: @grafana/grafana-oncall-frontend this test file passes locally when renamed to `PluginSetup.test.tsx`, however it fails on CI w/ a very cryptic error message. Any thoughts?

maskin25: no idea, when and why it was renamed to _skipped_ 🤔

grafana-plugin/src/state/rootBaseStore/index.ts
```diff
@@ -56,6 +56,9 @@ export class RootBaseStore {
   @observable
   initializationError = null;
 
+  @observable
+  currentlyUndergoingMaintenance = false;
```
maskin25, line 60: can we reuse `initializationError` instead of adding new property `currentlyUndergoingMaintenance` here?

maskin25 (approved):

grafana-plugin/src/state/rootBaseStore/index.ts
joeyorlando: I wanted to add this new attribute to distinguish between a setup error and maintenance mode. For maintenance mode, I think we should disable the "Retry" and "Configure Plugin" buttons ([here](https://github.com/grafana/oncall/pull/1917/files#diff-e3115759934c8107978a15e1ab02872bba90969b069b6c0d3dadec4a83c1b119R50-R61))

Issue #1103, opened 2023-01-07
Telegram integration error "Flood control exceeded"


aksenk:
Hi!

I have a problem with configuring telegram integration.

I have deployed grafana oncall via helm chart and installed telegram token and telegram webhookUrl (like `https://oncall.mydomain.com/`).

But when I open "ChatOps -> Telegram" page I see message: `To connect channel setup Telegram environment first, which includes connection to your bot and host URL.`

If I go to the page "Env Variables" I see that my variables are filled in, but in column `Status` for variable `TELEGRAM_WEBHOOK_HOST` I see error: `Telegram error: Flood control exceeded. Retry in 1.0 seconds`.

I don't understand what's wrong, because I created a new bot for a new installation (which doesn't even have a single alert yet).

Please help.

Sammyant:
Hi, we have the same issue with new version (v1.1.19) with docker compose

joanyko:
Hi, same issue in helm chart (v1.1.0)

Acetolyne:
please try removing the trailing / from the url I noticed when setting mine up that i could not have the trailing / in the url else I got the same message. If this fixes the issue for you please let us know and I can put in a PR to automatically remove the trailing slash so that the plugin handles these URLs in a better way.

kirpozh:
The problem still persists. I tried to proxy traffic through a separate endpoint, I also tried through ngrok, as described in the tutorial video - it does not work.
Removing trailing from the message above didn't help either.
As described in the first message, that only downgrade to a version lower than 1.0.51 works, then I can't do it because of the inability to put this version through compose.
Please help

vitvord:
We have the same problem. Removing the database and completely reinstalling redis/rabbitmq/oncall does not help.
webhookUrl is publicly available.

oncall app and plugin: v1.1.29

kuzaxak:
Version: 1.2.22

Same behaviour without tailing `/`

alexintech:
I had the same problem. I suppose if for some period `TELEGRAM_WEBHOOK_HOST` was wrong or publicly unavailable, such error may occur and oncall saves it into database.

When I change `TELEGRAM_WEBHOOK_HOST` in UI, it successfully registers webhook. It can be verified by using:

```
curl --location --request POST 'https://api.telegram.org/bot{{token}}/getWebhookInfo'
```
[...]

PR #1934, opened 2023-05-15
Fix bug #1103 in setting TELEGRAM_WEBHOOK_HOST causing `Flood control exceeded` error

alexintech:
# What this PR does

When `TELEGRAM_WEBHOOK_HOST` sets into database it calls Telegram's `setWebhook` twice. The first call on field validation goes correctly. But after saving the field `LiveSettings.validate_settings()` runs, which validates all settings. It calls Telegram's `setWebhook` again and in that case receives `Flood control exceeded` error from Telegram.

I have found that `LiveSettings.validate_settings` has been invented in commit https://github.com/grafana/oncall/commit/ef37e6a6fd85b4ff62aa8e351d76b788fb94471f like a replacement for twilio dependent settings.

This PR revalidates only updated dependent settings. So it does not flood Telegram API.

[#1103](https://github.com/grafana/oncall/issues/1103)

## Checklist


vadimkerr (changes requested):
Hi @alexintech 👋 Thank you for opening a PR!
Looks like this change disables live setting validation for everything except Twilio, and I'm not sure this would be a good idea. Please see [LiveSettingsValidator](https://github.com/grafana/oncall/blob/2905423bad3e3b11de60434cb7f13fde8e1b43d1/engine/apps/base/utils.py#L34-L179) for more info on live settings validation.

Have you considered removing the second call to `setWebhook` [here](https://github.com/grafana/oncall/blob/9e598385f492df6d7fd336c0ef441762fc4f3a72/engine/apps/api/views/live_setting.py#L73-L74)? The first call to `setWebhook` is performed on [serializer save](https://github.com/grafana/oncall/blob/9e598385f492df6d7fd336c0ef441762fc4f3a72/engine/apps/api/views/live_setting.py#L52), so the second call indeed seems to be redundant. Let me know if this makes sense 🙂

alexintech:
> Hi @alexintech 👋 Thank you for opening a PR! Looks like this change disables live setting validation for everything except Twilio, and I'm not sure this would be a good idea. Please see [LiveSettingsValidator](https://github.com/grafana/oncall/blob/2905423bad3e3b11de60434cb7f13fde8e1b43d1/engine/apps/base/utils.py#L34-L179) for more info on live settings validation.
> 
> Have you considered removing the second call to `setWebhook` [here](https://github.com/grafana/oncall/blob/9e598385f492df6d7fd336c0ef441762fc4f3a72/engine/apps/api/views/live_setting.py#L73-L74)? The first call to `setWebhook` is performed on [serializer save](https://github.com/grafana/oncall/blob/9e598385f492df6d7fd336c0ef441762fc4f3a72/engine/apps/api/views/live_setting.py#L52), so the second call indeed seems to be redundant. Let me know if this makes sense 🙂

Hi, @vadimkerr! Thank you for your attention. 

Actually, there are three possible calls to the `setWebhook` API when updating a live setting.

**First case (valid webhook URL entered):**
1. The first call to `setWebhook` is performed on [serializer save](https://github.com/grafana/oncall/blob/9e598385f492df6d7fd336c0ef441762fc4f3a72/engine/apps/api/views/live_setting.py#L52) as you mentioned. It sets a valid URL to Telegram and clears `error` field if one exists in the database.
2. The second call [here](https://github.com/grafana/oncall/blob/9e598385f492df6d7fd336c0ef441762fc4f3a72/engine/apps/api/views/live_setting.py#L73-L74) won't actually call the `setWebhook` endpoint because it checks [here](https://github.com/grafana/oncall/blob/5fae708d0acda4548bae466813a993d4f72cfef4/engine/apps/telegram/client.py#L42-L44) if the webhook URL has already been set. 
3. The second actual call to `setWebhook` is performed in [LiveSettings.validate_settings()](https://github.com/grafana/oncall/blob/5fae708d0acda4548bae466813a993d4f72cfef4/engine/apps/base/models/live_setting.py#L211-L213). This call results in a `Flood control exceeded` error, and this error is saved to the database.

**Second case (invalid webhook URL entered, for example, https://123.ngrok):**

1. The first call to `setWebhook` is performed on [serializer save](https://github.com/grafana/oncall/blob/9e598385f492df6d7fd336c0ef441762fc4f3a72/engine/apps/api/views/live_setting.py#L52). Telegram responds with the error `Bad Request: bad webhook: Failed to resolve host: Name or service not known` and unsets the webhook URL if one was set before. This error is saved to the database.
2. The second call to `setWebhook` happens [here](https://github.com/grafana/oncall/blob/9e598385f492df6d7fd336c0ef441762fc4f3a72/engine/apps/api/views/live_setting.py#L73-L74) after the webhook URL has been unset, and it sends a real request to Telegram. However, the result of this call does not change the error field in the database.
3. The third call to the `setWebhook` endpoint happens in [LiveSettings.validate_settings()](https://github.com/grafana/oncall/blob/5fae708d0acda4548bae466813a993d4f72cfef4/engine/apps/base/models/live_setting.py#L211-L213). This call results in a `Flood control exceeded` error, which overrides the more informative error message from the first step. This means that the actual cause of the error, the invalid webhook URL, is not saved in the database. Instead, the database will contain the `Flood control exceeded` error, which is not helpful in identifying and resolving the underlying issue.

I have considered removing [these lines](https://github.com/grafana/oncall/blob/9e598385f492df6d7fd336c0ef441762fc4f3a72/engine/apps/api/views/live_setting.py#L73-L74):
1. In the first case it doesn't change anything;
2. In the second case it also doesn't change anything.

Another option I have tried is modifing [_check_telegram_webhook_host()](https://github.com/grafana/oncall/blob/5fae708d0acda4548bae466813a993d4f72cfef4/engine/apps/base/utils.py#L139C9-L145) adding validation for existing url:
```python
            url = create_engine_url("/telegram/", override_base=telegram_webhook_host)
            bot = Bot(token=live_settings.TELEGRAM_TOKEN)

            webhook_info = bot.get_webhook_info()
            if webhook_info.url == url:
                return

            bot.set_webhook(url)
```
1. It actually resolves the issue in the first case;
2. However, in the second case, it shows the same Flood control exceeded error instead of the more informative Bad Request: bad webhook: Failed to resolve host: Name or service not known error. This happens even with the removal of [these lines](https://github.com/grafana/oncall/blob/9e598385f492df6d7fd336c0ef441762fc4f3a72/engine/apps/api/views/live_setting.py#L73-L74) 

In [LiveSettingsValidator](https://github.com/grafana/oncall/blob/2905423bad3e3b11de60434cb7f13fde8e1b43d1/engine/apps/base/utils.py#L34-L179) I see only Twilio settings use multiple settings to execute validation. Currently when I change a setting value, it validates one setting, and then runs validations for all existing settings.

When I update one setting value, basically I'm excepting that only that field and other dependent on it would be validated.
But if you really need to revalidate all settings, currently you need to change some value back and forth. Maybe for that case it would be better to have an extra button in UI `Revalidate all settings`?  I'm not sure if revalidation of _**all**_ live settings is required after every change.

vadimkerr:
The issue should be fixed in https://github.com/grafana/oncall/pull/2100, so I'm closing this PR. @alexintech thank you for the contribution!

PR #1951, opened 2023-05-17
Fix for creating Integration

Ukochka:
# What this PR does

Fix for creating Integration
After creating Integration we land on Integration page, not a table
Added error notification for TemplateFoEdit obj to avoid crashing
Loading state for AlertGroup List
Small style fixes

### DIFF
--- a/grafana-plugin/src/pages/integration_2/Integration2.tsx
+++ b/grafana-plugin/src/pages/integration_2/Integration2.tsx
@@ -579,10 +579,14 @@ class Integration2 extends React.Component<Integration2Props, Integration2State>
   getTemplatesList = (): CascaderOption[] => INTEGRATION_TEMPLATES_LIST;
 
   openEditTemplateModal = (templateName, channelFilterId?: ChannelFilter['id']) => {
-    this.setState({
-      isEditTemplateModalOpen: true,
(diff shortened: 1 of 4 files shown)


grafana-plugin/src/pages/integration_2/Integration2.tsx
```diff
@@ -579,10 +579,14 @@ class Integration2 extends React.Component<Integration2Props, Integration2State>
...
-      isEditTemplateModalOpen: true,
-      selectedTemplate: templateForEdit[templateName],
-    });
+    if (templateForEdit[templateName]) {
+      this.setState({
+        isEditTemplateModalOpen: true,
+        selectedTemplate: templateForEdit[templateName],
+      });
+    } else {
+      openErrorNotification('Template can not be edited. Please contact support.');
```
teodosii, line 588: How can we reach this state where the template is undefined?

Ukochka: For example if the name of the template will be changed from backend, and we did not change it on frontend side, because I am using TemplateForEdit config to have all the parameters I need
Also we have this problem when we update in OSS version and forget smth in private (happened twice already)

teodosii (approved):

PR #1958, opened 2023-05-17
Refactoring alert group metrics

matiasb:
Trying to:

 - move response_time to separate model (to avoid alert group table changes)
 - use same function to update cache in every case, to unify metrics cache update in one method
 -  do the metrics update after db is updated in all cases
 - always pass organization_id


engine/apps/alerts/migrations/0016_alertgroupmetrics.py
joeyorlando, line 1: is this to avoid having to add a column to the alertgroup table? after next Monday this should no longer be an issue

PR #1970, opened 2023-05-18
Fix inbound email bug when attaching files

vadimkerr:
# What this PR does
Fix inbound email endpoint bug when attaching files to email leads to HTTP 500.

## Checklist


engine/apps/email/inbound.py
```diff
@@ -69,7 +69,7 @@ def post(self, request, alert_receive_channel):
                 alert_receive_channel_pk=alert_receive_channel.pk,
                 image_url=None,
                 link_to_upstream_details=None,
-                integration_unique_data=request.data,
+                integration_unique_data=None,
```
vadimkerr, line 72: This data is not needed, the only fields available for templating are subject, message, and sender.

matiasb (approved):

PR #1976, opened 2023-05-19
Map country code to different Twilio resources 

mderynck:
Many countries are introducing different requirements for SMS senders to register and/or use alpha numeric ids, short codes or regional numbers or face being blocked.  The changes in this PR will give us more flexibility by allowing us to map to different resources in Twilio based on the phone number we are trying to reach.  For this first implementation the selection is made based on country code of the recipient.  Verification and phone calls were given the same treatment although the immediate need is for SMS.  Senders with no country code set can be used as catch-all defaults.  This also falls back to the configured live settings/environment variables if not configured.

Possible future additions:
- Move through list of trying multiple senders before failing notification
- Easily expanded to allow per-organization or per-user resources to let users and tenants configure their own Twilio
- Add UI + replace live settings so users can configure their own settings
- More selection criteria if needed

TODO:

- [x] Add+Fix Tests
- [x] Verify changes are compatible with #1713


engine/apps/twilioapp/models/twilio_sender.py
```diff
@@ -0,0 +1,37 @@
...
+    api_key_secret = models.CharField(max_length=64, null=True, default=None)
+
+    def get_twilio_api_client(self):
+        if self.api_key_sid and self.api_key_secret:
+            return Client(self.api_key_sid, self.api_key_secret, self.account_sid)
+        else:
+            return Client(self.account_sid, self.auth_token)
+
+
+class TwilioSender(PolymorphicModel):
```
vadimkerr, line 20: Maybe it would be easier to have this model as an abstract Django model? Having an abstract model here will keep things DRY without having to deal with any of the polymorphic stuff (which sometimes can make migrations more complex and introduce [random weirdness](https://github.com/grafana/oncall/pull/1652)).

engine/apps/twilioapp/phone_provider.py
```diff
@@ -217,24 +219,47 @@ def _parse_number(self, number: str):
...
             return Client(live_settings.TWILIO_ACCOUNT_SID, live_settings.TWILIO_AUTH_TOKEN)
 
+    @property
+    def _default_twilio_number(self):
+        return live_settings.TWILIO_NUMBER
+
+    def _twilio_sender(self, sender_type, to):
+        _, _, country_code = self._parse_number(to)
+        TwilioSender = apps.get_model("twilioapp", sender_type)
+        senders = list(TwilioSender.objects.filter(Q(country_code=country_code) | Q(country_code__isnull=True)))
```
matiasb, line 237: maybe you can use [order_by + asc or desc](https://docs.djangoproject.com/en/4.2/ref/models/expressions/#django.db.models.Expression.asc) here? (instead of sorting in Python) and then get the `.first()` from the resulting queryset ?

PR #1997, opened 2023-05-23
e2e tests on CI - actually await k8s resources to be ready before starting tests

joeyorlando:
Occasionally, the Playwright global setup step (which authenticates w/ the Grafana API + configures the plugin) would fail, leading to the CI job to instantly fail (playwright doesn't retry global setup if it fails).

My current hypothesis as to why this is happening is because the `oncall-engine` and `oncall-celery` pods aren't _actually_  ready in these cases based on the way the `jupyterhub/action-k8s-await-workloads` action await k8s workloads:

Screenshot 2023-05-23 at 18 24 36

By using the `kubectl rollout status deployment/<deployment-name> --timeout=300s` instead, we can be sure that these pods are _actually_ ready to receive traffic before we start the tests.
```bash
❯ kubectl rollout status --help
Show the status of the rollout.

 By default 'rollout status' will watch the status of the latest rollout until it's done. If you don't want to wait for
the rollout to finish then you can use --watch=false. Note that if a new rollout starts in-between, then 'rollout
status' will continue watching the latest revision. If you want to pin to a specific revision and abort if it is rolled
over by another revision, use --revision=N where N is the revision you need to watch for.
```

Lastly, even despite this, sometimes the `POST /api/internal/v1/plugin/sync` endpoint will return HTTP 500 ([example logs](https://github.com/grafana/oncall/actions/runs/5062712137/jobs/9088529416#step:19:2536) from failed CI job). In this case, let's setup the Playwright global setup to retry 3 times.

### DIFF
--- a/.github/workflows/linting-and-tests.yml
+++ b/.github/workflows/linting-and-tests.yml
@@ -289,6 +289,10 @@ jobs:
 
       - name: Collect Workflow Telemetry
         uses: runforesight/workflow-telemetry-action@v1
+        with:
+          comment_on_pr: false
+          proc_trace_chart_show: false
+          proc_trace_table_show: false
 
       - name: Create k8s Kind Cluster
         uses: helm/kind-action@v1.3.0
--- a/.github/workflows/linting-and-tests.yml
+++ b/.github/workflows/linting-and-tests.yml
@@ -344,8 +344,8 @@ jobs:
       # and complete them much faster by using multiple test processes
       # With just 1 engine/celery/grafana pod, the backend crawls to a halt when there is > 1 parallelized integration test process
       # NOTE: it appears that using > 1 grafana container w/ SQLite as the database sometimes leads to failed
-      # grafana database migrations (presumably because multiple containers are trying to write to the same
-      # mounted sqlite file @ the same time)
+      # grafana database migrations (this is documented in this GitHub issue
+      # https://github.com/bitnami/charts/issues/10905)
       #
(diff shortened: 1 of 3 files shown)


.github/workflows/linting-and-tests.yml
```diff
@@ -361,7 +364,7 @@ jobs:
             --set oncall.twilio.authToken="${{ secrets.TWILIO_AUTH_TOKEN }}" \
             --set oncall.twilio.phoneNumber="\"${{ secrets.TWILIO_PHONE_NUMBER }}"\" \
             --set oncall.twilio.verifySid="${{ secrets.TWILIO_VERIFY_SID }}" \
-            --set grafana.replicas=3 \
+            --set grafana.replicas=1 \
```
joeyorlando, line 367: Can't reliably use > 1 grafana replica when using SQLite as the grafana database
https://github.com/bitnami/charts/issues/10905

.github/workflows/linting-and-tests.yml
```diff
@@ -287,6 +287,9 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
 
+      - name: Collect Workflow Telemetry
+        uses: runforesight/workflow-telemetry-action@v1
```
joeyorlando, line 291: this is useful to see how much CPU/memory various steps in the workflow are using. It attaches an artifact to the build that looks something like this:
Screenshot 2023-05-23 at 19 33 13
Screenshot 2023-05-23 at 19 33 20

PR #2027, opened 2023-05-25
changed notification title and subtitle; removed body

imtoori:
# What this PR does

## Checklist

### DIFF
--- a/engine/apps/mobile_app/alert_rendering.py
+++ b/engine/apps/mobile_app/alert_rendering.py
@@ -31,7 +31,7 @@ def get_push_notification_subtitle(alert_group):
 
 
     subtitle = f"#{alert_group.inside_organization_number} {alert_title}\n" 
-    + emojize(f"via {alert_group.channel.short_name}", use_aliases=True)
-    + f"\n{alert_status}"
+    +emojize(f"via {alert_group.channel.short_name}", use_aliases=True)
+    +f"\n{alert_status}"
 
     return subtitle
--- a/engine/apps/mobile_app/alert_rendering.py
+++ b/engine/apps/mobile_app/alert_rendering.py
@@ -29,8 +29,10 @@ def get_push_notification_subtitle(alert_group):
 
     alert_status = f"Status: {status_verbose}, alerts: {alerts_count_str}"
 
-    subtitle = (f"#{alert_group.inside_organization_number} {alert_title}\n"
-    +emojize(f"via {alert_group.channel.short_name}", use_aliases=True)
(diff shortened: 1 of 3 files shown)


engine/apps/mobile_app/alert_rendering.py
```diff
@@ -9,11 +9,28 @@ def _render_for(self):
...
+        alert_count_rounded = (number_of_alerts // 10) * 10
+        alerts_count_str = f"{alert_count_rounded}+"
+
+    alert_status = f"Status: {status_verbose}, alerts: {alerts_count_str}"
+
+    subtitle = f"#{alert_group.inside_organization_number} {alert_title}\n"
+    +emojize(f"via {alert_group.channel.short_name}", use_aliases=True)
+    +f"\n{alert_status}"
+
+    return subtitle
```
joeyorlando, line 36: it looks like [some unit tests](https://github.com/grafana/oncall/actions/runs/5082276846/jobs/9131808338?pr=2027#step:5:530) that reference this function are failing. I haven't tested if this works but you may want to try something like this instead:
```suggestion
    message = (
      f"#{alert_group.inside_organization_number} {alert_title} "
      f"via {alert_group.channel.short_name} "
      f"Status: {status_verbose}, alerts: {alerts_count_str}"
    )

    return emojize(message, use_aliases=True)
```

imtoori: ok I fixed everything 💪🏼

engine/apps/mobile_app/alert_rendering.py
```diff
@@ -9,11 +9,30 @@ def _render_for(self):
...
+        alert_count_rounded = (number_of_alerts // 10) * 10
+        alerts_count_str = f"{alert_count_rounded}+"
+
+    alert_status = f"Status: {status_verbose}, alerts: {alerts_count_str}"
+
+    subtitle = (
+        f"#{alert_group.inside_organization_number} {alert_title}\n"
+        + emojize(f"via {alert_group.channel.short_name}", use_aliases=True)
+        + f"\n{alert_status}"
     )
```
joeyorlando, line 36: you may want to pass the entire string assigned to the `subtitle` variable to the `emojize` function. If `alert_title` contains any emoji codes they won't be properly rendered I believe (may want to test this).

imtoori: I didn't do that just because currently that's how we do it, we don't "emojize" the alert title. 
But actually I think it's just better to do it

joeyorlando (approved):

PR #2034, opened 2023-05-26
TwilioPhoneProvider optimizations

Konstantinov-Innokentii:
# What this PR does
This PR does three improvements in twilio_phone_provider:
1. [Speed-up](https://github.com/grafana/oncall/pull/2034/files#diff-7a311767169c024e60e2b4e35fd531dd6e2f1ea785cfc84263e11e7932d622af) query which calculates amount of phone_calls/sms left. 
2. Remove code which was needed only for backward compatibility during the release of PhoneProvider refactoring and improves logging for handling status/gather updates.
3. Add db_index to twilio_sid. We are doing lot of lookups by sid and with increasing amount of data it became resource consuming.


engine/apps/twilioapp/status_callback.py
```diff
@@ -25,58 +25,63 @@ def update_twilio_call_status(call_sid, call_status):
 
     if call_sid and call_status:
         logger.info(f"twilioapp.update_twilio_call_status: processing sid={call_sid} status={call_status}")
-        status = TwilioCallStatuses.DETERMINANT.get(call_status)
+        status_code = TwilioCallStatuses.DETERMINANT.get(call_status)
+
+        if status_code is None:
+            logger.warning(
```
iskhakov, line 31: how normal is receiving unexpected status? This code won't alert us if we start receiving unexpected statuses

engine/apps/user_management/subscription_strategy/free_public_beta_subscription_strategy.py
```diff
@@ -61,12 +61,10 @@ def _calculate_phone_notifications_left(self, user):
         day_start = now.replace(hour=0, minute=0, second=0, microsecond=0)
         calls_today = PhoneCallRecord.objects.filter(
             created_at__gte=day_start,
-            represents_alert_group__channel__organization=self.organization,
```
iskhakov, line 64: 👍

iskhakov: we remove it because user can be a member of single organization and organization filter is excess, isn't it?

iskhakov (approved):

engine/apps/user_management/subscription_strategy/free_public_beta_subscription_strategy.py
Konstantinov-Innokentii: right

engine/apps/twilioapp/status_callback.py
Konstantinov-Innokentii: It's ok, status is only needed for better timeline. The thing about checking status always was in twilio part, so I decided to at least log it

PR #2039, opened 2023-05-26
Add initial setup for local dev prometheus exporter

matiasb:


mderynck (approved):

prometheus.yml
joeyorlando, line 1: _nit_; consider moving this to the dev directory

joeyorlando (approved):

Ferril (approved):

joeyorlando (commented):
@matiasb do you also need to update `dev/.gitignore` so that `dev/prometheus.yml` stays source controlled? Currently that file is set to this, so I'm wondering how this file got added to the PR?
```gitignore
# Ignore everything
*

# Except
!.env.dev.example
!.env.mysql.dev
!.env.postgres.dev
!.env.sqlite.dev
!add_env_var.sh
!README.md
```

matiasb:
> @matiasb do you also need to update [`dev/.gitignore`](https://github.com/grafana/oncall/blob/dev/dev/.gitignore) so that `dev/prometheus.yml` stays source controlled? Currently that file is set to this, so I'm wondering how this file got added to the PR?

I see, right :+1:  I git added before moving it into the dev dir, so that's probably how it ended here.

Issue #562, opened 2022-09-26
Non-functional deployment with helm chart


maxime-sourdin:
Hello,
I tried to deploy Oncall with helm chart (on a managed Kubernetes cluster, via ArgoCD), using the built-in Mysql.

I am seeing problems as the database migration jobs are not going through:

`amqp.exceptions.AccessRefused: (0, 0): (403) ACCESS_REFUSED - Login was refused using authentication mechanism PLAIN. For details see the broker logfile.`

When I did the tests based on the provided docker-compose, I had the same problem.

Here is an extract from the values file:
```
base_url: example.com
image:
  repository: grafana/oncall
  tag: "v1.0.37"
  pullPolicy: IfNotPresent
service:
  enabled: false
  type: LoadBalancer
  port: 8080
  annotations: {}
engine:
  replicaCount: 1
  resources: {}
celery:
  replicaCount: 1
  resources: {}
oncall:
  slack:
    enabled: false
    command: ~
    clientId: ~
    clientSecret: ~
    apiToken: ~
    apiTokenCommon: ~
  telegram:
    enabled: false
    token: ~
    webhookUrl: ~
migrate:
  enabled: true
env: []
ingress:
  enabled: false
  annotations:
    kubernetes.io/ingress.class: "nginx"
    cert-manager.io/issuer: "letsencrypt-prod"
  tls: 
    - hosts:
        - "{{ .Values.base_url }}"
      secretName: certificate-tls
  extraPaths: []
ingress-nginx:
  enabled: false
cert-manager:
  enabled: false
  installCRDs: false
  webhook:
    timeoutSeconds: 30
    securePort: 10260
  podDnsPolicy: None
  podDnsConfig:
    nameservers:
      - 8.8.8.8
      - 1.1.1.1
mariadb:
  enabled: true
  persistence:
    enabled: true
    storageClass: "csi-ssd-disk-topology"  
  auth:
    database: oncall
  primary:
    persistence:
      enabled: true
      storageClass: "csi-ssd-disk-topology"  
    extraEnvVars:
    - name: MARIADB_COLLATE
      value: utf8mb4_unicode_ci
    - name: MARIADB_CHARACTER_SET
      value: utf8mb4
  secondary:
    persistence:
      enabled: true
      storageClass: "csi-ssd-disk-topology"  
    extraEnvVars:
    - name: MARIADB_COLLATE
      value: utf8mb4_unicode_ci
    - name: MARIADB_CHARACTER_SET
[...]

PR #856, opened 2022-11-15
Remove net.core.somaxconn in Helm Chart

MadEngineX:
**What this PR does**:

When this PR was accepted https://github.com/grafana/oncall/pull/88/files somebody removed maxconn property for docker-compose https://github.com/grafana/oncall/issues/84#issuecomment-1156672285 but forget about Helm.

**Which issue(s) this PR fixes**:

This issue can be fixed by removing this property.

https://github.com/grafana/oncall/issues/562

**Checklist**
- [ ] Tests updated
- [ ] Documentation added
- [ ] `CHANGELOG.md` updated

PR #2045, opened 2023-05-29
Helm chart: added configuration of uwsgi using environment variables

alexintech:
# What this PR does

Adds `uwsgi` configuration to helm chart.
Sets environment variables with name capitalized and prefixed with `UWSGI_`, and dashes are substituted with underscores like described [here](https://uwsgi-docs.readthedocs.io/en/latest/Configuration.html#environment-variables)

Sets `UWSGI_LISTEN=1024` by default, but can be overwritten or completely removed by `uwsgi: null`

Or is it better to not specify default value (it's not backward compatible)?

Also, small indentation fixes for postgresql configuration.

closes https://github.com/grafana/oncall/issues/562

Also, [this PR](https://github.com/grafana/oncall/pull/856) has been closed because of this PR

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,11 +12,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add models and framework to use different services (Phone, SMS, Verify) in Twilio depending on
 the destination country code by @mderynck ([#1976](https://github.com/grafana/oncall/pull/1976))
 - Prometheus exporter backend for alert groups related metrics
+- Helm chart: configuration of `uwsgi` using environment variables by @alexintech ([#2045](https://github.com/grafana/oncall/pull/2045))
 
 ### Fixed
 
 - Fix error when updating closed modal window in Slack by @vadimkerr ([#2019](https://github.com/grafana/oncall/pull/2019))
 - Fix final schedule export failing to update when ical imported events set start/end as date ([#2025](https://github.com/grafana/oncall/pull/2025))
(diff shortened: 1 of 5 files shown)


joeyorlando (approved):
nice! thanks for your continued contribution to improving the Helm charts @alexintech

jrossBah:
@alexintech Big question for ya. I am currently working with getting grafana oncall implemented and we had to turn this setting down to 128 to get it to work in our kubernetes cluster. I am wondering why you default set it to 1024. As We understand it the container the chart spins up defaults to 128. We found this by changing it on the ec2 instance and still seeing the error until we lowered this new uwsgi setting.

Do you know is there a way in these values that changes the container default? Thank you!

alexintech:
> @alexintech I am wondering why you default set it to 1024. 
> 
> Do you know is there a way in these values that changes the container default? Thank you!

@jrossBah That's not my decision. 1024 has been hardcoded in helm chart. I only made this value configurable. To maintain the same behavior default value remained the same.

jrossBah:
> > @alexintech I am wondering why you default set it to 1024.
> > Do you know is there a way in these values that changes the container default? Thank you!
> 
> @jrossBah That's not my decision. 1024 has been hardcoded in helm chart. I only made this value configurable. To maintain the same behavior default value remained the same.

That's so interesting. What we found is that this limit seems to conflict with the limitations of the kubernetes and also docker containers that set at 128. So unless you bring this down to 128 or under the engine pods will fail.

Issue #445, opened 2022-08-31
Helm - SecurityContext - cannot setgid() as non-root user


vptimmy:
We are trying to install a test instance of onCall via the helm chart to our K8 cluster.   We do not allow our container to run as root.

We are getting the following error from grafana-oncall-engine when trying to deploy the chart:

```
2022-08-31 11:33:46 current working directory: /etc/app
2022-08-31 11:33:46 writing pidfile to /tmp/project-master.pid
2022-08-31 11:33:46 detected binary path: /usr/local/bin/uwsgi
2022-08-31 11:33:46 cannot setgid() as non-root user
```

Here is the helm/values.yaml I have been trying to deploy with.  Any suggestions are greatly appreciated!

```
podSecurityContext:
    fsGroup: 2000

init:
  securityContext:
    capabilities:
      drop:
        - ALL
    runAsNonRoot: true
    runAsUser: 1000

securityContext:
  capabilities:
    drop:
      - ALL
  runAsNonRoot: true
  runAsUser: 1000

ingress-nginx:
  enabled: false

grafana:
  enabled: false

cert-manager:
  enabled: false

rabbitmq:
  enabled: true

redis:
  enabled: true

mariadb:
  enabled: true
  architecture: standalone
  auth:
    database: oncall
```

juris:
The issue still persists with chart version 1.0.5 and app version 1.0.35. I've just uncommented the default helm values:
```
podSecurityContext:
  fsGroup: 2000

securityContext:
  capabilities:
    drop:
    - ALL
  readOnlyRootFilesystem: false
  runAsNonRoot: true
  runAsUser: 1000

init:
  securityContext:
    allowPrivilegeEscalation: false
    capabilities:
      drop:
      - ALL
    privileged: false
    readOnlyRootFilesystem: false
    runAsGroup: 1337
    runAsNonRoot: true
    runAsUser: 1337
```

CactusZ:
Same issue for us, would be good to have a fix for that!

anlight:
<img width="779" alt="image" src="https://user-images.githubusercontent.com/12930183/200177249-dbde2260-7c53-4362-ada2-b794875a4aa0.png">

isnt:
I made it work with this configuration (chart and app version: `1.2.21`):
```
# just commented out
podSecurityContext:
  fsGroup: 2000

# some modifications
securityContext:
  capabilities:
    drop:
    - ALL
    add:
    - SETGID
    - SETUID
[...]

PR #1890, opened 2023-05-08
Feat(Dev): Improve Building of Grafana Plugin in Development Env + update node version

roock:
# What this PR does

- Improvement to the local development environment for the grafana plugin
  - Run initial yarn build inside the docker container with the same version that is later used for periodic rebuilds
  - Removes the requirement for having yarn/nodejs installed locally
  - Using a named volume for storing the node_modules, so they are only stored once
  - Remove the yarn install step from the Dockerfile
  - Ideally we store the node_modules only once inside the named volumes. Currently they are stored times
    - on the host system outside of dockerin grafana-plugins/node_modules
    - inside the docker image
    - inside the anonymous docker volume created at the start of a container
- update `node` to 18.16.0 (14.17.0 has reached end-of-life as of 3 weeks ago)

## Checklist

- [X] ~Unit, integration, and e2e (if applicable) tests updated~ N/A
- [ ] Documentation added (or `pr:no public docs` PR label added if not required)
- [ ] `CHANGELOG.md` updated (or `pr:no changelog` PR label added if not required)

PR #2053, opened 2023-05-30
Run containers as a non-root user

alexintech:
# What this PR does

Create a custom non-root user and use it to start an app. So uwsgi does not require to use `setUid` and `setGid` system calls.

It handles errors while starting in Kubernetes with `runAsNonRoot: true` check. 

closes https://github.com/grafana/oncall/issues/445

## Checklist


engine/Dockerfile
```diff
@@ -1,4 +1,10 @@
 FROM python:3.11.3-slim-buster AS base
+
+# Create a group and user to run an app
+ENV APP_USER=appuser
+RUN groupadd -r -g 1000 ${APP_USER} && \
```
joeyorlando, line 5: These flags are equivalent according to the [manpage](https://man7.org/linux/man-pages/man8/groupadd.8.html#:~:text=The%20groupadd%20command%20creates%20a,digits%2C%20underscores%2C%20or%20dashes.) for `groupadd`
```suggestion
RUN groupadd --system --gid 1000 ${APP_USER} && \
```
I would prefer the more verbose flags for readability

engine/Dockerfile
```diff
@@ -1,4 +1,10 @@
...
+    useradd --no-log-init -r -u 1000 -g ${APP_USER} ${APP_USER}
```
joeyorlando, line 6: same suggestion here. also, why the `--no-log-init` flag?
```suggestion
    useradd --no-log-init --system --uid 1000 --gid ${APP_USER} ${APP_USER}
```

engine/Dockerfile
joeyorlando, line 1: consider deduplicating `1000` to an `ENV` variable at the top of the `Dockerfile` so that it can be reused on lines 5, 6, and 57

docker-compose-developer.yml
```diff
@@ -3,6 +3,9 @@ version: "3.9"
 x-labels: &oncall-labels
   - "com.grafana.oncall.env=dev"
 
+x-user: &oncall-user
+  "appuser"
```
joeyorlando, line 7: is this needed? according to the `docker-compose` configuration [docs](https://docs.docker.com/compose/compose-file/05-services/#user):

> `user` overrides the user used to run the container process. Default is that set by image (i.e. Dockerfile `USER`), if not set, `root`. 

You are setting `USER` in `engine/Dockerfile` [here](https://github.com/grafana/oncall/pull/2053/files#diff-a25c2322baf692c37d4ecc9c3611772e78dfb8bbaf9cc4ef27b507e60cb3ddc9R57), so I think this may be redundant

joeyorlando (commented):
Overall looks great to me! Thanks for picking this up. Will approve once you've addressed the questions/concerned listed 😄 

@iskhakov do you mind giving this a second set of 👀 ?

docker-compose-developer.yml
```diff
@@ -52,7 +55,7 @@ services:
       MOBILE_APP_QR_INTERVAL_QUEUE: 290000 # 4 minutes and 50 seconds
     volumes:
       - ./grafana-plugin:/etc/app
-      - node_modules_dev:/etc/app/node_modules
+      - ./grafana-plugin/node_modules:/etc/app/node_modules
```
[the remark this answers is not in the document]
alexintech: When I run `make init`, `node_modules` folder on my host machine is empty and PyCharm cannot correctly understand the project. Or maybe I don't know how to tell PyCharm where to find node_modules?

I can revert these changes if they break developer experience 😄

engine/Dockerfile
alexintech: I considered deduplicate `1000`, but how `ENV` variable should be named? `APP_USER_UID_GID`? Logically there should be two variables `APP_USER_UID` and `APP_USER_GID`.

I didn't want to introduce extra `ENV` variables for the value, that used only 3 times and should not be reused again in the `Dockerfile`

engine/Dockerfile
alexintech: I don't remember where I found the recommendation to use `--no-log-init`. With this flag `useradd` does not write to `/var/log/faillog` and `/var/log/lastlog` and saves around `300Kb` 😄 It's not a lot but I think it's not a bad flag for containerized user.

iskhakov (changes requested):
overall looks good to me

I would like to make a final look before merging, so I leave my review in "Request changes" state

docker-compose-developer.yml
joeyorlando: ahh yes, that makes sense. thanks for the detailed explanation here 👍

docker-compose-developer.yml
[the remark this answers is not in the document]
joeyorlando: > `node_modules` folder on my host machine is empty

this is correct. [`make init`](https://github.com/grafana/oncall/blob/dev/Makefile#L91) spins up an ephemeral `oncall_ui` container, defined in `docker-compose-dev.yml`, and runs `yarn install && yarn build:dev`.

This installs `node_modules` to the `node_modules_dev` volume mount, not `./grafana-plugin/node_modules` on your host machine & builds the plugin dist to `./grafana-plugin/dist` (on your host-machine). There's more context in the PR description of #1890 on this change. tldr; now you no longer need to have `node` installed on your machine to worry about installing these deps.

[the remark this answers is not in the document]
joeyorlando: re: PyCharm, you'd have to point it to the volume mounted `node_modules`, which I suspect is not possible? what if you ran `yarn install` on your host machine (on top of `make init`), does this fix it?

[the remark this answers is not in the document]
alexintech: > what if you ran yarn install on your host machine (on top of make init), does this fix it?

Yes, that fixes. But in this case there are `node_modules` in `grafana_plugin/node_modules` and in a docker volume. And you're also need installed `node` to run `yarn install` 😄 

> Ideally we store the node_modules only once inside the named volumes.

I only do not understand, why `node_modules` inside the named volume is better than in `grafana_plugin/node_module` on host machine?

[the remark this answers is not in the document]
joeyorlando: > why node_modules inside the named volume is better than in grafana_plugin/node_module on host machine

in some sense it's opinionated, but by storing them on the volume, you don't require the developer to have `node` installed on their host machine (because the container takes care of this for you). The frontend will only run w/ specific versions of `node`, so this just makes it one less thing the dev needs to worry about 😄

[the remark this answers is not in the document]
alexintech: > you don't require the developer to have node installed on their host machine (because the container takes care of this for you).

Exactly the same in this case. You run `make init` and it downloads `node_modules` and exposes them to host machine. So PyCharm can find imports and you can dive into sources of installed dependencies if you need. And I suppose other IDEs also may benefit.

Actually, there is no need to mount  `./grafana-plugin/node_modules:/etc/app/node_modules`, as `./grafana-plugin:/etc/app` already can expose node_modules.

What code editor do you use?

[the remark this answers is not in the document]
joeyorlando: lets move this discussion to a separate GitHub issue so that we don't block the other work you've done here 😄 I've created #2097.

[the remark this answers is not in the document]
alexintech: Sure!

engine/Dockerfile
```diff
@@ -1,4 +1,10 @@
 FROM python:3.11.3-slim-buster AS base
+
+# Create a group and user to run an app
+ENV APP_USER=appuser
+RUN groupadd --system --gid 1000 ${APP_USER} && \
```
iskhakov, line 5: can we keep gid=2000 to make sure it is backwards-compatible?

alexintech: Yes, we can change to `gid=2000`. 

I think it's a common practice to set the same `uid` and `gid` for a custom user. I suppose that nothing would break, as process that starts application have the same user permissions. Something can be broken, if another process with some other uid and in the same group wants to use files with group permissions. But I suppose it's not applicable to dockerized application.

I'll change to `gid=2000` if you prefer that.

alexintech: Hi @iskhakov @joeyorlando ! Change to `gid=2000` or not ?

joeyorlando: I agree w/ @iskhakov , lets keep it to 2000 just to be safe

iskhakov (approved):

engine/Dockerfile
```diff
@@ -41,9 +55,7 @@ RUN pip install -r requirements-enterprise-docker.txt
 
 FROM base AS prod
 
-# This is required for prometheus_client to sync between uwsgi workers
-RUN mkdir -p /tmp/prometheus_django_metrics;
-RUN chown -R 1000:2000 /tmp/prometheus_django_metrics
-ENV prometheus_multiproc_dir "/tmp/prometheus_django_metrics"
+# Change to a non-root user (number is required by Kubernetes runAsNonRoot check)
+USER 1000
```
joeyorlando, line 59: @alexintech actually one thought came to mind. If we move `USER 1000` higher up to line 13, I _think_ we may not need any of the `chown` commands?

alexintech: > If we move USER 1000 higher up to line 13

That will break `RUN apt-get install -y sqlite3 default-mysql-client postgresql-client` [here](https://github.com/grafana/oncall/pull/2053#diff-a25c2322baf692c37d4ecc9c3611772e78dfb8bbaf9cc4ef27b507e60cb3ddc9R49) in `dev` target.

joeyorlando: that makes sense 👍 

I opened [a PR](https://github.com/grafana/oncall/pull/2132) to move `USER` to the end of the `dev` target so that it is present if a different build `target` is specified to `docker build`

Issue #1889, opened 2023-05-08
Bug: Alertmanager Templates not working when Slack or Telegram is disabled


roock:
Oncall OnPremise Dev branch

Steps to Reproduce:
- In `dev/.env.dev` set either `FEATURE_TELEGRAM_INTEGRATION_ENABLED=False` or `FEATURE_SLACK_INTEGRATION_ENABLED=False`
- Start Development Environment from dev branch according to the docs
- Open Integration Page and Add an Alertmanager Integration
- Open the Alert Templates Section
- The page doesn't load
- The logs contain the following lines

```
2023-05-08 12:47:32 source=engine:app google_trace_id=none logger=django.request Internal Server Error: /api/internal/v1/alert_receive_channel_templates/CYJ2WTTKI6S5I
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
               ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
    return view_func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/rest_framework/viewsets.py", line 125, in view
    return self.dispatch(request, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/rest_framework/views.py", line 509, in dispatch
    response = self.handle_exception(exc)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/rest_framework/views.py", line 469, in handle_exception
    self.raise_uncaught_exception(exc)
  File "/usr/local/lib/python3.11/site-packages/rest_framework/views.py", line 480, in raise_uncaught_exception
    raise exc
  File "/usr/local/lib/python3.11/site-packages/rest_framework/views.py", line 506, in dispatch
[...]

PR #2061, opened 2023-05-30
validation before  deleting item in list

kranthikirang:
# Before deleting the item in response_data we make sure that item is present first for alert manager templates

## https://github.com/grafana/oncall/issues/1889


Konstantinov-Innokentii:
Hello, @kranthikirang. Thanks for your contribution. I took a look in this part of code in debugger. Problem here is that `_get_templates_to_show` method is called on line 346, while any template values added to response on line 348, so it makes no sense to check if there are values in list, before deleting them - list will be always empty.

Konstantinov-Innokentii:
I continued this work in https://github.com/grafana/oncall/pull/2064. Thanks @kranthikirang. Code responsible for templates is quite confusing

iskhakov:
@kranthikirang thank you for the contribution, the issue was fixed by #2064

PR #2064, opened 2023-05-31
Fix 500 on templates when slack or tg integration is disabled

Konstantinov-Innokentii:
# What this PR does
Continue the work, started in https://github.com/grafana/oncall/pull/2061. 
Check if slack or telegram integration is enabled to include related templates in the response
https://github.com/grafana/oncall/issues/1889

## Checklist

### DIFF
--- a/engine/apps/api/serializers/alert_receive_channel.py
+++ b/engine/apps/api/serializers/alert_receive_channel.py
@@ -325,9 +325,7 @@ class AlertReceiveChannelTemplatesSerializer(EagerLoadingMixin, serializers.Mode
         """Update core templates if needed."""
         errors = {}
 
-        core_template_names = self.CORE_TEMPLATE_NAMES
-
-        for field_name in core_template_names:
+        for field_name in self.core_templates_names:
             value = data.get(field_name)
             validator = jinja_template_env.from_string
             if value is not None:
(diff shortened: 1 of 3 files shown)


engine/apps/api/serializers/alert_receive_channel.py
```diff
@@ -343,7 +341,7 @@ def _handle_core_template_updates(self, data, ret):
 
     def to_representation(self, obj):
         ret = super().to_representation(obj)
-        ret = self._get_templates_to_show(ret)
+        # ret = self._get_templates_to_show(ret)
```
iskhakov, line 344: left comment

if we don't need it _get_templates_to_show function can be removed

engine/apps/api/serializers/alert_receive_channel.py
```diff
@@ -343,7 +324,6 @@ def _handle_core_template_updates(self, data, ret):
 
     def to_representation(self, obj):
         ret = super().to_representation(obj)
-        ret = self._get_templates_to_show(ret)
```
iskhakov, line 346: do we need this function anymore?

https://github.com/grafana/oncall/pull/2064/files#diff-b8990ca4b389382d4e7b6f58fd7c1bf0773c85d3a86db2a4a4486e2459dcf8ffR337

iskhakov (approved):

PR #2078, opened 2023-06-01
remove LICENSING.md

joeyorlando:
this file is mostly empty, plus we have a `LICENSE` file. Do we still need `LICENSING.md`?


joeyorlando:
closing as it looks like other `grafana` org repos define a similar file (even if this one is empty for our purposes)

PR #2080, opened 2023-06-01
Trigger slack shift notifications on current shift change

matiasb:
Before this change, a diff ical check (which happens with frequency with imported ical), particularly with overrides in an API/terraform schedule would trigger unexpected slack notifications because the prev vs current ical comparison will flag a diff, but when comparing current and previous shifts, `current_shifts` will have the shift in progress while the `prev_shifts` calculated from the overrides-only diff will most of the time be empty (unless you set/change an override at current time).

Simplified the checks to always compare previous current shifts (ie. the ones in the schedule from the DB) vs the recalculated ones using the (refreshed) ical data from the schedule.


vadimkerr (approved):
Nice 👍 
Do I understand correctly this will fix repeated messages completely?

matiasb:
> Nice +1 Do I understand correctly this should fix repeated messages completely?

I hope so! :crossed_fingers:

PR #2116, opened 2023-06-07
Rares/templates tweaks 5

teodosii:
# What this PR does

- Removed duplicate copy within Routes
- Made drawer content of templates more responsive
- Fetch telegram info and display it accordingly
- Fixed pagination being reset when going back to table view
- Some other minor things

### DIFF
--- a/grafana-plugin/src/containers/IntegrationContainers/CollapsedIntegrationRouteDisplay/CollapsedIntegrationRouteDisplay.tsx
+++ b/grafana-plugin/src/containers/IntegrationContainers/CollapsedIntegrationRouteDisplay/CollapsedIntegrationRouteDisplay.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 
 import { ConfirmModal, HorizontalGroup, Icon, VerticalGroup } from '@grafana/ui';
 import cn from 'classnames/bind';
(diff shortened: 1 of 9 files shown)


grafana-plugin/src/containers/IntegrationContainers/CollapsedIntegrationRouteDisplay/CollapsedIntegrationRouteDisplay.tsx
```diff
@@ -84,15 +92,17 @@ const CollapsedIntegrationRouteDisplay: React.FC<CollapsedIntegrationRouteDispla
...
-                {IntegrationHelper.getChatOpsChannels(channelFilter).map((chatOpsChannel, key) => (
-                  <HorizontalGroup key={key}>
-                    <Text type="secondary">Publish to ChatOps</Text>
-                    <Icon name={chatOpsChannel.icon} />
-                    <Text type="primary" strong>
-                      {chatOpsChannel.name}
-                    </Text>
-                  </HorizontalGroup>
-                ))}
+                {IntegrationHelper.getChatOpsChannels(channelFilter, telegramInfo)
```
Ukochka, line 95: I think it was a check here (some times ago) to check if chatops is enabled in general, but it disappeared in some of the PR. Should we check it here? Or are we check it in some other place?
I mean if Slack feature is not enabled, should it be displayed?

Ukochka (approved):

Issue #2118, opened 2023-06-07
Pre-release schedule v2 bug collection


Matvey-Kuk:
✅
1) Schedule list, "Oncall" -> "On-call now"
<img width="188" alt="Screenshot 2023-06-07 at 12 19 44" src="https://github.com/grafana/oncall/assets/3284841/5b093cf9-e08d-4184-91ee-502035f39983">

✅
2) New schedule, "Additional notification settings" -> "Notification settings"
<img width="311" alt="Screenshot 2023-06-07 at 12 20 48" src="https://github.com/grafana/oncall/assets/3284841/30dd9f12-f332-4aaa-84c5-b91be75eda06">

✅
3) Top nav stick to the top of the schedule screen:
<img width="1215" alt="Screenshot 2023-06-07 at 12 25 33" src="https://github.com/grafana/oncall/assets/3284841/24a2c1ff-e8a9-47d9-b1a1-41cb83b9f7c7">

✅
4) Didn't we update texts here? I think we agreed on "Delete past shifts".
<img width="516" alt="Screenshot 2023-06-07 at 12 26 53" src="https://github.com/grafana/oncall/assets/3284841/3e660d83-f9a4-403b-82ad-38e5d69d13ba">

✅
5)  "Add rotation to the schedule to see users" -> "Add rotation to see users"
<img width="396" alt="Screenshot 2023-06-07 at 12 59 17" src="https://github.com/grafana/oncall/assets/3284841/2f42451e-9e63-41e7-8069-7b504333b2c6">

raphael-batte:
There are white visual artefacts in the modal
![image](https://github.com/grafana/oncall/assets/102970268/c6d68e62-0416-421b-bd69-c139adab4f4d)

raphael-batte:
Show warning when user tries to save rotation without users. Modal shouldn't closes.

raphael-batte:
Sometimes dropdown with recurrence option is empty 
![image](https://github.com/grafana/oncall/assets/102970268/51d90ad7-4884-45b4-ac62-3e4a809ea3b3)

raphael-batte:
Week start settings and days arrangement in the modal should be the same
![image](https://github.com/grafana/oncall/assets/102970268/3fcb948c-95fb-4511-a4f1-27199683a6a1)

Ferril:
Some weird prerendered shifts behavior when opening update shift window
[...]

Issue #1322, opened 2023-02-14
Update Rotations popup doesn't seem to reflect actual rotation values


teodosii:
Steps to reproduce
- Create a new schedule
- Add a new rotation with duration of 1 month for user 1 with recurring days of Monday, Wednesday and Friday
- Add a new rotation with duration of 1 month for user 2 with recurring days of Tuesday, Thursday and Saturday

All good until now, it should look like this 
![image](https://user-images.githubusercontent.com/40542072/218777223-f98583f7-75fb-476a-ab01-c291de6465c6.png)

Try opening the rotation for user 1 and you will notice it says the repeated days are Tuesday, Thursday and Sunday (!)

![image](https://user-images.githubusercontent.com/40542072/218777414-16e55799-abb6-48a8-98ff-2d24fcd63182.png)

Do the same for user 2 and you will notice it says Monday, Wednesday and Friday

![image](https://user-images.githubusercontent.com/40542072/218777540-e08e88f5-d978-422e-b845-3688e7c4bf9f.png)

Whenever I try opening a rotation it seems the wrong days are displayed.

Matvey-Kuk:
Good catch @teodosii ! Like your attention to details.

teodosii:
Just checked on this one to see if it's on the frontend side, but seems to be an issue in the api response

Here's what I see when I should be seeing Monday, Wednesday, Friday

![image](https://user-images.githubusercontent.com/40542072/220614582-16e558e2-1cc9-488c-817a-aab575cfe426.png)

Normal schedule view illustrates the correct days

![image](https://user-images.githubusercontent.com/40542072/220614729-3c5f17b2-5363-46c3-b90e-520422aa941a.png)

PR #2124, opened 2023-06-07
Update shift week_start when translating to UTC

matiasb:
This fixes scenario described [here](https://github.com/grafana/oncall/issues/2118#issuecomment-1580499754).

When a rotation is setup in UTC+1, and the shift starts at 00:00 with Monday as active day and a weekly frequency, the values are translated to UTC when submitting to the backend, so the shift data becomes something like: shift starting at 23:00 on Sunday, but since week_start is on Monday, the "first event" in the week belongs to the "previous week". This can be addressed by moving the week_start, so a weekly shift that was starting on a Monday but in UTC tz starts on Sunday, "translates" to a UTC week_start on Sunday:

rotation-example

(this is with the proposed changes; otherwise you get the same issue linked above where the first event in the week is assigned to the other user group).

About selected week days changed when editing a rotation, see inline comment (related to [this](https://github.com/grafana/oncall/issues/1322#issuecomment-1521787786))

### DIFF
--- a/grafana-plugin/src/containers/RotationForm/RotationForm.tsx
+++ b/grafana-plugin/src/containers/RotationForm/RotationForm.tsx
@@ -25,7 +25,13 @@ import { Schedule, Shift } from 'models/schedule/schedule.types';
 import { getTzOffsetString } from 'models/timezone/timezone.helpers';
 import { Timezone } from 'models/timezone/timezone.types';
 import { User } from 'models/user/user.types';
-import { getDateTime, getStartOfWeek, getUTCByDay, getUTCString } from 'pages/schedule/Schedule.helpers';
+import {
+  getDateTime,
+  getStartOfWeek,
+  getUTCByDay,
+  getUTCString,
+  getUTCWeekStart,
+} from 'pages/schedule/Schedule.helpers';
 import { SelectOption } from 'state/types';
 import { useStore } from 'state/useStore';
 import { getCoords, waitForElement } from 'utils/DOM';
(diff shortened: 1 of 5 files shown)


grafana-plugin/src/containers/RotationForm/RotationForm.tsx
matiasb, line 1: In the same way that we render the shift start / end datetimes (returned as UTC by the backend) in the selected TZ, I think we should map [`by_day` selected days](https://github.com/grafana/oncall/pull/2124/files#diff-bbaac1b12ad1ba869bb1fa17686974f6aa66bfdd64d60ff21665ff887def7255R233) accordingly, as mentioned [here](https://github.com/grafana/oncall/issues/1322#issuecomment-1521787786).

grafana-plugin/src/pages/schedule/Schedule.helpers.ts
```diff
@@ -52,6 +52,26 @@ export const getUTCByDay = (dayOptions: SelectOption[], by_day: string[], moment
...
+      // move one day after
+      week_start_index = (week_start_index + 1) % 7;
+    } else {
+      // move one day before
+      week_start_index = (((week_start_index - 1) % 7) + 7) % 7;
+    }
+  }
+  return byDayOptions[week_start_index];
+};
+
```
joeyorlando, line 74: _nit_ consider adding a unit test here

joeyorlando (approved):
LGTM 👍

PR #2129, opened 2023-06-08
Fixes for search, for last alert payload and some design changes

Ukochka:
Fixes for search, for last alert payload and some design changes


grafana-plugin/src/containers/TemplatesAlertGroupsList/TemplatesAlertGroupsList.tsx
```diff
@@ -43,6 +44,12 @@ const TemplatesAlertGroupsList = (props: TemplatesAlertGroupsListProps) => {
     return `${height}px`;
   };
 
+  const getCodeEditorHeightWithBadge = () => {
+    const mainDiv = document.getElementById('content-container-id');
+    const height = mainDiv?.getBoundingClientRect().height - 59 - 42;
```
teodosii, line 49: Could you please add some constants here so that the next person reading this knows what is 59 and what is 42? These magic numbers could mean anything, at least if you know which DOM element would that height represent is a good start.

teodosii (approved):

PR #2131, opened 2023-06-08
add user locale field to mobile app user settings table

joeyorlando:
# What this PR does

add user locale field to mobile app user settings table

## Checklist


engine/apps/mobile_app/tests/test_your_going_oncall_notification.py
```diff
@@ -47,6 +48,171 @@ def test_shift_starts_within_range(timing_window_lower, timing_window_upper, sec
...
+    multiple_day_shift_title = tasks._get_youre_going_oncall_notification_title(
+        schedule, seconds_until_going_oncall, multiple_day_shift, maus
+    )
+    multiple_day_shift_no_locale_title = tasks._get_youre_going_oncall_notification_title(
+        schedule, seconds_until_going_oncall, multiple_day_shift, maus_no_locale
+    )
+
+    assert (
+        multiple_day_shift_title
+        == f"You're going on call in {humanized_time_until_going_oncall} for schedule {schedule_name}, 2023-07-08 09 h 00 - 2023-07-12 17 h 00"
```
vadimkerr, line 126: `09 h 00` looks a bit weird IMO, maybe it should be `09:00` or at least `09h00`?
I'm not an expert in locales, but I can't see anything similar to this format in the [reference for fr-CA](https://leap.hcldoc.com/help/topic/SSS28S_8.2.1/XFDL_Specification/i_xfdl_r_formats_fr_CA.html).

engine/apps/mobile_app/tests/test_your_going_oncall_notification.py
```diff
@@ -47,6 +48,171 @@ def test_shift_starts_within_range(timing_window_lower, timing_window_upper, sec
...
+    )
+    assert (
+        multiple_day_shift_no_locale_title
+        == f"You're going on call in {humanized_time_until_going_oncall} for schedule {schedule_name}, 7/8/23, 9:00\u202fAM - 7/12/23, 5:00\u202fPM"
```
vadimkerr, line 130: why use code `\u20` instead of regular " " spaces?

joeyorlando:
@vadimkerr these two points you raised seem to be coming from the `babel` library itself. I've raised two questions over in that repo to clarify this behaviour.

**`\u202f` character**
https://github.com/python-babel/babel/issues/1012
We _could_ do a string replacement on this character, but that seems flaky imo.

**Odd spacing in time formatting w/ `fr_CA` locale**
https://github.com/python-babel/babel/issues/1013

It _appears_ like this is the default formatting behaviour for `fr_CA`?
```python3
>>> babel.dates.format_time(now, format="short", locale="fr_CA")
'15 h 27'
>>> babel.dates.format_time(now, format="short", locale="fr")
'15:27'
>>> babel.dates.format_time(now, format="short", locale="en")
'3:27\u202fPM'
```

I don't think overriding the datetime formatting [pattern syntax](https://babel.pocoo.org/en/latest/dates.html#pattern-syntax) would be useful here, because that would default the purpose of localizing the datetimes in the first place 😕

joeyorlando:
@vadimkerr these formatting decisions seem to be the default coming from [Unicode CLDR](https://cldr.unicode.org/). I suggest for now we stick with these and refine them over time if need be. Overriding them to handle different formats/locales, to suit our liking, seems like a lot of work.

Regarding the `\u202f` unicode characters that `babel` returns. I verified that Android renders these properly, I presume iOS does as well. For example, when `6/13/23, 10:00\u202fPM - 6/14/23, 10:00\u202fPM` is sent to FCM as the notification title:
Screenshot 2023-06-14 at 18 00 22

PR #2132, opened 2023-06-08
engine Dockerfile - move USER to base layer

joeyorlando:


engine/Dockerfile
```diff
@@ -47,6 +47,9 @@ RUN mkdir -p /tmp/prometheus_django_metrics;
 RUN chown -R ${APP_USER}:${APP_USER} /tmp/prometheus_django_metrics
 ENV prometheus_multiproc_dir "/tmp/prometheus_django_metrics"
 
+# Change to a non-root user (number is required by Kubernetes runAsNonRoot check)
+USER 1000
+
```
joeyorlando, line 52: should be present in `base` layer so that there are no permission issues regardless of which `target` is specified w/ `docker build`

alexintech: That will break `dev` target.

You can copy `USER 1000` to the end of `dev`, if it does not break `dev-enterprise` target. And do not remove at `prod` target

PR #2151, opened 2023-06-11
add mypy static type checker to backend codebase

joeyorlando:
# What this PR does

- Adds [`mypy` static type checking](https://mypy-lang.org/) to our CI pipeline. Currently there is still a **ton** of errors being returned by the tool, as we'll need to fix pre-existing errors. I think we can slowly chip away at these errors in small PRs, doing them all in one large PR is likely very risky.
  - Also, this PR starts chipping away at one of the main type errors that we have which is accessing the `datetime` class (from the `datetime` library) or `timedelta` function on the `django.utils.timezone` module. Basically we should be instead accessing these two objects from the native `datetime` module. This makes sense because the [`__all__` attribute](https://github.com/django/django/blob/main/django/utils/timezone.py#L14-L30) in `django.utils.timezone` does not re-export `datetime` or `timedelta`.
- splits `engine` dependencies out into `requirements.txt` and `requirements-dev.txt`

## Checklist


.pre-commit-config.yaml
```diff
@@ -8,28 +8,29 @@ repos:
...
         name: black - pd-migrator
         files: ^tools/pagerduty-migrator
+        args: [--config=engine/pyproject.toml]
       - id: black
         name: black - dev/scripts
         files: ^dev/scripts
+        args: [--config=engine/pyproject.toml]
 
   - repo: https://github.com/pycqa/flake8
-    rev: 3.9.2
```
joeyorlando, line 32: bump `flake8` so that we can use `py311` as our target version, which is the actual Python version we use (configured in `engine/pyproject.toml`

.pre-commit-config.yaml
```diff
@@ -39,25 +40,15 @@ repos:
...
         name: flake8 - dev/scripts
         files: ^dev/scripts
-        # Make sure config is compatible with black
-        # https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length
-        args:
-          [
-            --max-line-length=88,
-            "--select=C,E,F,W,B,B950",
-            "--extend-ignore=E203,E501",
-          ]
```
joeyorlando, line 60: use consistent formatting across all Python code. Simply use the `engine/tox.ini` for this configuration

engine/apps/alerts/escalation_snapshot/utils.py
```diff
@@ -1,29 +1,37 @@
+import datetime
+import typing
+
 import pytz
 from django.utils import timezone
 
 
-def eta_for_escalation_step_notify_if_time(from_time, to_time, current_time=None) -> timezone.datetime:
```
joeyorlando, line 5: this function is called once, and `current_time` is never passed in, getting rid of it

engine/apps/alerts/escalation_snapshot/utils.py
```diff
@@ -1,29 +1,33 @@
...
+
+    now = timezone.now()
+    current_date = now.date()
+    current_time = now.time()
+
     if from_time < to_time:
-        if from_time > current_time.time():
-            eta = timezone.datetime.combine(current_time.date(), from_time).astimezone(pytz.UTC)
-        elif current_time.time() >= to_time:
-            eta = timezone.datetime.combine((current_time + timezone.timedelta(days=1)).date(), from_time).astimezone(
```
joeyorlando, line 16: `current_time` was always previously equal to `timezone.now()` since the `current_time` arg passed into the function was always `None`, hence changing `current_time` to `now` in this line

engine/apps/alerts/escalation_snapshot/utils.py
```diff
@@ -1,29 +1,33 @@
...
-        if from_time > current_time.time():
-            eta = timezone.datetime.combine(current_time.date(), from_time).astimezone(pytz.UTC)
-        elif current_time.time() > to_time:
-            eta = timezone.datetime.combine((current_time + timezone.timedelta(days=1)).date(), from_time).astimezone(
-                pytz.UTC
-            )
+        if from_time > current_time:
+            eta = datetime.datetime.combine(current_date, from_time).astimezone(pytz.UTC)
+        elif current_time > to_time:
+            eta = datetime.datetime.combine((now + datetime.timedelta(days=1)).date(), from_time).astimezone(pytz.UTC)
```
joeyorlando, line 32: same thing here

engine/apps/public_api/serializers/maintenance.py
```diff
@@ -24,13 +24,14 @@ class Meta:
...
-        if obj.till_maintenance_timestamp is not None:
-            return serializers.DateTimeField().to_representation(
-                datetime.datetime.fromtimestamp(obj.till_maintenance_timestamp)
-            )
+    def get_maintenance_end_at(self, obj: MaintainableObject) -> typing.Optional[str]:
+        if obj.till_maintenance_timestamp is None:
+            return None
+        return serializers.DateTimeField().to_representation(
+            datetime.datetime.fromtimestamp(obj.till_maintenance_timestamp)
+        )
```
joeyorlando, line 37: logic is still the same here, just explicitly returning `None` now

engine/apps/alerts/grafana_alerting_sync_manager/grafana_alerting_sync.py
```diff
@@ -348,11 +349,7 @@ def _get_receiver_config_for_datasource(self, is_grafana_datasource, webhook_url
             }
         return receiver
 
-    def _create_contact_point_from_payload(
-        self,
-        payload,
-        datasource,
-    ) -> "apps.alerts.models.GrafanaAlertingContactPoint":
```
joeyorlando, line 355: `apps.alerts.models.GrafanaAlertingContactPoint` wasn't resolved properly for some reason. Explicitly importing `GrafanaAlertingContactPoint` above with:
```python
if TYPE_CHECKING:
    from apps.alerts.models import GrafanaAlertingContactPoint
```

engine/apps/alerts/escalation_snapshot/snapshot_classes/escalation_snapshot.py
```diff
@@ -43,7 +44,7 @@ def __init__(
         escalation_policies_snapshots: typing.List["EscalationPolicySnapshot"],
         slack_channel_id: str,
         pause_escalation: bool,
-        next_step_eta: typing.Optional[str],
+        next_step_eta: typing.Optional[datetime.datetime],
```
joeyorlando, line 47: the typing here was wrong. This can't possibly be a `str` since we are doing the following down on line 106:
```python
return self.next_step_eta > (timezone.now() - datetime.timedelta(minutes=5))
```

joeyorlando:
see [here](https://github.com/grafana/oncall/actions/runs/5241314930/jobs/9463679562#step:4:384) for a portion of the errors still existing (I _think_ `mypy` cuts off the reported errors after about 160).

engine/pyproject.toml
```diff
@@ -1,10 +1,62 @@
...
+    "pyroscope.*",
+    "ratelimit.*",
+    "recurring_ical_events.*",
+    "rest_polymorphic.*",
+    "slackclient.*",
+    "social_core.*",
+    "social_django.*",
+    "twilio.*",
+    "uwsgidecorators.*",
+    "whitenoise.*",
```
joeyorlando, line 57: for now lets tell `mypy` not to worry about static type checking for things coming from these modules.

The versions of these modules currently being used don't have type stubs. Either we can try upgrading them to a newer version, which maybe include type stubs, or write the relevant stubs ourselves. these should be separate PRs though.

Matvey-Kuk (approved):

PR #2167, opened 2023-06-13
Draft AlertManager integration v2

Konstantinov-Innokentii:
# What this PR does
Introduces AlertManagerV2 integration with better grouping and autoresolving, not intended for production use yet.


engine/apps/integrations/urls.py
```diff
@@ -33,6 +34,7 @@
     path("alertmanager/<str:alert_channel_key>/", AlertManagerAPIView.as_view(), name="alertmanager"),
     path("amazon_sns/<str:alert_channel_key>/", AmazonSNS.as_view(), name="amazon_sns"),
     path("heartbeat/<str:alert_channel_key>/", HeartBeatAPIView.as_view(), name="heartbeat"),
+    path("alertmanager/v2/<str:alert_channel_key>/", AlertManagerV2View.as_view(), name="alertmanager_v2"),
```
iskhakov, line 37: may be call it alertmanager_testing - so nobody starts using it

iskhakov, line 37:
```suggestion
    path("alertmanager_testing/<str:alert_channel_key>/", AlertManagerV2View.as_view(), name="alertmanager_v2"),
```

may be call it alertmanager_testing - so nobody starts using it

engine/apps/integrations/views.py
```diff
@@ -391,3 +391,39 @@ def _process_heartbeat_signal(self, request, alert_receive_channel):
...
+    """
+
+    def post(self, request, *args, **kwargs):
+        alert_receive_channel = self.request.alert_receive_channel
+        if not alert_receive_channel.integration == AlertReceiveChannel.INTEGRATION_ALERTMANAGER_V2:
+            return HttpResponseBadRequest(
+                f"This url is for integration with {alert_receive_channel.config.title}."
+                f"Key is for {alert_receive_channel.get_integration_display()}"
+            )
+        alerts = request.data.get("alerts", [])
```
iskhakov, line 408: consider limiting the maximum number of alerts

engine/apps/integrations/views.py
```diff
@@ -391,3 +391,39 @@ def _process_heartbeat_signal(self, request, alert_receive_channel):
...
+
+        data = request.data
+        if "numFiring" not in request.data:
+            # Count firing and resolved alerts manually if not present in payload
+            num_firing = len(list(filter(lambda a: a["status"] == "firing", alerts)))
+            num_resolved = len(list(filter(lambda a: a["status"] == "resolved", alerts)))
+            data = {**request.data, "numFiring": num_firing, "numResolved": num_resolved}
+
+        create_alert.apply_async(
```
iskhakov, line 417: technically we save alertmanager alert group as oncall alert. Need to understand how to represent that correctly

iskhakov (approved):
overall looks good to me

engine/apps/integrations/views.py
Konstantinov-Innokentii: Actually it is limited when receiving a request, by DATA_UPLOAD_MAX_MEMORY_SIZE, which is set to 1mb. Do we want to add more trimming here?

engine/apps/integrations/views.py
Konstantinov-Innokentii: I think it looks fine in OnCall terminology

PR #2190, opened 2023-06-13
Save only organizations with integrations in cache 

Ferril:
Save only organizations with integrations in cache  for collecting metrics

### DIFF
--- a/engine/apps/metrics_exporter/helpers.py
+++ b/engine/apps/metrics_exporter/helpers.py
@@ -21,17 +21,23 @@ from apps.metrics_exporter.constants import (
 )
 
 
-def get_organization_ids():
-    """Try getting organizations ids from cache, otherwise get from db and save values in cache"""
+def get_organization_ids_from_db():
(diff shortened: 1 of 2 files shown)


engine/apps/metrics_exporter/helpers.py
```diff
@@ -23,10 +23,14 @@
 
 def get_organization_ids():
     """Try getting organizations ids from cache, otherwise get from db and save values in cache"""
-    Organization = apps.get_model("user_management", "Organization")
+    AlertReceiveChannel = apps.get_model("alerts", "AlertReceiveChannel")
     organizations_ids = cache.get(METRICS_ORGANIZATIONS_IDS, [])
     if not organizations_ids:
-        organizations_ids = Organization.objects.all().values_list("id", flat=True)
+        organizations_ids = (
+            AlertReceiveChannel.objects.filter(organization__deleted_at__isnull=True)
```
matiasb, line 30: Since we are using the same query in 2 places, maybe we can move this to a separate helper?

PR #2220, opened 2023-06-13
Require users when creating a schedule rotation

matiasb:


engine/apps/api/serializers/on_call_shifts.py
```diff
@@ -189,11 +189,15 @@ def _correct_validated_data(self, event_type, validated_data):
...
         return validated_data
 
+    def _require_users(self, validated_data):
+        users = validated_data.get("rolling_users")
+        if not users:
+            raise serializers.ValidationError({"rolling_users": ["User(s) are required"]})
+
     def create(self, validated_data):
         validated_data = self._correct_validated_data(validated_data["type"], validated_data)
-        validated_data["name"] = CustomOnCallShift.generate_name(
```
Ferril, line 194: Isn't this field required to create a shift?

matiasb: yeah, fixing; bad merge from the original PR, good catch! :+1:

Ferril (approved):

PR #2239, opened 2023-06-14
Unify "name" and "title" field for oncall shift

matiasb:
Porting changes from https://github.com/grafana/oncall/pull/1833


matiasb:
> I believe I merged this changes to Maxim's branch [#1833 (comment)](https://github.com/grafana/oncall/pull/1833#issuecomment-1558929508)

Right, but we are merging backend changes which are not dependent on the frontend changes to dev so we can release them, while testing the plugin only (which we can deploy independently per stack; and hopefully we can limit the big PR to be frontend changes only)

mderynck (approved):

PR #2242, opened 2023-06-14
Change permissions used for finishing plugin setup

mderynck:
Fixes issue where user not having `plugins:install`permission were unable to complete setup of OnCall.

- Check multiple Grafana permissions to complete OnCall setup instead of `plugins:install` since the plugin is already installed at this point
- Use the following permissions
  - `plugins:write` - Plugin setup will write to plugin config
  - `users:read` - Grafana API key being granted to OnCall will be used to read users from Grafana
  - `teams:read` - Grafana API key being granted to OnCall will be used to read teams from Grafana
  - `apikeys:create` - If Grafana API key does not exist it will be created
  - `apikeys:delete` - If existing Grafana API key does not work it will be deleted and recreated

Closes https://github.com/grafana/oncall-private/issues/1925

TODO:
- [x] Fix tests


grafana-plugin/src/state/rootBaseStore/index.ts
```diff
@@ -184,22 +185,42 @@ export class RootBaseStore {
...
+          return this.setupPluginError(
+            '🚫 User is missing permission(s) ' +
+              missingPermissions.join(', ') +
+              ' to setup OnCall before it can be used'
+          );
+        } else {
+          return this.setupPluginError(
+            '🚫 User with Admin permissions in your organization must sign on and setup OnCall before it can be used'
+          );
+        }
```
joeyorlando, line 223: 👍

joeyorlando (approved):

joeyorlando (approved):

grafana-plugin/src/containers/PluginConfigPage/PluginConfigPage.tsx
```diff
@@ -228,8 +227,8 @@ const PluginConfigPage: FC<OnCallPluginConfigPageProps> = ({
       {pluginIsConnected ? (
         <>
           <p>
-            Plugin is connected! Continue to Grafana OnCall by clicking the{' '}
-            <img alt="Grafana OnCall Logo" src={logo} width={18} /> icon over there 👈
+            Plugin is connected! Continue to Grafana OnCall by clicking OnCall under Alerts & IRM in the navigation over
+            there 👈
```
joeyorlando, line 231: nice 👍

grafana-plugin/src/plugin/PluginSetup/index.tsx
```diff
@@ -34,7 +35,10 @@ const PluginSetupWrapper: FC<PluginSetupWrapperProps> = ({ text, children }) =>
 
 const PluginSetup: FC<PluginSetupProps> = observer(({ InitializedComponent, ...props }) => {
   const store = useStore();
-  const setupPlugin = useCallback(() => store.setupPlugin(props.meta), [props.meta]);
+  const setupPlugin = useCallback(
+    () => PluginState.getGrafanaPluginSettings().then((meta) => store.setupPlugin(meta)),
```
joeyorlando, line 39: for my own curiosity, how're `props.meta` and the results from `GET /api/plugins/grafana-oncall-app/settings` different here?

grafana-plugin/src/state/plugin/index.ts
```diff
@@ -300,11 +313,22 @@ class PluginState {
...
+      });
+    } catch (e) {
+      return this.getHumanReadableErrorFromOnCallError(
+        e,
+        onCallApiUrl,
+        'install',
+        onCallApiUrlIsConfiguredThroughEnvVar
+      );
+    }
   };
```
joeyorlando, line 332: 👍

joeyorlando (approved):

grafana-plugin/src/plugin/PluginSetup/index.tsx
mderynck: This handles the case mentioned in the comment for `triggerPluginReset` where configuration has been removed but the page has not been reloaded (Still need to update that comment).  Could also help with some edge cases where the provisioned configuration has changed and reload has not happened yet.

mderynck: So I ended up changing to reload the page when configuration is removed since it was confusing to create a second `meta` object and use it instead of the one passed in.  Was there a reason we stopped reloading the page when configuration was removed? I thought we used to do that.

grafana-plugin/src/containers/PluginConfigPage/PluginConfigPage.test.tsx
```diff
@@ -302,7 +302,7 @@ describe('PluginConfigPage', () => {
     // click the confirm button within the modal, which actually triggers the callback
     await userEvent.click(screen.getByText('Remove'));
 
-    await screen.findByTestId(successful ? PLUGIN_CONFIGURATION_FORM_DATA_ID : STATUS_MESSAGE_BLOCK_DATA_ID);
+    // await screen.findByTestId(successful ? PLUGIN_CONFIGURATION_FORM_DATA_ID : STATUS_MESSAGE_BLOCK_DATA_ID);
```
mderynck, line 305: Need some help on fixing this test properly now that the page is reloading on reset.

grafana-plugin/src/plugin/PluginSetup/index.tsx
joeyorlando: this comment.. doh 🤦 
```tsx
/**
   * NOTE: there is a possible edge case when resetting the plugin, that would lead to an error message being shown
   * (which could be fixed by just reloading the page)
   * This would happen if the user removes the plugin configuration, leaves the page, then comes back to the plugin
   * configuration.
   *
   * This is because the props being passed into this component wouldn't reflect the actual plugin
   * provisioning state. The props would still have onCallApiUrl set in the plugin jsonData, so when we make the API
   * call to check the plugin state w/ OnCall API the plugin-proxy would return a 502 Bad Gateway because the actual
   * provisioned plugin doesn't know about the onCallApiUrl.
   *
   * This could be fixed by instead of passing in the plugin provisioning information as props always fetching it
   * when this component renders (via a useEffect). We probably don't need to worry about this because it should happen
   * very rarely, if ever
   */
```

I _believe_ it was removed when the new navigation came into effect, and there wasn't a need to reload to have the OnCall icon show up on the left hand nav

grafana-plugin/src/containers/PluginConfigPage/PluginConfigPage.tsx
```diff
@@ -75,13 +74,14 @@ const PluginConfigPage: FC<OnCallPluginConfigPageProps> = ({
...
   const processEnvOnCallApiUrl = process.env.ONCALL_API_URL; // don't destructure this, will break how webpack supplies this
   const onCallApiUrl = pluginMetaOnCallApiUrl || processEnvOnCallApiUrl;
-  const licenseType = pluginIsConnected?.license;
+  const fallbackLicense = CLOUD_VERSION_REGEX.test(APP_VERSION) ? GRAFANA_LICENSE_CLOUD : GRAFANA_LICENSE_OSS;
+  const licenseType = pluginIsConnected?.license || fallbackLicense;
 
   const resetQueryParams = useCallback(() => removePluginConfiguredQueryParams(pluginIsEnabled), [pluginIsEnabled]);
 
   const triggerDataSyncWithOnCall = useCallback(async () => {
+    await resetMessages();
```
joeyorlando, line 83:
```suggestion
    resetMessages();
```

grafana-plugin/src/containers/PluginConfigPage/PluginConfigPage.test.tsx
joeyorlando: hmm, this type of assertion _may_ no longer be possible for the `successful` scenario.

We're expecting the configuration form to be present here, but that is reliant on a full page/state reload to arrive at this state. It might be beyond what `@testing-library/react` and `@testing-library/user-event` are capable of? Curious to hear what someone from @grafana/grafana-oncall-frontend thinks here?

grafana-plugin/src/state/rootBaseStore/index.ts
```diff
@@ -223,13 +236,31 @@ export class RootBaseStore {
...
+      return !contextSrv.hasAccess(permission, fallback);
+    });
+  }
+
   hasFeature(feature: string | AppFeature) {
     // todo use AppFeature only
     return this.features?.[feature];
   }
 
+  getLicense() {
```
joeyorlando, line 252: you could also do something like this (equivalent to `@property` in Python):
```tsx
get license() {
  ...
}

isOpenSource(): boolean {
  return this.license === GRAFANA_LICENSE_OSS;
}
```

grafana-plugin/src/containers/PluginConfigPage/PluginConfigPage.tsx
```diff
@@ -75,13 +74,14 @@ const PluginConfigPage: FC<OnCallPluginConfigPageProps> = ({
   const pluginMetaOnCallApiUrl = jsonData?.onCallApiUrl;
   const processEnvOnCallApiUrl = process.env.ONCALL_API_URL; // don't destructure this, will break how webpack supplies this
   const onCallApiUrl = pluginMetaOnCallApiUrl || processEnvOnCallApiUrl;
-  const licenseType = pluginIsConnected?.license;
+  const fallbackLicense = CLOUD_VERSION_REGEX.test(APP_VERSION) ? GRAFANA_LICENSE_CLOUD : GRAFANA_LICENSE_OSS;
```
joeyorlando, line 77: _nit_ maybe move this to `utils/consts.ts` with the rest of the version regex-y type logic

grafana-plugin/src/containers/PluginConfigPage/PluginConfigPage.tsx
```diff
@@ -196,16 +186,24 @@ const PluginConfigPage: FC<OnCallPluginConfigPageProps> = ({
...
       <>
         <StatusMessageBlock text={syncError} />
-        <Button variant="primary" onClick={triggerDataSyncWithOnCall} size="md">
-          Retry Sync
-        </Button>
+        <HorizontalGroup>
+          <Button variant="primary" onClick={triggerDataSyncWithOnCall} size="md">
+            Retry Sync
+          </Button>
+          {licenseType === GRAFANA_LICENSE_OSS ? <RemoveConfigButton /> : null}
```
joeyorlando, line 205: _nit_ consider deduplicating this block into a component that can be reused:
```tsx
const ReconfigurePluginButtons = () => (
  <HorizontalGroup>
    <Button variant="primary" onClick={triggerDataSyncWithOnCall} size="md">
      Retry Sync
    </Button>
    {licenseType === GRAFANA_LICENSE_OSS ? <RemoveConfigButton /> : null}
  </HorizontalGroup>
);
```
(probably nested inside of the `PluginConfigPage` function so that it has access to `licenseType` and `triggerDataSyncWithOnCall`)

joeyorlando (approved):

PR #2261, opened 2023-06-15
Fix views causing drf-spectacular exceptions

iskhakov:
# What this PR does

## Checklist


joeyorlando:
nice! 😃

engine/apps/api/urls.py
```diff
@@ -69,7 +72,7 @@
 router.register(r"oncall_shifts", OnCallShiftView, basename="oncall_shifts")
 
 urlpatterns = [
-    path("", include(router.urls)),
+    # path("", include(router.urls)), # TODO: this can be removed
```
Konstantinov-Innokentii, line 75: Why it can be removed? Isn't it required to serve routers registered in router? https://www.django-rest-framework.org/api-guide/routers/

iskhakov: because it is also done on line 
https://github.com/grafana/oncall/pull/2261/files#diff-d87ee8df9c3c7eed2c4b729461f3831c5318b69907c58cdc0a2e566c170c2e9bL119

engine/apps/api/urls.py
```diff
@@ -84,8 +87,8 @@
         name="api-get-channel-verification-code",
     ),
     optional_slash_path("current_subscription", SubscriptionView.as_view(), name="subscription"),
-    optional_slash_path("terraform_file", TerraformGitOpsView.as_view(), name="terraform_file"),
-    optional_slash_path("terraform_imports", TerraformStateView.as_view(), name="terraform_imports"),
+    # optional_slash_path("terraform_file", TerraformGitOpsView.as_view(), name="terraform_file"),
+    # optional_slash_path("terraform_imports", TerraformStateView.as_view(), name="terraform_imports"),
```
joeyorlando, line 91: these views have been removed in #2530

PR #2273, opened 2023-06-16
Optmise autoresolve query

iskhakov:
# What this PR does

## Checklist

### DIFF
--- a/engine/apps/alerts/tasks/resolve_alert_group_by_source_if_needed.py
+++ b/engine/apps/alerts/tasks/resolve_alert_group_by_source_if_needed.py
@@ -23,7 +23,7 @@ def resolve_alert_group_by_source_if_needed(alert_group_pk):
         )
     else:
 
-        is_more_than_max_alerts_in_group, _ = alert_group.alerts_count_gt(
+        is_more_than_max_alerts_in_group = alert_group.alerts_count_gt(
             AlertGroupForAlertManager.MAX_ALERTS_IN_GROUP_FOR_AUTO_RESOLVE
         )
         if is_more_than_max_alerts_in_group:
--- a/engine/apps/alerts/tasks/resolve_alert_group_by_source_if_needed.py
+++ b/engine/apps/alerts/tasks/resolve_alert_group_by_source_if_needed.py
@@ -22,11 +22,9 @@ def resolve_alert_group_by_source_if_needed(alert_group_pk):
             alert_group.active_resolve_calculation_id
         )
     else:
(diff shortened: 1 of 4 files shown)


engine/apps/alerts/tasks/resolve_alert_group_by_source_if_needed.py
```diff
@@ -22,9 +22,12 @@ def resolve_alert_group_by_source_if_needed(alert_group_pk):
             alert_group.active_resolve_calculation_id
         )
     else:
+        # alert_group.alerts.count() is too expensive, we only need to know if there are more than N alerts in group
+        max_alerts_number = AlertGroupForAlertManager.MAX_ALERTS_IN_GROUP_FOR_AUTO_RESOLVE + 1
+        is_more_than_max_alerts_in_group = len(alert_group.alerts.values("pk")[:max_alerts_number]) >= max_alerts_number
```
iskhakov, line 27:
```suggestion
        is_more_than_max_alerts_in_group = len(alert_group.alerts.values("pk")[:max_alerts_number]) == max_alerts_number
```

engine/apps/alerts/tasks/resolve_alert_group_by_source_if_needed.py
```diff
@@ -22,9 +22,12 @@ def resolve_alert_group_by_source_if_needed(alert_group_pk):
             alert_group.active_resolve_calculation_id
         )
     else:
+        # alert_group.alerts.count() is too expensive, we only need to know if there are more than N alerts in group
+        max_alerts_number = AlertGroupForAlertManager.MAX_ALERTS_IN_GROUP_FOR_AUTO_RESOLVE + 1
+        is_more_than_max_alerts_in_group = len(alert_group.alerts.values("pk")[:max_alerts_number]) == max_alerts_number
```
vadimkerr, line 27:
```suggestion
        is_more_than_max_alerts_in_group = alert_group.alerts.all()[:max_alerts_number].count() == max_alerts_number
```

It's probably a good idea to count on the DB level as we only care about the number of alerts.

vadimkerr: `alert_group.alerts.values("pk")[:max_alerts_number]` will limit the queryset and will return up to 500 IDs.

`alert_group.alerts.all()[:max_alerts_number].count()` will limit the queryset and return a single number.
 
Both queries serve the same purpose, but the second one is more efficient / appropriate for the use case.

Konstantinov-Innokentii: @vadimkerr thanks, I'll polish this on monday. Probably I will method to the alert group to handle such counter

PR #2278, opened 2023-06-16
Fix duplicate orders for user notification policies

vadimkerr:
# What this PR does

Fixes an issue when multiple user notification policies have duplicated order values, leading to the following unexpected behaviours:
1. Not possible to rearrange notification policies that have duplicated orders.
2. The notification system only executes the first policy from each order group. For example, if there are policies with orders `[0, 0, 0, 0]`, only the first policy will be executed, and all others will be skipped. So the user will see four policies in the UI, while only one of them will be actually executed.

This PR fixes the issue by adding a unique index on `(user_id, important, order)` for `UserNotificationPolicy` model. However, it's not possible to add that unique index using the ordering library that we use due to it's implementation details.
I added a new abstract Django model `OrderedModel` that's able to work with such unique indices + under concurrent load.

Important info on this new `OrderedModel` abstract model:
- Orders are unique on the DB level
- Orders are allowed to be non-consecutive, for example order sequence `[100, 150, 400]` is valid
- When deleting an instance, orders of other instances don't change. This is a notable difference from the library we use. I think it's better to only delete the instance without changing any other orders, because it reduces the number of dependencies between instances (e.g. Terraform drift will be much smaller this way if a policy is deleted via the web UI).

Related to https://github.com/grafana/oncall-private/issues/1680

## Checklist


engine/apps/base/models/ordered_model.py
```diff
@@ -0,0 +1,157 @@
...
+                        raise
+                    attempts += 1
+                    time.sleep(random.random())
+
+        return wrapper
+
+    return _retry_with_params
+
+
+class OrderedModel(models.Model):
```
vadimkerr, line 51: Main focus of the PR

engine/apps/api/serializers/user_notification_policy.py
```diff
@@ -80,36 +80,16 @@ class UserNotificationPolicySerializer(UserNotificationPolicyBaseSerializer):
         default=NotificationChannelAPIOptions.DEFAULT_NOTIFICATION_CHANNEL,
     )
 
-    class Meta(UserNotificationPolicyBaseSerializer.Meta):
-        fields = [*UserNotificationPolicyBaseSerializer.Meta.fields, "prev_step"]
-        read_only_fields = ("order",)
-
     def create(self, validated_data):
-        prev_step = validated_data.pop("prev_step", None)
```
vadimkerr, line 88: `prev_step` param seems to be deprecated, I couldn't find any usages

docker-compose-developer.yml
```diff
@@ -208,7 +208,7 @@ services:
     container_name: mysql
     labels: *oncall-labels
     image: mysql:8.0.32
-    command: --default-authentication-plugin=mysql_native_password --character-set-server=utf8mb4 --collation-server=utf8mb4_unicode_ci
+    command: --default-authentication-plugin=mysql_native_password --character-set-server=utf8mb4 --collation-server=utf8mb4_unicode_ci --max_connections=1024
```
vadimkerr, line 211: Increasing `max_connections` for local concurrency tests that open more connections than default number (151)

engine/tox.ini
```diff
@@ -9,6 +9,6 @@ banned-modules =
 [pytest]
 # https://pytest-django.readthedocs.io/en/latest/configuring_django.html#order-of-choosing-settings
 # https://pytest-django.readthedocs.io/en/latest/database.html
-addopts = --color=yes --showlocals
+addopts = --no-migrations --color=yes --showlocals
```
vadimkerr, line 12: This is needed to allow creating models on demand in tests, e.g. `TestOrderedModel`.
Without `--no-migrations` it would be required to store migration for such test models.

matiasb (approved):
LGTM

mderynck (approved):

docker-compose-developer.yml
joeyorlando: what do you think about adding a `./dev/mysql.ini` file?

We have four containers (two in both this file + `docker-compose-mysql-rabbitmq.yml`) which run the `mysql` image. We could de-dupe this config and just volume mount this new file into all of these containers.

joeyorlando (approved):

docker-compose-developer.yml
vadimkerr: All 4 mysql containers share only these two settings: `--character-set-server=utf8mb4` and  `--collation-server=utf8mb4_unicode_ci`. Not sure if deduping those is worth of the extra complexity added by volume mounting the `ini` file. I'll take a closer look at this, I think it should be outside of the scope of this PR.

PR #2289, opened 2023-06-20
add is_restricted column to alert_group table

joeyorlando:
# What this PR does

Plus mark `alert_receive_channel.restricted_at` column as deprecated. This column will be removed in a future release.

## Checklist


engine/apps/alerts/models/alert_receive_channel.py
```diff
@@ -108,6 +112,8 @@ class AlertReceiveChannel(IntegrationOptionsMixin, MaintainableObject):
     Channel generated by user to receive Alerts to.
     """
 
+    alert_groups: models.QuerySet["AlertGroup"]
```
vadimkerr, line 115: AFAIK types for this kind of related objects is not `QuerySet` but `ReverseManyToOneDescriptor` or `RelatedManager`.

```python3
>>> AlertReceiveChannel.alert_groups
<django.db.models.fields.related_descriptors.ReverseManyToOneDescriptor object at 0x11ee4b650>

>>> AlertReceiveChannel.objects.last().alert_groups
<django.db.models.fields.related_descriptors.create_reverse_many_to_one_manager.<locals>.RelatedManager object at 0x107342450>
```

Is this not set correctly by `django-stubs`? I'm not sure if we want to manually add this kind of type annotations for models on the reverse side of FKs.

vadimkerr (approved):

engine/apps/alerts/models/alert_receive_channel.py
joeyorlando: good point. I can't seem to find the docs suggesting to do this. I know that adding this type annotation appeased things in my IDE at least.

I'll remove these type annotations from this PR and move them to #2180 so as to not block this PR

PR #2293, opened 2023-06-20
Grouping templating polishing11

Ukochka:
- [x] Alert group scroll
- [x] Result scroll right padding
- [x] Cheatsheet scroll right padding
- [x] Template editor and cheatsheet title alight center
- [x] Edit custom payload - left padding
- [x] Change background of Use custom payload title like Template Editor
- [x] No alert groups is missing right border
- [x] Save template - should not close the drawer if there is an error
- [x] Width cheatsheet code 100%
- [x] Edit/close Alert payload margin 8px
- [x] Background secondary Result - when Select alert group message


grafana-plugin/src/components/CheatSheet/CheatSheet.tsx
```diff
@@ -30,7 +30,7 @@ const CheatSheet = (props: CheatSheetProps) => {
           <IconButton name="times" onClick={onClose} />
         </HorizontalGroup>
         <Text type="secondary">{cheatSheetData.description}</Text>
-        <div>
+        <div style={{ width: '100%' }}>
```
teodosii, line 33: You can use class `u-width-100` instead (`cx('u-width-100')`)

teodosii (approved):

PR #2300, opened 2023-06-21
Fixes for templates&grouping

teodosii:
# What this PR does

- Fixed a few issues when using the light theme
- Show/Hide Escalation - Search Escalation - Reload Escalation (all fixes)
- Inbound Email show in http endpoint
- Allow opening the route template in collapsed state (added new button for it)

### DIFF
--- a/grafana-plugin/src/containers/IntegrationContainers/ExpandedIntegrationRouteDisplay/ExpandedIntegrationRouteDisplay.tsx
+++ b/grafana-plugin/src/containers/IntegrationContainers/ExpandedIntegrationRouteDisplay/ExpandedIntegrationRouteDisplay.tsx
@@ -10,6 +10,7 @@ import {
   Tooltip,
   ConfirmModal,
   LoadingPlaceholder,
+  Select,
 } from '@grafana/ui';
 import cn from 'classnames/bind';
 import { observer } from 'mobx-react';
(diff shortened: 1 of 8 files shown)


grafana-plugin/src/containers/IntegrationContainers/ExpandedIntegrationRouteDisplay/ExpandedIntegrationRouteDisplay.tsx
```diff
@@ -195,28 +199,34 @@ const ExpandedIntegrationRouteDisplay: React.FC<ExpandedIntegrationRouteDisplayP
...
                         className={cx('select', 'control')}
                         placeholder="Select escalation chain"
                         isLoading={isRefreshingEscalationChains}
-                        displayField="name"
                         onChange={onEscalationChainChange}
-                        value={channelFilter.escalation_chain}
+                        options={Object.keys(escalationChainStore.items).map(
+                          (eschalationChainId: EscalationChain['id']) => ({
+                            id: escalationChainStore.items[eschalationChainId].id,
+                            value: escalationChainStore.items[eschalationChainId].name,
```
teodosii, line 213: Some weirdness here as it relies on value apparently. For now I'll let both value and label to be the same, it needs `value` for searching

grafana-plugin/src/style/vars.css
```diff
@@ -38,7 +38,7 @@
 }
 
 .theme-light {
-  --cards-background: var(--blue-2);
+  --cards-background: rgba(244, 245, 245);
```
teodosii, line 41: This was correct for theme-dark but wrong value for theme-light as it was pointing to blue instead. It had no use prior to this, changed the value of it

Ukochka (approved):

PR #2317, opened 2023-06-23
Remove url from sms notification, change format

iskhakov:
# What this PR does
Remove link from sms notification to avoid difficulties with different countries anti-spam regulations

## Checklist


joeyorlando:
➕ for @Dieterbe's comment, I had the same question.

mderynck:
Only the title for SMS is editable, but if a URL gets put as the title it could still be blocked. A warning in the template editor would be good or we can completely strip out the URL in the phone provider.

iskhakov:
Thank you for the suggestions!

grafana-plugin/src/components/AlertTemplates/CommonAlertTemplatesForm.config.ts
```diff
@@ -46,7 +46,8 @@ export const commonTemplateForEdit: { [id: string]: TemplateForEdit } = {
   sms_title_template: {
     name: TemplateOptions.SMS.key,
     displayName: 'Sms title',
-    description: '',
+    description:
+      "Result of this template will be used as title of SMS message. Please don't include any urls to avoid SMS message being blocked by carriers.",
```
joeyorlando, line 50: It seems some countries also don't like when phone numbers are included:
image


```suggestion
      "Result of this template will be used as title of SMS message. Please don't include any urls, or phone numbers, to avoid SMS message being blocked by carriers.",
```

PR #2339, opened 2023-06-26
add zvonok integration

sreway:
Added integration with [zvonok.com](https://zvonok.com) service.

Features:
- Phone number validation
- Test calls
- Selection of pre-recorded audio
- Making calls
- Processing call status
- Acknowledgment alert group (optional)

To process the call status, it is required to add a postback with the GET method on the side of the zvonok.com service with the following format ([more info here](https://zvonok.com/ru-ru/guide/guide_postback/)):

```${ONCALL_BASE_URL}/zvonok/call_status_events?campaign_id={ct_campaign_id}&call_id={ct_call_id}&status={ct_status}&user_choice={ct_user_choice}```

The names of the transmitted parameters can be redefined through environment variables.

### CODE REVIEW

File: engine/apps/zvonok/models/phone_call.py [hunk 1]
```diff
@@ -0,0 +1,78 @@
...
+        blank=True,
+        max_length=50,
+    )
+
+    campaign_id = models.CharField(
+        max_length=50,
+    )
+
+
+
```
joeyorlando, line 78: extra newlines here

joeyorlando (commented):
thanks for the contribution! 😄 

do you mind adding trailing newlines to the new files + adding back the trailing newlines to modified files. thank you!

engine/apps/zvonok/models/phone_call.py
sreway: done

sreway: done

engine/apps/zvonok/models/phone_call.py
```diff
@@ -0,0 +1,77 @@
...
+        "in_process": IN_PROCESS,
+        "pincode_nook": PINCODE_NOOK,
+        "synth_error": SYNTH_ERROR,
+        "user": USER,
+    }
+
+
+class ZvonokPhoneCall(ProviderPhoneCall, models.Model):
+    class Meta:
+        db_table = "zvonok_phonecall"
```
Konstantinov-Innokentii, line 59: There is no need to specify db_table here. It was specified for  PhoneCallRecord only for migration purposes.

engine/apps/zvonok/urls.py
```diff
@@ -0,0 +1,9 @@
+from django.urls import path
+from .views import CallStatusCallback, HealthCheck
+
+app_name = "zvonok"
```
Konstantinov-Innokentii, line 4: How `app_name` is used? Could it be removed?

engine/apps/zvonok/views.py
```diff
@@ -0,0 +1,48 @@
...
+    def get(self, request):
+        call_id = request.GET.get(live_settings.ZVONOK_POSTBACK_CALL_ID)
+        call_status = request.GET.get(live_settings.ZVONOK_POSTBACK_STATUS)
+        user_choice = request.GET.get(live_settings.ZVONOK_POSTBACK_USER_CHOICE)
+        update_zvonok_call_status(call_id=call_id, call_status=call_status, user_choice=user_choice)
+        return Response(data="", status=status.HTTP_204_NO_CONTENT)
+
+
+class HealthCheck(APIView):
+    def get(self, request):
```
Konstantinov-Innokentii, line 47: Is this view required by zvonok?  If not, let's remove it

engine/settings/base.py
```diff
@@ -720,5 +721,18 @@ class BrokerTypes:
 PHONE_PROVIDERS = {
     "twilio": "apps.twilioapp.phone_provider.TwilioPhoneProvider",
     # "simple": "apps.phone_notifications.simple_phone_provider.SimplePhoneProvider",
+    "zvonok": "apps.zvonok.phone_provider.ZvonokPhoneProvider",
 }
 PHONE_PROVIDER = os.environ.get("PHONE_PROVIDER", default="twilio")
+
+if PHONE_PROVIDER == "zvonok":
```
Konstantinov-Innokentii, line 728: I believe we don't need this check, because currently we have no way to make it work as expected.

PHONE_PROVIDER value can be changed via UI, but at this point it still  will be twilio, since value is changed through live_settings mechanism and db, not via env.
So, it makes no sense to check value of settings.PHONE_PROVIDER, but live_settings.PHONE_PROVIDER should be checked. However it's impossible, since django  checks settings files before apps are loaded. You can try and will receive:
```django.core.exceptions.AppRegistryNotReady: Apps aren't loaded yet.``` Exception.

Also, I think that force-disabling GRAFANA_CLOUD_NOTIFICATIONS_ENABLED is not good idea, since someone might want to switch to cloud notifications after zvonok is configured.

File: engine/settings/base.py [hunk 6]
```diff
@@ -233,6 +233,7 @@ class DatabaseTypes:
     "fcm_django",
     "django_dbconn_retry",
     "apps.phone_notifications",
+    "apps.zvonok"
```
Konstantinov-Innokentii, line 236: Could you please move enabling of zvonok app to line 685, where we plug-in apps required only for oss?

engine/engine/urls.py
```diff
@@ -67,6 +67,11 @@
         path("api/internal/v1/", include("apps.oss_installation.urls", namespace="oss_installation")),
     ]
 
+if settings.PHONE_PROVIDER == "zvonok":
```
Konstantinov-Innokentii, line 70: It is needed to check live_settings.PHONE_PROVIDER here. If it won't work ( I'm not remember if apps are loaded before urls, you can simply enable zvonok urls if settings.IS_OPEN_SOURCE (line 65)

Konstantinov-Innokentii (changes requested):
LGTM, great work. Left some comments.

Besides that comments it is needed to:
1. Remove links to non-english docs
2. Add paragraph about this phone provider to https://grafana.com/docs/oncall/latest/open-source/#twilio-setup ? It's in the docs folder.

sreway:
> LGTM, great work. Left some comments.
> 
> Besides that comments it is needed to:
> 
> 1. Remove links to non-english docs
> 2. Add paragraph about this phone provider to https://grafana.com/docs/oncall/latest/open-source/#twilio-setup. It's in the docs folder.

Fixed

PR #2359, opened 2023-06-27
update rabbitmq docker containers to v3.12.0

### LINKED ISSUE #695: Docker-compose - Using rabbitmq version from 2 years ago.
Hi Everyone,
This issue is both a question and a suggestion.
Looks like the example in the docker-compose file (see below) is using [rabbitmq:3.7.15-management](https://hub.docker.com/layers/library/rabbitmq/3.7.15-management/images/sha256-b00ca90bc3ae85867317ab624ee476bd5d8d2a19f9ccc414e5baf685c770120d?context=explore) which is a 3 year old version (see attached screenshot).

I was wondering if the rabbitmq version can be replaced with 3.11 and if it can be we should update the tag to [3.11-management](https://hub.docker.com/layers/library/rabbitmq/3.11-management/images/sha256-72a686d14b4869a1c02fa8675a711a99b64ae894258b0f135b2d0e5bf7972985?context=explore) which is 5 days old and will automatically update with newer 3.11.x versions.

Thank you !

https://github.com/grafana/oncall/blob/7cc01e0585d5e9dd9a4f6195212832c41d12e5e9/docker-compose.yml#L102-L104

image

joeyorlando:
# What this PR does

Update `rabbitmq` Docker containers used in the `docker-compose` config files, Drone pipelines, and GitHub Actions to use version 3.12.0.

FWIW, we're already using v12.0.0 of the bitnami `rabbitmq` `helm` chart which, by default, uses the `3.12.0-debian-11-r0` tag for the `rabbitmq` image ([chart docs](https://artifacthub.io/packages/helm/bitnami/rabbitmq/12.0.0)).

closes #695 

## Checklist

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -144,7 +144,7 @@ steps:
 # Services for Unit Test Backend
 services:
   - name: rabbit_test
-    image: rabbitmq:3.7.19
+    image: rabbitmq:3.12.0
     environment:
       RABBITMQ_DEFAULT_USER: rabbitmq
       RABBITMQ_DEFAULT_PASS: rabbitmq
@@ -418,6 +418,6 @@ kind: secret
 name: drone_token
 ---
 kind: signature
-hmac: 59ad498428731001484be9ac6b9d10685b2cd49f30eb3d97ed7dc99f031dca22
+hmac: e4d25ad9f26d0a2444033ec077029da50818d912fd24ba49c0a915947adffbab
 
 ...
--- a/.github/workflows/linting-and-tests.yml
+++ b/.github/workflows/linting-and-tests.yml
@@ -96,7 +96,7 @@ jobs:
       SLACK_CLIENT_OAUTH_ID: 1
     services:
       rabbit_test:
-        image: rabbitmq:3.7.19
+        image: rabbitmq:3.12.0
         env:
           RABBITMQ_DEFAULT_USER: rabbitmq
           RABBITMQ_DEFAULT_PASS: rabbitmq
@@ -148,7 +148,7 @@ jobs:
       ONCALL_TESTING_RBAC_ENABLED: ${{ matrix.rbac_enabled }}
     services:
       rabbit_test:
-        image: rabbitmq:3.7.19
+        image: rabbitmq:3.12.0
         env:
           RABBITMQ_DEFAULT_USER: rabbitmq
           RABBITMQ_DEFAULT_PASS: rabbitmq
@@ -192,7 +192,7 @@ jobs:
       ONCALL_TESTING_RBAC_ENABLED: ${{ matrix.rbac_enabled }}
     services:
       rabbit_test:
-        image: rabbitmq:3.7.19
+        image: rabbitmq:3.12.0
         env:
           RABBITMQ_DEFAULT_USER: rabbitmq
           RABBITMQ_DEFAULT_PASS: rabbitmq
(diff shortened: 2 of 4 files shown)


Konstantinov-Innokentii (approved):

roock:
@joeyorlando I think it would make sense to include this in the changelog for multiple reasons:
- Users should be informed to update rabbitmq to a more recent version
- Existing installations will need to perform several intermediate steps to upgrade from 3.7 to 3.12 (see https://rabbitmq.com/upgrade.html#rabbitmq-version-upgradability).

joeyorlando:
@roock thanks for pointing this out.

this _shouldn't_ have any user facing impact. the two modified `docker-compose` config files are only used for local development purposes AFAIK. Changes to the `rabbitmq` version in the `helm` chart were added to the changelog under [v1.2.42](https://github.com/grafana/oncall/blob/dev/CHANGELOG.md#v1242-2023-06-12).

wdyt?

roock:
@joeyorlando yes, helm chart is not affected.

I don't agree on the usage of the docker compose files. It's not clear to me that the `docker-compose-mysql-rabbitmq.yml` version of the docker-compose setup should be considered for development only, for me it is something in between the "hobby" and the "production"(helm) setup.

When I started setting up our Grafana OSS stack somewhen in 2022, there was only one docker-compose file with mysql and rabbitmq included (see https://github.com/grafana/oncall/pull/611), that's why I still use this stack and ran into the above issue when upgrading to a recent rabbitmq version.

Probably the `docker-compose-mysql-rabbitmq.yml` file needs proper documentation on its purpose.

joeyorlando:
@roock you raise a fair point 👍 retroactively added this to the changelog in [this commit](https://github.com/grafana/oncall/blob/dev/CHANGELOG.md#v132-2023-06-29).

re: purpose of `docker-compose-mysql-rabbitmq.yml`. I believe this was originally the main `docker-compose.yml` but got refactored in #611 to it's current file name. I don't 100% recall why it stuck around tbh.

Personally, I've been advocating to get rid of this file, as maintaining three separate, but very similar, `docker-compose` configuration files leads to a larger surface area, and more potential to break something. imo `docker-compose-developer.yml` should be renamed to `docker-compose.yml`,  as it's the most "fleshed-out" and battle-tested config file. We could then achieve what these three files do, all via the `COMPOSE_PROFILES` env var. But I digress 🤓 

> let the dev team know, that they need to recreate their local stack after pulling in your change

We don't have a formal process for this atm. I suppose you could run into issues in the scenario you mentioned, where if you reuse the same rabbitmq Docker volume mount, but upgrade image versions, you'd have to follow those upgrade steps you pointed out. At least on the core team, when people hit these types of problems for local development, they just run `make cleanup && make init start`. But if you have any suggestions for some, lmk!

roock:
> @roock you raise a fair point 👍 retroactively added this to the changelog in [this commit](https://github.com/grafana/oncall/blob/dev/CHANGELOG.md#v132-2023-06-29).

👌

> re: purpose of `docker-compose-mysql-rabbitmq.yml`. I believe this was originally the main `docker-compose.yml` but got refactored in #611 to it's current file name. I don't 100% recall why it stuck around tbh.
> 
> Personally, I've been advocating to get rid of this file, as maintaining three separate, but very similar, `docker-compose` configuration files leads to a larger surface area, and more potential to break something. imo `docker-compose-developer.yml` should be renamed to `docker-compose.yml`, as it's the most "fleshed-out" and battle-tested config file. We could then achieve what these three files do, all via the `COMPOSE_PROFILES` env var. But I digress 🤓

Sounds like a good plan to migrate the two different docker compose files into one with profiles (although I'm not sure docker compose is sophisticated enough to handle such a complex setup).

PR #2375, opened 2023-06-28
use docker buildx native multiarch image support

joeyorlando:
# What this PR does

`docker buildx` natively supports multiarch builds. This means we no longer need to have duplicate `docker build` Drone steps + there is no longer a need for the manual "manifesting" step. This [blog post](https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/) from Docker perfectly summarizes the difference between what we do now vs. what this PR would do.

This would also get rid of pushing `vX.Y.Z-{architecture}` tagged images. These are redundant because modern `docker` clients can pull the appropriate image for the host OS, from a multiarch image:
Screenshot 2023-06-28 at 12 54 09

## Checklist

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -224,10 +224,8 @@ trigger:
 ---
 kind: pipeline
 type: docker
-name: OSS engine release (amd64)
-platform:
-  os: linux
-  arch: amd64
+name: OSS engine release
+
 steps:
   - name: set engine version
     image: alpine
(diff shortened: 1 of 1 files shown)


.drone.yml
```diff
@@ -286,7 +242,8 @@ steps:
       DOCKER_BUILDKIT: 1
     settings:
       repo: grafana/oncall
-      tags: ${DRONE_TAG}-arm64-linux
+      tags: ${DRONE_TAG}
+      platform: linux/arm64/v8,linux/amd64
```
joeyorlando, line 246: main focus of this PR

Konstantinov-Innokentii (approved):

PR #2385, opened 2023-06-28
OnCall plugin use service accounts instead of api keys

### LINKED ISSUE #1806: Transition OnCall plugin to use Grafana service account
This issue is for tracking transitioning the OnCall plugin to use a service account instead of API key for communication with the OnCall backend. We should still keep the API key code and use it for older versions that do not support service accounts.

More info about API key deprecation: https://github.com/grafana/grafana/issues/53567

mderynck:
# What this PR does
Changes OnCall plugin to use service accounts and api tokens instead of api keys. API keys will continue to work but if the plugin ever replaces them it will use a service account instead.  Previously this was thought to be unnecessary but it was missing the case where the API key was converted to a service account which it could no longer find when searching the `/api/auth/keys` endpoint.  That key would not be deleted and it would conflict with a newly created one of the same name.  

Now the behaviour is as follows: 
1. Anytime a new token is needed all API keys and tokens under the service account matching the defined names will be deleted
2. A service account will be created named `sa-autogen-OnCall` if one does not already exist
3. An api token will be created under that service account named `OnCall`

#1806 

## Checklist


joeyorlando:
@mderynck this closes #1806, right?

Konstantinov-Innokentii:
@mderynck will this changes work with older grafana versions ( for example OSS) with no service accounts?

joeyorlando:
@Konstantinov-Innokentii w/ this PR it might be worthwhile bumping our [minimum required grafana version](https://github.com/grafana/oncall/blob/dev/grafana-plugin/src/plugin.json#L628-L631) in `.grafana-plugin/src/plugin.json`. API keys are slated to be deprecated in the "near future":

Screenshot 2023-06-29 at 10 07 43

grafana-plugin/src/state/plugin/index.ts
joeyorlando, line 1: _nit_  `PluginState.getGrafanaToken` is called in the `PluginState.syncDataWithOnCall` method, there's a small comment there mentioning "API keys", we might to remove/update that comment

joeyorlando (approved):

PR #2390, opened 2023-06-29
re-enable mypy GitHub Actions CI job

joeyorlando:
# What this PR does

```bash
❯ mypy .
Success: no issues found in 595 source files
```

- re-enable the mypy CI check
- fixes all `django-manager-missing` mypy errors
- disable all other rules currently giving mypy errors
  - changing the approach here. rather than enforcing that backend contributors fix >= 1 `mypy` error on their PR, lets simply disable all the rules that're currently returning errors and slowly re-enable these one at a time #2392 

## Checklist


engine/apps/auth_token/models/api_auth_token.py
```diff
@@ -8,6 +8,8 @@
 
 
 class ApiAuthToken(BaseAuthToken):
+    objects: models.QuerySet["ApiAuthToken"]
```
vadimkerr, line 11: nit: `objects` is not an instance of `QuerySet` but `Manager`

```python3
>>> from django.db.models import QuerySet, Manager
>>> isinstance(ApiAuthToken.objects, QuerySet)
False
>>> isinstance(ApiAuthToken.objects, Manager)
True
```

vadimkerr (approved):

Issue #1586, opened 2023-03-21
Integration test for Users screen


teodosii:
Our frontend tests lack plenty of integration tests, it would be super nice to start adding a few in order to test the most used flows. This first integration test on Users should ideally cover

- permissions (Viewer/Editor/Admin when accessing links or tabs from View My Profile)
- actions on the UI do not throw exceptions

Ideally we should only mock the stores to have data in it, anything else should work in integration with other components part of the Users tree.

joeyorlando:
@teodosii 👋 wdyt about re-describing the scope of this issue, to make the list of deliverables very clear? imo, it's a bit open-ended, so it's tough to know when this is task is complete.

Example deliverables could be something like:
- As an Admin or Editor, I should be able to do...
- As a Viewer, I should be not be able to..

teodosii:
Good idea, I'll rephrase this task once I get back to it. Or if you wanna take this instead feel free to do it as well.

maskin25:
Currently I added two tests:
- admin is allowed to edit other profiles
- editor is not allowed to edit other profiles

(see `grafana-plugin/integration-tests/users/usersActions.test.ts`)

teodosii:
I'm thinking of these scenarios
- **Viewers** are not allowed to view other users
- **Viewers** cannot access restricted tabs in the view my profile
- **Editors** cannot edit other users' profiles
- **Editors** cannot view other users' data
- **Editors** cannot access restricted tabs in the view my profile
- **Admin** can do all the above

PR #2417, opened 2023-07-04
add first multi-role e2e tests

joeyorlando:
# What this PR does

Lays ground work for #1586. Adds three new fixtures, `adminRolePage`, `editorRolePage`, and `viewerRolePage`. These fixtures can be easily accessed in a `test` context and allow the test to be run as a user authenticated with one of these Grafana basic roles.

The bulk of the changes in the PR are to the "global setup" step. There is a bit of logic + communication with the Grafana instance's API, in order to create all the necessary authentication credentials.

Lastly, adds the first basic role authorization test, asserting that Admin/Editors can view the list of OnCall users, whereas Viewers cannot.

## Checklist


grafana-plugin/integration-tests/users/viewUsers.test.ts
joeyorlando, line 1: main change in this PR

grafana-plugin/integration-tests/utils/constants.ts
```diff
@@ -1,6 +1,13 @@
...
 export const ONCALL_API_URL = process.env.ONCALL_API_URL || 'http://host.docker.internal:8080';
-export const GRAFANA_USERNAME = process.env.GRAFANA_USERNAME || 'oncall';
-export const GRAFANA_PASSWORD = process.env.GRAFANA_PASSWORD || 'oncall';
 export const MAILSLURP_API_KEY = process.env.MAILSLURP_API_KEY;
+
+export const GRAFANA_VIEWER_USERNAME = process.env.GRAFANA_VIEWER_USERNAME || 'viewer';
+export const GRAFANA_VIEWER_PASSWORD = process.env.GRAFANA_VIEWER_PASSWORD || 'viewer';
+export const GRAFANA_EDITOR_USERNAME = process.env.GRAFANA_EDITOR_USERNAME || 'editor';
+export const GRAFANA_EDITOR_PASSWORD = process.env.GRAFANA_EDITOR_PASSWORD || 'editor';
+export const GRAFANA_ADMIN_USERNAME = process.env.GRAFANA_ADMIN_USERNAME || 'oncall';
```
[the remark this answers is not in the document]
joeyorlando: false positive, this is only used for testing purposes.

grafana-plugin/integration-tests/utils/constants.ts
```diff
@@ -1,6 +1,13 @@
...
+export const GRAFANA_ADMIN_PASSWORD = process.env.GRAFANA_ADMIN_PASSWORD || 'oncall';
```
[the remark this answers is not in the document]
joeyorlando: false positive, this is only used for testing purposes.

grafana-plugin/integration-tests/.env.example
```diff
@@ -0,0 +1,10 @@
+# copy this file to ./.env and fill out the values according to your local setup
```
iskhakov, line 1: is this comment still relevant?

iskhakov (approved):

grafana-plugin/integration-tests/globalSetup.ts
```diff
@@ -1,37 +1,57 @@
...
+  config: FullConfig,
+  userName: string,
+  password: string,
+  storageStateFileLocation: string,
+  userCreationSettings?: UserCreationSettings,
+  closeContext = false
+): Promise<BrowserContext> => {
+  if (userCreationSettings !== undefined && IS_OPEN_SOURCE) {
+    const { adminAuthedRequest, role } = userCreationSettings;
+    await grafanaApiClient.idempotentlyCreateUserWithRole(adminAuthedRequest, userName, password, role);
```
teodosii, line 37: Hmm shouldn't this go in a try catch clause?

grafana-plugin/integration-tests/globalSetup.ts
```diff
@@ -66,61 +86,55 @@ const configureOnCallPlugin = async (page: Page): Promise<void> => {
...
+  const adminBrowserContext = await generateLoginStorageStateAndOptionallCreateUser(
+    config,
+    GRAFANA_ADMIN_USERNAME,
+    GRAFANA_ADMIN_PASSWORD,
+    ADMIN_USER_STORAGE_STATE
+  );
+  const adminPage = await adminBrowserContext.newPage();
+  const { request: adminAuthedRequest } = adminBrowserContext;
+
+  await generateLoginStorageStateAndOptionallCreateUser(
```
teodosii, line 110: So these calls basically open a new window for each call and perform the login, and in tests you just get a reference on that rolePage?

grafana-plugin/integration-tests/globalSetup.ts
joeyorlando: I think it's okay here. The `playwright.APIRequestContext` object that makes the HTTP requests operates a bit different from `axios` that we use for the actual plugin UI. It doesn't throw exceptions for HTTP 412 (user already exists), whereas `axios` probably does?

grafana-plugin/integration-tests/globalSetup.ts
joeyorlando: correct 👍 on setup, it will basically generate three files:
- `integration-tests/.auth/admin.json`
- `integration-tests/.auth/editor.json`
- `integration-tests/.auth/viewer.json`

each of the fixtures then create a "browser context" object that is authenticated using one of these three files, depending on which role you select

grafana-plugin/integration-tests/globalSetup.ts
teodosii: Hmm that's odd, as you're also throwing an exception from that function if you get some other response back.

teodosii (approved):

grafana-plugin/integration-tests/globalSetup.ts
joeyorlando: ah fair point 👍 in that case I think the tests should purposefully fail, because something is not setup the way we would expect

PR #2421, opened 2023-07-04
address mobile device push notification delivery issue when user had > 1 registered device

joeyorlando:
# What this PR does

Address issue where if the user had multiple registered devices w/ FCM, doing django queries like `.first()` could potentially pick the wrong device. Do this in two ways:
1. set the `DELETE_INACTIVE_DEVICES` `fcm_django` setting to `True`. According to the [docs](https://github.com/xtrinch/fcm-django/blob/20e275618b47cea4e1b4dd3b1a8c640a164a7ec5/README.rst?plain=1#L127-L130), this works as follows:

    > devices to which notifications cannot be sent, are deleted upon receiving error response from FCM
2. Customizing the `FCMDevice` model provided by `fcm_django`. Add a new method, `get_active_device_for_user`, so that we can centralize the logic for this rather than duplicating `FCMDevice.objects.filter(user=user).first()`

https://raintank-corp.slack.com/archives/C0229FD3CE9/p1688461915752119

## Checklist


engine/apps/mobile_app/models.py
```diff
@@ -16,6 +17,19 @@ def get_expire_date():
...
+    def filter(self, *args, **kwargs):
+        return super().filter(*args, **kwargs, active=True)
+
+
+class FCMDevice(BaseFCMDevice):
+    active_objects = ActiveFCMDeviceQuerySet.as_manager()
+
+    @classmethod
+    def get_active_device_for_user(cls, user: User) -> "FCMDevice" | None:
+        return cls.active_objects.filter(user=user).first()
```
iskhakov, line 30: I'm not in the context of FCM, but shouldn't we consider the last connected device as active?

joeyorlando: There is explicitly an `active` column for this (the setting of this to true/false is all managed by the `fcm_django` library).

**More context**
We have `ONE_DEVICE_PER_USER` set to `True` ([here](https://github.com/grafana/oncall/blob/dev/engine/settings/base.py#L621)) which according to the docs:
> true if you want to have only one active device per registered user at a time

There was an edge case with this setting though that I'm trying to solve for here. If the user had one active and one inactive device in `fcm_django_fcmdevice` table, using `.first()` could possibly return the inactive device

iskhakov: if we are sure there is the only one device, can we use `.get()`?

iskhakov:
great job, approved

engine/apps/mobile_app/models.py
joeyorlando: good point 👍 updated this and added some tests around this logic. also verified our databases that this is currently the case.

Issue #2463, opened 2023-07-07
internal api returns 404 for schedules?page=2 which is the page it told me to query next


imtoori:
### Discussed in https://github.com/grafana/oncall/discussions/2308

<div type='discussions-op-text'>

<sup>Originally posted by **Dieterbe** June 22, 2023</sup>
a request to /oncall/api/internal/v1/schedules?page=1
includes this bit:
```
next: '(...)/api/internal/v1/schedules?page=2'
```
but when i query this URL, i get an http 404</div>

The `next` link for paginating schedules is wrong. It doesn't include the prefix `/oncall`

current (wrong): `next: '(...)/api/internal/v1/schedules?page=2'` 
correct: `next: '(...)/oncall/api/internal/v1/schedules?page=2'`

PR #2467, opened 2023-07-07
fix pagination next and previous links for schedules and integrations GET endpoints

joeyorlando:

closes https://github.com/grafana/oncall/issues/2463

## Checklist

### DIFF
--- a/engine/apps/api/tests/test_schedules.py
+++ b/engine/apps/api/tests/test_schedules.py
@@ -147,7 +147,7 @@ def test_get_list_schedules(
 def test_get_list_schedules_pagination(
     schedule_internal_api_setup, make_escalation_chain, make_escalation_policy, make_user_auth_headers
 ):
-    user, token, calendar_schedule, ical_schedule, web_schedule, slack_channel = schedule_internal_api_setup
+    user, token, calendar_schedule, ical_schedule, web_schedule, _ = schedule_internal_api_setup
 
     # setup escalation chain linked to web schedule
     escalation_chain = make_escalation_chain(user.organization)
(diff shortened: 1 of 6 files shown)


mderynck (approved):

engine/apps/api/tests/test_schedules.py
```diff
@@ -219,7 +219,7 @@ def test_get_list_schedules_pagination(
     client = APIClient()
 
     schedule_list_url = reverse("api-internal:schedule-list")
-    absolute_url = create_engine_url(schedule_list_url, override_base="http://testserver")
```
joeyorlando, line 222: this should be fine. `override_base` has no effect here because the _actual_ absolute URL that is being generated is overriden here on the actual `request` object itself in `PathPrefixedCursorPagination.paginate_queryset`:
```py3
class PathPrefixedPagination(PageNumberPagination):
    def paginate_queryset(self, queryset, request, view=None):
        request.build_absolute_uri = lambda: create_engine_url(request.get_full_path())
        return super().paginate_queryset(queryset, request, view)
```

engine/apps/api/views/alert_receive_channel.py
```diff
@@ -29,26 +28,11 @@
...
-    max_page_size = 50
-
-    def paginate_queryset(self, queryset, request, view=None):
-        """Override to apply pagination only if ?page= is present in query params
-        Required for backwards compatibility with older versions
-        """
-        page_number = request.query_params.get(self.page_query_param, None)
-        if not page_number:
-            return None
-        return super().paginate_queryset(queryset, request, view)
```
joeyorlando, line 49: I think this can be removed. @iskhakov wdyt?

This was _likely_ here to handle the transition when the paginated integration's page was release, and to handle the case in Cloud where UI clients running the previous version would've received a different API response.

iskhakov: I agree

PR #2471, opened 2023-07-07
augment API response pagination attributes

joeyorlando:
# What this PR does

This PR:
- adds a few attributes to paginated API responses
- removes channel filter "send demo alert" internal API endpoint + tests (this endpoint was marked as deprecated + not consumed by the web UI)

With the new paginated API response schema, the web UI will no longer need to:
- hardcode `ITEMS_PER_PAGE` for each table
- manually calculate total number of pages

(these two things ☝️  will be done in https://github.com/grafana/oncall/issues/2476)

For `GET /api/internal/v1/alertgroups` the response will now look like this:
```diff
{
    "next": <url> | None,
    "previous": <url> | None,
    "results": [],
++  "page_size": <int>
}
```

For all other paginated API responses, the response will now look like:
```diff
{
    "count": <int>,
    "next": <url> | None,
    "previous": <url> | None,
    "results": [],
++  "page_size": <int>,
++  "current_page_number": <int>,
++  "total_pages": <int>
}
```

## TODO
- [x] update public API docs to include these new attributes

## Checklist


engine/common/api_helpers/paginators.py
joeyorlando, line 1: main change for the backend

engine/apps/alerts/models/alert_receive_channel.py
```diff
@@ -527,8 +527,8 @@ def heartbeat_module(self):
         return getattr(heartbeat, self.INTEGRATIONS_TO_REVERSE_URL_MAP[self.integration], None)
 
     # Demo alerts
-    def send_demo_alert(self, force_route_id=None, payload=None):
```
joeyorlando, line 530: `force_route_id` seemed to _only_ be passed in when this method was called from `ChannelFilter. send_demo_alert` (which was already deprecated and is now removed in this PR)

engine/apps/oss_installation/views/cloud_users.py
```diff
@@ -60,20 +68,7 @@ def get(self, request):
...
-                [
-                    ("count", self.page.paginator.count),
-                    ("matched_users_count", matched_users_count),
-                    ("next", self.get_next_link()),
-                    ("previous", self.get_previous_link()),
-                    ("results", data),
-                ]
-            )
-        )
+        return self.get_paginated_response(data, len(cloud_identities))
```
joeyorlando, line 71: `HundredPageSizePaginator` inherits from `PathPrefixedPagePagination`  and therefore all of these attributes, except for `matched_users_count` would be set. `matched_users_count` is being set above in the new `CloudUsersPagination` paginator class.

matiasb (approved):
Looks good

PR #2474, opened 2023-07-08
pagination cleanup - frontend changes

joeyorlando:
**Changes**
- modify table components + stores to use the pagination parameters that're returned from the API, rather than having to hardcode `ITEMS_PER_PAGE` in the UI
- update `make_request` method to allow second argument, `RequestConfig`, to be optional (no need to pass in `{}` instead)


grafana-plugin/src/components/Policy/NotificationPolicy.tsx
```diff
@@ -32,9 +32,8 @@ export interface NotificationPolicyProps {
   slackUserIdentity?: User['slack_user_identity'];
   onChange: (id: NotificationPolicyType['id'], value: NotificationPolicyType) => void;
   onDelete: (id: string) => void;
-  notificationChoices: any[];
-  channels?: any[];
```
joeyorlando, line 36: `channels` is not used

PR #2493, opened 2023-07-11
Exclude schedules from deleted organizations from notification list

Ferril:
# What this PR does
On notifying about shifts change in slack we don't check whether organization was deleted. This PR excludes schedules from deleted organizations from shift notification list

## Checklist


joeyorlando (commented):
overall looks good 👍 

looks like we have pre-existing unit tests for these two tasks. We want want to add another test case to ensure it filters out deleted orgs.

vadimkerr (approved):

Issue #458, opened 2022-09-01
Remove "v" from Grafana OnCall versions for more consistency with the rest of grafana products


iskhakov:
Otherwise it leads to following output
```
✔ Downloaded grafana-oncall-app vv1.0.21 zip successfully
```

maskin25:
@iskhakov is it still reproducing?

Matvey-Kuk:
@maskin25 I think so: https://github.com/grafana/oncall/releases/tag/v1.1.23

kuzaxak:
double `v` doe not allow to use grafana operator, as this version wouldn't pass CRD validation. Could it be fixed?

kuzaxak:
https://github.com/grafana/oncall/issues/1153

kuzaxak:
https://github.com/grafana/oncall/issues/1858

kuzaxak:
@Matvey-Kuk Could you please have a look into that issue as it was reported multiple times and create a confusion.

PR #2540, opened 2023-07-14
Remove v prefix from plugin version on build

matiasb:
Related to https://github.com/grafana/oncall/issues/458

Change version naming for plugin and docker releases to drop the `v` prefix (ie. `1.2.3` instead of `v1.2.3`).
Update both to keep consistency between plugin and engine versions, which are also checked to match.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -10,7 +10,7 @@ steps:
       - apt-get update
       - apt-get --assume-yes install jq
       - cd grafana-plugin/
-      - if [ -z "$DRONE_TAG" ]; then echo "No tag, not modifying version"; else jq '.version="${DRONE_TAG#v}"' package.json > package.new && mv package.new package.json && jq '.version' package.json; fi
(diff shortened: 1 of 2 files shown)


.drone.yml
```diff
@@ -10,7 +10,7 @@ steps:
       - apt-get update
       - apt-get --assume-yes install jq
       - cd grafana-plugin/
-      - if [ -z "$DRONE_TAG" ]; then echo "No tag, not modifying version"; else jq '.version="${DRONE_TAG}"' package.json > package.new && mv package.new package.json && jq '.version' package.json; fi
+      - if [ -z "$DRONE_TAG" ]; then echo "No tag, not modifying version"; else jq ".version=\"${DRONE_TAG#v}\"" package.json > package.new && mv package.new package.json && jq '.version' package.json; fi
```
joeyorlando, line 13: interesting bash variable substitution trick. TIL

joeyorlando (approved):

PR #2550, opened 2023-07-17
Maxim/bring heartbeats back to UI

maskin25:
# What this PR does

Bring heartbeats back to UI

## Checklist


joeyorlando (commented):
thanks for helping out with this 🙏 

I checked things locally and it mostly looks good to me. Few comments/questions.
1. Can we add some sort of confirmation popup notification when you click "Update" in the heartbeat settings drawer? imo, at the moment it's not very clear to the user that anything has happened.
2. I think this is worth adding basic e2e tests for. Without something like this, there's nothing to prevent us from regressing into the broken state again. For example we could test that you're able to update an integration's heartbeat interval + are able to send a request to the integration's heartbeat URL + see that in the UI this is represented to you as the green heart symbol. Just some ideas. There are similar e2e tests [here](https://github.com/grafana/oncall/blob/dev/grafana-plugin/integration-tests/integrations/maintenanceMode.test.ts) for testing an integration's maintenance mode.

maskin25:
> 1. Can we add some sort of confirmation popup notification when you click "Update" in the heartbeat settings drawer? imo, at the moment it's not very clear to the user that anything has happened.

Which type of confirmation are you asking about? Like below or just green notification at the right top corner?


Screenshot 2023-07-17 at 17 15 29

> 2. adding basic e2e tests

Good idea, I'll add

grafana-plugin/src/models/alert_receive_channel/alert_receive_channel.ts
```diff
@@ -91,11 +91,14 @@ export class AlertReceiveChannelStore extends BaseStore {
   async loadItem(id: AlertReceiveChannel['id'], skipErrorHandling = false): Promise<AlertReceiveChannel> {
     const alertReceiveChannel = await this.getById(id, skipErrorHandling);
 
+    // @ts-ignore
     this.items = {
       ...this.items,
-      [id]: alertReceiveChannel,
+      [id]: omit(alertReceiveChannel, 'heartbeat'),
```
joeyorlando, line 97: can we store `heartbeat` on the objects within `items? why do we need `alertReceiveChannelToHeartbeat`? I believe a heartbeat _always_ belongs to a single alert receive channel.

maskin25:
@joeyorlando I've added a couple of integration tests for heartbeats and green notification on heartbeat settings update

grafana-plugin/src/models/alert_receive_channel/alert_receive_channel.ts
maskin25: in my mind heartbeat is a separate entity (it even has it's own id and ideally it should have separate endpoint), that's why we have 'models/heartbeat', we need `alertReceiveChannelToHeartbeat` to save appropriate heartbeat Id for each alert receive channel, it's one to one mapping. Yes we can store heartbeat id  just in alert receive channel object, but I did not want to pollute  alert receive channel object

joeyorlando (approved):
LGTM (pending the failing tests on the build)! kudos for the added tests ❤️

grafana-plugin/src/containers/IntegrationContainers/IntegrationHeartbeatForm/IntegrationHeartbeatForm.tsx
```diff
@@ -27,88 +32,93 @@ const IntegrationHeartbeatForm = observer(({ alertReceveChannelId, onClose }: In
...
+                <IntegrationInputField value={heartbeat?.link} showEye={false} isMasked={false} />
+              </Field>
+            </div>
+            <Collapse isOpen={false} label="Instruction">
+              <p className={cx('instruction')}>
+                To send periodic heartbeat alerts from <Emoji text={alertReceiveChannel?.verbal_name || ''} /> to
+                OnCall, do the following:
+                <span
+                  dangerouslySetInnerHTML={{
+                    __html: heartbeat.instruction,
```
iskhakov, line 87: This instructions are moved to the docs. No need to add them here as they will be removed from the API https://grafana.com/docs/oncall/latest/integrations/alertmanager/#configuring-oncall-heartbeats-optional

maskin25: fixed

PR #2564, opened 2023-07-18
mark manual_severity and resolution_note_ts columns on alertgroup table as deprecated

joeyorlando:
# What this PR does

These columns don't appear to be referenced anywhere in the codebase. Marking them as deprecated and will remove them/drop the columns from the table in a subsequent release

## Checklist


engine/apps/user_management/migrations/0013_alter_organization_acknowledge_remind_timeout.py
joeyorlando, line 1: this appears to have been included as a result of `python manage.py makemigrations`

Issue #1735, opened 2023-04-12
ChatOps tab not loading [v1.2.7]


AlbertasG:
ChatOps page does not load or perform any network requests when selecting ChatOps tab. Tested on both Firefox and Chrome.

Edit: Other functionality works: Schedules, Escalation chains, Integrations, etc.

<img width="1512" alt="image" src="https://user-images.githubusercontent.com/27416848/231472069-371b3a3e-232e-4e60-acf6-0559c8b7a904.png">

- [OpenSource ]
- [v1.2.7] 
- [`part:chartops`]
- [bug]

AlbertasG:
Tried with `v1.2.9` also same behaviour.

xom4ek:
Same problem with 1.1.41 
![telegram-cloud-photo-size-2-5251325628442069318-y](https://user-images.githubusercontent.com/29627495/233379013-e74fd777-9c50-4cc0-92ea-10e0ce8c8e61.jpg)

Downgrade to 1.1.40 fix problem
![telegram-cloud-photo-size-2-5251325628442068764-y](https://user-images.githubusercontent.com/29627495/233379071-92c82574-a6fc-4020-a4cb-1b62ed3f5ddd.jpg)

baracoder:
Still occurs on 1.2.41

teodosii:
@baracoder Hi there! Can you post your response for this request? `[your-instance]/api/plugin-proxy/grafana-oncall-app/api/internal/v1/features/`? I just checked what the ChatOps tab renders and if you don't have any chatops feature enabled (such as Slack/Telegram), then that will not show up there, most likely resulting in an empty display

baracoder:
@teodosii I totally missed the feature toggle :see_no_evil:   - the slack feature was not enabled. 
```
["grafana_cloud_connection","grafana_cloud_notifications"]
```
With the feature enabled, it works

Ukochka:
Seems this bug is not relevant anymore.

teodosii:
We will now be showing this banner with links to slack/telegram docs if the env vars are set to false 
![image](https://github.com/grafana/oncall/assets/40542072/1d4a186b-7ec5-456f-b068-aae4bcc80b68)

PR #2566, opened 2023-07-18
Alert with information has been added to chatOPs tab

Ukochka:
# What this PR does
If FEATURES Slack and Telegram are not enabled for OSS, User will see nothing at the chatops tab. Alert has been added for this situation
https://github.com/grafana/oncall/issues/1735


grafana-plugin/src/pages/settings/tabs/ChatOps/ChatOps.tsx
```diff
@@ -105,6 +114,9 @@ const TabsContent = (props: TabsContentProps) => {
   const { activeTab } = props;
   const store = useStore();
 
+  console.log('FEATURES', store.features);
+  console.log('HAS SLACK', store.hasFeature(AppFeature.Slack));
+  console.log('HAS TELEGRAM', store.hasFeature(AppFeature.Telegram));
```
joeyorlando, line 119:
```suggestion
```

grafana-plugin/src/pages/settings/tabs/ChatOps/ChatOps.tsx
```diff
@@ -44,9 +45,17 @@ class ChatOpsPage extends React.Component<ChatOpsProps, ChatOpsState> {
...
   render() {
     const { activeTab } = this.state;
+    const { store } = this.props;
 
     return (
       <div className={cx('root')}>
+        {!store.hasFeature(AppFeature.Slack) && !store.hasFeature(AppFeature.Telegram) && (
+          <div>
+            <Alert severity="warning" title="No chatops found">
+              Chatops is disabled because no chat integration is configured. Please check docs for open source!
```
joeyorlando, line 55: Should we make a mention to "docs for open source"? This would also be shown on cloud instances. What about "Please check the documentation here" ("here" could be a link pointing people to the docs).

joeyorlando (changes requested):
I would recommend adding 1 or 2 unit tests here to assert that this alert shows/doesn't show based on the `!store.hasFeature(AppFeature.Slack) && !store.hasFeature(AppFeature.Telegram)` logic.

grafana-plugin/src/pages/settings/tabs/ChatOps/ChatOps.tsx
Ukochka: @Matvey-Kuk these features are only for OSS, or am I mistaken?

teodosii (approved):

grafana-plugin/src/pages/settings/tabs/ChatOps/ChatOps.tsx
teodosii: Added link for both telegram and slack as each separate page mentions the needed env variable to enable the given chatops feature

teodosii:
I don't think adding a unit test here brings any value whatsoever. The reason is mostly that the asserted condition is too trivial and there's just too much mocking needed to actually make the test run for just a simple test.

PR #2571, opened 2023-07-18
Fix slack channels sync

Ferril:
# What this PR does
- Fixes issue with slack channels sync periodic tasks when we get slack rate limit exception.
- Adds check for active task id to avoid starting multiple tasks for one slack team.

Collecting channels for slack for some teams causes rate limit exception, which causes the task to restart and start collecting slack channels from the beginning. This PR adds new paginated api call and refactors the slack channel sync task to continue collect data after rate limit from the step before it was raised using `cursor` value from the slack response.

## Checklist

### DIFF
--- a/engine/apps/slack/tasks.py
+++ b/engine/apps/slack/tasks.py
@@ -1,6 +1,8 @@
 import logging
 import random
+from typing import Optional
 
+from celery import uuid as celery_uuid
 from celery.utils.log import get_task_logger
 from django.apps import apps
 from django.conf import settings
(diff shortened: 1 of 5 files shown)


engine/apps/slack/tasks.py
```diff
@@ -580,6 +613,20 @@ def populate_slack_channels_for_team(slack_team_identity_id):
...
+            # save collected channels ids to cache and restart the task with the current pagination cursor
+            cache.set(collected_channels_key, collected_channels, timeout=3600)
+            delay = random.randint(1, 3) * 60
+            logger.warning(
+                f"'conversations.list' slack api error: rate_limited. SlackTeamIdentity pk: {slack_team_identity_id}. "
+                f"Delay populate_slack_channels_for_team task for {delay//60} min."
+            )
+            start_populate_slack_channels_for_team(slack_team_identity_id, delay, cursor)
+        else:
+            # delete excess channels
```
vadimkerr, line 626: Consider adding `assert collected_channels` just before deleting. It might be redundant, but this way it's harder to accidentally wipe out all channels when making changes to this task.

vadimkerr (approved):

Issue #2589, opened 2023-07-19
Add public documentation for Shift Swap Requests


joeyorlando:
### What would you like to see!

**Deliverable**

Add public documentation which clearly explains:
- what Shift Swap Requests (SSR) are. maybe give some example scenarios of where they could come in handy
- explain the create SSR workflow
- explain the "Take" SSR workflow

### Product Area

Schedules

joeyorlando:
Also, be sure to update the Insights logs part of the documentation

PR #2597, opened 2023-07-19
shift swap requests model + CRUD endpoints

### LINKED ISSUE #2587: Add the ability to CRUD Shift Swap Requests

**Deliverables**
1. A data model that supports the feature set as outlined in the latest design doc/mockups.
2. I should be able to CRUD a `ShiftSwapRequest` object via the internal API. 

**Non-deliverables for this task**
The following will be done in subsequent tasks.

1. Accepting open shift swap requests
2. ChatOps notifications
3. "deadlines"
4. UI work
 
**Resources**

- [Design doc](https://docs.google.com/document/d/1BLIWkfd5fNKVzQ342wU7mQ9IwXiEvt-xCmJ-GFldlKw/edit?usp=sharing)
- [Mockups](https://www.figma.com/file/sJ4RvHvnUq6AZAiHYYiB5j/Grafana-OnCall?type=design&node-id=8116%3A136607&mode=design&t=5IA3tWes8OoxBM1w-1) 

### Product Area

Schedules

### Anything else to add?

_No response_

joeyorlando:
# What this PR does

This PR should allow us to start working on _most_ of the remaining tasks for this feature set.
- Adds a basic `ShiftSwapRequest` model + CRUD endpoints. 
- Adds a `POST /api/internal/v1/shift_swaps/<id>/take` endpoint which allows a benefactor to take a request (only when certain conditions about the ssr are met)

Closes #2587 

## Checklist

- [x] Documentation added (or `pr:no public docs` PR label added if not required) will be done in #2589
- [x] `CHANGELOG.md` updated (or `pr:no changelog` PR label added if not required) (will update once we ship the finalized feature set)


engine/apps/api/permissions.py
```diff
@@ -265,7 +265,9 @@ def has_object_permission(self, request: AuthenticatedRequest, view: ViewSetOrAP
             for permission_class, actions in rbac_object_permissions.items():
                 if action in actions:
                     return permission_class.has_object_permission(request, view, obj)
-            return False
+
+            # Note: if an endpoint is not found within the rbac_object_permissions dictionary,
+            # that means object permissions are not relevant to this endpoint. Return True (authorized)
```
joeyorlando, line 270: I guess when we've used `rbac_object_permissions` on a `ViewSet` before, we always specify _all_ the endpoints in this object. See [here](https://github.com/grafana/oncall/pull/2597/files#diff-0520dcc9ebfb340cb8390d8fc135fb45364c323abe5eb6bc508e140b58d38ac4R42-R46) for what I'm trying to achieve. tldr; only the beneficiary should be able to PUT/PATCH/DELETE an SSR

engine/common/tests/test_ordered_model.py
```diff
@@ -8,6 +8,8 @@
 
 
 class TestOrderedModel(OrderedModel):
+    __test__ = False
+
```
joeyorlando, line 12: fixes a warning I saw when running the tests locally

engine/apps/webhooks/__init__.py
joeyorlando, line 1: for some reason, when trying to run the tests locally, I was getting the following errors w/o this file present:
```bash
❯ make test
COMPOSE_PROFILES=engine,oncall_ui,rabbitmq,grafana,mysql,redis DB=mysql BROKER_TYPE=rabbitmq docker compose -f docker-compose-developer.yml run --rm oncall_engine_commands pytest --ds=settings.ci-test
[+] Building 0.0s (0/0)
[+] Building 0.0s (0/0)
================================================================================================ test session starts =================================================================================================
platform linux -- Python 3.11.3, pytest-7.3.1, pluggy-1.2.0
django: settings: settings.ci-test (from option)
rootdir: /etc/app
configfile: tox.ini
plugins: django-4.5.2, factoryboy-2.5.1, Faker-19.1.0
collected 2493 items / 3 errors

======================================================================================================= ERRORS =======================================================================================================
_______________________________________________________________________ ERROR collecting apps/webhooks/tests/test_alert_group_status_change.py _______________________________________________________________________
ImportError while importing test module '/etc/app/apps/webhooks/tests/test_alert_group_status_change.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/local/lib/python3.11/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
E   ModuleNotFoundError: No module named 'tests.test_alert_group_status_change'
____________________________________________________________________________ ERROR collecting apps/webhooks/tests/test_trigger_webhook.py ____________________________________________________________________________
ImportError while importing test module '/etc/app/apps/webhooks/tests/test_trigger_webhook.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/local/lib/python3.11/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
E   ModuleNotFoundError: No module named 'tests.test_trigger_webhook'
________________________________________________________________________________ ERROR collecting apps/webhooks/tests/test_webhook.py ________________________________________________________________________________
ImportError while importing test module '/etc/app/apps/webhooks/tests/test_webhook.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/local/lib/python3.11/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
E   ModuleNotFoundError: No module named 'tests.test_webhook'
```

Makefile
```diff
@@ -121,9 +127,11 @@ install-precommit-hook: install-pre-commit
...
 
 test:  ## run backend tests
-# always use settings.ci-test django settings file when running the tests
-# if we use settings.dev it's very possible that some fail just based on the settings alone
-	$(call run_engine_docker_command,pytest --ds=settings.ci-test)
+	$(call run_backend_tests)
+
+test-dev:  ## very similar to `test` command, but allows you to pass arbitray args to pytest
+           ## for example, `make test-dev ARGS="--last-failed --pdb"
+	$(call run_backend_tests,$(ARGS))
```
joeyorlando, line 134: I wanted to be able to do something like this, without having to manually modify this file:
```bash
$ make test-dev ARGS="apps/api/tests/test_shift_swaps.py --last-failed --pdb"
```

engine/apps/shift_swaps/models/shift_swap_request.py
```diff
@@ -0,0 +1,140 @@
...
+    deleted_at = models.DateTimeField(null=True)
+
+    schedule = models.ForeignKey(
+        to="schedules.OnCallSchedule", null=False, on_delete=models.CASCADE, related_name="shift_swap_requests"
+    )
+
+    time_zone = models.CharField(max_length=100)
+
+    swap_start = models.DateTimeField()
+    swap_end = models.DateTimeField()
```
joeyorlando, line 65: @matiasb based on your experience working w/ schedules/overrides. is there a benefit to storing datetimes as timezone aware/naive?

an alternative to what I did here is not have a `time_zone` column and simply have that as a required field when creating/updating objects, and we make `swap_start` and `swap_end` timezone aware before persisting them

engine/apps/shift_swaps/models/shift_swap_request.py
```diff
@@ -0,0 +1,140 @@
...
+    )
+
+    OPEN = "open"
+    TAKEN = "taken"
+    PAST_DUE = "past_due"
+    DELETED = "deleted"
+
+    def __str__(self) -> str:
+        return (
+            f"{str(self.schedule.name)} {str(self.beneficiary.username)} {str(self.swap_start)} - {str(self.swap_end)}"
```
vadimkerr, line 85: calling `str` seems to be excessive here, `f` strings should handle that internally

engine/apps/shift_swaps/models/shift_swap_request.py
```diff
@@ -0,0 +1,140 @@
...
+    )
+    # benefactor - the person taking on shift workload from the beneficiary
+    benefactor = models.ForeignKey(
+        to="user_management.User", null=True, on_delete=models.CASCADE, related_name="taken_shift_swap_requests"
+    )
+
+    OPEN = "open"
+    TAKEN = "taken"
+    PAST_DUE = "past_due"
+    DELETED = "deleted"
```
vadimkerr, line 81: Maybe use an enum here?

engine/settings/base.py
```diff
@@ -65,6 +65,7 @@
 FEATURE_INBOUND_EMAIL_ENABLED = getenv_boolean("FEATURE_INBOUND_EMAIL_ENABLED", default=False)
 FEATURE_PROMETHEUS_EXPORTER_ENABLED = getenv_boolean("FEATURE_PROMETHEUS_EXPORTER_ENABLED", default=False)
 FEATURE_WEBHOOKS_2_ENABLED = getenv_boolean("FEATURE_WEBHOOKS_2_ENABLED", default=True)
+FEATURE_SHIFT_SWAPS_ENABLED = getenv_boolean("FEATURE_SHIFT_SWAPS_ENABLED", default=True)
```
vadimkerr, line 68: Shouldn't this be disabled by default?

```suggestion
FEATURE_SHIFT_SWAPS_ENABLED = getenv_boolean("FEATURE_SHIFT_SWAPS_ENABLED", default=False)
```

vadimkerr (approved):
Great work, LGTM!

engine/apps/shift_swaps/models/shift_swap_request.py
matiasb: FWIW, since `USE_TZ` setting is set to `True` all datetime fields should be already timezone aware (using `UTC` timezone per the `TIME_ZONE` setting). But I think we don't really trust on this (I guess because the fields were not originally tz-aware?), and that's why there is a timezone field for shifts, for example. And then, we are saving a timestamp value without any tz conversion as an UTC-aware datetime working as a naive value, since if there is an associated timezone value, we assume the tz of that datetime value is based on that one, defaulting to `UTC` otherwise (see `convert_dt_to_schedule_timezone` in the `CustomOnCallShift` model). So we are in an intermediate scenario between using tz-aware and naive datetimes.

I think we should eventually migrate everything to fully use timezone aware datetimes because it should make things simpler/more readable when dealing with multiple TZs. Not really sure if introducing a different way to interpret datetime values will make things more complex or not (ie. we will need to know when to convert or not a datetime value?).

If I understand correctly, passing a tz-aware datetime here in a non-UTC tz should be converting the datetime value to UTC and store that in the DB, right? (so we will be working with converted to UTC datetime values later in our logic involving this).

engine/apps/webhooks/__init__.py
mderynck: Odd we didn't see this error pop up earlier, I guess I run my tests via IDE so that somehow avoids it.

engine/apps/shift_swaps/models/shift_swap_request.py
joeyorlando: > I think we should eventually migrate everything to fully use timezone aware datetimes because it should make things simpler/more readable when dealing with multiple TZs

➕ ➕ 

> If I understand correctly, passing a tz-aware datetime here in a non-UTC tz should be converting the datetime value to UTC and store that in the DB, right

correct. it appears everything is persisted in the db in UTC. Django must handle the tz conversions when reading/writing:

Screenshot 2023-07-20 at 19 38 15


```python3
>>> from apps.shift_swaps.models import ShiftSwapRequest
>>> from apps.user_management.models import User
>>> from apps.schedules.models import OnCallSchedule
>>> u = User.objects.first()
>>> s = OnCallSchedule.objects.first()
>>> ssr_utc = ShiftSwapRequest.objects.create(beneficiary=u, schedule=s, time_zone="Europe/Luxembourg", swap_start="2023-07-20T23:00:00Z", swap_end="2023-07-20T23:30:00Z")
>>> ssr_lux = ShiftSwapRequest.objects.create(beneficiary=u, schedule=s, time_zone="Europe/Luxembourg", swap_start="2023-07-20T23:00:00-04:00", swap_end="2023-07-20T23:30:00-04:00")
>>> ssr = ShiftSwapRequest.objects.create(beneficiary=u, schedule=s, time_zone="Europe/Luxembourg", swap_start="2023-07-20 23:00:00", swap_end="2023-07-20 23:30:00")
>>> ssr_utc.swap_end
'2023-07-20T23:30:00Z'
>>> ssr.swap_end
'2023-07-20 23:30:00'
>>> ssr_lux.swap_end
'2023-07-20T23:30:00-04:00'
```

joeyorlando: actually, come to think of it, maybe we can get rid of this column here and just use the schedule's `time_zone`, since a shift swap will always have a schedule related to it. wdyt?

matiasb: FYI, `time_zone` information is not "persisted":

```
>>> ssr_lux.swap_end
'2023-07-20T23:30:00-04:00'
>>> ssr_lux.refresh_from_db()
>>> ssr_lux.swap_end
datetime.datetime(2023, 7, 21, 3, 30, tzinfo=<UTC>)
```

everything is stored in `UTC` in the DB, and next time you fetch from DB, you don't know which was the original tz/datetime value that was provided (which is not necessarily an issue, just for clarification).

> actually, come to think of it, since a shift swap will always have a schedule related to it, we can probably get rid of this column, and just use the schedule's time_zone (maybe add a time_zone @property on this model for easier access).

Hm... I don't think we can use the schedule TZ for this because (as things work today for schedules/shifts, as mentioned in the previous comment) that would mean the datetime you get from the DB was specified using that TZ (which may not be the case?), and directly setting that TZ as the `swap_start` or `swap_end` values will potentially give unexpected results.

So... I think we could keep this information as UTC-aware datetimes (Django will do the right conversion if a TZ-aware datetime is given), and work with these fields always knowing/expecting UTC timestamps, or we could keep the existing approach where if a TZ is provided, save that value in the model and use datetime values as naive ones (this is the intermediate "awareness" scenario that we have I was referring to).

I think I would prefer the first option in this case (unless there are any objections about moving to a full Django tz aware datetimes gradually), although it diverges from what we have elsewhere (and we lose information about the original TZ in which the datetimes were defined).

Makes sense?

engine/common/timezones.py
```diff
@@ -20,11 +19,3 @@ def raise_exception_if_not_valid_timezone(timezone, exception_class=BadRequest):
...
     if not is_valid_timezone(timezone):
         raise exception_class(detail="Invalid timezone")
-
-
-class TimeZoneField(serializers.CharField):
-    def _validator(self, value: str):
-        raise_exception_if_not_valid_timezone(value, serializers.ValidationError)
-
-    def __init__(self, **kwargs):
-        super().__init__(validators=[self._validator], **kwargs)
```
joeyorlando, line 30: moved to `engine/common/api_helpers/custom_fields.py` with the rest of the custom DRF serializer fields.

engine/common/tests/test_timezones.py
```diff
@@ -42,56 +41,3 @@ class MyCustomException(APIException):
...
-        with pytest.raises(serializers.ValidationError, match="This field is required"):
-            serializer = MySerializer(data={})
-            serializer.is_valid(raise_exception=True)
-
-        try:
-            serializer = MySerializer(data={"tz": "UTC"})
-            serializer.is_valid(raise_exception=True)
-            assert serializer.validated_data["tz"] == "UTC"
-        except Exception:
-            pytest.fail()
```
joeyorlando, line 97: relocated to a different test file, `engine/common/tests/test_custom_fields.py`, w/ other tests related to custom DRF serializer fields

engine/apps/shift_swaps/models/shift_swap_request.py
joeyorlando: @matiasb see [this commit](https://github.com/grafana/oncall/pull/2597/commits/399ede4767bcc4fdab1f9f1308bb8735643ca8af). 

based on our discussion, clients will be _forced_ to specify timezone when creating/updating `swap_start` and/or `swap_end`. I used ISO-8601 as a starting point, which allows specifying time zone as either UTC (ie. `Z` at the end) or UTC offset (ex. `+00:00` or `-07:00). However, ISO-8601 also allows not specifying tz (ex. 2023-07-20 12:00:00), I got rid of this.

There's a new custom DRF serializer field, [`TimeZoneAwareDatetimeField`](https://github.com/grafana/oncall/pull/2597/files#diff-27f90cb3c156840972f72d16e217d9808cd80ec2f7783188100d33d534040abbR167-R206), which handles all the input validation + formatting for us.

It might be best to check out the test cases:
- [serializer field test cases](https://github.com/grafana/oncall/pull/2597/files#diff-9f7a24d79d62efdf8bf51c93c9ca8a2074f1ac4f019f9e8fb1b46fe6477008bbR64-R102)
- API related test cases ([create](https://github.com/grafana/oncall/pull/2597/files#diff-de6c9e0686c9bd63dabc4b2824a0f2a381d799ad35add03705892b1ddcdce67fR187-R242) and [update](https://github.com/grafana/oncall/pull/2597/files#diff-de6c9e0686c9bd63dabc4b2824a0f2a381d799ad35add03705892b1ddcdce67fR307-R359))

let me know what you think.

engine/apps/shift_swaps/models/shift_swap_request.py
```diff
@@ -0,0 +1,151 @@
...
+    """
+    the person who is relieved from (part of) their shift(s)
+    """
+
+    benefactor = models.ForeignKey(
+        to="user_management.User", null=True, on_delete=models.CASCADE, related_name="taken_shift_swap_requests"
+    )
+    """
+    the person taking on shift workload from the beneficiary
+    """
```
joeyorlando, line 89: these docstrings may look a bit odd, but they follow [PEP-224](https://peps.python.org/pep-0224/) (Attribute Docstrings). VS Code gives you a nice tooltip w/ this info:
Screenshot 2023-07-20 at 23 44 38

engine/apps/api/serializers/shift_swap.py
```diff
@@ -0,0 +1,73 @@
...
+            "status",
+            "schedule",
+            "swap_start",
+            "swap_end",
+            "description",
+            "beneficiary",
+            "benefactor",
+        ]
+        read_only_fields = [
+            "status",
```
iskhakov, line 41: are these fields readonly too?
```suggestion
            "id",
            "created_at",
            "updated_at",
            "status",
```

joeyorlando: these three fields are marked `read_only=True`, inside of their fields above. I _think_ that should infer the same meaning, no?

matiasb (approved):
LGTM

engine/apps/shift_swaps/models/shift_swap_request.py
```diff
@@ -0,0 +1,151 @@
...
+
+class ShiftSwapRequestManager(models.Manager):
+    def get_queryset(self):
+        return ShiftSwapRequestQueryset(self.model, using=self._db).filter(deleted_at=None)
+
+    def hard_delete(self):
+        return self.get_queryset().hard_delete()
+
+
+class ShiftSwapRequest(models.Model):
```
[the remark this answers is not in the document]
joeyorlando: that's a good idea 👍 

I guess technically we _could_ get rid of the entire `shift_swaps` django app and move everything to the `schedules` one. However, I think the API endpoints should be a separate resource and not nested under schedules (ex. `/api/internal/v1/schedules/<schedule_id>/shift_swaps/<id>/take`)

Issue #2447, opened 2023-07-06
Alert Groups table filters are not persisted when open alert group and return to the list


iskhakov:
### What went wrong?

**What happened**:
  - Filters got reset, because the query on alert group page doesn't persist filters. Possibly introduced by this [change](https://github.com/grafana/oncall/pull/2155/files#diff-69480f166a0bbcdec27b309ac1abb9f2bacb99290c631ab6c5efc0e7382fe215R21) 

**What did you expect to happen**:
  - Filters should remain the same 

https://github.com/grafana/oncall/assets/2262529/2ef095ac-d0ba-4a69-a428-e51b879d7d39

### How do we reproduce it?

1. Open Grafana OnCall and go to Alert Groups
2. Set filter `mine=True`
3. Click on one of the titles of Alert Group to open it
4. Click Arrow Back to go back to the list of Alert Groups
5. Filter is reset

### Grafana OnCall Version

v1.3.6

### Product Area

Alert Flow & Configuration

### Grafana OnCall Platform?

I use Grafana Cloud

### User's Browser?

Chrome, Safari

fculpo:
Same here, Oncall OSS since 1.3.7 I believe

PR #2605, opened 2023-07-20
added all queries to incident page

Ukochka:
# What this PR does
When you click 'Back' from Alert Group page user returns to the Alert Gorups page with prev filters

https://github.com/grafana/oncall/issues/2447


maskin25 (approved):

joeyorlando (commented):
is it worthwhile adding some sort of test to assert that this is the case? maybe a unit test to start? right now we don't have a way to ensure this expected behaviour doesn't regress.

PR #2606, opened 2023-07-20
Remove unnecessary team checks

vadimkerr:
# What this PR does

Removes unnecessary team checks enforced by public API, as they seem to be outdated and not aligned with the web UI and docs.

From public [docs](https://grafana.com/docs/oncall/latest/user-and-team-management/#manage-teams-in-grafana-oncall):
> Resources from different teams can be connected with one another. For instance, you can create an integration in one team, set up multiple routes for the integration, and utilize escalation chains from other teams. Users, schedules, and outgoing webhooks from other teams can also be included in the escalation chain.

## Checklist

### DIFF
--- a/engine/apps/public_api/serializers/schedules_calendar.py
+++ b/engine/apps/public_api/serializers/schedules_calendar.py
@@ -35,17 +35,7 @@ class ScheduleCalendarSerializer(ScheduleBaseSerializer):
         }
 
     def validate_shifts(self, shifts):
-        # Get team_id from instance, if it exists, otherwise get it from initial data.
-        if self.instance and self.instance.team:
-            team_id = self.instance.team.public_primary_key
-        else:
-            # Terraform sends empty string instead of None. In this case change team_id value to None.
-            team_id = self.initial_data.get("team_id") or None
-
         for shift in shifts:
-            shift_team_id = shift.team.public_primary_key if shift.team else None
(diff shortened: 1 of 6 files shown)


engine/apps/public_api/serializers/schedules_calendar.py
```diff
@@ -35,17 +35,7 @@ class Meta:
...
-        if self.instance and self.instance.team:
-            team_id = self.instance.team.public_primary_key
-        else:
-            # Terraform sends empty string instead of None. In this case change team_id value to None.
-            team_id = self.initial_data.get("team_id") or None
-
         for shift in shifts:
-            shift_team_id = shift.team.public_primary_key if shift.team else None
-            if shift_team_id != team_id:
-                raise BadRequest(detail="Shifts must be assigned to the same team as the schedule")
```
iskhakov, line 48: Is there a case when shift belongs to the other team from the schedule?

vadimkerr: I don't think it's possible now but should be possible after merging this PR. I tried to create a shift for team X and a schedule for team Y using that shift, there were no issues with it. So unless we have a particular reason to enforce team checks here, I'd propose to remove them, so it's consistent with other models/resources.

mderynck (approved):

PR #2611, opened 2023-07-21
add rbac_permissions to current-user response + rbac_enabled to current-org response

joeyorlando:
# What this PR does

unblocks https://github.com/grafana/oncall-mobile-app/issues/152

## Checklist


joeyorlando:
@imtoori you can use these as such:

`GET /api/internal/v1/current_team/` (not aptly named imo, actually represents the currently authenticated user's _organization_)

```json
{
    "pk": "OJLC8P3BPHABN",
    "name": "Self-Hosted Organization",
    "slack_team_identity": null,
    "slack_channel": null,
    "rbac_enabled": true,
    "is_resolution_note_required": false,
    "env_status": {
        "telegram_configured": true,
        "phone_provider": {
            "configured": true,
            "test_sms": false,
            "test_call": true,
            "verification_call": false,
            "verification_sms": true
        }
    },
    "banner": {
        "body": null,
        "title": null
    }
}
```

`GET /api/internal/v1/user/` - I believe you're already using this one

```json
{
    ...<other stuff>
    "role": 0,
    "rbac_permissions": [
        {
            "action": "grafana-oncall-app.alert-groups:read"
        },
        {
            "action": "grafana-oncall-app.alert-groups:write"
        },
        ...<big list of permissions>
    ]
}
```

vadimkerr (approved):
LGTM

vadimkerr:
@joeyorlando I think `current_team` is used as the endpoint path because `organization` was called `team` in the past. I think a good solution would be to register the same view under an additional `current_organization` (or just `organization`) path, and update plugin & mobile app accordingly. It's probably out of the scope of this PR, but if it's a one line change consider adding it here so the mobile app uses the correct route right away.

joeyorlando:
> I think a good solution would be to register the same view under an additional current_organization (or just organization) path, and update plugin & mobile app accordingly

I like this idea, especially because it's easier to deprecate `current_team` if the mobile app isn't already consuming it. will give it a go 👍

engine/apps/api/urls.py
```diff
@@ -106,5 +108,3 @@
     path(r"login/<backend>/", auth.overridden_login_slack_auth, name="slack-auth"),
     path(r"complete/<backend>/", auth.overridden_complete_slack_auth, name="complete-slack-auth"),
 ]
-
-urlpatterns += router.urls
```
joeyorlando, line 110: this line is essentially a duplicate of [line 69](https://github.com/grafana/oncall/pull/2611/files#diff-d87ee8df9c3c7eed2c4b729461f3831c5318b69907c58cdc0a2e566c170c2e9bR69) and can be removed:
Screenshot 2023-07-21 at 15 49 00

joeyorlando:
@grafana/grafana-oncall-frontend can I get some 👀 ?

the main changes here are renaming `teamStore` -> `organizationStore`. the naming of `teamStore` was a bit misleading as it represents an organization, just needed to be updated (`grafanaTeamStore` === "team store"). Also updated the API calls to move calls to `/current_team/` to `/organization/`. `/current_team/` will be deprecated in the near future.

grafana-plugin/src/models/organization/organization.ts
joeyorlando, line 1: moved from `TeamStore`. There were a few things on `TeamStore` that were also removed as they were no longer used:
```tsx
  @observable
  redirectingToProperTeam = false;

  @observable.shallow
  teams: { [id: number]: Team[] } = {};
```

grafana-plugin/src/models/organization/organization.ts
```diff
@@ -0,0 +1,30 @@
...
+
+import { Organization } from './organization.types';
+
+export class OrganizationStore extends BaseStore {
+  @observable
+  currentOrganization?: Organization;
+
+  constructor(rootStore: RootStore) {
+    super(rootStore);
+    this.path = '/organization/';
```
joeyorlando, line 15: this was changed from `/current_team/`, see [this change](https://github.com/grafana/oncall/pull/2611/files#diff-d87ee8df9c3c7eed2c4b729461f3831c5318b69907c58cdc0a2e566c170c2e9bR72-L72) on the backend

grafana-plugin/src/models/user/user.types.ts
```diff
@@ -36,7 +35,6 @@ export interface User {
     slack_login: string;
   } | null;
   post_onboarding_entry_allowed: any;
-  teams: Team[];
```
joeyorlando, line 39: this is not used

grafana-plugin/src/state/rootBaseStore/index.ts
```diff
@@ -80,29 +80,29 @@ export class RootBaseStore {
...
-  slackStore: SlackStore = new SlackStore(this);
-  slackChannelStore: SlackChannelStore = new SlackChannelStore(this);
-  heartbeatStore: HeartbeatStore = new HeartbeatStore(this);
-  scheduleStore: ScheduleStore = new ScheduleStore(this);
-  userGroupStore: UserGroupStore = new UserGroupStore(this);
-  alertGroupStore: AlertGroupStore = new AlertGroupStore(this);
-  resolutionNotesStore: ResolutionNotesStore = new ResolutionNotesStore(this);
-  apiTokenStore: ApiTokenStore = new ApiTokenStore(this);
-  globalSettingStore: GlobalSettingStore = new GlobalSettingStore(this);
-  filtersStore: FiltersStore = new FiltersStore(this);
```
joeyorlando, line 105: typing here is redundant. these attributes are properly typed w/o the explicit typing.

vadimkerr (approved):

PR #2619, opened 2023-07-23
`apps.get_model` -> `import`

vadimkerr:
# What this PR does

Remove [`apps.get_model`](https://docs.djangoproject.com/en/3.2/ref/applications/#django.apps.apps.get_model) invocations and use inline `import` statements in places where models are imported within functions/methods to avoid circular imports.

I believe `import` statements are more appropriate for most use cases as they allow for better static code analysis & formatting, and solve the issue of circular imports without being unnecessarily dynamic as `apps.get_model`. With `import` statements, it's possible to:

- Jump to model definitions in most IDEs
- Automatically sort inline imports with `isort`
- Find import errors faster/easier (most IDEs highlight broken imports)
- Have more consistency across regular & inline imports when importing models

This PR also adds a flake8 rule to ban imports of `django.apps.apps`, so it's harder to use `apps.get_model` by mistake (it's possible to ignore this rule by using `# noqa: I251`). The rule is not enforced on directories with migration files, because `apps.get_model` is often used to get a historical state of a model, which is useful when writing migrations ([see this SO answer for more details](https://stackoverflow.com/a/37769213)). So `apps.get_model` is considered OK in migrations (even necessary in some cases).

## Checklist


engine/apps/alerts/incident_log_builder/incident_log_builder.py
```diff
@@ -32,8 +31,7 @@ def get_log_records_list(self, with_resolution_notes=False):
         return all_log_records_sorted
 
     def _get_log_records_for_after_resolve_report(self):
-        EscalationPolicy = apps.get_model("alerts", "EscalationPolicy")
-        AlertGroupLogRecord = apps.get_model("alerts", "AlertGroupLogRecord")
+        from apps.alerts.models import AlertGroupLogRecord, EscalationPolicy
```
vadimkerr, line 34: example of `isort` doing its thing

engine/apps/slack/scenarios/distribute_alerts.py
```diff
@@ -525,7 +524,8 @@ class CustomButtonProcessStep(AlertGroupActionsMixin, scenario_step.ScenarioStep
     REQUIRED_PERMISSIONS = [RBACPermission.Permissions.CHATOPS_WRITE]
 
     def process_scenario(self, slack_user_identity, slack_team_identity, payload):
-        CustomButtom = apps.get_model("alerts", "CustomButton")
```
vadimkerr, line 528: CustomButto**m**

engine/apps/twilioapp/phone_provider.py
```diff
@@ -260,11 +266,10 @@ def _default_twilio_api_client(self):
     def _default_twilio_number(self):
         return live_settings.TWILIO_NUMBER
 
-    def _twilio_sender(self, sender_type, to):
+    def _twilio_sender(self, sender_model, to):
```
vadimkerr, line 269: I had to change this a bit to make sure models are not referenced using strings

engine/tox.ini
```diff
@@ -5,6 +5,7 @@ extend-exclude = */migrations/*
 ban-relative-imports = parents
 banned-modules =
   jinja2.Template = Use apply_jinja_template instead
+  django.apps.apps = Import models directly instead
```
vadimkerr, line 8: Adding a rule to ban `django.apps.apps`

engine/apps/grafana_plugin/tests/test_app_config.py
```diff
@@ -2,7 +2,7 @@
 from unittest.mock import patch
 
 import pytest
-from django.apps import apps
+from django.apps import apps  # noqa: I251
```
vadimkerr, line 5: This is the only place in the entire repo using `noqa: I251`, because `django.apps.apps` is used for something different than `get_model` here.

Matvey-Kuk (changes requested):
A few years ago “apps.get_model” was a best-practice in the Django world which was helping us to avoid circular dependencies. What changed and how should we address circular deps in the future?

joeyorlando (approved):
nice! 🙂 

while you're at it, may want to consider the [7 instances ](https://github.com/search?q=repo%3Agrafana%2Foncall-private+apps.get_model&type=code) of this in the private repo.

engine/tox.ini
iskhakov: get_model is required in the migrations, would this rule allow that?

engine/apps/alerts/grafana_alerting_sync_manager/grafana_alerting_sync.py
```diff
@@ -357,7 +356,8 @@ def _create_contact_point_from_payload(self, payload, datasource) -> "GrafanaAle
 
         receiver_config = self._get_receiver_config(is_grafana_datasource, payload)
 
-        GrafanaAlertingContactPoint = apps.get_model("alerts", "GrafanaAlertingContactPoint")
+        from apps.alerts.models import GrafanaAlertingContactPoint
```
iskhakov, line 359: should we move the imports to the top?

iskhakov (approved):
Nice job Vadim, I agree it would improve the quality of life

engine/tox.ini
vadimkerr: yes, migration directories are ignored [here](https://github.com/grafana/oncall/pull/2619/files#diff-e1f096d06b7cdece59d58be9c27f1aee24c15c02c84b0063b35edcfb0d6fb528R4)

engine/apps/alerts/grafana_alerting_sync_manager/grafana_alerting_sync.py
vadimkerr: Yes, I think imports should be at the top, I'll try to come up with an additional PR for this (+adding a rule to enforce that in CI)

PR #2624, opened 2023-07-24
Remove checks that slow down plugin load and cause "Initializing plugin..."

iskhakov:
# What this PR does

* Removes "Initializing plugin.." message during load
* Removes black screen when plugin loads
* Removes wait for syncs between OnCall and Grafana
* Deprecates GET /status, POST /sync, GET /sync in favour of single POST /status

## Checklist


grafana-plugin/src/state/rootBaseStore/index.ts
```diff
@@ -211,25 +207,17 @@ export class RootBaseStore {
         }
       }
     } else {
-      const syncDataResponse = await PluginState.syncDataWithOnCall(this.onCallApiUrl);
```
joeyorlando, line 214: we can remove everything associated with `PluginState.syncDataWithOnCall`, it's only used here. As well as `PluginState.pollOnCallDataSyncStatus` and `PluginState._pollOnCallDataSyncStatus`.

I _believe_ this also means we can mark the `POST /api/internal/v1/plugin/sync` endpoint as deprecated as well. Which means the whole `PluginSyncView` is deprecated 🥳  (double check me on this, but I think this is the case)

joeyorlando (approved):

engine/apps/auth_token/auth.py
```diff
@@ -87,6 +87,12 @@ def authenticate_credentials(self, token_string: str, request: Request) -> Tuple
         except InvalidToken:
             raise exceptions.AuthenticationFailed("Invalid token.")
 
+        # TODO: decide if we want to add this check instead of doing in plugin sync
```
mderynck, line 90: I don't think we should check these here, auth is currently checking Grafana -> OnCall is allowed.  I don't think that should fail if Grafana <- OnCall isn't working.

engine/apps/grafana_plugin/views/status.py
```diff
@@ -1,19 +1,81 @@
...
+        # Check if current user is in OnCall database
+        user_is_present_in_org = PluginAuthentication.is_user_from_request_present_in_organization(
+            request, organization
+        )
+        # If user is not present in OnCall database, set token_ok to False, which will trigger reinstall
+        if not user_is_present_in_org:
+            token_ok = False
+
+        # Start task to refresh organization data in OnCall database with Grafana
+        plugin_sync_organization_async.apply_async((organization.pk,))
```
mderynck, line 62: The task itself does not set `api_token_status` so we should set it to PENDING here so that if multiple users are loading the plugin at the same time we don't build up unnecessary tasks.

iskhakov: add same as it was in POST /sync

grafana-plugin/src/containers/PluginConfigPage/PluginConfigPage.tsx
```diff
@@ -206,15 +211,26 @@ const PluginConfigPage: FC<OnCallPluginConfigPageProps> = ({
...
-      <ConfigurationForm onSuccessfulSetup={triggerDataSyncWithOnCall} defaultOnCallApiUrl={processEnvOnCallApiUrl} />
+      <ConfigurationForm onSuccessfulSetup={triggerUpdatePluginStatus} defaultOnCallApiUrl={processEnvOnCallApiUrl} />
     );
   } else {
     // plugin is fully connected and synced
+    const pluginLink = (
+      <LinkButton href={`/a/grafana-oncall-app/`} variant="primary">
+        Open Grafana OnCall
+      </LinkButton>
+    );
```
joeyorlando, line 222: 👍

grafana-plugin/src/state/rootBaseStore/index.ts
```diff
@@ -130,56 +128,63 @@ export class RootBaseStore {
...
    * Finally, try to load the current user from the OnCall backend
    */
   async setupPlugin(meta: OnCallAppPluginMeta) {
-    this.appLoading = true;
     this.initializationError = null;
     this.onCallApiUrl = meta.jsonData?.onCallApiUrl;
 
+    if (!FaroHelper.faro) {
+      FaroHelper.initializeFaro(this.onCallApiUrl);
+    }
```
joeyorlando, line 159: what's this for?

iskhakov: [Tracing](https://grafana.com/oss/faro/)

joeyorlando: right. I just wasn't sure why this line was all of a sudden being added here but now I see it was removed [here](https://github.com/grafana/oncall/pull/2624/files#diff-c75a16456e6f502bc9eeb90ae7c381d244ded9046ac876186f5834734947b566L270). makes sense 👍

teodosii (approved):

PR #2625, opened 2023-07-24
Refactoring schedule final events for reusability

matiasb:
First step towards reusing `schedule.final_events` where current / upcoming schedule shifts information is needed (eg. escalations, shift notifications, etc)

### DIFF
--- a/engine/apps/schedules/tests/test_quality_score.py
+++ b/engine/apps/schedules/tests/test_quality_score.py
@@ -190,7 +190,7 @@ def test_get_schedule_score_weekdays(
     assert response.json() == {
         "total_score": 86,
         "comments": [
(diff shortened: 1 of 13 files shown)


engine/apps/schedules/tests/test_quality_score.py
```diff
@@ -190,7 +190,7 @@ def test_get_schedule_score_weekdays(
     assert response.json() == {
         "total_score": 86,
         "comments": [
-            {"type": "warning", "text": "Schedule has gaps (29% not covered)"},
+            {"type": "warning", "text": "Schedule has gaps (28% not covered)"},
```
matiasb, line 193: This value slightly changed because the (start, end) limits for the period taken into account has changed minimally (by ~1 second) when refactoring the backend implementation.

Ferril (approved):
LGTM

Issue #2172, opened 2023-06-13
update isort and black tools to use Python 3.11 as their "target version"


joeyorlando:
Currently [they're configured](https://github.com/grafana/oncall/blob/dev/engine/pyproject.toml#L4-L11) to use Python 3.9 as the target version. We've since upgraded to Python 3.11, and should update this.

Additionally, we should consider centralizing the configuration for `isort`, `black`, and `flake8`. Currently we duplicate the configuration to handle `engine`, `dev/scripts`, and `tools/pagerduty-migrator`.

this exists as of v1.2.43

Matvey-Kuk:
@joeyorlando please format according to requirements or close.

vadimkerr:
Done in https://github.com/grafana/oncall/pull/2641 ([isort](https://github.com/grafana/oncall/pull/2641/files#diff-262ba8727e625e7c53accd7457e696deb98c4328d153085b25166f83900714eeR5), [black](https://github.com/grafana/oncall/pull/2641/files#diff-262ba8727e625e7c53accd7457e696deb98c4328d153085b25166f83900714eeR10))

PR #2641, opened 2023-07-25
Minor formatting changes

vadimkerr:
# What this PR does

- Updates `black` and `flake8` to latest
- Removes `F541` from flake8 ignore (`F541 f-string is missing placeholders`)
- Enables ["float to top" option](https://pycqa.github.io/isort/docs/configuration/options.html#float-to-top) for `isort`


engine/settings/prod_without_db.py
```diff
@@ -13,7 +15,6 @@ def on_uwsgi_worker_exit():
     # Only works under uwsgi web server environment
     pass
```
joeyorlando, line 16: this isn't at all related to your PR, but I don't quite understand this block between lines 3 - 16. I don't think it's _actually_ doing anything. I _think_ the original intent may have been this:
```python
import uwsgi
from prometheus_client import multiprocess

def on_uwsgi_worker_exit():
  multiprocess.mark_process_dead(os.getpid())

uwsgi.atexit = on_uwsgi_worker_exit # note how this is unindented, if it's nested within the function definition it doesn't do anything as this function is never called
```

joeyorlando (approved):
nice! LGTM 👍 

There is #2172 which is vaguely related to this. Goal of that one though is to bump the `target` version for isort/black to match our version of Python (3.11). May be out of scope for this PR here.

PR #2652, opened 2023-07-27
Fix race condition in plugin auth

iskhakov:
# What this PR does

## Checklist


engine/apps/grafana_plugin/helpers/gcom.py
```diff
@@ -50,7 +50,8 @@ def check_gcom_permission(token_string: str, context) -> GcomToken:
             name="allow_plugin_organization_signup", defaults={"boolean_value": True}
         )[0].boolean_value
         if allow_signup:
-            organization = Organization.objects.create(
+            # Get org from db or create a new one
+            organization = Organization.objects.get_or_create(
```
joeyorlando, line 54: what exactly causes the race condition? a comment could be useful

PR #2659, opened 2023-07-27
Add instruction on removing nullable fields from Django models

vadimkerr:
Adds an instruction on removing nullable fields without downtime.


dev/README.md
```diff
@@ -427,3 +427,69 @@ backwards compatible
 
 See [django-migration-linter checklist](https://github.com/3YOURMIND/django-migration-linter/blob/main/docs/incompatibilities.md)
 for the common mistakes and best practices
+
+### Removing a nullable field from a model
```
vadimkerr, line 431: Change preview [here](https://github.com/grafana/oncall/blob/vadimkerr/remove-field-readme/dev/README.md#removing-a-nullable-field-from-a-model)

dev/README.md
```diff
@@ -427,3 +427,69 @@ backwards compatible
...
+
+    ```python
+    operations = [
+        migrations.SeparateDatabaseAndState(
+            state_operations=[
+                # Put the RemoveField operation generated by Django here
+            ],
+            database_operations=[],
+        ),
+    ]
```
iskhakov, line 453: Do we need to ignore the migration linter here?

dev/README.md
```diff
@@ -427,3 +427,69 @@ backwards compatible
 
 See [django-migration-linter checklist](https://github.com/3YOURMIND/django-migration-linter/blob/main/docs/incompatibilities.md)
 for the common mistakes and best practices
+
+### Removing a nullable field from a model
+
+> This only works for nullable fields (fields with `null=True` in the field definition).
+>
+> DO NOT USE THIS APPROACH FOR NON-NULLABLE FIELDS, IT WILL BREAK THINGS!
```
iskhakov, line 435: what if the field has the default value on the database level? We had a check for that in migration linter

dev/README.md
vadimkerr: Don't need to, check out [this migration](https://github.com/grafana/oncall/blob/ad0ed6714ac9993582e78b50d855aaf69df17619/engine/apps/alerts/migrations/0027_remove_alertreceivechannel_restricted_at_from_state.py#L10-L20)

dev/README.md
vadimkerr: If a field is non-nullable and has a default value on the DB level, it should be fine to use the same approach. This section is only about nullable fields, so I don't want to include this info here. I'll change `IT WILL BREAK THINGS` to `IT CAN BREAK THINGS` so it's more precise.

joeyorlando (approved):
nice! 🔥

dev/README.md
iskhakov: so it means it is reversible?

engine/engine/management/commands/remove_field.py
```diff
@@ -0,0 +1,64 @@
...
+from django.db import connection
+from django.db.migrations import Migration
+from django.db.migrations.autodetector import MigrationAutodetector
+from django.db.migrations.loader import MigrationLoader
+from django.db.migrations.writer import MigrationWriter
+
+from common.migrations.remove_field import RemoveFieldDB, RemoveFieldState
+
+
+class Command(BaseCommand):
```
vadimkerr, line 11: Example: `python manage.py remove_field alerts AlertReceiveChannel restricted_at` (assuming that field exists)

This generates two migrations:

`0031_remove_alertreceivechannel_restricted_at_state.py`

```python
# Generated by Django 3.2.20 on 2023-08-07 20:26

import common.migrations.remove_field
from django.db import migrations


class Migration(migrations.Migration):

    dependencies = [
        ('alerts', '0030_auto_20230731_0341'),
    ]

    operations = [
        common.migrations.remove_field.RemoveFieldState(
            model_name='AlertReceiveChannel',
            name='restricted_at',
        ),
    ]
```

`0032_remove_alertreceivechannel_restricted_at_db.py`

```python
# Generated by Django 3.2.20 on 2023-08-07 20:26

import common.migrations.remove_field
from django.db import migrations


class Migration(migrations.Migration):

    dependencies = [
        ('alerts', '0031_alertreceivechannel_restricted_at'),
    ]

    operations = [
        common.migrations.remove_field.RemoveFieldDB(
            model_name='AlertReceiveChannel',
            name='restricted_at',
            remove_state_migration=('alerts', '0031_remove_alertreceivechannel_restricted_at_state'),
        ),
    ]
```

dev/README.md
vadimkerr: Yes, the first migration of the two is reversible

dev/README.md
vadimkerr: Non-nullable fields are a bit trickier, my intention is to scope this PR purely to nullable fields for now.

joeyorlando (approved):
great work, this looks like awesome tooling to have 😄

PR #2671, opened 2023-07-28
Require basic data loaded for schedule page

iskhakov:
# What this PR does

## Checklist

### DIFF
--- a/grafana-plugin/src/plugin/GrafanaPluginRootPage.tsx
+++ b/grafana-plugin/src/plugin/GrafanaPluginRootPage.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect } from 'react';
+import React, { useEffect, useState } from 'react';
 
 import classnames from 'classnames';
 import dayjs from 'dayjs';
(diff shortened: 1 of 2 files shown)


grafana-plugin/src/plugin/GrafanaPluginRootPage.tsx
```diff
@@ -153,7 +156,7 @@ export const Root = observer((props: AppRootProps) => {
               <Schedules query={query} />
             </Route>
             <Route path={getRoutesForPage('schedule')} exact>
-              <Schedule query={query} />
+              <Schedule query={query} basicDataLoaded={basicDataLoaded} />
```
joeyorlando, line 159: this seems odd to me.. why not just only conditionally render the root `<Switch>` component if `basicDataLoaded` is true?

teodosii: Indeed, this looks a bit odd. I'll change it so that it's at the top level instead, a few other pages also seem to depend on the data loaded from there.

PR #2677, opened 2023-07-28
Apply shift swap requests to schedule events

matiasb:
Reflect swap requests details in schedule events.

### DIFF
--- a/engine/apps/schedules/ical_utils.py
+++ b/engine/apps/schedules/ical_utils.py
@@ -61,7 +61,6 @@ IcalEvents = typing.List[IcalEvent]
 def users_in_ical(
     usernames_from_ical: typing.List[str],
     organization: "Organization",
-    users_to_filter: typing.Optional["UserQuerySet"] = None,
 ) -> typing.Sequence["User"]:
     """
     This method returns a sequence of `User` objects, filtered by users whose username, or case-insensitive e-mail,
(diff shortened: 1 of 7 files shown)


engine/apps/schedules/ical_utils.py
```diff
@@ -378,35 +360,11 @@ def get_oncall_users_for_multiple_schedules(
...
-                current_usernames, _ = get_usernames_from_ical_event(event)
-                usernames.update(current_usernames)
-
-    # Fetch relevant users from the db
-    emails = [username.lower() for username in usernames]
-    users = organization.users.filter(
-        Q(**User.build_permissions_query(RBACPermission.Permissions.SCHEDULES_WRITE, organization))
-        & (Q(username__in=usernames) | Q(email__lower__in=emails))
-    )
-
```
matiasb, line 402: Iterating over schedule events and getting users with an `OR` query doesn't seem better than going the original path (did some testing). Also, when applying swaps, if a user wasn't part of the original schedule, we don't get that user information.

engine/apps/schedules/models/shift_swap_request.py
```diff
@@ -122,7 +127,8 @@ def take(self, benefactor: "User") -> None:
         self.benefactor = benefactor
         self.save()
 
-        # TODO: implement the actual override logic in https://github.com/grafana/oncall/issues/2590
+        # make sure final schedule ical representation is updated
+        refresh_ical_final_schedule.apply_async((self.schedule.pk,))
```
joeyorlando, line 131: ❤️

joeyorlando (approved):

engine/apps/schedules/ical_utils.py
```diff
@@ -61,7 +60,6 @@
 def users_in_ical(
     usernames_from_ical: typing.List[str],
     organization: "Organization",
-    users_to_filter: typing.Optional["UserQuerySet"] = None,
 ) -> typing.Sequence["User"]:
```
joeyorlando, line 63: we can probably change the return type here. I believe it was like this to accommodate the union between a list and a queryset

PR #2710, opened 2023-08-01
Update schedule slack notifications

Ferril:
# What this PR does

Update schedule slack notifications to use schedule final events instead of getting events from iCal

## Checklist


engine/apps/schedules/tests/test_notify_about_empty_shifts_in_schedule.py
```diff
@@ -0,0 +1,170 @@
...
+
+    assert mock_slack_api_call.called
+
+    schedule.refresh_from_db()
+    assert empty_shifts_report_sent_at != schedule.empty_shifts_report_sent_at
+    assert schedule.has_empty_shifts
+
+
+@pytest.mark.django_db
+def test_empty_non_empty_shifts_trigger_notification(
```
matiasb, line 111: I guess this is the original behavior, but wondering if this should be a warning in the schedule UI but not triggering a notification (since there is still an active user for the shift). In any case, something to discuss later.

engine/apps/alerts/tasks/notify_ical_schedule_shift.py
```diff
@@ -183,160 +26,137 @@ def notify_ical_schedule_shift(schedule_pk):
...
-        if calendar is not None:
-            current_shifts_result, current_users_result = get_current_shifts_from_ical(
-                calendar,
-                schedule,
-                overrides_priority,
+    current_shifts = schedule.final_events(now, now, False, False)
+
+    prev_shifts = json.loads(schedule.current_shifts) if not schedule.empty_oncall else []
+    prev_shifts_updated = False
+    # convert prev_shifts to new events format for compatibility with the previous version of this task
```
matiasb, line 51: It will be nice to leave a task for later to drop this conversion. Optional, if you think it makes sense, I would separate this "translation" in a helper function in this module (so it is simpler to follow the task logic ignoring this compatibility thing)

engine/apps/alerts/tasks/notify_ical_schedule_shift.py
```diff
@@ -183,160 +26,137 @@ def notify_ical_schedule_shift(schedule_pk):
...
+    # get days_to_lookup for next shifts
+    if len(new_shifts) != 0:
+        days_to_lookup = (new_shifts[-1]["end"].date() - now.date()).days + 1
+        days_to_lookup = max([days_to_lookup, MIN_DAYS_TO_LOOKUP_FOR_THE_END_OF_EVENT])
+    else:
+        days_to_lookup = MIN_DAYS_TO_LOOKUP_FOR_THE_END_OF_EVENT
+
+    datetime_end = now + datetime.timedelta(days=days_to_lookup)
+
+    next_shifts_unfiltered = schedule.final_events(now, datetime_end, False, False)
```
matiasb, line 113: Not now, but maybe we could consider having only [one call](https://github.com/grafana/oncall/pull/2710/files?diff=unified&w=0#diff-ea1a6a0a4efe49256b2c05a6f516f0978db29859e0400bd4b758abb5919c1118R47) to `final_events`?

engine/apps/alerts/tasks/notify_ical_schedule_shift.py
```diff
@@ -183,160 +26,137 @@ def notify_ical_schedule_shift(schedule_pk):
...
+    # drop events that already started
+    next_shifts = []
+    for next_shift in next_shifts_unfiltered:
+        if now < next_shift["start"]:
+            next_shifts.append(next_shift)
+
+    next_shifts = sorted(next_shifts, key=lambda shift: shift["start"])
```
matiasb, line 120: `final_events` should already return events sorted by `shift["start"]`

engine/apps/alerts/tasks/notify_ical_schedule_shift.py
```diff
@@ -183,160 +26,137 @@ def notify_ical_schedule_shift(schedule_pk):
...
+        for shift in next_shifts[1:]:
+            if shift["start"] == earliest_shift["start"]:
+                upcoming_shifts.append(shift)
+
+    empty_oncall = len(current_shifts) == 0
+    if empty_oncall:
+        schedule.empty_oncall = True
+    else:
+        schedule.empty_oncall = False
+        schedule.current_shifts = json.dumps(current_shifts, default=str)
```
matiasb, line 137: nit, I think this could be changed to something like
```
schedule.empty_oncall = len(current_shifts) == 0
if not schedule.empty_oncall:
    schedule.current_shifts = json.dumps(current_shifts, default=str)
```
?

matiasb (approved):
LGTM

A few comments added.
Great you added more tests covering different notification trigger (or not) scenarios :+1: 
And **really** nice to see the notify task simplification!

engine/apps/alerts/tasks/notify_ical_schedule_shift.py
Ferril: It would be great, but in this case we need to decide how to calculate days_to_lookup

engine/apps/schedules/tests/test_notify_about_gaps_in_schedule.py
```diff
@@ -0,0 +1,433 @@
...
+        "frequency": CustomOnCallShift.FREQUENCY_DAILY,
+        "schedule": schedule_no_gaps,
+    }
+    on_call_shift = make_on_call_shift(
+        organization=organization, shift_type=CustomOnCallShift.TYPE_ROLLING_USERS_EVENT, **data
+    )
+    on_call_shift.add_rolling_users([[user1]])
+    schedule_no_gaps.refresh_ical_file()
+
+    assert schedule_no_gaps.check_gaps_for_next_week() is False
```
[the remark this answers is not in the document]
Ferril: makes sense 👍

engine/apps/schedules/tests/test_notify_about_empty_shifts_in_schedule.py
Ferril: Agree, we should discuss it later. I think, the main point is make customer know that something is wrong with one of their shifts despite the presence of an active user.

engine/apps/schedules/tasks/notify_about_gaps_in_schedule.py
```diff
@@ -79,20 +80,21 @@ def notify_about_gaps_in_schedule(schedule_pk):
         task_logger.info(f"Tried to notify_about_gaps_in_schedule for non-existing schedule {schedule_pk}")
         return
 
-    today = timezone.now().date()
-    gaps = list_of_gaps_in_schedule(schedule, today, today + timezone.timedelta(days=7))
-    schedule.gaps_report_sent_at = today
+    now = timezone.now()
+    events = schedule.final_events(now, now + datetime.timedelta(days=7), with_empty=False)
```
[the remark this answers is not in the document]
Ferril: You are right, we can filter it in the task. It makes sense since we need to filer the output anyway.

I think, if we change using of `check_gaps_for_next_week` we should also change using `check_empty_shifts_for_next_week` since they have pretty similar logic. I would propose to do it in separate PR

matiasb (approved):

PR #2744, opened 2023-08-03
fix make start command when using mysql/postgres as db

### LINKED ISSUE #2492: Unable to start local development using Docker Compose 2.19.0
### What went wrong?

**What happened**:
  - `make start` fails with an error: `service postgres is required by grafana but is disabled. Can be enabled by profiles [postgres]`

  - starting development containers

### How do we reproduce it?

1. Use docker-compose v2.19.1
2. Use `COMPOSE_PROFILES=oncall_ui,redis,grafana,engine`
3. `make init`
4. `make start`

```
COMPOSE_PROFILES=oncall_ui,redis,grafana,engine DB=sqlite BROKER_TYPE=redis SQLITE_DB_FILE=./engine/oncall.db docker compose -f docker-compose-developer.yml up --remove-orphans -d
service postgres is required by grafana but is disabled. Can be enabled by profiles [postgres]
make: *** [start] Error 1
```

### Grafana OnCall Version

v1.3.1

### Product Area

Other

### Grafana OnCall Platform?

Docker

### User's Browser?

_No response_

### Anything else to add?

I suppose it is caused by [this fix](https://github.com/docker/compose/pull/10602) and now `depends_on` is required by default.

As I understand `required` field would be implemented introduced in https://github.com/compose-spec/compose-spec/pull/382 

But now using docker compose v2.18.1 is the only option.

joeyorlando:
# What this PR does

Tested `make start` w/ both `mysql` and `postgres` in `COMPOSE_PROFILES` and things spin up properly.

- Fixes some `yaml` formatting issues introduced [here](https://github.com/grafana/oncall/pull/2728/files#diff-f5d10b03472abe3719098ae8a8855468e92524ebe790c39a34d2c632f3f0486d) in #2728 which were causing the `mysql` container to fail to start up when running `make start`
- Addresses #2492 by specifying `required: false` under `depends_on`, for containers in `docker-compose-developer.yml` which have container(s) which may be conditionally spun-up based on `COMPOSE_PROFILES`. Basically in v2.20.2 they introduced [this change](https://github.com/docker/compose/releases/tag/v2.20.2#:~:text=Add%20support%20of%20depends_on.required%20attribute%20by) which was a breaking change for our setup and preventing us from upgrading to >= 2.19.0.

Closes #2492 

‼️ with this PR you will need to make sure you are running `docker-compose` >= `v2.20.2`, otherwise 👇  ‼️ 

image

## Checklist

### DIFF
--- a/docker-compose-developer.yml
+++ b/docker-compose-developer.yml
@@ -144,12 +144,16 @@ services:
     depends_on:
       postgres:
         condition: service_healthy
+        required: false
       mysql:
         condition: service_healthy
+        required: false
       rabbitmq:
         condition: service_healthy
+        required: false
       redis:
         condition: service_healthy
+        required: false
     profiles:
       - engine
 
@@ -208,7 +212,7 @@ services:
     container_name: mysql
     labels: *oncall-labels
     image: mysql:8.0.32
-    command: >
+    command: >-
       --default-authentication-plugin=mysql_native_password --character-set-server=utf8mb4
       --collation-server=utf8mb4_unicode_ci --max_connections=1024
     restart: always
@@ -236,7 +240,7 @@ services:
     container_name: mysql_to_create_grafana_db
     labels: *oncall-labels
     image: mysql:8.0.32
-    command: >
+    command: >-
       bash -c "mysql -h mysql -uroot -pempty
       -e 'CREATE DATABASE IF NOT EXISTS grafana CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;'"
     depends_on:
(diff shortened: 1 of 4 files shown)


docker-compose-developer.yml
```diff
@@ -276,7 +280,7 @@ services:
     container_name: postgres_to_create_grafana_db
     labels: *oncall-labels
     image: postgres:14.4
-    command: >
```
joeyorlando, line 279: `>-` removes trailing newlines.. I suppose this causes an issue for `mysql`? 

https://stackoverflow.com/a/21699210

joeyorlando:
how to update `docker-compose` to `v2.20.2`:

1. Download the `docker-compose` binary from the v2.20.2 release [here](https://github.com/docker/compose/releases/tag/v2.20.2) (select your OS's architecture)
2. Run the following commands:
```bash
$ mv <docker-compose-binary> ~/.docker/cli-plugins/docker-compose
$ chmod +x ~/.docker/cli-plugins/docker-compose
$ docker-compose --version
Docker Compose version v2.20.2
$ make start # should work
```

(or you might be able to use `brew upgrade docker-compose` on Mac)

vadimkerr (approved):

PR #2748, opened 2023-08-03
Filter out untaken swaps from final schedule and shift notifications

matiasb:
Avoid creating (or notifying) about potential event splits resulting from untaken swap requests.

### DIFF
--- a/engine/apps/schedules/models/on_call_schedule.py
+++ b/engine/apps/schedules/models/on_call_schedule.py
@@ -344,6 +344,7 @@ class OnCallSchedule(PolymorphicModel):
         with_gap: bool = False,
         filter_by: str | None = None,
         all_day_datetime: bool = False,
+        ignore_untaken_swaps: bool = False,
         from_cached_final: bool = False,
     ) -> ScheduleEvents:
         """Return filtered events from schedule."""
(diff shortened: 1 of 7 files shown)


vadimkerr (approved):
LGTM

engine/apps/schedules/models/on_call_schedule.py
```diff
@@ -411,10 +414,16 @@ def final_events(
         datetime_end: datetime.datetime,
         with_empty: bool = True,
         with_gap: bool = True,
+        ignore_untaken_swaps: bool = False,
```
vadimkerr, line 417: Might be worth checking all the usages of `final_events` to make sure other parts are consistent. For example `apps.schedules.ical_utils.list_users_to_notify_from_ical_for_period` looks like it should ignore SSRs as well.

Alternatively, maybe `ignore_untaken_swaps` should be `True` by default?

vadimkerr: There's a check for `swap.is_past_due` [here](https://github.com/grafana/oncall/commit/bb9f647608b2b3aad51e6f5e13e6131843277975#diff-4037b7bbef9fc16d9b541beb3ed46f760916d7cd720847c3123adf7afb5ab4b4R679), so this doesn't affect current on-calls in any way, only future events. So `ignore_untaken_swaps` is only for when dealing with future events.

PR #2759, opened 2023-08-04
Support mobile app notification sounds with custom extensions

vadimkerr:
# What this PR does

Instead of always adding `.aiff` or `.mp3` at the end of notifications sound names depending on the platform (iOS vs Android), add them only if no extension is present already. This should make it possible to use sounds with custom extensions.

## Checklist


engine/apps/mobile_app/models.py
```diff
@@ -175,3 +176,22 @@ class VolumeType(models.TextChoices):
...
+    def get_notification_sound_name(self, message_type: MessageType, platform: Platform) -> str:
+        sound_name = {
+            MessageType.DEFAULT: self.default_notification_sound_name,
+            MessageType.IMPORTANT: self.important_notification_sound_name,
+            MessageType.INFO: self.info_notification_sound_name,
+        }[message_type]
+
+        # If sound name already contains an extension, return it as is
+        if "." in sound_name:
+            return sound_name
```
vadimkerr, line 189: Main change of the PR

engine/apps/mobile_app/models.py
```diff
@@ -10,6 +10,7 @@
 
 from apps.auth_token import constants, crypto
 from apps.auth_token.models import BaseAuthToken
+from apps.mobile_app.tasks import MessageType, Platform
```
joeyorlando, line 13: _nit_ maybe create a new file, `apps/mobile_app/types.py`, since these types aren't specific to the celery tasks

joeyorlando (approved):
LGTM!

engine/apps/mobile_app/models.py
vadimkerr: Good call! Addressed in https://github.com/grafana/oncall/pull/2759/commits/ac895e6bc1521ec92cf9cf7ffde05a4161c5cb5c

Issue #2756, opened 2023-08-04
Not receiving acknowledge reminders in Slack thread


vadimkerr:
### What went wrong?

**What happened**:

I configured the Slack integration to post acknowledgment reminders every hour, but no reminders are posted for acknowledged alert groups.

**What did you expect to happen**:

I expected to see acknowledgment reminders getting posted every 1 hour in the alert group's Slack message thread, while the alert group is acknowledged and not resolved.

### How do we reproduce it?

1. Install the Slack integration & configure it to "Remind every 1 hour and never unack"
2. Send a demo alert
3. Wait for the message to be posted to the configured Slack channel
4. Wait for more than 1 hour and verify no acknowledgment reminder is posted to the thread

### Grafana OnCall Version

r203-v1.3.22

### Product Area

Alert Flow & Configuration, Chatops

### Grafana OnCall Platform?

I use Grafana Cloud

vadimkerr:
This should be fixed in https://github.com/grafana/oncall/pull/2769 and released in [v1.3.24](https://github.com/grafana/oncall/releases/tag/v1.3.24).

PR #2769, opened 2023-08-09
Fix Slack acknowledgment reminders

vadimkerr:
# What this PR does

Fixes a bug with Slack acknowledgment reminders not being sent (+ some refactoring and unit tests).

https://github.com/grafana/oncall/issues/2756

## Checklist


engine/apps/alerts/models/alert_group.py
```diff
@@ -1599,28 +1594,20 @@ def bulk_silence(user: User, alert_groups: "QuerySet[AlertGroup]", silence_delay
...
-                countdown=delay,
-            )
-            self.last_unique_unacknowledge_process_id = unique_unacknowledge_process_id
-            self.save(update_fields=["last_unique_unacknowledge_process_id"])
+        # Check if the "Remind every N hours" setting is enabled
+        countdown = Organization.ACKNOWLEDGE_REMIND_DELAY[self.channel.organization.acknowledge_remind_timeout]
+        if not countdown:
+            return
+
+        self.last_unique_unacknowledge_process_id = celery_uuid()
```
vadimkerr, line 1608: The main change of the PR: instead of using `uuid1` to generate Celery task IDs, use `celery.uuid`.

The method stopped working after we upgraded to [kombu 5.3.0](https://github.com/celery/kombu/releases/tag/v5.3.0) (a transitive dependency of celery). Previously, when `UUID` objects were passed to Celery tasks as arguments, they were converted to `str` and it was safe to compare it with strings. After upgrading to kombu 5.3.0, UUIDs are passed to tasks as `UUID` type, thus comparing them with strings always returns `False`.

matiasb (approved):
LGTM

Issue #2678, opened 2023-07-28
Public API for Shift Swap Requests


joeyorlando:
### What would you like to see!

**Deliverables**
1. I should be able to CRUD a `ShiftSwapRequest` object via the public API. 
2. I should be able to "take" a shift shift swap request via the public API.
3. These endpoints should be well documented in the public documentation

### Product Area

Schedules, Other

PR #2775, opened 2023-08-10
Add shift swaps endpoints to public API

matiasb:
Related to https://github.com/grafana/oncall/issues/2678


engine/apps/public_api/views/shift_swap.py
```diff
@@ -0,0 +1,98 @@
...
+
+    # public API customizations
+    throttle_classes = [UserThrottle]
+
+    def get_queryset(self):
+        schedule_id = self.request.query_params.get("schedule_id", None)
+        beneficiary = self.request.query_params.get("beneficiary", None)
+        benefactor = self.request.query_params.get("benefactor", None)
+        starting_after = self.request.query_params.get("starting_after", None)
+        untaken = self.request.query_params.get("untaken", "false") == "true"
```
vadimkerr, line 37: nit: maybe `taken` would work better here? `untaken=False` double negations can be hard on the brain 😄

engine/apps/public_api/views/shift_swap.py
```diff
@@ -0,0 +1,98 @@
...
+            raise NotFound
+
+    def _get_user(self, field_name: str):
+        """Return user from ID given by field_name.
+
+        Fallback to request.user if no ID is provided.
+        """
+        user_pk = self.request.data.pop(field_name, None)
+        if not user_pk:
+            user = self.request.user
```
vadimkerr, line 82: `self.request.user` is it the user that created an API token? If so, maybe it makes sense to make the `beneficiary` and `benefactor` fields required, without any fallbacks?
I think we don't show API key creators anywhere, so having a fallback could introduce extra confusion.

Screenshot 2023-08-11 at 16 18 48

vadimkerr (approved):
LGTM

engine/apps/public_api/views/shift_swap.py
matiasb: Makes sense :+1: I guess it will be good to discuss end-user/roles API tokens at some point.

PR #2783, opened 2023-08-11
Don't send notifications about past SSRs when turning on info notifications

vadimkerr:
# What this PR does

Fixes a bug when a user is receiving push notifications about past SSRs after turning on the info notifications in the mobile app.

## Checklist

### DIFF
--- a/engine/apps/mobile_app/tasks.py
+++ b/engine/apps/mobile_app/tasks.py
@@ -595,11 +595,11 @@ def _should_notify_user_about_shift_swap_request(
     except MobileAppUserSettings.DoesNotExist:
         return False  # don't notify if the app is not configured
 
-    return (
-        mobile_app_user_settings.info_notifications_enabled  # info notifications must be enabled
-        and user.is_in_working_hours(now, mobile_app_user_settings.time_zone)  # user must be in working hours
-        and not _has_user_been_notified_for_shift_swap_request(shift_swap_request, user)  # don't notify twice
-    )
+    return user.is_in_working_hours(
+        now, mobile_app_user_settings.time_zone
+    ) and not _has_user_been_notified_for_shift_swap_request(  # user must be in working hours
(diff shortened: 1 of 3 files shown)


engine/apps/mobile_app/tests/test_shift_swap_request.py
```diff
@@ -270,6 +270,32 @@ def test_notify_user_about_shift_swap_request(make_organization, make_user, make
...
+):
+    organization = make_organization()
+    beneficiary = make_user(organization=organization)
+    benefactor = make_user(organization=organization)
+    schedule = make_schedule(organization, schedule_class=OnCallScheduleWeb)
+
+    FCMDevice.objects.create(user=benefactor, registration_id="test_device_id")
+    MobileAppUserSettings.objects.create(user=benefactor, info_notifications_enabled=False)
+
+    now = timezone.datetime(2023, 8, 1, 19, 38, tzinfo=timezone.utc)
```
matiasb, line 285: Double-checking, is it ok to have this hardcoded datetime here?

engine/apps/mobile_app/tasks.py
```diff
@@ -595,10 +595,10 @@ def _should_notify_user_about_shift_swap_request(
     except MobileAppUserSettings.DoesNotExist:
         return False  # don't notify if the app is not configured
 
-    return (
-        mobile_app_user_settings.info_notifications_enabled  # info notifications must be enabled
```
matiasb, line 599: If I understand it ok, the idea is to flag the swap request as notified if a notification should have been sent while notifications are disabled, so when enabling we don't try to send it later, right?

matiasb (approved):

engine/apps/mobile_app/tasks.py
vadimkerr: Yes, that's the idea

PR #2791, opened 2023-08-15
Return API URL as part of status

mderynck:
# What this PR does
Provide API URL when returning status to inform plugin or mobile app if it should be talking to a different backend in case of migration.

## Checklist

### DIFF
--- a/engine/apps/grafana_plugin/views/status.py
+++ b/engine/apps/grafana_plugin/views/status.py
@@ -14,7 +14,7 @@ from common.api_helpers.mixins import GrafanaHeadersMixin
 
 
 class StatusView(GrafanaHeadersMixin, APIView):
-    permission_classes = PluginTokenVerified
+    permission_classes = (PluginTokenVerified,)
 
     def post(self, request: Request) -> Response:
         """
--- a/engine/apps/grafana_plugin/views/status.py
+++ b/engine/apps/grafana_plugin/views/status.py
@@ -14,7 +14,7 @@ from common.api_helpers.mixins import GrafanaHeadersMixin
 
 
 class StatusView(GrafanaHeadersMixin, APIView):
-    permission_classes = (PluginTokenVerified,)
+    permission_classes = PluginTokenVerified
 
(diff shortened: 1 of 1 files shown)


engine/apps/mobile_app/auth.py
```diff
@@ -43,9 +42,4 @@ def authenticate_credentials(self, token_string: str) -> Tuple[Optional[User], O
         except InvalidToken:
             return None, None
 
-        if auth_token.organization.is_moved:
```
mderynck, line 46: This will no longer be needed if the mobile app updates its backend URL based on the status endpoint.

engine/apps/grafana_plugin/views/status.py
```diff
@@ -96,7 +98,7 @@ def get(self, _request: Request) -> Response:
                 "is_installed": is_installed,
                 "token_ok": token_ok,
                 "allow_signup": allow_signup,
-                "is_user_anonymous": self.grafana_context["IsAnonymous"],
+                "is_user_anonymous": self.grafana_context.get("IsAnonymous", _request.user is None),
```
joeyorlando, line 101: this endpoint is marked as deprecated, is that still the case?

engine/apps/grafana_plugin/views/sync.py
joeyorlando, line 1: nice!

engine/apps/mobile_app/auth.py
joeyorlando: what about for older versions of the mobile app that don't do that?

joeyorlando (approved):
lgtm 👍 just some minor questions

engine/apps/mobile_app/auth.py
mderynck: This was the incomplete first attempt at fixing the issue, so it was not really useful. Since we paused the migration of orgs with mobile users we will need make sure the mobile app version that uses this status endpoint is rolled out before we start migration again.

PR #2794, opened 2023-08-15
Improve template to handle empty group labels

Konstantinov-Innokentii:
– Fix error with missing "groupLabels" or "alertname"
– Support rendering alerts from migrated legacy alertmanagers

### DIFF
--- a/engine/config_integrations/alertmanager.py
+++ b/engine/config_integrations/alertmanager.py
@@ -22,26 +22,27 @@ acknowledge_condition = None
 
 # Web
 web_title = """\
-{%- set groupLabels = payload.groupLabels.copy() -%}
-{%- set alertname = groupLabels.pop('alertname') | default("") -%}
+{%- set groupLabels = payload.get("groupLabels", {}).copy() -%}
+{%- set alertname = groupLabels.pop('alertname', "")  -%}
 
 
-[{{ payload.status }}{% if payload.status == 'firing' %}:{{ payload.numFiring }}{% endif %}] {{ alertname }} {% if groupLabels | length > 0 %}({{ groupLabels|join(", ") }}){% endif %}
-"""  # noqa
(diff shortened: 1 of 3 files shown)


engine/config_integrations/alertmanager.py
```diff
@@ -22,26 +22,27 @@
...
+{%- set severity = groupLabels.severity -%}
 
-{% set severity = payload.groupLabels.severity -%}
 {% if severity %}
 {%- set severity_emoji = {"critical": ":rotating_light:", "warning": ":warning:" }[severity] | default(":question:") -%}
 Severity: {{ severity }} {{ severity_emoji }}
 {% endif %}
 
-{%- set status = payload.status | default("Unknown") %}
+{%- set status = payload.get("status", "Unknown") %}
```
iskhakov, line 42: just curious, what is the difference here? I think previous version is more straightforward and clear for non-python developers

engine/config_integrations/alertmanager.py
```diff
@@ -22,26 +22,27 @@
...
 {%- set status_emoji = {"firing": ":fire:", "resolved": ":white_check_mark:"}[status] | default(":warning:") %}
 Status: {{ status }} {{ status_emoji }} (on the source)
-{% if status == "firing" %}
+{% if status == "firing" and payload.numFiring %}
```
iskhakov, line 45: should we add check for numResolved?

iskhakov (approved):
overall looks good to me, see minor comments

engine/config_integrations/alertmanager.py
Konstantinov-Innokentii: I'm checking the numFiring here to not to render empty ':', since alerts from migrated legacy alertmanager have no numFiring field. I'm still thinking how to do it better.

engine/config_integrations/alertmanager.py
Konstantinov-Innokentii: Just aligned it with other `.get`usages. ( It was used in previous templates also)

joeyorlando (commented):
is there any way to add some sort of integration tests for these templates?

Issue #2679, opened 2023-07-28
Introduce shift swap request notification follow-ups


joeyorlando:
### What would you like to see!

This is a follow-up task to #2630 and #2588. **Note**: both of those issues should be merged first before working on this.

Those tasks were focused on sending out the initial notifications when a shift swap request is first created.

**Deliverables**
In this task, we should implement a strategy to send intelligent follow-up Slack, and mobile app, notifications.

First we need to come to a consensus on a notification follow-up schedule (ex. First notification <= 1 week before the first shift starts. Second X time before shift starts, etc).

### Product Area

Chatops, Mobile App, Schedules

joeyorlando:
@vadimkerr and I decided to first merge #2630 and #2588 before working on follow-ups because Slack and the mobile-app will have a lot of shared logic for calculating when to send these notifications

PR #2798, opened 2023-08-15
Shift swap request Slack follow-ups

vadimkerr:
# What this PR does

Add Slack follow-up messages for shift swap requests:

Screenshot 2023-08-15 at 20 19 49

https://github.com/grafana/oncall/issues/2679

## Checklist

### DIFF
--- a/engine/apps/schedules/tasks/shift_swaps/slack_followups.py
+++ b/engine/apps/schedules/tasks/shift_swaps/slack_followups.py
@@ -47,7 +47,7 @@ def _get_shift_swap_request_pks_in_followup_window(now: datetime.datetime) -> li
 
 
 @shared_dedicated_queue_retry_task(
-    autoretry_for=(Exception,), retry_backoff=True, max_retries=1 if settings.DEBUG else None
+    autoretry_for=(Exception,), retry_backoff=True, max_retries=1 if settings.DEBUG else 10
 )
 def send_shift_swap_request_slack_followup(shift_swap_request_pk: int) -> None:
     try:
--- a/engine/apps/schedules/tasks/shift_swaps/slack_followups.py
+++ b/engine/apps/schedules/tasks/shift_swaps/slack_followups.py
@@ -11,6 +11,7 @@ from common.custom_celery_tasks import shared_dedicated_queue_retry_task
 
 task_logger = get_task_logger(__name__)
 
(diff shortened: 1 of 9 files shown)


engine/settings/prod_without_db.py
```diff
@@ -77,8 +77,10 @@ def on_uwsgi_worker_exit():
     "apps.schedules.tasks.notify_about_empty_shifts_in_schedule.start_notify_about_empty_shifts_in_schedule": {
         "queue": "default"
     },
-    "apps.schedules.tasks.shift_swaps.slack_messages.post_shift_swap_request_creation_message": {"queue": "default"},
```
joeyorlando, line 80: good catch 👍

engine/apps/schedules/tasks/shift_swaps/slack_followups.py
```diff
@@ -0,0 +1,92 @@
...
+        if not _has_followup_been_sent(shift_swap_request):
+            send_shift_swap_request_slack_followup.delay(shift_swap_request.pk)
+            _mark_followup_sent(shift_swap_request)
+
+
+def _get_shift_swap_requests_in_followup_window(now: datetime.datetime) -> list[ShiftSwapRequest]:
+    """Get all SSRs that are in the followup window."""
+
+    shift_swap_requests_in_notification_window = []
+    for shift_swap_request in ShiftSwapRequest.objects.filter(benefactor__isnull=True, swap_start__gt=now):
```
joeyorlando, line 46: _nit_ on the query. We already have this same query [here](https://github.com/grafana/oncall/blob/dev/engine/apps/mobile_app/tasks.py#L514). wdyt about moving it as a method on the `QuerySet` (or `Manager`, not sure which one is more appropriate)? Something like, `get_open_requests`.

This way if we need to change the definition of this query, it only needs to be done once.

joeyorlando (approved):
LGTM! Just one minor nit comment.

engine/apps/schedules/tasks/shift_swaps/slack_followups.py
vadimkerr: Good call, addressed in https://github.com/grafana/oncall/pull/2798/commits/7fe839dd39ceb9706f52ecf1cda17257fbf14555

PR #2800, opened 2023-08-16
Add fallback if escalate_alert_group task was not scheduled

iskhakov:
# What this PR does

## Checklist

### DIFF
--- a/engine/apps/alerts/escalation_snapshot/escalation_snapshot_mixin.py
+++ b/engine/apps/alerts/escalation_snapshot/escalation_snapshot_mixin.py
@@ -246,13 +246,19 @@ class EscalationSnapshotMixin:
             self.build_raw_escalation_snapshot() if not self.pause_escalation else self.raw_escalation_snapshot
         )
         task_id = celery_uuid()
-
-        AlertGroup.objects.filter(pk=self.pk).update(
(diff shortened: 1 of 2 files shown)


engine/apps/alerts/escalation_snapshot/escalation_snapshot_mixin.py
```diff
@@ -246,13 +246,19 @@ def start_escalation_if_needed(self, countdown=START_ESCALATION_DELAY, eta=None)
...
-        AlertGroup.objects.filter(pk=self.pk).update(
+        AlertGroup.objects.get(pk=self.pk).update(
             active_escalation_id=task_id,
             is_escalation_finished=False,
             raw_escalation_snapshot=raw_escalation_snapshot,
         )
-        escalate_alert_group.apply_async((self.pk,), countdown=countdown, immutable=True, eta=eta, task_id=task_id)
+        try:
+            escalate_alert_group.apply_async((self.pk,), countdown=countdown, immutable=True, eta=eta, task_id=task_id)
+        except Exception:
```
joeyorlando, line 256: what sort of `Exception`s do we tend to see here? Would it be safer to catch a specific exception(s), rather than all?

PR #2821, opened 2023-08-17
Add direct paging events when checking logs for license

matiasb:


devinrcheevers:
Would this affect the monthly active users we report to billing?

joeyorlando (approved):

Issue #2781, opened 2023-08-11
Escalation step "Continue escalation if current time is in range" doesn't work correctly


Ferril:
### What went wrong?

**What happened**:
The escalation step's wording says that it uses UTC timezone. In PUT request on updating time range and in escalation logs I can see that the time range for this step is different than I selected in the step settings.
<img width="1042" alt="Screenshot 2023-08-11 at 13 16 27" src="https://github.com/grafana/oncall/assets/22730923/b4248d09-6345-43fe-a33a-c5f3adc8d886">

<img width="417" alt="Screenshot 2023-08-11 at 13 16 55" src="https://github.com/grafana/oncall/assets/22730923/3c6b3b5d-8f84-4172-b678-8a97f7c2bdc5">

### How do we reproduce it?

1. Set non-UTC local timezone
2. Create escalation step "Continue escalation if current time is in range"
3. Select time range in local timezone, that doesn't cover current time
4. Send demo alert in any integration using escalation chain with this step
5. Check escalation timeline

### Grafana OnCall Version

r213-v1.3.23

### Product Area

Alert Flow & Configuration

### Grafana OnCall Platform?

I use Grafana Cloud

matiasb:
FWIW, besides the UI and logs mentioning expected UTC times, [public API](https://grafana.com/docs/grafana-cloud/alerting-and-irm/oncall/oncall-api-reference/escalation_policies/) also expects UTC values for from/to.

As noted, the conversion is being done in the frontend before pushing the data, so I think we should disable this TZ conversion there and be consistent with the expectations everywhere else (or alternatively, make sure to display in the UI which is the timezone assumed before the conversion).

FTR, this is also reported [here](https://github.com/grafana/support-escalations/issues/7069).

PR #2826, opened 2023-08-18
Prevent conversion of escalation start/end times to UTC

teodosii:
# What this PR does

~~Adds a boolean `convertToUTC` to the `TimeRange` component.~~ 

- Discard conversion to UTC of strings within `TimeRange` component

For this use case on escalation chains since the user inputted fields are strings they will not be further converted to UTC and instead left as it is, which previously resulted in setting a wrong start/end combination.

https://github.com/grafana/oncall/issues/2781

## Checklist


vadimkerr (approved):
Backend changes LGTM

Issue #2799, opened 2023-08-15
Add ability for users to download a csv report from iCal


eshward23:
### What would you like to see!

Currently, API can be used to pull schedules for all users over a period like a month. It works only for web schedules and not for iCal. More specifically, the endpoint is currently only active for web schedules. It will return HTTP 400 for schedules defined via Terraform or iCal.
Related documentation: https://grafana.com/docs/oncall/latest/oncall-api-reference/schedules/#export-a-schedules-final-shifts
Customer is using iCal to track shift schedules and unable to pull a report. It is needed for how payments are made to the team and currently using excel script reading the iCal. 

### Product Area

Schedules

Matvey-Kuk:
Hi, thank you for the report! The core team doesn't have it in a backlog, so not sure when it will be implemented. Let's keep it open to see if smb else needs this feature.

Matvey-Kuk:
Hey, sorry for closing it without the context. With the latest update, this script should work for iCal schedules: https://grafana.com/docs/oncall/latest/oncall-api-reference/schedules/#export-a-schedules-final-shifts

PR #2863, opened 2023-08-22
Enable shifts export endpoint for all schedule types

matiasb:
Related to https://github.com/grafana/oncall/issues/2799


joeyorlando (approved):
LGTM 👍

Isn't there a caveat with Terraform/iCal schedules that if you change a rotation, it will change the historical shifts as well?

Issue #2843, opened 2023-08-21
Switching pages on the Integration table doesn't fully change the page


teodosii:
### What went wrong?

**What happened**:
-

**What did you expect to happen**:
-

### How do we reproduce it?

1. Go to Integrations page
2. Switch the page (e.g. go to 2)
3. You'll see a quick list refresh but it takes you back to page 1 (sometimes even displaying the data from page 2 but showing that's page 1 or vice-versa)

Expected behavior is that you don't see any flicker and the table should reflect whatever values are for that page you're on.

### Grafana OnCall Version

all

### Product Area

Other

### Grafana OnCall Platform?

None

PR #2869, opened 2023-08-23
Tweaks for integrations

teodosii:
# What this PR does

This PR adds a few tweaks to the integrations page

- prevent calling `/counters` on the integration page, instead call it with `counters/[id]`
- call to `/counters` will be done only once when we load all the integrations and prevent calling it again once we switch pages. Prior to this change every page resulted in a call to `/counters`, and due to the fact the request took on average `3-5s` it was being stalled by the browser on subsequent calls.
- Call to `/connected_contact_points` is now conditioned by the integration type
- No more page flicking when going back from integration page to the list of integrations
- No more page flickering when switching pages in table view if there's more requests ongoing
- Prevent discarding filters when going back from Integration view to the table view
- Fixed routes not moving up/down
- Fixed routes flickering when adding/removing a route

These changes should result in slightly loader times because `/counters` was time-consuming request, and less UX issues.

- described at https://github.com/grafana/oncall/issues/2843 but incompletely

## Checklist


grafana-plugin/src/models/user/user.ts
```diff
@@ -111,35 +111,35 @@ export class UserStore extends BaseStore {
...
+        {}
+      ),
+    };
+
+    this.searchResult = {
+      count,
+      results: results.map((item: User) => item.pk),
+    };
+
+    return response;
```
teodosii, line 142: this part is not touched, I only removed the wrapped `new Promise (...)` and instead returned the response which is a `Promise` too.

grafana-plugin/src/pages/integration/Integration.tsx
```diff
@@ -697,6 +692,17 @@ class Integration extends React.Component<IntegrationProps, IntegrationState> {
       })
       .finally(() => this.setState({ isLoading: false }));
   }
+
+  async loadContactPointsMaybe(id: AlertReceiveChannel['id']) {
+    const { alertReceiveChannelStore } = this.props.store;
```
teodosii, line 697: the page will not wait until contact points are loaded, but instead render them when they're available so that we don't slow down the page load.

grafana-plugin/src/pages/integration/Integration.tsx
```diff
@@ -697,6 +692,17 @@ class Integration extends React.Component<IntegrationProps, IntegrationState> {
...
+
+    if (IntegrationHelper.isGrafanaAlerting(alertReceiveChannelStore.items[id])) {
+      // this will be delayed and not awaitable so that we don't delay the whole page load
+      return await alertReceiveChannelStore.updateConnectedContactPoints(id).catch(noop);
```
maskin25, line 701: `.catch(noop);` I think this needs to be removed, because we don't call contact points endpoint if integration is non-GrafanaAlerting

grafana-plugin/src/pages/integration/Integration.tsx
```diff
@@ -697,6 +692,17 @@ class Integration extends React.Component<IntegrationProps, IntegrationState> {
       })
       .finally(() => this.setState({ isLoading: false }));
   }
+
+  async loadContactPointsMaybe(id: AlertReceiveChannel['id']) {
```
maskin25, line 696: let's maybe use more general name like `loadExtraData`, and inside that method we will check which data do we need depending on integration type and load it

maskin25 (approved):

grafana-plugin/src/models/filters/filters.ts
```diff
@@ -17,6 +17,9 @@ export class FiltersStore extends BaseStore {
   @observable.shallow
   public values: { [page: string]: FiltersValues } = {};
 
+  @observable
+  public isLoading = true;
```
teodosii, line 21: easier to track now if the filters are loading or not

grafana-plugin/src/pages/integration/Integration.tsx
```diff
@@ -697,6 +705,17 @@ class Integration extends React.Component<IntegrationProps, IntegrationState> {
       })
       .finally(() => this.setState({ isLoading: false }));
   }
+
+  async loadExtraDataMaybe(id: AlertReceiveChannel['id']) {
```
maskin25, line 709: just `loadExtraData` is better

grafana-plugin/src/models/filters/filters.ts
maskin25: so don't see the reason of having  `isLoading` flag

maskin25 (approved):

grafana-plugin/src/state/rootBaseStore/index.ts
```diff
@@ -79,6 +80,9 @@ export class RootBaseStore {
   @observable
   incidentsPage: any = this.initialQuery.p ? Number(this.initialQuery.p) : 1;
 
+  @observable
+  currentPage: { [key: string]: number } = {};
```
maskin25, line 84: we can use `src/pages/index.tsx`  -> `getMatchedPage` to get current page

grafana-plugin/src/pages/integrations/Integrations.tsx
```diff
@@ -486,19 +497,23 @@ class Integrations extends React.Component<IntegrationsProps, IntegrationsState>
...
-    return alertReceiveChannelStore.updatePaginatedItems(integrationsFilters).then(() => {
-      this.setState({ page: 1 });
-      LocationHelper.update({ p: 1 }, 'partial');
-    });
+    const newPage = isOnMount ? store.getCurrentPage(PAGE.Integrations) : 1;
+
+    return alertReceiveChannelStore
+      .updatePaginatedItems(integrationsFilters, newPage, false, () => this.invalidateRequestFn(newPage))
+      .then(() => {
+        store.setCurrentPage(PAGE.Integrations, newPage);
```
maskin25, line 514: did not get why do we need `store.setCurrentPage` to navigate we use ` history.push(`${PLUGIN_ROOT}/foo` or react router links, to get current page we can use `getMatchedPage`

grafana-plugin/src/state/rootBaseStore/index.ts
teodosii: this gets you the page number, not the page string

grafana-plugin/src/pages/integrations/Integrations.tsx
teodosii: `store.setCurrentPage` is not for navigation, is for updating the store to know which was the last page we hit (the page num, not the page string)

PR #2913, opened 2023-08-30
Exclude the latest alert groups from escalation finished check

Ferril:
# What this PR does
Exclude the latest alert groups from escalation finished check to give them time for building escalation snapshot

https://github.com/grafana/irm/issues/4910

## Checklist

### DIFF
--- a/engine/apps/alerts/tasks/check_escalation_finished.py
+++ b/engine/apps/alerts/tasks/check_escalation_finished.py
@@ -114,7 +114,7 @@ def check_escalation_finished_task() -> None:
     """
     from apps.alerts.models import AlertGroup
 
-    now = timezone.now() - datetime.timedelta(seconds=30)
+    now = timezone.now() - datetime.timedelta(minutes=5)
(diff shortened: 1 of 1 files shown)


engine/apps/alerts/tasks/check_escalation_finished.py
```diff
@@ -114,7 +114,7 @@ def check_escalation_finished_task() -> None:
     """
     from apps.alerts.models import AlertGroup
 
-    now = timezone.now()
+    now = timezone.now() - datetime.timedelta(seconds=30)
```
joeyorlando, line 117: should we make this interval a bit bigger (maybe 30 minutes) just to be on the safe side in case the celery tasks take a bit longer to execute?

Ferril: We may miss something important. Maximum time between creation and failing auditing that I've seen was 12 sec (alert group looked healthy and was escalating in parallel). I'd propose increasing this time maximum to 5 minute, wdyt?

joeyorlando (approved):

PR #2919, opened 2023-08-30
Show build log in the local env

Matvey-Kuk:
Added "progress: plain" to show build log while building local env. Otherwise it's just hanging with "RUN pip install" and you have no idea what's going on.


joeyorlando (approved):

joeyorlando:
@Matvey-Kuk out of curiosity, how long is the image build time for the engine image? We recently switched from a debian based image to alpine (to address some CVEs). Alpine is much slower from what I remember, but I'm curious how slow.

Matvey-Kuk:
@joeyorlando it's about 7 minutes on my Mac Air M2. I guess it depends on the internet mostly

Issue #2915, opened 2023-08-30
Build 'When I am on-call' for web UI


raphael-batte:
### What would you like to see!

![image](https://github.com/grafana/oncall/assets/102970268/067106e8-d807-4c58-8155-7cec19378359)
![image](https://github.com/grafana/oncall/assets/102970268/01349a02-6d5f-444e-8a8d-984e9bc35e27)

### Product Area

Schedules

raphael-batte:
I think it makes sense to add more clear message context in the block when it will be possible from BE. 

At the moment in user is not on-call we are providing message 'There are no schedules relevant to user' in all the cases.
![image](https://github.com/grafana/oncall/assets/102970268/c2e5e865-64e6-47bd-a85f-5b50ddc0424b)


**Possible direction**

1. User don't have any relevant schedule -  then message is 'There are no schedules relevant to user'

2. User is not on-call but added to schedules - empty block w/o any message 

3. There are relevant on-call shifts - then shifts blocks

PR #2928, opened 2023-08-31
Add endpoint for getting schedules events for current user

Ferril:
# What this PR does

https://github.com/grafana/oncall/issues/2915

## Checklist


engine/apps/schedules/tests/test_on_call_schedule.py
```diff
@@ -2552,3 +2509,165 @@ def test_filter_events_ical_duplicated_uid(make_organization, make_user_for_orga
...
+    assert len(upcoming_shifts) == 5
+    for shift in upcoming_shifts:
+        users = {u["pk"] for u in shift["users"]}
+        assert current_user.public_primary_key in users
+
+    passed_shifts, current_shifts, upcoming_shifts = schedule.shifts_for_user(user2, start_date, days)
+    assert len(passed_shifts) > 0
+    assert len(current_shifts) > 0
+    assert len(upcoming_shifts) > 0
+    for shifts in (passed_shifts, current_shifts, upcoming_shifts):
```
matiasb, line 2627: Besides checking the right user is there, it would be nice to test the past shifts are ended events, current are in progress and upcoming are in the future.

engine/apps/schedules/models/on_call_schedule.py
```diff
@@ -505,33 +505,32 @@ def refresh_ical_final_schedule(self):
...
-                    current_shift = e
-                    continue
-                upcoming_shift = e
-                break
+                if event["end"] <= now:
+                    passed_shifts.append(event)
+                elif event["start"] <= now < event["end"]:
+                    current_shifts.append(event)
+                else:
+                    upcoming_shifts.append(event)
```
matiasb, line 531: It will be nice if it is not too complex to keep the events sorted in each list.

matiasb (approved):
LGTM, with a couple of minor suggestions.

engine/apps/schedules/models/on_call_schedule.py
matiasb: Sorry, I was confused (I was thinking in the API response, if we combine the events from different schedules... but right, we keep events separated by schedule, so they should be sorted :+1: )

PR #2933, opened 2023-08-31
fix IntegrityErrors occuring when creating ResolutionNoteSlackMessage objects

joeyorlando:
# What this PR does

Closes https://github.com/grafana/oncall-private/issues/1822

## Checklist


engine/apps/alerts/models/resolution_note.py
```diff
@@ -83,6 +83,11 @@ class ResolutionNoteSlackMessage(models.Model):
     class Meta:
         unique_together = ("thread_ts", "ts")
 
+        indexes = [
+            models.Index(fields=["ts", "thread_ts", "alert_group_id"]),
+            models.Index(fields=["ts", "thread_ts", "slack_channel_id"]),
+        ]
+
```
joeyorlando, line 90: added indices to speed up querying these objects. These are the two filter patterns we use when querying these objects.

engine/apps/slack/scenarios/slack_channel_integration.py
```diff
@@ -82,46 +82,32 @@ def save_thread_message_for_resolution_note(
...
-                        "chat.postEphemeral",
-                        channel=channel,
-                        user=slack_user_identity.slack_id,
-                        text=":warning: Unable to update the <{}|message> in Resolution Note: the message is too long ({}). "
-                        "Max length - 2900 symbols.".format(permalink, len(text)),
-                    )
-                return
-            slack_thread_message.text = text
-            slack_thread_message.save()
+        slack_thread_message, created = ResolutionNoteSlackMessage.objects.get_or_create(
```
Ferril, line 85: I think it's better to check `len(text)` before creating ResolutionNoteSlackMessage. Too long text may cause issues either with saving it to the db or displaying it in Slack 🤔

joeyorlando: 👍 

do you happen to know why there is this condition?:
```python3
if slack_thread_message.added_to_resolution_note:
```
if that can be removed, we can still use `get_or_create`

joeyorlando: alternative solution is to keep the separate `get`/`create` calls as it was, and add _another_ `try/except` when creating, and catch any possible `IntegrityError`s (but this seems a bit messier imo)

engine/apps/slack/scenarios/slack_channel_integration.py
```diff
@@ -82,46 +82,31 @@ def save_thread_message_for_resolution_note(
...
-            slack_thread_message = ResolutionNoteSlackMessage.objects.get(
-                ts=message_ts,
-                thread_ts=thread_ts,
-                alert_group=alert_group,
+        if len(text) > 2900:
+            self._slack_client.api_call(
+                "chat.postEphemeral",
+                channel=channel,
+                user=slack_user_identity.slack_id,
+                text=":warning: Unable to update the <{}|message> in Resolution Note: the message is too long ({}). "
```
Ferril, line 90: I would change the message a bit. Something like "Unable to show the message in Resolution Notes..."

Ferril (approved):

PR #2947, opened 2023-09-01
update slack_sdk dependency to latest version

joeyorlando:
# What this PR does

- update `slackclient` dependency to latest version. The version we were using was 5 years old 😲
  - first followed the v2 migration guide [here](https://github.com/slackapi/python-slack-sdk/wiki/Migrating-to-2.x) followed by the v3 migration guide [here](https://slack.dev/python-slack-sdk/v3-migration/). The main changes were:
    - The PyPI project was renamed from `slackclient` to `slack_sdk`
    - it is discouraged/harder to call `api_call` and encouraged to call the helper methods (ex. `chat_postMessage`; [note](https://github.com/slackapi/python-slack-sdk/wiki/Migrating-to-2.x#web-client-api-changes) in migration guide docs)
    - In 1.x, a failed api call would return the error payload to you and have you handle the error. In 2.x, a failed api call will throw an exception. To handle this in your code, you will have to wrap api calls with a try except block. Since we overload `WebClient.api_call` this was an easy change and only required a one line change
- remove `apps.slack.slack_client.slack_server.SlackClientServer` class. The new version of `slack_sdk` handles the case that we needed to overload for in the first place.
- merged `apps/slack/slack_client/slack_client.py` and  `apps/slack/slack_client/exceptions.py` into `apps/slack/client.py`

## Checklist


engine/apps/slack/client.py
```diff
@@ -3,38 +3,37 @@
...
 
-        proxies = kwargs.get("proxies")
 
-        if self.refresh_token:
-            if callable(self.token_update_callback):
-                token = None
-            else:
-                raise TokenRefreshError("Token refresh callback function is required when using refresh token.")
-        # Slack app configs
-        self.server = SlackClientServer(token=token, connect=False, proxies=proxies)
```
joeyorlando, line 36: `self.server` is the _only_ spot where `SlackClientServer` is referenced, and `SlackClientWithErrorHandling.server` is never referenced, hence we can remove the class.

Additionally, `SlackClientWithErrorHandling.__init__` was only ever overriden to override `self.server`, with that said we can remove the overriden `__init__`.

vadimkerr: btw we're using `slackclient==1.3.0`, and the latest version is `3.21.3`. Looks like the scenario when `json.loads` fails [is handled better](https://github.com/slackapi/python-slack-sdk/blob/main/slack_sdk/web/base_client.py#L298-L302) in the latest version. should we consider upgrading?

vadimkerr: [1.3.0](https://github.com/slackapi/python-slack-sdk/releases/tag/v1.3.0) is 5 years old 😮

joeyorlando: good catch 👍 I'll take a look at migrating us to the latest version

engine/apps/slack/models/slack_usergroup.py
```diff
@@ -112,11 +111,7 @@ def update_oncall_members(self):
     def update_members(self, slack_ids):
         sc = SlackClientWithErrorHandling(self.slack_team_identity.bot_access_token)
 
-        sc.api_call(
-            "usergroups.users.update",
-            usergroup=self.slack_id,
-            users=slack_ids,
-        )
```
joeyorlando, line 119: in `api_call()` the `timeout` parameter has been removed and is now passed at the client level ([migration guide docs](https://github.com/slackapi/python-slack-sdk/wiki/Migrating-to-2.x#web-client-api-changes))

engine/apps/slack/client.py
```diff
@@ -87,7 +90,10 @@ def paginated_api_call_with_ratelimit(self, *args, **kwargs) -> Tuple[dict, Opti
         return cumulative_response, cursor, rate_limited
 
     def api_call(self, *args, **kwargs):
-        response = super(SlackClientWithErrorHandling, self).api_call(*args, **kwargs)
+        try:
+            response = super(SlackClientWithErrorHandling, self).api_call(*args, **kwargs)
+        except SlackApiError as err:
+            response = err.response
```
joeyorlando, line 96: In 1.x, a failed api call would return the error payload to you and have you handle the error. In 2.x, a failed api call will throw an exception ([migration guide docs](https://github.com/slackapi/python-slack-sdk/wiki/Migrating-to-2.x#error-handling-changes))

joeyorlando: even though we are no longer _directly_ calling `SlackClientWithErrorHandling.api_call`, all of the helper methods (ex. `chat_postMessage`) call `self.api_call` internally so this error handling change here will catch any newly thrown exceptions

vadimkerr (approved):
LGTM, great job upgrading the Slack SDK 🙌

PR #2957, opened 2023-09-04
Refactor `AlertGroup.slack_message`

vadimkerr:
# What this PR does

Refactors the foreign key relationship between `AlertGroup` and `SlackMessage` models (see https://github.com/grafana/oncall-private/issues/1867 for more info).

https://github.com/grafana/oncall-private/issues/1867

## Checklist


engine/apps/alerts/migrations/0032_remove_alertgroup_slack_message_state.py
```diff
@@ -0,0 +1,18 @@
...
+    dependencies = [
+        ('alerts', '0031_auto_20230831_1445'),
+    ]
+
+    operations = [
+        common.migrations.remove_field.RemoveFieldState(
+            model_name='AlertGroup',
+            name='slack_message',
+        ),
+    ]
```
vadimkerr, line 18: Remove `AlertGroup.slack_message` so the only FK left is `SlackMessage.alert_group`

engine/apps/slack/alert_group_slack_service.py
```diff
@@ -61,22 +56,8 @@ def update_alert_group_slack_message(self, alert_group: "AlertGroup") -> None:
...
-                    slack_id=result["ts"],
-                    organization=slack_message.organization,
-                    _slack_team_identity=slack_message.slack_team_identity,
-                    channel_id=slack_message.channel_id,
-                    alert_group=alert_group,
-                )
-                slack_message_updated.save()
-                alert_group.slack_message = slack_message_updated
-                alert_group.save(update_fields=["slack_message"])
-                logger.info(f"Message has been posted for alert_group {alert_group.pk}")
```
vadimkerr, line 79: Removing this logic completely because it's hard to maintain with the new FK structure.
I think it makes more sense to just ignore deleted AG Slack messages instead of trying to post a new message (why post a message one more time if the user already decided they don't want to see it and deleted it?)

matiasb (approved):
LGTM.

I see you are removing the field following the guideline :+1: , so I guess there will be a PR later removing the column from the DB, right?

vadimkerr:
@matiasb yes, the DB field will be dropped in a future release.

PR #2965, opened 2023-09-05
Fix escalation step "Notify if num alerts in time window"

Ferril:
# What this PR does
Fix escalation step "Notify if num alerts in time window" when escalation policy was deleted during the escalation

https://github.com/grafana/oncall-private/issues/2017

## Checklist


engine/apps/alerts/escalation_snapshot/snapshot_classes/escalation_policy_snapshot.py
```diff
@@ -411,11 +411,11 @@ def _escalation_step_notify_if_num_alerts_in_time_window(
 
         last_alert = alert_group.alerts.last()
 
-        time_delta = datetime.timedelta(minutes=self.escalation_policy.num_minutes_in_window)
```
Ferril, line 414: We keep these params in escalation snapshot and don't need to call EscalationPolicy object here

PR #2999, opened 2023-09-08
Update escalation policies public API to handle new webhooks

### LINKED ISSUE #2998: Escalation policies public endpoint does not support advanced webhooks
### What went wrong?

**What happened**:
- Trying to set up an escalation policy using the API to set an existing webhook step returns:
`{"action_to_trigger":["Object does not exist"]}`

- The escalation policy is created with the given webhook as action

### How do we reproduce it?

Create a webhook. Try to set it up as an escalation policy using the public API:

```
curl "https://xxx/api/v1/escalation_policies/" \
  --request POST \
  --header "Authorization: xxxx" \
  --header "Content-Type: application/json" \
  --data '{
  "escalation_chain_id": "some-id",
  "type": "trigger_action",
  "action_to_trigger": "<webhook-id>",
  "position": 0
}'
```

### Grafana OnCall Version

r231-v1.3.36

### Product Area

Other

### Grafana OnCall Platform?

I use Grafana Cloud

### User's Browser?

_No response_

### Anything else to add?

_No response_

matiasb:
Support new webhooks when creating or updating.
Return existing information if still using previous custom button actions.

Fixed  #2998


Ferril (approved):
LGTM

Issue #1906, opened 2023-05-09
Unify breadcrumbs behaviour with other Grafana Apps and main core


raphael-batte:
I noticed that the existing behavior of breadcrumbs differs from other applications and the grafana core.

- For some reason we are not showing the current page
- Grafana and other applications - display

I propose to unify the behavior with Grafana. Additionally, this will allow you to remove the back arrow from the title

<img width="745" alt="image" src="https://github.com/grafana/oncall/assets/102970268/d31c7adc-7253-4646-8f9c-b3e4674f7616">
<img width="619" alt="image" src="https://github.com/grafana/oncall/assets/102970268/74078a85-fbd7-4619-a282-036d2420551f">

PR #3003, opened 2023-09-11
Unify breadcrumbs behaviour with other Grafana Apps and main core

maskin25:
# What this PR does

Unify breadcrumbs behaviour with other Grafana Apps and main core

https://github.com/grafana/oncall/issues/1906

## Checklist


grafana-plugin/src/pages/incident/Incident.tsx
```diff
@@ -89,6 +93,12 @@ class IncidentPage extends React.Component<IncidentPageProps, IncidentPageState>
     store.alertGroupStore.updateSilenceOptions();
   }
 
+  componentWillUnmount(): void {
```
teodosii, line 96: Wouldn't it be better to make a generic React component that handles this cleanup so that we end with less duplicated lines in the affected components?
Like a `BaseContext<P, S>` component that extends `React.Component<P, S>` where you add the `contextType`, `context`, `cleanup` function and `componentWillUnmount` (and if ever we're going to add more to the componentWillUnmount for the extended component we can just  call `cleanup` on the base class instead)

teodosii (approved):

CHANGELOG.md
```diff
@@ -20,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Notify user via Slack/mobile push-notification when their shift swap request is taken by @joeyorlando ([#2992](https://github.com/grafana/oncall/pull/2992))
+- Unify breadcrumbs behaviour with other Grafana Apps and main core# ([1906](https://github.com/grafana/oncall/issues/1906))
```
joeyorlando, line 27: @maskin25 can you retroactively update the `CHANGELOG` to remove this line? Seems to be duplicated, correct?

Issue #2764, opened 2023-08-08
Allow specifying a Redis db number


itspngu:
### What would you like to see!

Currently, the Redis db used by Oncall is [hardcoded as 1](https://github.com/grafana/oncall/blob/a1a9e0c33c0c2fe5a78e68af3409dcc3ce9d5f1c/engine/settings/base.py#L188). It would be nice if this was configurable, allowing users to safely re-use the same Redis deployment across multiple GrafanaLabs products (Grafana, Oncall, Loki, etc.) in smaller setups.

How it's done in Loki: https://grafana.com/docs/loki/latest/configuration/#cache_config
How it's done in Grafana: https://grafana.com/docs/grafana/latest/setup-grafana/configure-grafana/#remote_cache

### Product Area

Other

MrEcco:
There are actually two Redis databases in use: 0 (broker) and 1 (cache). But it is possible to run the whole application in a single number, celery and cache don't conflict (due to different types of keys and different naming).

@itspngu I've prepared PR (https://github.com/grafana/oncall/pull/3029) with the fix. Could you please leave a comment or like in there, just to get maintainers' attention?

PR #2644, opened 2023-07-26
Added REDIS_PORT mapping from values

deepminimal:
# What this PR does
Now we can use custom redis port with specific database number

## Checklist

- [ ] Unit, integration, and e2e (if applicable) tests updated
- [ ] Documentation added (or `pr:no public docs` PR label added if not required)
- [ ] `CHANGELOG.md` updated (or `pr:no changelog` PR label added if not required)

PR #3029, opened 2023-09-16
feature: Hardening the Helm deployment with Redis and Postgres TLS

MrEcco:
# What this PR does

Short summary: this PR improves security and configuration management for Helm deployment. Please take a look at the details below.

Issues:
- Cannot explicitly define redis database (only 0 and 1 numbers are used)
- Cannot securely use TLS for Redis (cannot set CA certificate; cannot set client certificates)
- Cannot securely use TLS for Postgres (cannot set CA certificate; cannot set client certificates; cannot set `verify-full` validation)
- ~~Chart option `securityContext.readOnlyRootFilesystem: true` issues CrashLoopBack pod state~~ will be moved to new PR

## Checklist

- [x] ~~Unit, integration, and e2e (if applicable) tests updated~~ (not required)

- [x] Helm tests are fixed and updated
- [x] Manually verified the features:
  - [x] postgres TLS connection with `verify-full` validation
  - [x] redis TLS connection with `cert_required` validation
  - [x] redis protocol and database number controls
  - [x] all containers properly work in read-only root filesystem
  - [x] all changes are backward compatible (doesn't break old deployments)

## Changelog

- Fixed helm tests
- Added configuration options for secure TLS communication with dependencies like Redis, MySQL, and Postgres
- ~~Added configuration option for relocating `celerybeat` database file (read-only root filesystem issue)~~ will be moved to new PR
- Improved redis database configuration options
- Now only single redis database is used
- Added ability to mount custom volumes (with CA certificates, for example) into Helm chart
- ~~Fixed issue with read-only root filesystem for Helm chart~~ will be moved to new PR
- Add ability to work with Redis ACL (and AWS ElastiCache)


MrEcco:
I would really appreciate it if you take it to review and release as soon as possible :)

MrEcco:
These PRs and issues might be relevant:
- https://github.com/grafana/oncall/pull/2644
- https://github.com/grafana/oncall/issues/2764

itspngu:
This is very useful, I was planning to look into the Redis DB number thing but haven't had the chance yet, thank you!

MrEcco:
@iskhakov @Matvey-Kuk I'm very sorry for the direct mention of maintainers, but could you please take a look and probably leave a comment or accept the PR?

matiasb (commented):
Makes sense to me, added a couple of minor comments.

engine/settings/base.py
```diff
@@ -443,6 +478,10 @@ class BrokerTypes:
     "ALERT_GROUP_ESCALATION_AUDITOR_CELERY_TASK_HEARTBEAT_URL", None
 )
 
+SCHEDULE_FILENAME = os.getenv("SCHEDULE_FILENAME")
```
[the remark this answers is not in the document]
MrEcco: I cannot use `CELERY_*` names because these variables are namespace by Celery library (see here: https://github.com/grafana/oncall/blob/dev/engine/engine/celery.py#L26-L30). So the only idea for this name is `BEAT_SCHEDULE_DB_FILENAME`, I'll prepare this change. If you have other ideas you are welcome :)

This option already has default value. If this variable isn't set, it will never be passed to Celery so the library will choose the internal default (`celerybeat-schedule`, in the working directory of the process).

engine/settings/base.py
```diff
@@ -186,10 +219,10 @@ class DatabaseTypes:
             REDIS_URI,
         ],
         "OPTIONS": {
-            "DB": 1,
+            "DB": REDIS_DATABASE,
```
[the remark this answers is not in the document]
MrEcco: Please take a close look at these lines of code: [REDIS_URI definition](https://github.com/grafana/oncall/blob/dev/engine/settings/base.py#L177-L179), [remote cache](https://github.com/grafana/oncall/blob/dev/engine/settings/base.py#L182-L200), [broker](https://github.com/grafana/oncall/blob/dev/engine/settings/base.py#L417-L422). In case of a Redis-based broker for celery, this project uses both `0` and `1` databases by default. There are two default databases, so I just chose one that is default for redis library.

But perhaps you are right, the project uses rabbitmq broker instead of redis by default, but uses redis cache with 1. Was there a reason to use a non-default value for library database number?

In addition, the default number `1` may confuse users because most part of projects use `0` when not set.

So what should we use as default then? :)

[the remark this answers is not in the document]
MrEcco: I set the default REDIS_DATABASE to `1` for both broker and cache. But for helm-based installation, it is still `0` because the chart always defines this value.

MrEcco:
@matiasb Could you please set `pr:no public docs` and probably `pr:no changelog` labels as default PR requirements require? I have no permission to do it.

MrEcco:
Aaaaaaand I have no idea why one of the pipeline checks has failed :) I have no access to logs of `Grafana: ${{ inputs.grafana-image-tag }}` job. Can anybody help?

matiasb (approved):
Thanks! Changes look good to me, maybe @iskhakov or @joeyorlando can take an extra look on the helm side?

engine/settings/base.py
```diff
@@ -443,6 +480,10 @@ class BrokerTypes:
     "ALERT_GROUP_ESCALATION_AUDITOR_CELERY_TASK_HEARTBEAT_URL", None
 )
 
+BEAT_SCHEDULE_DB_FILENAME = os.getenv("BEAT_SCHEDULE_DB_FILENAME", "celerybeat-schedule")
+if BEAT_SCHEDULE_DB_FILENAME:
+    CELERY_BEAT_SCHEDULE_FILENAME = BEAT_SCHEDULE_DB_FILENAME
```
joeyorlando, line 485: is there a need for the `if` clause here?

```suggestion
CELERY_BEAT_SCHEDULE_FILENAME = os.getenv("CELERY_BEAT_SCHEDULE_FILENAME", "celerybeat-schedule")
```

helm/oncall/templates/_env.tpl
```diff
@@ -517,15 +553,43 @@
...
+{{- with .Values.externalRedis.username }}
+- name: REDIS_USERNAME
+  value: {{ . | quote }}
+{{- end }}
 - name: REDIS_PASSWORD
   valueFrom:
     secretKeyRef:
       name: {{ include "snippet.redis.password.secret.name" . }}
       key: {{ include "snippet.redis.password.secret.key" . | quote}}
+{{- if and (not .Values.postgresql.enabled) .Values.externalRedis.ssl_options.enabled }}
```
joeyorlando, line 573: referencing `.Values.postgresql.enabled` seems like a typo here

helm/oncall/templates/_env.tpl
```diff
@@ -484,6 +504,22 @@
...
+{{- end }}
+
+{{- define "snippet.redis.database" -}}
+{{ if not .Values.redis.enabled -}}
+  {{ default 0 .Values.externalRedis.database | quote }}
+{{- else -}}
+  "0"
+{{- end }}
+{{- end }}
+
```
joeyorlando, line 522: for these three new `define` blocks, I would remove the leading spaces before the values (ex. ` "0"` vs `"0"`

joeyorlando (commented):
thanks for the contribution! 😄  some minor changes/questions I left

helm/oncall/templates/celery/deployment.yaml
```diff
@@ -82,6 +84,22 @@ spec:
...
         {{- tpl . $ | nindent 8 }}
       {{- end }}
+      volumes:
+        - name: tempdir
+          emptyDir: {}
+        - name: celery
+          emptyDir: {}
+        {{- with .Values.celery.extraVolumes }}
+        {{- . | toYaml | nindent 8 }}
+        {{- end }}
```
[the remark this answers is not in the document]
MrEcco: There is an issue with performing in the `readOnlyRootFilesystem: true` mode. With this restriction, the container application cannot start at all. To make it working I need to allow read/write from some directories:
1. `tempdir` - cannot start without it, every container
2. `celery` - celery is starting but has many (bulky) errors in the log. Celery "beat" feature requires for creation of the file right in the directory of the code. The code should be immutable (or why use readOnlyRootFilesystem?), thus we need to relocate this file to `celery` volume.

The only reason is make `readOnlyRootFilesystem` working.

helm/oncall/templates/_env.tpl
MrEcco: I just follow the existing code style here. The indent will not be placed to manifest after templating. In `"snippet.rabbitmq.password.secret.key"` default value is set with indent. Should I replace it everywhere in the chart?

helm/oncall/templates/_env.tpl
```diff
@@ -517,15 +553,43 @@
...
   value: {{ include "snippet.redis.host" . }}
 - name: REDIS_PORT
-  value: "6379"
+  value: {{ include "snippet.redis.port" . }}
+- name: REDIS_DATABASE
+  value: {{ include "snippet.redis.database" . }}
+{{- with .Values.externalRedis.username }}
+- name: REDIS_USERNAME
+  value: {{ . | quote }}
+{{- end }}
```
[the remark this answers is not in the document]
MrEcco: I hid it because it is another rarely used environment variable in the deployment manifest. With default `""` value will work as expected. I'm preparing the fix. Thanks :)

helm/oncall/templates/_env.tpl
MrEcco: It is not a typo. The option `.Values.postgresql.enabled` is used for deploying PostgreSQL as the chart dependency (new dedicated cluster in the same namespace). This cluster has no SSL options. The option `.Values.externalRedis` defines the external cluster, and SSL configuration is possible only for it. If someone set to use internal Postgres deployment (for development, of course) and set `.Values.externalRedis.ssl_options` then SSL options should not be picked to manifest.

Thus, this is the right expression.

engine/settings/base.py
MrEcco: I will fix it.

TLDR (and link for history): https://docs.celeryq.dev/en/stable/userguide/configuration.html#new-lowercase-settings

helm/oncall/templates/_env.tpl
MrEcco: Yes, it is typo, thank you :)

lucasfcnunes:
@MrEcco thanks for your good work!

Closes #3063

MrEcco:
@matiasb @joeyorlando Guys, could anyone start the pipeline and perform the final review? :)

helm/oncall/templates/celery/deployment.yaml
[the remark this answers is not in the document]
joeyorlando: I would suggest alternatively providing these to `Values.celery.extraVolumeMounts`/`Values.celery.extraVolumes` in your deployment's values/settings. I don't think there is a need to include these as default for _everyone_

engine/settings/base.py
```diff
@@ -123,6 +123,15 @@ class DatabaseTypes:
 DATABASE_HOST = os.getenv("DATABASE_HOST") or os.getenv("MYSQL_HOST")
 DATABASE_PORT = os.getenv("DATABASE_PORT") or os.getenv("MYSQL_PORT")
 
+DATABASE_OPTIONS = os.getenv("DATABASE_OPTIONS") or os.getenv("MYSQL_OPTIONS")
```
joeyorlando, line 126: I would remove `or os.getenv("MYSQL_OPTIONS")` here

helm/oncall/templates/celery/deployment.yaml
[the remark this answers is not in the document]
MrEcco: These changes provide the solution for `readOnlyRootFilesystem` in the container (that by the best security practices should be enabled). These changes in defaults don't violate _everyone_'s deployments, just update the configuration, and don't require manual actions. Why it is preferable to keep current defaults for celery and a user should find the solution by himself?
I'm sorry for this question, it just looks like a bug. You use some security considerations like a non-root user, fs groups, capabilities, etc., but don't use read-only fs. It looks unfinished. I don't force everyone to use RO FS but also don't want anyone to spend time for searching a solution for the mentioned above error.

Anyway, I will accept any decision about this change, just leave a comment with the decision. If you say drop this change then I can enable it for my deployment exactly like you said. And thank you for your attentive review :)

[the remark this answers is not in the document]
joeyorlando: I believe I _roughly_ understand the issue you mentioned w/ `readOnlyRootFilesystem` in relation to Celery, but why do you need two `emptyDir` volumes/volumeMounts in that case? Wouldn't the `celery` one alone suffice?

If you want, you can also feel free to move this `readOnlyRootFilesystem` topic into a separate issue/PR and I can approve the current PR without this change. I don't want to keep dragging you along with questions but just want to make sure I understand fully before we merge this 🙂 up to you.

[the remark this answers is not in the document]
MrEcco: Done, this part is removed. Will create PR for `readOnlyRootFilesystem` later after the merge.

helm/oncall/templates/celery/deployment.yaml
```diff
@@ -68,6 +68,8 @@ spec:
             {{- include "snippet.db.env" . | nindent 12 }}
             {{- include "snippet.broker.env" . | nindent 12 }}
             {{- include "oncall.extraEnvs" . | nindent 12 }}
+            - name: BEAT_SCHEDULE_DB_FILENAME
+              value: /var/run/celery/celerybeat.db
```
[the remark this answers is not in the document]
MrEcco: Resolved. https://github.com/grafana/oncall/pull/3029#discussion_r1338725912

MrEcco:
https://github.com/grafana/oncall/pull/3029#issuecomment-1744619128 I'm also waiting too..

helm/oncall/templates/_env.tpl
```diff
@@ -476,6 +488,14 @@
 {{- end }}
 {{- end }}
 
+{{- define "snippet.redis.protocol" -}}
+{{ if not .Values.redis.enabled -}}
+  {{ default "redis" .Values.externalRedis.protocol | quote }}
+{{- else -}}
+  "redis"
+{{- end }}
```
joeyorlando, line 496: can this be simplified to the following?
```suggestion
{{ default "redis" .Values.externalRedis.protocol | quote }}
```

helm/oncall/templates/_env.tpl
joeyorlando, line 522: similar to above question

joeyorlando (commented):
few last comments. thanks for your patience!

helm/oncall/templates/_env.tpl
MrEcco: Just trying to follow existing structures and code style 😄 Sure I can but please take a look at the chart. It is full of similar solutions: https://github.com/grafana/oncall/blob/dev/helm/oncall/templates/_env.tpl#L271-L301

Let me 15 minutes. Please don't go too far 😄

MrEcco: DONE. @joeyorlando please start the tests again, after changes and rebase

joeyorlando (approved):

joeyorlando:
this will be released when v1.3.41 goes out. Keep an eye out for that (likely this week). Thanks again for the contribution and your patience @MrEcco 🙂

MrEcco:
Thank you too, @joeyorlando, and @matiasb 😄 

@lucasfcnunes, https://github.com/grafana/oncall/pull/3029#issuecomment-1744977948, now it is not so long to wait 😄

PR #3123, opened 2023-10-05
Add truncation for content beyond 2 lines and show tooltip with original value if content was truncated

teodosii:
# What this PR does

- Removed `MAX_LINE_LENGTH` which was used to truncate Integrations Name
- Added new component `TextEllipsisTooltip` that determines if its content was truncated, and if so, it will show a tooltip with the original value. Integrations Name and Alert Group Name fields inside tables have been changed so that after filling in 2 lines, the content gets truncated and a tooltip is shown on hover.

Before
image
After
image

As you can see this results in more space being taken for the integrations name (but no more than what it should take), and the end result is always truncated after 2 lines. Same applies to the Integrations

Before
image
After
image

The `Don't delete` integration wasn't even showing 100% of its full width in the before screen, but it shows in the after because now there's no more limit on the characters.


grafana-plugin/src/components/TooltipBadge/TooltipBadge.tsx
```diff
@@ -17,21 +17,33 @@ interface TooltipBadgeProps {
...
 
 const TooltipBadge: FC<TooltipBadgeProps> = (props) => {
-  const { borderType, text, tooltipTitle, tooltipContent, onHover, addPadding, icon, customIcon, className, ...rest } =
-    props;
+  const {
+    borderType,
+    text,
+    tooltipTitle,
+    tooltipContent,
+    placement,
```
maskin25, line 33: maybe `placement = 'bottom-start'` to use 'bottom-start' as default

grafana-plugin/src/components/TextEllipsisTooltip/TextEllipsisTooltip.tsx
```diff
@@ -0,0 +1,59 @@
...
+  className,
+  content: textContent,
+  placement,
+  children,
+}) => {
+  const [isEllipsis, setIsEllipsis] = useState(true);
+  const elContentRef = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
+    setEllipsis();
```
maskin25, line 29: it's called only onMount, but textContent and children can be changed later

grafana-plugin/src/components/TextEllipsisTooltip/TextEllipsisTooltip.tsx
```diff
@@ -0,0 +1,59 @@
...
+  }, []);
+
+  const elContent = (
+    <div className={cx(className)} ref={elContentRef} onMouseOver={setEllipsis}>
```
maskin25, line 33: ok, we'll display tooltip (if element is too long) `onMouseOver`, but what about calling `setEllipsis ` on `onMouseOut` ?

grafana-plugin/src/components/TextEllipsisTooltip/TextEllipsisTooltip.tsx
teodosii: In current use cases the content cannot change, the only issue is on resizing as that will determine truncation, but that's sorted out by `onMouseHover`.

grafana-plugin/src/components/TextEllipsisTooltip/TextEllipsisTooltip.tsx
teodosii: Why would we need to call on mouseleave?

maskin25: ok, just looks a bit weird to update the state on `onMouseHover`, in theory mouse hovering shouldn't affect this state...

grafana-plugin/src/components/TextEllipsisTooltip/TextEllipsisTooltip.tsx
maskin25: maybe it's better to add [children] to useEffect deps right now?

grafana-plugin/src/components/TextEllipsisTooltip/TextEllipsisTooltip.tsx
teodosii: IMO that's a good spot for checking if anything changed. It's like a trigger to show the tooltip, instead we just update the state which will show the tooltip... I don't like setting it on `resize`, maybe there could some other actions triggering truncation which I'm not aware of right now, `onMouseOver` catches every case.

maskin25 (approved):

grafana-plugin/src/pages/schedules/Schedules.tsx
```diff
@@ -369,14 +369,14 @@ class SchedulesPage extends React.Component<SchedulesPageProps, SchedulesPageSta
...
-                      <Avatar size="medium" src={user.avatar} />
-                    </div>
-                    <MatchMediaTooltip placement="top" content={user.username} maxWidth={TABLE_COLUMN_MAX_WIDTH}>
-                      <span className="table__email-content">{user.username}</span>
-                    </MatchMediaTooltip>
-                  </div>
+                  <HorizontalGroup>
+                    <TextEllipsisTooltip placement="top" content={user.username}>
+                      <Text type="secondary" className={cx('overflow-child')}>
+                        <Avatar size="small" src={user.avatar} />{' '}
```
teodosii, line 375: Changed from `medium` to `small` so that it's consistent with other places (see incident's users or the teams icon)

PR #3144, opened 2023-10-11
fixes 3069

### LINKED ISSUE #3069: Default values for ingress annotations in helm chart conflicts with cluster issuers and custom nginx ingress classes
### What went wrong?

**What happened**:
Tried to install Grafana OnCall with Helm into our existing infrastructure. Added annotations on the ingress to account for our cluster-issuer and custom nginx ingress class. Did not include cert-manager or nginx ingress controller in the helm template as this is already present in our cluster. Got error messages on install as issuer and cluster-issuer annotations cannot co-exist.  The default value for ingressClassName annotation is also problematic if you want to use custom ingressClassName.

- Helm install to successfully complete.

### How do we reproduce it?

1. Install the helm chart with an `cert-manager.io/cluster-issuer: "custom"` annotation in the ingress and with cert-manager and nginx ingress controller disabled.
3. Observe error messages when installing.

### Grafana OnCall Version

v1.3.38

### Product Area

Helm

### Grafana OnCall Platform?

Kubernetes

### User's Browser?

_No response_

### Anything else to add?

Would be happy to help with a PR. But need input on how to handle this.

Options:
1. Remove default values for ingress annotations
2. Only add the default values for the issuer and nginxClassName annotations if cert-manager and nginx ingress controller is enabled respectively.

git4fred:
# What this PR does
Comment some bad default values for ingress manifest generation

#3069 

## Checklist

### DIFF
--- a/helm/oncall/values.yaml
+++ b/helm/oncall/values.yaml
@@ -365,9 +365,9 @@ env: {}
 ingress:
   enabled: true
   #  className: ""
-  annotations:
-    kubernetes.io/ingress.class: "nginx"
-    cert-manager.io/issuer: "letsencrypt-prod"
+  annotations: {}
+  #  kubernetes.io/ingress.class: "nginx"
+  #  cert-manager.io/issuer: "letsencrypt-prod"
   tls:
     - hosts:
         - "{{ .Values.base_url }}"


helm/oncall/values.yaml
```diff
@@ -365,9 +365,9 @@ env: {}
 ingress:
   enabled: true
   #  className: ""
-  annotations:
-    kubernetes.io/ingress.class: "nginx"
-    cert-manager.io/issuer: "letsencrypt-prod"
+  annotations: {}
+  #  kubernetes.io/ingress.class: "nginx"
+  #  cert-manager.io/issuer: "letsencrypt-prod"
```
joeyorlando, line 370: can you elaborate on this change? what about simply modifying `ingress.annotations` in your `values.yml` (or via `--set`) when these defaults are not applicable?

git4fred: I want to remove them because:
- I use oncall helm chart as a dependency in my deployment, so the default `values.yaml` file is take into account 
- I use `ingress.ClassName` key. So it enters in conflict with the ingress annotation, and kubernetes does not allow to have the annotation  `kubernetes.io/ingress.class` with `.spec.ingressClassName` in a manifest.
- I use a `clusterissuer` for certmanager. So I add the annotation  `cert-manager.io/cluster-issuer` and the other cert-manager annotation stands.  cert-manger does not deal with an `issuer` and a `cluster-issuer` annotations.

For the ingressClassName, you should also modify your template `templates/ingress-regular.yaml` as if you define a `ingress.className` key, then you have the annotation and `.spec.ingressClassName` that are defined, and the generated manifest does not work with latest kubernetes versions.

jgourmelen:
any update ?

PR #3146, opened 2023-10-12
fix(helm): Remove redis replica by default

adinhodovic:
I don't think the replica is even used. Less pods.

# What this PR does

Remove redis replica by default.

## Checklist


joeyorlando (commented):
imo it might be even more beneficial to instead set `sentinel.enabled` to `true` by default. This would essentially promote a new replica to master in the event that the master node fails. From the [docs](https://artifacthub.io/packages/helm/bitnami/redis/16.13.2#master-replicas-with-sentinel) for the `bitnami/redis@16.13.2` helm chart:

> Master-Replicas with Sentinel
> When installing the chart with architecture=replication and sentinel.enabled=true, it will deploy a Redis® master StatefulSet (only one master allowed) and a Redis® replicas StatefulSet. In this case, the pods will contain an extra container with Redis® Sentinel. This container will form a cluster of Redis® Sentinel nodes, which will promote a new master in case the actual one fails. In addition to this, only one service is exposed:

I would be hesitant to make this change in this PR a default just because when people upgrade their chart version it may have unintended changes. Note that if you prefer to run things in `standalone` mode, you can update your helm deployment's `values.yaml` (or pass as a `--set` flag to your `helm` command).

PR #3147, opened 2023-10-12
Handle None role when syncing users from Grafana

matiasb:
Fixes https://github.com/grafana/oncall-private/issues/2201


joeyorlando (commented):
this handles the sync portion of things but do you know how authorization checks will work if a user has a basic role of None?

(my guess is it won't make a difference since this scenario is only applicable when RBAC is enabled?)

matiasb:
> this handles the sync portion of things but do you know how authorization checks will work if a user has a basic role of None?
> 
> (my guess is it won't make a difference since this scenario is only applicable when RBAC is enabled?)

Not really sure why we got this user with the 'None' role, or what that means (at Grafana level). Can this user log in? Shouldn't get access to anything? Fallbacking to `Viewer` in OnCall seemed ok (we require a value for the `role` field), but maybe we need to consider adding the `None` option?

joeyorlando:
@matiasb take a look at [this issue](https://github.com/grafana/oncall-private/issues/2169) (tldr; having a basic role of "None" is, as of recently, a valid scenario)

matiasb:
Reworked the changes by adding the `None` role to OnCall

joeyorlando (approved):

Issue #3154, opened 2023-10-17
Ical schedules are missing in GET /schedules/current_user_events/ response


maskin25:
![Image](https://github.com/grafana/oncall/assets/3278022/6b420df7-f743-4143-8ea3-650c67c4367a)

matiasb:
This should have been fixed by #3156

PR #3156, opened 2023-10-17
Update ical schedule creation/update to trigger final schedule refresh

matiasb:
Otherwise iCal schedules only refresh their cached final representation once a day (via periodic task).
Related to https://github.com/grafana/oncall/issues/3154


vadimkerr (approved):
LGTM

PR #3172, opened 2023-10-20
Added mechanism to discard older network requests

teodosii:
# What this PR does

- Added mechanism to prevent concurrency conflicts when multiple network requests are fired and the order of resolving is reversed. This resulted in flickering the results data, thus showing discrepancy between the filters (e.g. the search input) and the table data.

Within `RemoteFilters` and `UsersFilters` we're setting an ID for each network result, and whenever we get to hit the store, we'll compare that ID with the latest generated ID, thus if they match, we're all good, if they don't, that means that network request should be discarded and its data shouldn't be passed to the store.

This was implemented for:
- Users page
- Schedules page
(and more to come)

Fixes https://github.com/grafana/oncall-private/issues/2108


maskin25 (approved):

grafana-plugin/e2e-tests/users/usersActions.test.ts
```diff
@@ -61,6 +61,8 @@ test.describe('Users screen actions', () => {
     const { page } = adminRolePage;
     await goToOnCallPage(page, 'users');
 
+    await page.waitForTimeout(1000);
```
joeyorlando, line 64: is there a more deterministic way of waiting here? Waiting for a static amount of time feels like it could lead to flaky tests (ex. if for example the tests are run in a Cloud Grafana env and for whatever reasons network latency is a bit higher).

teodosii: From what I tested seems like there's a very fast flicker in the search response right in the beginning, but it is not noticeable no matter how many attempts you give. The timeout is kind of needed cause otherwise it will discard the search results, thus instead of like 1 or 2 results it will show you all, failing the tests. Maybe it could be less than 1000ms.

Worth investigating it at some point but I couldn't figure out what's causing that really

PR #3189, opened 2023-10-24
Allow OnCall API to use Grafana Service Accounts

mderynck:
# What this PR does
Allows public OnCall API to use Grafana service accounts for authorization.  In cloud requests using a Grafana service account token also needs to provide headers for  `X-Grafana-Org-Slug` and `X-Grafana-Instance-Slug`

This is **alpha** functionality, it may break or be removed in the future.  Going to use this on one endpoint (resolution notes) before we consider the implications across all of public API.

## Checklist


engine/apps/public_api/views/resolution_notes.py
joeyorlando, line 1: not sure if we already have auth related tests for the public API, if so I would recommend updating them to test that they work with both authn methods + maybe test that authorization works as we'd expect for service account tokens

engine/apps/grafana_plugin/helpers/client.py
```diff
@@ -274,6 +274,21 @@ def get_alerting_notifiers(self):
...
+        return self.api_post("api/serviceaccounts", {"name": name, "role": role})
+
+    def create_service_account_token(self, service_account_id, name, seconds_to_live=None):
+        token_config = {"name": name}
+        if seconds_to_live:
+            token_config["secondsToLive"] = seconds_to_live
+        return self.api_post(f"api/serviceaccounts/{service_account_id}/tokens", token_config)
+
+    def get_service_account_token_permissions(self):
+        return self.api_get("api/access-control/user/permissions")
```
joeyorlando, line 290: if you're feeling ambitious you could update the return types similar to [this](https://github.com/grafana/oncall/pull/3189/files#diff-910e9ef3a0f314a531500237fae412769bc6e4b9400ff604192d2a61f9336ec6R243) to strongly type the return payload

engine/apps/auth_token/grafana/grafana_auth_token.py
joeyorlando, line 1: _nit_ consider updating the type hints here

joeyorlando (approved):

engine/apps/auth_token/auth.py
```diff
@@ -262,3 +264,70 @@ def authenticate_credentials(
...
+        return user, auth_token
+
+    def determine_role_from_permissions(self, permissions):
+        if "plugins:write" in permissions:
+            return LegacyAccessControlRole.ADMIN
+        if "dashboards:write" in permissions:
+            return LegacyAccessControlRole.EDITOR
+        if "dashboards:read" in permissions:
+            return LegacyAccessControlRole.VIEWER
+        return LegacyAccessControlRole.NONE
```
joeyorlando, line 333: I would double check with the authnz team about this (+ consider adding a comment on why we need to do this)

engine/apps/public_api/views/resolution_notes.py
```diff
@@ -5,16 +5,27 @@
...
+
+    rbac_permissions = {
+        "metadata": [RBACPermission.Permissions.ALERT_GROUPS_READ],
+        "list": [RBACPermission.Permissions.ALERT_GROUPS_READ],
+        "retrieve": [RBACPermission.Permissions.ALERT_GROUPS_READ],
+        "create": [RBACPermission.Permissions.ALERT_GROUPS_WRITE],
+        "update": [RBACPermission.Permissions.ALERT_GROUPS_WRITE],
+        "partial_update": [RBACPermission.Permissions.ALERT_GROUPS_WRITE],
+        "destroy": [RBACPermission.Permissions.ALERT_GROUPS_WRITE],
+    }
```
joeyorlando, line 28: _nit_ consider adding some public api authnz tests

joeyorlando (approved):

Issue #2462, opened 2023-07-06
UI making too many requests


mderynck:
### What went wrong?

**What happened**:
  - Alert groups page keeps making new requests for stats and updates when previous ones are still pending
  - These requests are made every 15s, in the case any of these requests takes longer than 15s due to DB performance or network issues more requests can keep stacking up making the problem worse.

**What did you expect to happen**:
  - Alert groups page should not make new requests for updates while the same requests are already pending
  - Alert groups page should not make requests if it is not visible (not active tab or minimized)

### How do we reproduce it?

1. Open Grafana OnCall and navigate to Alert groups
2. Open dev tools, network page to watch requests
3. Simulate requests taking longer than 15s and observe

### Grafana OnCall Version

v1.3.7

### Product Area

Alert Flow & Configuration

### Grafana OnCall Platform?

None

joeyorlando:
Hosted Grafana had an incident recently w/ # of egress connections in Azure, and [part of the problem](https://raintank-corp.slack.com/archives/C05H0LEGCE5/p1689585617895639) was that we make too many requests.

joeyorlando:
Also, I think we should revisit the [`RootBaseStore.updateBasicData` function](https://github.com/grafana/oncall/blob/dev/grafana-plugin/src/state/rootBaseStore/index.ts#L108-L130). It makes 12 HTTP requests on initial plugin load. Most of them are for things that are not even relevant to the currently loaded page. Instead we should make these requests as needed.

PR #3205, opened 2023-10-27
Pagination refactoring/cleanup

### LINKED ISSUE #2931: Move pagination state to filters
Right now the pagination is stored separately for each page (alert gorups, schedules and so on). With the last iteration on Integration tweaks we migrated the pagination to the store, but the remaining is to do the same for all remaining pages and let the pagination live in the filters instead of the global store.

teodosii:
# What this PR does

- Pagination cleanup so that frontend no longer hardcodes page size
- Moved pagination to the filters store instead for all pages that have pagination enabled
- Prevent UI triggering polling if window is inactive or if the previous request didn't complete

#2931 
#2462

## Checklist


grafana-plugin/src/pages/incidents/Incidents.tsx
```diff
@@ -78,27 +76,29 @@ class Incidents extends React.Component<IncidentsPageProps, IncidentsPageState>
 
     const cursor = cursorQuery || undefined;
     const start = !isNaN(startQuery) ? Number(startQuery) : 1;
-    const itemsPerPage = !isNaN(perpageQuery) ? Number(perpageQuery) : ITEMS_PER_PAGE;
-
+    const end = !isNaN(perpageQuery) ? Number(perpageQuery) : undefined;
```
teodosii, line 79: This will set it to either the query's values or to 1 and undefined, where the end will be determined when fetching the results

grafana-plugin/src/pages/incidents/Incidents.tsx
```diff
@@ -78,27 +76,29 @@ class Incidents extends React.Component<IncidentsPageProps, IncidentsPageState>
...
-    store.alertGroupStore.updateBulkActions();
-    store.alertGroupStore.updateSilenceOptions();
   }
 
   private pollingIntervalId: NodeJS.Timer = undefined;
 
+  componentDidMount() {
+    const { alertGroupStore } = this.props.store;
+
+    alertGroupStore.updateBulkActions();
```
teodosii, line 98: These belong in the cDU and not in the constructor.

grafana-plugin/src/models/alertgroup/alertgroup.ts
```diff
@@ -44,7 +44,12 @@ export class AlertGroupStore extends BaseStore {
   incidentsItemsPerPage?: number;
 
   @observable
-  alertsSearchResult: any = {};
+  alertsSearchResult: {
+    prev?: string;
+    next?: string;
+    results?: any;
+    page_size?: number;
+  } = {};
```
brojd, line 52: Correct me if I'm wrong but the only place where we update this observable is this one:
image

So the code above assumes it is nested object like `Record<string, {...}>`. So I would say that either we update this type to reflect reality or we refactor updating this observable and getting its value to not use nested but a flat structure.

If I miss sth here or you want to discuss, pls ping me:)

brojd: And additionally seems that the only usage puts array of strings as `results`, can we reflect that in type instead of using `any`?

brojd (commented):
Few comments / questions about the code itself, I'm not ready to assess any aspect of the business logic though:)

grafana-plugin/src/pages/incidents/Incidents.tsx
```diff
@@ -526,33 +486,29 @@ class Incidents extends React.Component<IncidentsPageProps, IncidentsPageState>
...
-            next={next}
-            onChange={this.onChangeCursor}
-            onChangeItemsPerPage={this.handleChangeItemsPerPage}
-          />
-        </div>
+        {this.shouldShowPagination() && (
+          <div className={cx('pagination')}>
+            <CursorPagination
+              current={`${pagination.start}-${pagination.end}`}
+              itemsPerPage={alertGroupStore.incidentsItemsPerPage}
```
maskin25, line 495: hmm, why don't we use `page_size` from `alertGroupStore.getAlertSearchResult('default');` like


` const { results, prev, next, page_size } = alertGroupStore.getAlertSearchResult('default');`

brojd (approved):

grafana-plugin/src/pages/incidents/Incidents.tsx
```diff
@@ -649,6 +602,71 @@ class Incidents extends React.Component<IncidentsPageProps, IncidentsPageState>
...
+
+  getTableColumns(): Array<{ width: string; title: string; key: string; render }> {
+    const { store } = this.props;
+
+    return [
+      {
+        width: '140px',
+        title: 'Status',
+        key: 'time',
+        render: withSkeleton(this.renderStatus),
```
maskin25, line 623: AFAIK we can get rid of `withSkeleton` fn, because  alert can't contain attribute `short` anymore, backend removed this

grafana-plugin/src/pages/incidents/Incidents.tsx
```diff
@@ -719,13 +737,28 @@ class Incidents extends React.Component<IncidentsPageProps, IncidentsPageState>
...
   }
 
   setPollingInterval(filters: IncidentsFiltersType = this.state.filters, isOnMount = false) {
-    this.pollingIntervalId = setInterval(() => {
-      this.fetchIncidentData(filters, isOnMount);
-    }, POLLING_NUM_SECONDS * 1000);
+    const startPolling = (delayed = false) => {
+      this.pollingIntervalId = setTimeout(
+        async () => {
+          const isBrowserWindowInactive = document.hidden;
```
maskin25, line 747: nice idea! but are you stopping/starting polling on `document.hidden` change?

grafana-plugin/src/models/alertgroup/alertgroup.ts
```diff
@@ -294,21 +280,30 @@ export class AlertGroupStore extends BaseStore {
     // @ts-ignore
     this.alerts = new Map<number, Alert>([...this.alerts, ...newAlerts]);
 
+    this.incidentsItemsPerPage = page_size;
```
maskin25, line 283: do we really need to store this in alertgroupstore as a separate field? can we get it as `getSearchResult().page_size`?

grafana-plugin/src/pages/incidents/Incidents.tsx
teodosii: I could have done that but I just didn't want to add more logic (subscribe to event, unsubscribe etc) in the class. If the tab is inactive, it won't call the backend, if it becomes active again it will detect that in the `setTimeout`.

grafana-plugin/src/models/alertgroup/alertgroup.ts
teodosii: I guess you're right, let me try it :thinking:

grafana-plugin/src/pages/incidents/Incidents.tsx
```diff
@@ -526,33 +486,29 @@ class Incidents extends React.Component<IncidentsPageProps, IncidentsPageState>
...
-            onChange={this.onChangeCursor}
-            onChangeItemsPerPage={this.handleChangeItemsPerPage}
-          />
-        </div>
+        {this.shouldShowPagination() && (
+          <div className={cx('pagination')}>
+            <CursorPagination
+              current={`${pagination.start}-${pagination.end}`}
+              itemsPerPage={alertGroupStore.incidentsItemsPerPage}
+              itemsPerPageOptions={[
```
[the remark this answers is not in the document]
teodosii: Hmm unfortunately the paging on alert group is slightly different than what we have traditionally in the other places. You kind of need the ends of the pagination :thinking:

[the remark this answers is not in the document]
teodosii: Ahh nvm, that can be changed. I thought you were referring to the pagination state

grafana-plugin/src/pages/incidents/Incidents.tsx
teodosii: It's exactly the same value

grafana-plugin/src/models/alertgroup/alertgroup.ts
maskin25: weird but ok

PR #3208, opened 2023-10-27
address some minor direct paging backend issues

joeyorlando:

- Fixes an issue where if the user does not appear in the `UserHasNotification` query, we don't actually unpage the user and therefore they still show up in the `paged_users` array. (unpaging == creating a `AlertGroupLogRecord.TYPE_UNPAGE_USER` log record)
- Fixes an issue where if a user is paged multiple times, they would currently show up in `paged_users` > 1

## Checklist


engine/apps/alerts/paging.py
```diff
@@ -163,15 +163,16 @@ def unpage_user(alert_group: AlertGroup, user: User, from_user: User) -> None:
             ).select_for_update()[0]
```
joeyorlando, line 163: I'm not entirely sure the purpose of `UserHasNotification` but when testing locally there were several cases where `POST /unpage_user` wouldn't actually "unpage" the user because the log record was never created due to an `IndexError` being thrown.

This change ensures that this log record is always created.

clord (approved):

matiasb (approved):

PR #3209, opened 2023-10-27
Add GitHub action cron job to triage stale GH issues

joeyorlando:


.github/workflows/triage-stale-pull-requests.yml
```diff
@@ -25,3 +26,30 @@ jobs:
...
+          days-before-issue-close: 30
+          days-before-pr-stale: -1
+          days-before-pr-close: -1
+          ascending: true # start processing older issues first
+          stale-issue-label: "pr:stale"
+          exempt-issue-labels: "not-stale"
+          stale-issue-message: >
+            This issue has been automatically marked as stale because it has not had activity in the last
+            120 days. It will be closed in 30 days if no further activity occurs.
+          close-issue-message: >
```
Matvey-Kuk, line 54: Please don’t close them. I don’t think we had an agreement to close them automatically.

Matvey-Kuk (changes requested):

.github/workflows/triage-stale-pull-requests.yml
joeyorlando: I can't recall the exact consensus we reached. In simple terms this would:
- mark an issue stale after 4 months of inactivity, by adding the `pr:stale` label
- if after 30 more days (5 months total), it is still active, it would be closed **unless** we add the `not-stale` label to issues (these would be exempt)

PR #3128, opened 2023-10-05
Add responders improvements

joeyorlando:
# What this PR does

https://www.loom.com/share/c5e10b5ec51343d0954c6f41cfd6a5fb

## Summary of backend changes
- Add `AlertReceiveChannel.get_orgs_direct_paging_integrations` method and `AlertReceiveChannel.is_contactable` property. These are needed to be able to (optionally) filter down teams, in the `GET /teams` internal API endpoint ([here](https://github.com/grafana/oncall/pull/3128/files#diff-a4bd76e557f7e11dafb28a52c1034c075028c693b3c12d702d53c07fc6f24c05R55-R63)), to just teams that have a "contactable" Direct Paging integration
- `engine/apps/alerts/paging.py`
  -  update these functions to support new UX. In short `direct_paging` no longer takes a list of `ScheduleNotifications` or an `EscalationChain` object
  - add `user_is_oncall` helper function
  - add `_construct_title` helper function. In short if no `title` is provided, which is the case for Direct Pages originating from OnCall (either UI or Slack), then the format is `f"{from_user.username} is paging <team.name (if team is specified> <comma separated list of user.usernames> to join escalation"`
- `engine/apps/api/serializers/team.py` - add `number_of_users_currently_oncall` attribute to response schema ([code](https://github.com/grafana/oncall/pull/3128/files#diff-26af48f796c9e987a76447586dd0f92349783d6ea6a0b6039a2f0f28bd58c2ebR45-R52))
- `engine/apps/api/serializers/user.py` - add `is_currently_oncall` attribute to response schema ([code](https://github.com/grafana/oncall/pull/3128/files#diff-6744b5544ebb120437af98a996da5ad7d48ee1139a6112c7e3904010ab98f232R157-R162))
- `engine/apps/api/views/team.py` - add support for two new optional query params `only_include_notifiable_teams` and `include_no_team` ([code](https://github.com/grafana/oncall/pull/3128/files#diff-a4bd76e557f7e11dafb28a52c1034c075028c693b3c12d702d53c07fc6f24c05R55-R70))
- `engine/apps/api/views/user.py`
  - in the `GET /users` internal API endpoint, when specifying the `search` query param now also search on `teams__name` ([code](https://github.com/grafana/oncall/pull/3128/files#diff-30309629484ad28e6fe09816e1bd226226d652ea977b6f3b6775976c729bf4b5R223); this is a new UX requirement) 
[...]

PR #3220, opened 2023-10-30
few add responders patches

joeyorlando:

Closes https://github.com/grafana/support-escalations/issues/8143

Fix a few minor issues introduced in #3128:

- Fix slow `GET /users` internal API endpoint related to [this change](https://github.com/grafana/oncall/blob/dev/engine/apps/api/views/user.py#L239)
- Fix slow `GET /teams` internal API endpoint. Introduced a `short` query parameter that only invokes `apps.schedules.ical_utils.get_oncall_users_for_multiple_schedules` when `short=false`.
- Order results from `GET /teams` internal API endpoint by name (ascending)
- Fix search issue when searching for teams in the add responders popup window (this was strictly a frontend issue)
- CSS changes to add responders dropdown to fix lonnnggg results list:
  **Before**
  Screenshot 2023-10-31 at 10 06 20
  **After**
  Screenshot 2023-10-31 at 10 48 12

## Still todo

The `apps.schedules.ical_utils.get_oncall_users_for_multiple_schedules` method is still very slow when an instance has a lot of users (ex. `ops`). Ideally we should refactor this method to be more efficient because we still need to call this method under some circumstances. Ex. to populate this dropdown when Direct Paging a user (note that it didn't finish loading here on `ops`):
Screenshot 2023-10-30 at 18 14 59

## Checklist

### DIFF
--- a/engine/apps/api/views/user.py
+++ b/engine/apps/api/views/user.py
@@ -240,18 +240,18 @@ class UserView(
     def _is_currently_oncall_request(self) -> bool:
         return self._get_is_currently_oncall_query_param() in ["true", "false"]
 
-    def _is_short_oncall_request(self) -> bool:
+    def _is_long_request(self) -> bool:
         return self.request.query_params.get("short", "true").lower() == "false"
 
-    def _is_currently_oncall_or_short_request(self) -> bool:
-        return self._is_currently_oncall_request() or self._is_short_oncall_request()
+    def _is_currently_oncall_or_long_request(self) -> bool:
+        return self._is_currently_oncall_request() or self._is_long_request()
 
     def get_serializer_context(self):
         context = super().get_serializer_context()
         context.update(
             {
                 "schedules_with_oncall_users": self.schedules_with_oncall_users
-                if self._is_currently_oncall_or_short_request()
+                if self._is_currently_oncall_or_long_request()
                 else {}
             }
         )
@@ -268,7 +268,7 @@ class UserView(
 
         if is_list_request and is_filters_request:
             return self.get_filter_serializer_class()
-        elif is_list_request and self._is_currently_oncall_or_short_request():
+        elif is_list_request and self._is_currently_oncall_or_long_request():
             return UserLongSerializer
 
         is_users_own_data = kwargs.get("pk") is not None and kwargs.get("pk") == user.public_primary_key
--- a/engine/apps/api/views/user.py
+++ b/engine/apps/api/views/user.py
@@ -234,9 +234,27 @@ class UserView(
         """
         return get_oncall_users_for_multiple_schedules(self.request.user.organization.oncall_schedules.all())
 
+    def _get_is_currently_oncall_query_param(self) -> str:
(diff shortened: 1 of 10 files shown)


engine/apps/api/views/user.py
```diff
@@ -234,9 +234,27 @@ def schedules_with_oncall_users(self):
...
+        return self._is_currently_oncall_request() or self._is_short_oncall_request()
+
     def get_serializer_context(self):
         context = super().get_serializer_context()
-        context.update({"schedules_with_oncall_users": self.schedules_with_oncall_users})
+        context.update(
+            {
+                "schedules_with_oncall_users": self.schedules_with_oncall_users
+                if self._is_currently_oncall_or_short_request()
+                else {}
```
joeyorlando, line 255: main change, _only_ set `context["schedules_with_oncall_users"]` for certain requests. Previously we were _always_ setting it even if the serializer didn't need it (`UserLongSerializer` is the only serializer that needs this [here](https://github.com/grafana/oncall/pull/3220/files#diff-30309629484ad28e6fe09816e1bd226226d652ea977b6f3b6775976c729bf4b5R271-R272)).

engine/apps/api/views/user.py
```diff
@@ -234,9 +234,27 @@ def schedules_with_oncall_users(self):
...
         return get_oncall_users_for_multiple_schedules(self.request.user.organization.oncall_schedules.all())
 
+    def _get_is_currently_oncall_query_param(self) -> str:
+        return self.request.query_params.get("is_currently_oncall", "").lower()
+
+    def _is_currently_oncall_request(self) -> bool:
+        return self._get_is_currently_oncall_query_param() in ["true", "false"]
+
+    def _is_short_oncall_request(self) -> bool:
+        return self.request.query_params.get("short", "true").lower() == "false"
```
matiasb, line 244: This seems confusing..? `_is_short_oncall_request` will return `True` if the `short` param value is `"false"`?

joeyorlando: ahh good catch, thanks 👍

matiasb (approved):

grafana-plugin/src/models/grafana_team/grafana_team.ts
```diff
@@ -29,37 +31,32 @@ export class GrafanaTeamStore extends BaseStore {
...
   }
 
-  getSearchResult(query = '') {
-    if (!this.searchResult[query]) {
-      return [];
-    }
-
-    return this.searchResult[query].map((teamId: GrafanaTeam['id']) => this.items[teamId]);
+  getSearchResult() {
+    console.log('yoooo');
```
teodosii, line 59: :rofl:

teodosii (approved):

grafana-plugin/src/models/grafana_team/grafana_team.ts
joeyorlando: oops lol, good catch 😆

PR #3256, opened 2023-11-02
remove AlertGroup.is_restricted column

joeyorlando:
# What this PR does

Removes deprecated/not-referenced `AlertGroup.is_restricted` column. Per our dev docs on "[Removing a nullable field from a model](https://github.com/grafana/oncall/blob/dev/dev/README.md#removing-a-nullable-field-from-a-model)" this is the 1st of two migration files. The [second migration file](https://github.com/grafana/oncall/files/13255276/0038_remove_alertgroup_is_restricted_db.py.zip) will be added in a PR in a subsequent PR/release.


engine/apps/alerts/models/alert_group.py
```diff
@@ -389,9 +389,6 @@ def status(self):
     # https://code.djangoproject.com/ticket/28545
     is_open_for_grouping = models.BooleanField(default=None, null=True, blank=True)
 
-    # TODO: drop this column in an upcoming release
-    is_restricted = models.BooleanField(default=False, null=True)
```
matiasb, line 393: Should we probably try the suggested steps?
https://github.com/grafana/oncall/blob/dev/dev/README.md#removing-a-nullable-field-from-a-model

joeyorlando: good point 👍

matiasb (approved):

PR #3274, opened 2023-11-03
Simplify integrations cache on startupprobe

matiasb:
Revert logic to detect integrations server on startupprobe and simplify logic (and rework failing test having issues with urlpatterns cache and reset), since cache is the same for engine and integrations, and it is populated once.


engine/engine/tests/test_views.py
```diff
@@ -39,4 +27,4 @@ def test_startupprobe_populates_integrations_cache(settings, detached_integratio
         response = client.get("/startupprobe/")
 
     assert response.status_code == status.HTTP_200_OK
-    assert mock_update_cache.called == is_cache_updated
+    assert mock_update_cache.called
```
vadimkerr, line 30: _nit_: maybe use `assert_called_once`? 🤓

vadimkerr (approved):

engine/engine/tests/test_views.py
```diff
@@ -39,4 +27,4 @@ def test_startupprobe_populates_integrations_cache(settings, detached_integratio
         response = client.get("/startupprobe/")
 
     assert response.status_code == status.HTTP_200_OK
-    assert mock_update_cache.called == is_cache_updated
+    mock_update_cache.assert_called_once
```
joeyorlando, line 30: I believe `assert_called_once` needs to be invoked as it is a method ([docs](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called_once))

Issue #1468, opened 2023-03-07
Fix "Grafana Cloud OnCall" wording


Konstantinov-Innokentii:
We have different naming for "connect to cloud" feature in Phone Verification (With cloud notifications enabled) and in mobile app connection. For cloud phone notifications we suggest user to connect/use/etc. to Grafana Cloud, but for the mobile app we are naming that Cloud OnCall. In env variables tab we have setting name "GRAFANA_CLOUD_NOTIFICATIONS_ENABLED" and description "Enable SMS/call notifications via Grafana Cloud OnCall".
  So, we have three different names for one thing ( Grafana Cloud, Cloud OnCall, Grafana Cloud OnCall).
  Upd: In docs I found that for mobile app we are asking users to connect to Grafana Cloud ([link](https://grafana.com/docs/oncall/latest/mobile-app/#connect-to-your-open-source-grafana-oncall-account)) And in docs [about cloud phone calls](https://grafana.com/docs/oncall/latest/mobile-app/#connect-to-your-open-source-grafana-oncall-account) we use both Cloud OnCall and Grafana Cloud.
  
  **So, the task is to name it "Grafana Cloud OnCall" everywhere**

alyssawada:
Closing as complete

PR #3279, opened 2023-11-06
Unify Grafana Cloud OnCall wording

brojd:
# What this PR does

- Unify naming in the app of Grafana Cloud / Cloud OnCall / Grafana Cloud OnCall so that it's always **Grafana Cloud OnCall** 
- Unify naming the same way in docs

https://github.com/grafana/oncall/issues/1468

## Checklist


grafana-plugin/src/pages/settings/tabs/Cloud/CloudPage.tsx
```diff
@@ -15,7 +15,7 @@ import { WithStoreProps } from 'state/types';
 import { useStore } from 'state/useStore';
 import { withMobXProviderContext } from 'state/withStore';
 import { openErrorNotification } from 'utils';
-import { determineRequiredAuthString, UserActions } from 'utils/authorization';
+import { UserActions, determineRequiredAuthString } from 'utils/authorization';
```
maskin25, line 18: just curious...do you intentionally change import order? (There are several cases in your PR)

maskin25 (approved):

grafana-plugin/src/pages/settings/tabs/Cloud/CloudPage.tsx
brojd: VS Code has now builtin capability to automatically sort & remove imports on save and I have it enabled. 
https://dev.to/shanesc/how-to-sort-and-cleanup-imports-on-save-in-vs-code-52p1 

I guess it might be worth to run it across the whole codebase in one PR at some point and then maybe agree that we all use it. Then we will have all the imports automatically sorted on save and such diffs will not be present in PRs

maskin25: not everybody in our team uses VS Code,  we use  eslint `import` plugin for that (see details here https://github.com/grafana/oncall/blob/dev/grafana-plugin/.eslintrc.js#L13), I propose to keep using eslint for such things

Konstantinov-Innokentii (approved):

grafana-plugin/src/pages/settings/tabs/Cloud/CloudPage.tsx
brojd: I agree if not everyone is using VS Code, we should definitely go for solution that will be consistent for everyone, no matter what IDE we're using.

But maybe inn this case maybe worth to consider https://www.npmjs.com/package/prettier-plugin-organize-imports ? 

I mean typically you format code with Prettier on save. You don't run `eslint --fix` on save, right? And I guess it's better to have those imports organized/sorted/deleted when you code rather than on some precommit hook or on CI

maskin25: I use precommit hook for formatting, it' ok for me...yes, we can start using https://www.npmjs.com/package/prettier-plugin-organize-imports but as for me it is optional, if you think this will improve our DX then feel free to add a PR for that =)

brendamuir (approved):

PR #3280, opened 2023-11-06
add get_cached_oncall_users_for_multiple_schedules method + add DELETE /alertgroups/<id> internal API endpoint

joeyorlando:
# What this PR does

This PR:
- adds a new method `apps.schedules.ical_utils.get_cached_oncall_users_for_multiple_schedules`. In short this method basically adds a layer of caching on top of `apps.schedules.ical_utils.get_oncall_users_for_multiple_schedules`. We store one cache value for each schedule. Cache results are stored for 15 minutes. To me this feels like a good balance between improving performance + returning stale results. Cache values are stored as:
  - key = `f"schedule_{schedule.public_primary_key}_oncall_users"`
  - value = `[user.public_primary_key for user in schedule.currently_oncall_users]`
- adds a `DELETE /alertgroups/<id>` internal API endpoint (needed by Grafana Incident for the Add Responders integration)
- updates the `is_currently_oncall` query parameter for the `GET /users` internal API endpoint to return ALL users when the query param value `== "all"`

https://github.com/grafana/oncall-private/issues/2264

## Checklist


grafana-plugin/src/containers/AddResponders/parts/AddRespondersPopup/AddRespondersPopup.tsx
```diff
@@ -79,7 +79,7 @@ const AddRespondersPopup = observer(
         }
         setVisible(false);
       },
-      [isCreateMode, userStore, directPagingStore, setShowUserConfirmationModal, setVisible]
```
joeyorlando, line 82: - `userStore` is not consumed within the method passed to `useCallback` here
- the method accessed on `directPagingStore` will not change and thereby doesn't need to be passed as a dep here

joeyorlando:
**TODO**
- [ ] finish WIP tests for `apps.schedules.ical_utils.get_cached_oncall_users_for_multiple_schedules`

engine/apps/api/views/alert_group.py
```diff
@@ -469,6 +470,11 @@ def enrich(self, alert_groups):
 
         return alert_groups
 
+    def destroy(self, request, *args, **kwargs):
+        instance = self.get_object()
+        delete_alert_group.apply_async((instance.pk, request.user.pk))
```
joeyorlando, line 475: follows similar logic to our `DELETE /alertgroups/<id>` [public API endpoint handler](https://github.com/grafana/oncall/blob/dev/engine/apps/public_api/views/incidents.py#L112) (don't think we need to implement the concept of "wipe" on the internal API endpoint for now)

mderynck (approved):

PR #3281, opened 2023-11-06
Alert group column/label selector

teodosii:
# What this PR does

Adds new functionality to enable which columns should show on the alert group page

image


grafana-plugin/src/utils/decorators.ts
```diff
@@ -0,0 +1,27 @@
+import { LoaderStore } from 'models/loader/loader';
+
+export function AutoLoadingState(actionKey: string) {
```
teodosii, line 3: This is newly added, I think it will really enhance UX experience if we start providing feedback when we wait for backend. This is to wrap class methods

grafana-plugin/src/utils/consts.ts
```diff
@@ -54,3 +54,5 @@ export enum PAGE {
 }
 
 export const TEXT_ELLIPSIS_CLASS = 'overflow-child';
+
+export const INCIDENT_HORIZONTAL_SCROLLING_STORAGE = 'isIncidentalTableHorizontalScrolling';
```
teodosii, line 58: Constant for storing the key (whether horizontal scrolling is enabled or not) in local storage. Can be moved to some specific filename if we add more I guess

grafana-plugin/jest.setup.ts
```diff
@@ -19,3 +19,35 @@ Object.defineProperty(window, 'matchMedia', {
     dispatchEvent: jest.fn(),
   })),
 });
+
+const global = window as any;
+
+global.ResizeObserver = class ResizeObserver {
```
brojd, line 25: We're already using another way of mocking window props that are not supported by JSDom so I suggest we use the same approach which is:
```
Object.defineProperty(window, 'ResizeObserver', {
  writable: true,
  value: class ResizeObserver {
    constructor(callback: ResizeObserverCallback) {
      setTimeout(() => {
        callback(
          [
            {
              contentRect: {
                x: 1,
                y: 2,
                width: 500,
                height: 500,
                top: 100,
                bottom: 0,
                left: 100,
                right: 0,
              },
              target: {},
            } as ResizeObserverEntry,
          ],
          this
        );
      });
    }
    observe() {}
    disconnect() {}
    unobserve() {}
  },
});
```

grafana-plugin/src/models/loader/action-keys.ts
```diff
@@ -0,0 +1,5 @@
+export enum ActionKey {
+  IS_ADDING_NEW_COLUMN_TO_ALERT_GROUP = 'IS_ADDING_NEW_COLUMN_TO_ALERT_GROUP',
```
teodosii, line 2: In the near future we can probably group this by the page name or something simillar

grafana-plugin/src/assets/style/utils.css
```diff
@@ -117,6 +117,10 @@
  * Icons
  */
 
+.loader {
+  margin-bottom: 0 !important;
```
brojd, line 121: Could you explain why do we need !important? Where we have a problem with precedence?

grafana-plugin/src/components/WithContextMenu/WithContextMenu.tsx
```diff
@@ -59,4 +63,8 @@ export const WithContextMenu: React.FC<WithContextMenuProps> = ({
       )}
     </div>
   );
+
+  function isContextMenuOpen() {
+    return isOpen === undefined ? isMenuOpen : isOpen;
+  }
```
brojd, line 69: This is very confusing :D We have now `isMenuOpen`, `forceIsOpen` and `isOpen` (that is another way of forcing opening/closing). If we're not able to simplify that in scope of this PR, can we create an issue right away to simplify that and put it in the sprint backlog to avoid longlasting TODOs?

teodosii: I'll revert these changes, these were some experiments in attempting to use this component instead

grafana-plugin/src/assets/style/utils.css
teodosii: So that it overwrites styling from grafana

grafana-plugin/src/containers/ColumnsSelector/ColumnsSelector.styles.ts
```diff
@@ -0,0 +1,93 @@
+import { css } from '@emotion/css';
+import { GrafanaTheme2 } from '@grafana/data';
+
+export const getColumnsSelectorStyles = (_theme: GrafanaTheme2) => {
```
brojd, line 4: 1) Can we use object literal syntax like css({ .... })? just as here: https://github.com/grafana/grafana/blob/main/contribute/style-guides/styling.md 
2) Shouldn't we use theme values for whitespaces (margins, paddings, gaps etc)?

grafana-plugin/src/containers/ColumnsSelector/ColumnsSelector.tsx
```diff
@@ -0,0 +1,273 @@
...
+      <div className={styles.columnItem} ref={columnElRef}>
+        <span className={styles.columnName}>{column.name}</span>
+
+        {column.type === AGColumnType.LABEL && (
+          <Tooltip content="Label Column">
+            <Icon aria-label="Label" name="tag-alt" className={styles.labelIcon} />
+          </Tooltip>
+        )}
+
+        {column.isVisible ? (
```
brojd, line 69: can we pls avoid nested ternaries? Also what can be used is `<RenderConditionally />` component, wdyt?

grafana-plugin/src/containers/ColumnsSelector/ColumnsSelector.tsx
```diff
@@ -0,0 +1,273 @@
...
+    const canResetData = useMemo(() => !isEqual(columns, getDefaultData()), [alertGroupStore.columns]);
+
+    const sensors = useSensors(
+      useSensor(PointerSensor),
+      useSensor(KeyboardSensor, {
+        coordinateGetter: sortableKeyboardCoordinates,
+      })
+    );
+
+    const isResetLoading = LoaderStore.isLoading(ActionKey.IS_RESETING_COLUMNS_FROM_ALERT_GROUP);
```
brojd, line 127: We shouldn't consume the store directly by importing the class but rather through `useStore`. We can also create a custom hook to simplify that. Take a look here: https://github.com/brojd/todo-app/blob/main/frontend/src/store/store.hooks.ts

grafana-plugin/src/containers/ColumnsSelector/ColumnsSelector.tsx
```diff
@@ -0,0 +1,273 @@
...
+        </div>
+      </div>
+    );
+
+    async function onReset() {
+      const columnsDefaultValues = getDefaultData();
+
+      return alertGroupStore
+        .updateTableSettings(columnsDefaultValues, true)
+        .then(() => alertGroupStore.fetchTableSettings());
```
brojd, line 215: maybe we can leverage `await` instead of `then`?

grafana-plugin/src/containers/ColumnsSelector/ColumnsSelector.styles.ts
teodosii: 1. I feel like it's more closer to native css this way... we can discuss this further but it just feels more naturally writing it this way to me :thinking: 
2. I don't think we need to rely on these, the plan was to stick to emotion so that we fetch theme variables that are most important like colors, background colors etc and not hardcode them in css files. Margins don't scale all the time, could be 4, could be 6, could be a multiple of whatever X. For example there could be cases where you just need to adjust your margins because grafana appends too much on some elements, that margin may very well not be a multiple of a standard X value

grafana-plugin/src/containers/ColumnsSelectorWrapper/ColumnsModal.tsx
```diff
@@ -0,0 +1,228 @@
...
+
+    async function onAddNewColumns() {
+      const mergedColumns = [
+        ...store.alertGroupStore.columns,
+        ...searchResults
+          .filter((item) => item.isChecked)
+          .map(
+            (it): AGColumn => ({
+              id: it.id,
+              name: it.name,
```
brojd, line 188: Small thing and general notice. It's a good practice to use more explicit names and prevent such shortcuts. Here we're iterating over "results", then in filter we call a single entry an "item" and then an "it". Maybe we can call it a "result"?

grafana-plugin/src/models/loader/action-keys.ts
brojd, line 2: I think we should simplify the format for example "ALERT_GROUP-ADD_NEW_COLUMN". The "IS_" prefix makes it difficult to reuse this enum for other stuff, for example an error of the same action.

grafana-plugin/src/models/loader/loader.ts
```diff
@@ -0,0 +1,20 @@
...
+
+interface LoadingResult {
+  [key: string]: boolean;
+}
+
+class LoaderStoreClass {
+  @observable
+  items: LoadingResult = {};
+
+  setLoadingAction(actionKey: string, isLoading: boolean) {
```
brojd, line 11: This should be annotated as @action

brojd: I think we should support LoadingState as union, not boolean. We might have more statuses, e.g. error. Example here: https://github.com/brojd/todo-app/blob/main/frontend/src/store/store.types.ts

grafana-plugin/src/pages/incidents/Incidents.tsx
```diff
@@ -423,27 +448,46 @@ class Incidents extends React.Component<IncidentsPageProps, IncidentsPageState>
...
-              variant="primary"
-              onClick={this.onIncidentsUpdateClick}
-            >
-              Refresh
-            </Button>
           </div>
-        )}
+
+          <div className={cx('fields-dropdown')}>
+            {hasInvalidatedAlert && (
```
brojd, line 454: just a small alternative to use RenderConditionally in this file as well

grafana-plugin/src/pages/incidents/Incidents.tsx
```diff
@@ -423,27 +448,46 @@ class Incidents extends React.Component<IncidentsPageProps, IncidentsPageState>
...
+              <HorizontalGroup spacing="xs">
+                <Text type="secondary">Results out of date</Text>
+                <Button
+                  style={{ marginLeft: '8px' }}
```
brojd, line 458: do we need to use inline styles?

grafana-plugin/src/pages/incidents/Incidents.tsx
```diff
@@ -423,27 +448,46 @@ class Incidents extends React.Component<IncidentsPageProps, IncidentsPageState>
...
+
+            {store.hasFeature(AppFeature.Labels) && (
+              <RadioButtonGroup
+                options={[
+                  { value: false, icon: 'wrap-text' },
+                  {
+                    value: true,
+                    icon: 'arrow-from-right',
+                  },
+                ]}
```
brojd, line 476: Doesn't matter in one case but doing that across the codebase matters. In general if we have value of nonprimitive like object/array that doesn't depend on anything from within the component body, it's good to move it outside the render. Otherwise the array is recreated on each render and can cause additional unneeded rerenders of the child

grafana-plugin/src/containers/ColumnsSelector/ColumnsSelector.styles.ts
brojd: So regarding point 2) - as we are not using theme vars now, let's leave it. We will get back to this topic after full migration to emotion.

Regarding 1st point, I understand that at first it might seem more naturally to write it with string syntax because we're using plain css in the project right now but I believe after the switch it's only the matter of time to get use to writing styles in js/ts way. Two reasons why I would suggest the latter are:
1) Intellisense + type safety
image
2) At the end we're a part of Grafana and such style guides are created for a reason. If it is explicitly recommended to use a functional way and unless we have a solid argument to not follow it, why not to do it.

@maskin25 maybe you can share your thoughts?

Btw. it doesn't matter that much for this PR what we decide but we need to have this decision before migrating the whole codebase

grafana-plugin/package.json
```diff
@@ -98,9 +98,9 @@
     "moment-timezone": "^0.5.35",
     "plop": "^2.7.4",
     "postcss-loader": "^7.0.1",
-    "react": "17.0.2",
-    "react-dom": "17.0.2",
-    "react-test-renderer": "^17.0.2",
+    "react": "18.2.0",
```
brojd, line 101: awesome:)

grafana-plugin/src/pages/incidents/Incidents.tsx
teodosii: Nope, but it was already there :D

grafana-plugin/src/containers/ColumnsSelector/ColumnsSelector.tsx
```diff
@@ -0,0 +1,273 @@
...
+  };
+
+  return (
+    <div ref={setNodeRef} style={{ ...style }} className={styles.columnRow}>
+      <div className={styles.columnItem} ref={columnElRef}>
+        <span className={styles.columnName}>{column.name}</span>
+
+        {column.type === AGColumnType.LABEL && (
+          <Tooltip content="Label Column">
+            <Icon aria-label="Label" name="tag-alt" className={styles.labelIcon} />
```
brojd, line 65: if I'm not wrong removing this class and related !importants change nothing and the icon looks the same:
image

grafana-plugin/src/components/WithContextMenu/WithContextMenu.tsx
```diff
@@ -5,6 +5,8 @@ import { ContextMenu } from '@grafana/ui';
 export interface WithContextMenuProps {
   children: (props: { openMenu: React.MouseEventHandler<HTMLElement> }) => JSX.Element;
   renderMenuItems: ({ closeMenu }: { closeMenu?: () => void }) => React.ReactNode;
+  isOpen?: boolean;
```
brojd, line 8: I guess it's not needed anymore

grafana-plugin/src/containers/ColumnsSelector/ColumnsSelector.styles.ts
brojd: Also first point in best practices from Emotion docs: https://emotion.sh/docs/best-practices#use-typescript-and-object-styles

brojd (approved):

engine/apps/user_management/utils.py
```diff
@@ -0,0 +1,35 @@
...
+if typing.TYPE_CHECKING:
+    from apps.user_management.models import User
+
+
+def default_columns() -> typing.List[AlertGroupTableColumn]:
+    columns = [
+        {"name": column.label, "id": column.value, "type": AlertGroupTableColumnTypeChoices.DEFAULT.value}
+        for column in AlertGroupTableDefaultColumnChoices
+    ]
+    return columns
```
vadimkerr, line 19: maybe this function would better fit in `apps/user_management/constants.py` along with `AlertGroupTableDefaultColumnChoices`?

engine/apps/user_management/utils.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def alert_group_table_user_settings(user: "User") -> AlertGroupTableColumns:
+    organization_columns = user.organization.alert_group_table_columns
+    visible_columns: typing.List[AlertGroupTableColumn]
```
vadimkerr, line 24: Maybe type [`user.alert_groups_table_selected_columns`](https://github.com/grafana/oncall/pull/3281/files#diff-c569330343b5bb0d5520d3c300e93b6d3d0fd07c73509ab3354161eda10e2be4R240) to be `list[AlertGroupTableColumns]` so this additional annotation wouldn't be needed?

engine/apps/user_management/utils.py
```diff
@@ -0,0 +1,35 @@
...
+    if user.alert_groups_table_selected_columns:
+        visible_columns = [
+            column for column in user.alert_groups_table_selected_columns if column in organization_columns
+        ]
+    else:
+        visible_columns = default_columns()
+    user.update_alert_group_table_columns_settings(visible_columns)
+    hidden_columns: typing.List[AlertGroupTableColumn] = [
```
vadimkerr, line 32: same as the comment above

engine/apps/user_management/utils.py
```diff
@@ -0,0 +1,35 @@
...
+def alert_group_table_user_settings(user: "User") -> AlertGroupTableColumns:
+    organization_columns = user.organization.alert_group_table_columns
+    visible_columns: typing.List[AlertGroupTableColumn]
+    if user.alert_groups_table_selected_columns:
+        visible_columns = [
+            column for column in user.alert_groups_table_selected_columns if column in organization_columns
+        ]
+    else:
+        visible_columns = default_columns()
+    user.update_alert_group_table_columns_settings(visible_columns)
```
vadimkerr, line 31: question: why it's necessary to make an update here? is it possible to avoid DB writes for this function?

engine/apps/user_management/utils.py
vadimkerr, line 19:
```suggestion
    return [
        {"name": column.label, "id": column.value, "type": AlertGroupTableColumnTypeChoices.DEFAULT.value}
        for column in AlertGroupTableDefaultColumnChoices
    ]
```

engine/apps/user_management/migrations/0018_auto_20231115_1206.py
```diff
@@ -0,0 +1,27 @@
...
+        ('user_management', '0017_alter_organization_maintenance_author'),
+    ]
+
+    operations = [
+        linter.IgnoreMigration(),
+        migrations.AddField(
+            model_name='organization',
+            name='alert_group_table_columns',
+            field=models.JSONField(default=apps.user_management.utils.default_columns),
+        ),
```
vadimkerr, line 21: does this field have to be non-nullable? maybe it makes sense to make it nullable?

```python
models.JSONField(null=True, default=apps.user_management.utils.default_columns)
```

engine/apps/api/views/alert_group_table_settings.py
```diff
@@ -0,0 +1,52 @@
...
+
+    rbac_permissions = {
+        "get_columns": [RBACPermission.Permissions.ALERT_GROUPS_READ],
+        "update_columns_settings": [RBACPermission.Permissions.ALERT_GROUPS_READ],
+        "update_columns_list": [RBACPermission.Permissions.OTHER_SETTINGS_WRITE],
+    }
+
+    def get_columns(self, request: Request) -> Response:
+        user = request.user
+        return Response(alert_group_table_user_settings(user))
```
vadimkerr, line 27:
```suggestion
        return Response(alert_group_table_user_settings(request.user))
```

engine/apps/api/views/alert_group_table_settings.py
```diff
@@ -0,0 +1,52 @@
...
+        organization = request.auth.organization
+        serializer = AlertGroupTableColumnsListSerializer(
+            data=request.data, context={"request": request, "is_org_settings": True}
+        )
+        serializer.is_valid(raise_exception=True)
+        columns: typing.List[AlertGroupTableColumn] = serializer.validated_data.get(
+            "visible", []
+        ) + serializer.validated_data.get("hidden", [])
+        organization.update_alert_group_table_columns(columns)
+        return Response(alert_group_table_user_settings(user))
```
vadimkerr, line 41:
```suggestion
        serializer = AlertGroupTableColumnsListSerializer(
            data=request.data, context={"request": request, "is_org_settings": True}
        )
        serializer.is_valid(raise_exception=True)
        columns: typing.List[AlertGroupTableColumn] = serializer.validated_data.get(
            "visible", []
        ) + serializer.validated_data.get("hidden", [])
        request.auth.organization.update_alert_group_table_columns(columns)
        return Response(alert_group_table_user_settings(request.user))
```

engine/apps/api/serializers/alert_group_table_settings.py
```diff
@@ -0,0 +1,65 @@
...
+
+        `is_org_settings=True` means that organization alert group table columns list should be updated.
+        Validate that all default columns are in the list.
+
+        `is_org_settings=False` means that list of visible columns for user should be updated.
+        Validate that all columns exist in organization alert group table columns list.
+        """
+        is_org_settings = self.context.get("is_org_settings") is True
+        organization = self.context["request"].auth.organization
+        columns_list = data["visible"] + data["hidden"]
```
vadimkerr, line 52:
```suggestion
        columns = data["visible"] + data["hidden"]
```

engine/apps/api/serializers/alert_group_table_settings.py
```diff
@@ -0,0 +1,65 @@
+from rest_framework import serializers
+from rest_framework.exceptions import ValidationError
+
+from apps.user_management.constants import AlertGroupTableColumnTypeChoices, AlertGroupTableDefaultColumnChoices
+
+
+class ColumnIdField(serializers.Field):
```
Konstantinov-Innokentii, line 7: Let's make column_id string everywhere

engine/apps/api/serializers/alert_group_table_settings.py
```diff
@@ -0,0 +1,65 @@
...
+        `is_org_settings=False` means that list of visible columns for user should be updated.
+        Validate that all columns exist in organization alert group table columns list.
+        """
+        is_org_settings = self.context.get("is_org_settings") is True
+        organization = self.context["request"].auth.organization
+        columns_list = data["visible"] + data["hidden"]
+        request_columns_ids = [column["id"] for column in columns_list]
+        if len(data["visible"]) == 0:
+            raise ValidationError("At least one column should be selected as visible")
+        if is_org_settings:
```
vadimkerr, line 56: looks like the variable `is_org_settings` is not used anywhere else?

```suggestion
        if self.context.get("is_org_settings"):
```

engine/apps/api/serializers/alert_group_table_settings.py
```diff
@@ -0,0 +1,65 @@
...
+    name = serializers.CharField(max_length=200)
+    id = ColumnIdField()
+    type = serializers.ChoiceField(choices=AlertGroupTableColumnTypeChoices.choices)
+
+    def validate(self, data):
+        self._validate_id(data)
+        return data
+
+    def _validate_id(self, data):
+        if (
```
Konstantinov-Innokentii, line 28: Could you please add a comment here

engine/apps/api/serializers/alert_group_table_settings.py
```diff
@@ -0,0 +1,65 @@
...
+        request_columns_ids = [column["id"] for column in columns_list]
+        if len(data["visible"]) == 0:
+            raise ValidationError("At least one column should be selected as visible")
+        if is_org_settings:
+            if not set(request_columns_ids) >= set(AlertGroupTableDefaultColumnChoices.values):
+                raise ValidationError("Default column cannot be removed")
+            elif len(request_columns_ids) > len(set(request_columns_ids)):
+                raise ValidationError("Duplicate column")
+        else:
+            organization_columns_ids = [column["id"] for column in organization.alert_group_table_columns]
```
vadimkerr, line 62:
```suggestion
            organization_columns_ids = [column["id"] for column in self.context["request"].auth.organization.alert_group_table_columns]
```

engine/apps/api/serializers/alert_group_table_settings.py
```diff
@@ -0,0 +1,65 @@
...
+        and validate that at least one column is selected as visible.
+
+        `is_org_settings=True` means that organization alert group table columns list should be updated.
+        Validate that all default columns are in the list.
+
+        `is_org_settings=False` means that list of visible columns for user should be updated.
+        Validate that all columns exist in organization alert group table columns list.
+        """
+        is_org_settings = self.context.get("is_org_settings") is True
+        organization = self.context["request"].auth.organization
```
vadimkerr, line 51:
```suggestion
```

engine/apps/api/serializers/alert_group_table_settings.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def _validate_id(self, data):
+        if (
+            data["type"] == AlertGroupTableColumnTypeChoices.DEFAULT.value
+            and data["id"] not in AlertGroupTableDefaultColumnChoices.values
+        ):
+            raise ValidationError("Invalid column id format")
+
+
+class AlertGroupTableColumnsListSerializer(serializers.Serializer):
```
Konstantinov-Innokentii, line 35: I propose to split this serializer in two.

grafana-plugin/src/containers/ColumnsSelector/ColumnsSelector.styles.ts
maskin25: 1. object literal syntax looks more convenient, in general we should use it to make Intellisense and ts checking work
2. I think we need to arrange a session with @raphael-batte to discuss margins/paddings unification to be consistent with Grafana

grafana-plugin/src/containers/TeamsList/TeamsList.tsx
```diff
@@ -170,7 +170,7 @@ const TeamModal = ({ teamId, onHide }: TeamModalProps) => {
                   { label: 'Team members and admins', value: '0' },
                 ]}
                 value={shareResourceToAll}
-                onChange={setShareResourceToAll}
+                onChange={setShareResourceToAll as (res: string) => void}
```
maskin25, line 173: can we don't use type casting here?

grafana-plugin/src/models/alertgroup/alertgroup.ts
```diff
@@ -72,6 +74,9 @@ export class AlertGroupStore extends BaseStore {
   @observable
   liveUpdatesPaused = false;
 
+  @observable
+  columns: AGColumn[] = [];
```
maskin25, line 78: it's better to avoid abbreviations (`AlertGroupColumn`)

grafana-plugin/src/utils/decorators.ts
maskin25: hmm, why is function name capitalised?

engine/apps/user_management/migrations/0018_auto_20231115_1206.py
Ferril: In this case migration will set value `None` instead of default for existing organizations

engine/apps/user_management/utils.py
Ferril: it's used for lazy update user settings. Another option - update it for all users at the background on changing organization settings

engine/apps/user_management/utils.py
Ferril: Sorry, didn't get it, where do you mean to type it?

vadimkerr:
```python3
alert_groups_table_selected_columns: list[AlertGroupTableColumn] | None = models.JSONField(default=None, null=True)
```

engine/apps/user_management/migrations/0018_auto_20231115_1206.py
Ferril: Oh, got it, thanks. It makes sense to add lazy update for organization settings then, since default value works incorrectly with nullable field

Konstantinov-Innokentii (approved):

grafana-plugin/src/containers/TeamsList/TeamsList.tsx
teodosii: the setter is a React.Dispatch, while the `onChange` expects something else

PR #3289, opened 2023-11-07
Update test to avoid failure depending on when it is run

matiasb:


joeyorlando (approved):

engine/apps/schedules/tests/test_on_call_schedule.py
```diff
@@ -2698,12 +2698,13 @@ def test_shifts_for_user_only_two_users_with_shifts(
 
     now = timezone.now()
     today = now.replace(hour=0, minute=0, second=0, microsecond=0)
+    tomorrow = today + timezone.timedelta(days=1)
     start_date = today - timezone.timedelta(days=2)
     days = 7
 
     data = {
-        "start": now + timezone.timedelta(hours=1),
```
matiasb, line 2705: When now was too close to EOD, +1h would start the shift the next day.

PR #3291, opened 2023-11-07
Improve user permissions query

matiasb:
The query for checking a user permission (used to get users from a Slack usergroup, for example) is timing out (and generating retries, besides affecting some use cases: [logs](https://ops.grafana-ops.net/explore?panes=%7B%22FCQ%22:%7B%22datasource%22:%22c-R8UWvVk%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22%7Bnamespace%3D%5C%22amixr-prod%5C%22,%20cluster%3D%5C%22prod-us-central-0%5C%22%7D%20%7C%3D%20%5C%22Timeout%20exceeded%20in%20regular%20expression%20match.%5C%22%22,%22queryType%22:%22range%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22c-R8UWvVk%22%7D,%22editorMode%22:%22code%22%7D%5D,%22range%22:%7B%22from%22:%22now-6h%22,%22to%22:%22now%22%7D%7D%7D&schemaVersion=1&orgId=1)):

`django.db.utils.OperationalError: (3699, 'Timeout exceeded in regular expression match.')`

Change to a `contains` query except for SQLite (not supported), where a simplified version of the original regex query is used.

### DIFF
--- a/engine/apps/user_management/models/user.py
+++ b/engine/apps/user_management/models/user.py
@@ -1,6 +1,7 @@
 import datetime
 import json
 import logging
+import re
 import typing
 from urllib.parse import urljoin
 
@@ -32,6 +33,10 @@ if typing.TYPE_CHECKING:
     from apps.user_management.models import Organization, Team
 
 logger = logging.getLogger(__name__)
+
+
+class PermissionsQuery(typing.TypedDict):
+    permissions__contains: typing.Dict
 
 
 class PermissionsRegexQuery(typing.TypedDict):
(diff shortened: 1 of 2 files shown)


joeyorlando (approved):
_nit_ consider adding a small unit test for this, otherwise LGTM 👍

matiasb:
> _nit_ consider adding a small unit test for this, otherwise LGTM 👍

There are already tests for this, not changing behavior with the update (so existing tests should keep passing without changes). Let me check if I can add some specific test triggering the described issue (it is DB level, mysql specific).

PR #3307, opened 2023-11-09
Inheritable integration labels

vadimkerr:
# What this PR does

Allows selecting integration labels that will be passed down to alert groups

Related to https://github.com/grafana/oncall-private/issues/2179

## Checklist


engine/apps/labels/models.py
```diff
@@ -105,6 +105,9 @@ class AlertReceiveChannelAssociatedLabel(AssociatedLabel):
         "alerts.AlertReceiveChannel", on_delete=models.CASCADE, related_name="labels"
     )
 
+    # If inherit is True, then the label will be passed down to alert groups
+    inherit = models.BooleanField(default=True, null=True)
```
matiasb, line 109: Wondering if `inherit` is the right word here? Maybe something more like `inheritable`, `transmit`, `spread`...? (not really sure, but it seemed confusing for me at first because this flags means that the alert group inherits the label, right? it may be a language issue though :-))

engine/apps/alerts/models/alert_receive_channel.py
```diff
@@ -638,6 +642,21 @@ def insight_logs_metadata(self):
             result["team"] = "General"
         return result
 
+    @property
+    def alert_group_labels(self) -> IntegrationAlertGroupLabels:
+        """
+        Alert group labels configuration for the integration used by AlertReceiveChannelSerializer.
+        See AlertReceiveChannelAssociatedLabel.inherit for more details.
+        """
+        return {"inherit": {label.key_id: label.inherit for label in self.labels.all()}}
```
matiasb, line 651: Any reason not to just keep a list of the transmissible labels?

engine/apps/labels/models.py
vadimkerr: Yes I like `inheritable` more as well

engine/apps/alerts/models/alert_receive_channel.py
vadimkerr: No particular reason AFAICT, I agree a list could work better here

engine/apps/labels/models.py
vadimkerr: addressed in https://github.com/grafana/oncall/pull/3307/commits/929fdcfc3e8525d61733109b4afff7c2847a9580

matiasb (approved):
LGTM

Ferril (approved):
LGTM

grafana-plugin/src/containers/IntegrationLabelsForm/IntegrationLabelsForm.tsx
```diff
@@ -0,0 +1,92 @@
...
+const IntegrationLabelsForm = observer((props: IntegrationLabelsFormProps) => {
+  const { id, onHide, onSubmit } = props;
+
+  const store = useStore();
+
+  const { alertReceiveChannelStore } = store;
+
+  const alertReceiveChannel = alertReceiveChannelStore.items[id];
+
+  const [alertGroupLabels, setAlertGroupLabels] = useState(alertReceiveChannel.alert_group_labels);
```
brojd, line 30: Maybe we can destructure right away?
```
const {
    alertReceiveChannelStore: { items, saveAlertReceiveChannel },
  } = useStore();
  const alertReceiveChannel = items[id];
  ...
  ```

File: grafana-plugin/src/containers/IntegrationLabelsForm/IntegrationLabelsForm.tsx [hunk 4]
```diff
@@ -0,0 +1,92 @@
...
+            </Tooltip>
+          </HorizontalGroup>
+          <ul className={cx('labels-list')}>
+            {alertReceiveChannel.labels.map((label) => (
+              <li key={label.key.id}>
+                <HorizontalGroup spacing="xs">
+                  {/* @ts-ignore */}
+                  <Input width="300px" value={label.key.name} disabled />
+                  {/* @ts-ignore */}
+                  <Input width="250px" value={label.value.name} disabled />
```
brojd, line 66: Can we avoid `@ts-ignore`, pls? Preferred way is to apply this width with css so just add a class. Second rescue would be to pass a number here instead of string)

grafana-plugin/src/pages/incident/Incident.tsx
```diff
@@ -339,6 +342,25 @@ class IncidentPage extends React.Component<IncidentPageProps, IncidentPageState>
...
+              {Boolean(store.hasFeature(AppFeature.Labels) && incident.labels.length) && (
+                <TooltipBadge
+                  tooltipTitle=""
+                  borderType="secondary"
+                  icon="tag-alt"
+                  addPadding
+                  text={incident.labels.length}
+                  tooltipContent={
+                    <VerticalGroup spacing="sm">
+                      {incident.labels.length
```
brojd, line 354: We have the same check of `length` in line 345. So at this point it will always be truthy which means that `'No labels attached'` will never be displayed. I guess this condition is redundant?

grafana-plugin/src/pages/incidents/Incidents.tsx
```diff
@@ -583,6 +587,40 @@ class Incidents extends React.Component<IncidentsPageProps, IncidentsPageState>
...
+                    <Button
+                      size="sm"
+                      icon="filter"
+                      tooltip="Apply filter"
+                      variant="secondary"
+                      onClick={this.getApplyLabelFilterClickHandler(label)}
+                    />
+                  </HorizontalGroup>
+                ))
+              : 'No labels attached'}
```
brojd, line 617: Same here, in line 591 we're already checking `item.labels.length` and we're returning `null` if it's falsy. So second check is redundant I believe

grafana-plugin/src/pages/incidents/Incidents.tsx
```diff
@@ -583,6 +587,40 @@ class Incidents extends React.Component<IncidentsPageProps, IncidentsPageState>
...
   }
 
+  renderLabels(item: AlertType) {
+    if (!item.labels.length) {
+      return null;
+    }
+
+    return (
+      <TooltipBadge
+        tooltipTitle=""
```
brojd, line 597: I can see that we're passing empty string in multiple places to this component:
image

How about we make this prop optional in `TooltipBadge` and then we get rid of passing empty string in those cases?

grafana-plugin/src/pages/incidents/Incidents.tsx
```diff
@@ -591,6 +629,29 @@ class Incidents extends React.Component<IncidentsPageProps, IncidentsPageState>
...
+      const labelToAddString = `${label.key.id}:${label.value.id}`;
+      if (oldLabelFilter.some((label) => label === labelToAddString)) {
+        return;
+      }
+
+      const newLabelFilter = [...oldLabelFilter, labelToAddString];
+
+      LocationHelper.update({ label: newLabelFilter }, 'partial');
+
+      filtersStore.needToParseFilters = true;
```
brojd, line 651: Let's pls avoid changing observable without an action. Newer versions of MobX will log a warning in every case like that. Using actions have also some other benefits:
image

grafana-plugin/src/pages/integration/Integration.tsx
```diff
@@ -731,7 +732,8 @@ const IntegrationActions: React.FC<IntegrationActionsProps> = ({
   isLegacyIntegration,
   changeIsTemplateSettingsOpen,
 }) => {
-  const { alertReceiveChannelStore } = useStore();
+  const store = useStore();
+  const { alertReceiveChannelStore } = store;
```
brojd, line 736: How about 
```
const { alertReceiveChannelStore, hasFeature } = useStore();
```
?

grafana-plugin/src/containers/IntegrationLabelsForm/IntegrationLabelsForm.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+  const [alertGroupLabels, setAlertGroupLabels] = useState(alertReceiveChannel.alert_group_labels);
+
+  const handleSave = useCallback(() => {
+    alertReceiveChannelStore.saveAlertReceiveChannel(id, { alert_group_labels: alertGroupLabels });
+
+    onSubmit();
+
+    onHide();
+  }, [alertGroupLabels]);
```
teodosii, line 38: Why use `useCallback` for such simple usages?

https://kentcdodds.com/blog/usememo-and-usecallback#why-is-usecallback-worse

brojd: I think it's the most famous article about hooks in the web :D

grafana-plugin/src/models/filters/filters.ts
```diff
@@ -35,6 +35,15 @@ export class FiltersStore extends BaseStore {
...
   }
 
+  @action
+  set needToParseFilters(value: boolean) {
+    this._needToParseFilters = value;
+  }
+
+  get needToParseFilters() {
+    return this._needToParseFilters;
+  }
```
brojd, line 45: Setters and getters have a different purpose in MobX. Getters are used to define `computeds` - they derive a value based on other observables. Setters are used for some really rare case, tbh I've never had such (to override computed mentioned before). More in docs: https://mobx.js.org/computeds.html#computed-setter 

Here we just need observable and action that can update it. Also this observable is a public property, I think underscore in the name is redundant. The name itself should also indicate that it is updating something, the verb should be used.

So to sum up I would suggest:
```
@observable
needToParseFilters = false;
...
 @action
 setNeedToParseFilters(value: boolean) {
    this.needToParseFilters = value;
  }

```

maskin25: fixed

PR #3311, opened 2023-11-09
fix(3093) Return timezone field of the user via public API

### LINKED ISSUE #3093: Return a Users Specified Time Zone in the API Response

When listing users from the API it would be very helpful to us to know what the users current timezone is for automating some schedule management.

### Product Area

Schedules

### Anything else to add?

_No response_

ravishankar15:
# What this PR does
Return the `timezone` field for the users GET API call

Closes #3093 

## Checklist


joeyorlando (commented):
hi @ravishankar15! thank you for your contribution 😄 

Just two small asks and then I will go ahead and approve/merge this in:
- do you mind adding a small note to the `CHANGELOG.md` about this?
- can you update the public API docs [here](https://github.com/grafana/oncall/blob/dev/docs/sources/oncall-api-reference/users.md) to include this new field in the response schemas?

Again, thanks!

PR #3322, opened 2023-11-10
Add test ensuring ingestion works without db access

matiasb:
Handling alert payloads should work without db access (but still requires cached integrations information)

### DIFF
--- a/engine/apps/integrations/tests/test_views.py
+++ b/engine/apps/integrations/tests/test_views.py
@@ -2,11 +2,24 @@ from unittest.mock import call, patch
 
 import pytest
 from django.core.files.uploadedfile import SimpleUploadedFile
+from django.db import OperationalError
 from django.urls import reverse
+from pytest_django.plugin import _DatabaseBlocker
 from rest_framework import status
 from rest_framework.test import APIClient
 
 from apps.alerts.models import AlertReceiveChannel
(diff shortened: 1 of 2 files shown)


engine/apps/labels/tests/factories.py
```diff
@@ -10,15 +10,15 @@
 
 
 class LabelKeyFactory(factory.DjangoModelFactory):
-    id = UniqueFaker("sentence", nb_words=3)
+    id = UniqueFaker("pystr", max_chars=36)
```
matiasb, line 13: There is a max length limit for IDs in the cache model; force that in tests generated values ([failing test](https://github.com/grafana/oncall/actions/runs/6827306305/job/18569140244?pr=3322#step:5:644))

engine/apps/integrations/tests/test_views.py
```diff
@@ -2,11 +2,24 @@
...
+
+
+class DatabaseBlocker(_DatabaseBlocker):
+    """Customize pytest_django db blocker to raise OperationalError exception."""
+
+    def _blocking_wrapper(*args, **kwargs):
+        __tracebackhide__ = True
+        __tracebackhide__  # Silence pyflakes
+        # mimic DB unavailable error
+        raise OperationalError("Database access disabled")
```
joeyorlando, line 22: nice 👍

joeyorlando (approved):

Issue #3321, opened 2023-11-10
Add responders freeze


teodosii:
### What went wrong?

Slack link for context https://raintank-corp.slack.com/archives/C04JCU51NF8/p1699542573237099

Noticed 2 issues with add responders feature while trying to use it in the dev/ops environments
- in the dev env the add responders search is not working, it's closing the modal instead (same if you click users)
- in the ops environment it causes total freeze (at least for me) - could be due to performing too many requests when searching for the users, probably worth caching the images. Throttling could also be increased I guess

### Grafana OnCall Version

all

### Product Area

Alert Flow & Configuration

### Grafana OnCall Platform?

None

PR #3325, opened 2023-11-10
Improve performance on user search results in add responders dropdown

joeyorlando:

Closes https://github.com/grafana/oncall/issues/3321

## Checklist

### DIFF
--- a/grafana-plugin/src/containers/AddResponders/parts/AddRespondersPopup/AddRespondersPopup.tsx
+++ b/grafana-plugin/src/containers/AddResponders/parts/AddRespondersPopup/AddRespondersPopup.tsx
@@ -10,6 +10,7 @@ import GTable from 'components/GTable/GTable';
 import Text from 'components/Text/Text';
 import { Alert as AlertType } from 'models/alertgroup/alertgroup.types';
 import { GrafanaTeam } from 'models/grafana_team/grafana_team.types';
+import { PaginatedUsersResponse } from 'models/user/user';
 import { UserCurrentlyOnCall } from 'models/user/user.types';
 import { useStore } from 'state/useStore';
 import { useDebouncedCallback, useOnClickOutside } from 'utils/hooks';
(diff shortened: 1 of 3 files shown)


grafana-plugin/src/containers/AddResponders/parts/AddRespondersPopup/AddRespondersPopup.tsx
```diff
@@ -97,11 +97,17 @@ const AddRespondersPopup = observer(
...
+
+      const onCallUserSearchResultsPromise = _searchForUsers('true').then((resp) =>
+        setOnCallUserSearchResults(resp.results)
+      );
+      const notOnCallUserSearchResultsPromise = _searchForUsers('false').then((resp) =>
+        setNotOnCallUserSearchResults(resp.results)
+      );
+
+      await Promise.all([onCallUserSearchResultsPromise, notOnCallUserSearchResultsPromise]);
     }, [searchTerm]);
```
brojd, line 111: Right now we have many places where we mix usage of `async/await` and `then()` but this is something that it would be good to unify in nearest future and switch fully to `async/await`. Before we do it across the whole codebase, maybe we can start to use `async/await` in new code?  

Here we can probably achieve exactly same result with:
```
const searchForUsers = useCallback(async () => {
      const _search = async (is_currently_oncall: 'true' | 'false' | 'all') => {
        const response = await userStore.search<PaginatedUsersResponse<UserCurrentlyOnCall>>({
          searchTerm,
          is_currently_oncall,
        });
        return response.results;
      };

      const [onCallUserSearchResults, notOnCallUserSearchResults] = await Promise.all([
        _search('true'),
        _search('false'),
      ]);

      setOnCallUserSearchResults(onCallUserSearchResults);
      setNotOnCallUserSearchResults(notOnCallUserSearchResults);
    }, [searchTerm]);
    ```
    
    What do you think?

File: grafana-plugin/src/containers/AddResponders/parts/AddRespondersPopup/AddRespondersPopup.tsx [hunk 2]
```diff
@@ -97,11 +97,17 @@ const AddRespondersPopup = observer(
     );
 
     const searchForUsers = useCallback(async () => {
-      /**
-       * specifying is_currently_oncall=all will tell the backend not to paginate the results
-       */
-      const userResults = await userStore.search<UserCurrentlyOnCall[]>({ searchTerm, is_currently_oncall: 'all' });
-      setUserSearchResults(userResults);
+      const _searchForUsers = (is_currently_oncall: string) =>
```
teodosii, line 100: can't this work with a boolean instead of a string?

teodosii (approved):

grafana-plugin/src/containers/AddResponders/parts/AddRespondersPopup/AddRespondersPopup.tsx
joeyorlando: possibly, I haven't tried. I guess it depends on how our HTTP network package serializes booleans when used as query param value

grafana-plugin/src/containers/AddResponders/parts/AddRespondersPopup/AddRespondersPopup.tsx
joeyorlando: great point, thanks for the suggestion. updated this 👍

grafana-plugin/src/containers/AddResponders/parts/AddRespondersPopup/AddRespondersPopup.tsx
joeyorlando: _update_: tested locally, it looks like that works just fine. I've updated this. Thanks for pointing that out

brojd (approved):

PR #3364, opened 2023-11-15
Fix broken openapi schema + add integration test

joeyorlando:

- Fix issue that was causing our openapi schema to return HTTP 500 + add an integration test which fetches the `.yaml` schema and validates that the endpoint returns HTTP 200 (should hopefully prevent this from happening again).
- add a few more type hints along the way

## Checklist

### DIFF
--- a/engine/apps/api/serializers/alert_group.py
+++ b/engine/apps/api/serializers/alert_group.py
@@ -91,7 +91,7 @@ class ShortAlertGroupSerializer(AlertGroupFieldsCacheSerializerMixin, serializer
             },
         )
     )
-    def get_render_for_web(self, obj):
+    def get_render_for_web(self, obj: "AlertGroup"):
         last_alert = obj.alerts.last()
         if last_alert is None:
             return {}
@@ -103,7 +103,9 @@ class ShortAlertGroupSerializer(AlertGroupFieldsCacheSerializerMixin, serializer
         )
 
 
-class AlertGroupListSerializer(EagerLoadingMixin, AlertGroupFieldsCacheSerializerMixin, serializers.ModelSerializer):
+class AlertGroupListSerializer(
+    EagerLoadingMixin, AlertGroupFieldsCacheSerializerMixin, serializers.ModelSerializer[AlertGroup]
+):
     pk = serializers.CharField(read_only=True, source="public_primary_key")
     alert_receive_channel = FastAlertReceiveChannelSerializer(source="channel")
     status = serializers.ReadOnlyField()
@@ -178,7 +180,7 @@ class AlertGroupListSerializer(EagerLoadingMixin, AlertGroupFieldsCacheSerialize
             },
         )
     )
-    def get_render_for_web(self, obj):
+    def get_render_for_web(self, obj: "AlertGroup"):
         if not obj.last_alert:
             return {}
         return AlertGroupFieldsCacheSerializerMixin.get_or_set_web_template_field(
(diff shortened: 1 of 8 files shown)


engine/apps/api/serializers/alert_group.py
```diff
@@ -189,17 +188,6 @@ def get_render_for_web(self, obj):
...
-        """Deprecated. TODO: remove"""
-        if not obj.last_alert:
-            return {}
-        return AlertGroupFieldsCacheSerializerMixin.get_or_set_web_template_field(
-            obj,
-            obj.last_alert,
-            AlertGroupFieldsCacheSerializerMixin.RENDER_FOR_CLASSIC_MARKDOWN_FIELD_NAME,
-            AlertGroupClassicMarkdownRenderer,
-        )
-
```
joeyorlando, line 202: this was marked as deprecated. It doesn't appear to be referenced on the frontend, removing it.

engine/apps/api/serializers/alert_group.py
```diff
@@ -250,14 +238,13 @@ def get_last_alert_at(self, obj) -> datetime.datetime:
...
-    def get_limited_alerts(self, obj):
+    def get_limited_alerts(self, obj: "AlertGroup"):
         """
         Overriding default alerts because there are alert_groups with thousands of them.
         It's just too slow, we need to cut here.
         """
         alerts = obj.alerts.order_by("-pk")[:100]
         return AlertSerializer(alerts, many=True).data
 
-    @extend_schema_field(PagedUserSerializer(many=True))
```
joeyorlando, line 261: this line seemed to be the issue causing HTTP 500 🤷

engine/apps/api/views/alert_group.py
joeyorlando, line 1: updating the endpoint descriptions, either via `extend_schema_view` (for inherited methods like `list` and `preview_template`) or via method docstrings is nice because _otherwise_ `drf_spectacular` will set the default endpoint description to the class's docstring.

In this case `AlertGroupView` does not define a docstring, but one of it's subclasses, `AlertGroupTeamFilteringMixin`, does. `AlertGroupTeamFilteringMixin` actually doesn't have a docstring, but its subclass, `TeamFilteringMixin`, does:
```
This mixin returns 403 and {"error_code": "wrong_team", "owner_team": {"name", "id", "email", "avatar_url"}}
    in case a requested instance doesn't belong to user's current_team.
```
_tldr_; `drf_spectacular` was setting all of the endpoint descriptions to that ☝️ for endpoints which did not explicitly list a `description` or have a method docstring.

engine/apps/api/views/alert_group.py
```diff
@@ -290,8 +296,6 @@ class AlertGroupView(
         "filters": [RBACPermission.Permissions.ALERT_GROUPS_READ],
         "silence_options": [RBACPermission.Permissions.ALERT_GROUPS_READ],
         "bulk_action_options": [RBACPermission.Permissions.ALERT_GROUPS_READ],
-        "create": [RBACPermission.Permissions.ALERT_GROUPS_WRITE],
-        "update": [RBACPermission.Permissions.ALERT_GROUPS_WRITE],
```
joeyorlando, line 294: we don't have a `create` or `update` method defined on `AlertGroupView`, no need for these permissions here

engine/apps/api/tests/test_openapi_schema.py
```diff
@@ -0,0 +1,14 @@
...
+from rest_framework.test import APIClient
+
+
+@pytest.mark.django_db
+def test_fetching_the_openapi_schema_works(settings):
+    client = APIClient()
+    response = client.get(reverse("schema"))
+
+    assert response.status_code == status.HTTP_200_OK
+    assert yaml.safe_load(response.content)["info"]["title"] == settings.SPECTACULAR_SETTINGS["TITLE"]
```
vadimkerr, line 14: 👍

vadimkerr (approved):
LGTM

PR #3380, opened 2023-11-17
fix missing users in rotations when RBAC is enabled

joeyorlando:
1. Enable RBAC
2. Create a schedule rotation layer which includes a user whom is Viewer + has role `Notifications Receiver` (this is the RBAC role we use to filter which users show up in the user dropdown in the rotations modal when creating a rotation)
3. The user _sorta_ shows up in the schedule but they are listed in `missing_users`

Screenshot 2023-11-17 at 10 12 30
Screenshot 2023-11-17 at 10 15 04

## Checklist

### DIFF
--- a/engine/apps/schedules/tests/test_ical_utils.py
+++ b/engine/apps/schedules/tests/test_ical_utils.py
@@ -128,23 +128,20 @@ def test_users_in_ical_basic_role(make_organization_and_user, make_user_for_orga
     ],
 )
 def test_users_in_ical_rbac(make_organization_and_user, make_user_for_organization, permission, included):
-    organization, user = make_organization_and_user()
+    organization, _ = make_organization_and_user()
     organization.is_rbac_permissions_enabled = True
     organization.save()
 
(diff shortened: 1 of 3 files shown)


engine/apps/schedules/tests/test_ical_utils.py
```diff
@@ -93,17 +93,62 @@ def test_users_in_ical_email_case_insensitive(make_organization_and_user, make_u
...
+
+    expected_result = {user}
+    if included:
+        expected_result.add(viewer)
+    assert set(users_in_ical(usernames, organization)) == expected_result
 
 
 @pytest.mark.django_db
-def test_list_users_to_notify_from_ical_viewers_inclusion(
+def test_list_users_to_notify_from_ical_viewers_exclusion(
```
joeyorlando, line 151: changed name here.. if you check down on line 172 this is actually testing that viewers are excluded (as would be expected by the "fallback role" when RBAC is disabled)

Ferril (approved):

PR #3394, opened 2023-11-21
helm: fix conditions create svc for detached integrations

### LINKED ISSUE #3393: Helm: svc is created for integrations when detached_integrations.enabled = False
### What went wrong?

**What happened**:
- svc is created for integrations when detached_integrations.enabled = False
 
- svc will not be created

### How do we reproduce it?

1. Download default values.yaml
2. `helm template --release-name oncall  grafana/oncall --values helm/values.yaml --namespace oncall > ./helm/templated/templated.yaml`
3. see templated.yaml

### Grafana OnCall Version

chart version 1.3.59

### Product Area

Helm

### Grafana OnCall Platform?

Kubernetes

### User's Browser?

_No response_

### Anything else to add?

_No response_

sentoz:
# What this PR does

Added a condition to the service that checks the variable detached_integrations.enabled

#3393


matiasb (approved):
Thanks :+1:

PR #3397, opened 2023-11-21
Fix deleting integrations with duplicate names

vadimkerr:
# What this PR does

Fixes a bug when it's not possible to delete two or more integrations having the same name at once.

https://github.com/grafana/oncall-private/issues/2313

## Checklist


engine/apps/alerts/models/alert_receive_channel.py
```diff
@@ -206,14 +206,6 @@ class AlertReceiveChannel(IntegrationOptionsMixin, MaintainableObject):
     rate_limited_in_slack_at = models.DateTimeField(null=True, default=None)
     rate_limit_message_task_id = models.CharField(max_length=100, null=True, default=None)
 
-    class Meta:
-        constraints = [
-            models.UniqueConstraint(
-                fields=["organization", "verbal_name", "deleted_at"],
-                name="unique integration name",
-            )
-        ]
```
vadimkerr, line 215: This unique constraint doesn't make sense. The only time it's violated it's when an org has two integrations with the same name and they are both deleted at the same time (so the update tries to set the same values for `(organization, verbal_name, deleted_at)` for both integrations and the update fails). This use case is valid and shouldn't fail.
The index name `"unique integration name"` is misleading as it actually doesn't forbid organizations from having integrations with duplicate names at the DB level.

matiasb (approved):

PR #3412, opened 2023-11-22
Delete duplicate direct paging integrations

vadimkerr:
# What this PR does

Deletes duplicate direct paging integrations (i.e. keeps only the first direct paging integration per team).
Also adds a unique constraint that will make such duplicates impossible at the DB level.

Related to https://github.com/grafana/oncall-private/issues/2302

## Checklist


engine/apps/alerts/models/alert_receive_channel.py
```diff
@@ -206,6 +206,20 @@ class AlertReceiveChannel(IntegrationOptionsMixin, MaintainableObject):
     rate_limited_in_slack_at = models.DateTimeField(null=True, default=None)
     rate_limit_message_task_id = models.CharField(max_length=100, null=True, default=None)
 
+    class Meta:
+        constraints = [
+            # This monstrous constraint ensures that there's at most one active direct paging integration per team
+            # This should work with SQLite, PostgreSQL and MySQL >= 8.0.13
```
Konstantinov-Innokentii, line 212: What will happen if some OSS customer using Mysql <8.0.13? Will migration fail?

vadimkerr: From [the docs](https://docs.djangoproject.com/en/4.2/ref/models/indexes/#expressions):

> Functional indexes are ignored with MySQL < 8.0.13 and MariaDB as neither supports them.

I'll update the comment to reflect this 👍

engine/apps/api/tests/test_alert_receive_channel.py
```diff
@@ -852,35 +852,6 @@ def test_update_alert_receive_channels_direct_paging(
     assert response.json()["detail"] == AlertReceiveChannel.DuplicateDirectPagingError.DETAIL
 
 
-@pytest.mark.django_db
-def test_delete_alert_receive_channel_direct_paging_duplicate(
-    make_organization_and_user_with_plugin_token, make_team, make_alert_receive_channel, make_user_auth_headers
-):
-    """Check that it's possible to delete direct paging integration even if there is a duplicate for the team."""
```
vadimkerr, line 859: Duplicate direct paging integrations are now impossible, so removing this test

engine/apps/alerts/migrations/0041_alertreceivechannel_unique_direct_paging_integration_per_team.py
```diff
@@ -0,0 +1,56 @@
...
+logger = logging.getLogger(__name__)
+
+
+def delete_duplicate_direct_paging_integrations(apps, schema_editor):
+    AlertReceiveChannel = apps.get_model("alerts", "AlertReceiveChannel")
+
+    # get (organization_id, team_id) pairs for teams that have more than one direct paging integration
+    duplicate_rows = AlertReceiveChannel.objects.values_list(
+        "organization_id", "team_id"
+    ).annotate(count=models.Count("id")).filter(integration="direct_paging", deleted_at__isnull=True, count__gt=1)
```
matiasb, line 16: Would it be better to filter before annotating (to avoid annotating over all integrations)? ie.
```
duplicate_rows = AlertReceiveChannel.objects.filter(
    integration="direct_paging", deleted_at__isnull=True
).values_list(
    "organization_id", "team_id"
).annotate(count=models.Count("id")).filter(count__gt=1)
```

engine/apps/api/tests/test_alert_receive_channel.py
matiasb: Maybe add some test which raises a DB error when trying to create a duplicate entry?

matiasb (approved):
LGTM

engine/apps/alerts/migrations/0041_alertreceivechannel_unique_direct_paging_integration_per_team.py
joeyorlando, line 1: _total nit_ - just as a general note, you should check-out [window functions](https://docs.djangoproject.com/en/4.2/ref/models/expressions/#window-functions). I _think_ you can accomplish this all in one SQL query w/ something like this (disclaimer: I didn't test this):

```python3
from django.db.models.expressions import Window
from django.db.models.functions import RowNumber
from django.db.models import F

duplicate_direct_paging_integrations = AlertReceiveChannel.objects.filter(
    integration="direct_paging",
    deleted_at__isnull=True
).annotate(
    row_number=Window(
        expression=RowNumber(),
        partition_by=[F("organization_id"), F("team_id")],
        order_by="id", # defaults to ASC
    )
).filter(
    row_number__gt=1,
).update(
    deleted_at=timezone.now()
)
```

joeyorlando (approved):
LGTM 👍

engine/apps/alerts/migrations/0041_alertreceivechannel_unique_direct_paging_integration_per_team.py
vadimkerr: Both produce equivalent SQL queries, so I think the filter order doesn't really matter:

```mysql
SELECT organization_id, team_id, COUNT(id) AS count
FROM alerts_alertreceivechannel
WHERE (deleted_at IS NULL AND integration = 'direct_paging') 
GROUP BY organization_id, team_id
HAVING COUNT(id) > 1 
ORDER BY NULL
```

engine/apps/alerts/migrations/0041_alertreceivechannel_unique_direct_paging_integration_per_team.py
vadimkerr: I'll check this out, thank you!

PR #3413, opened 2023-11-23
Rework alert group internal API team filter

matiasb:
Related to https://github.com/grafana/oncall-private/issues/2177

### DIFF
--- a/engine/apps/api/serializers/alert_group.py
+++ b/engine/apps/api/serializers/alert_group.py
@@ -125,6 +125,7 @@ class AlertGroupListSerializer(
     PREFETCH_RELATED = [
         "dependent_alert_groups",
         "log_records__author",
+        "labels",
     ]
 
     SELECT_RELATED = [
--- a/engine/apps/api/views/alert_group.py
+++ b/engine/apps/api/views/alert_group.py
@@ -28,12 +28,7 @@ from apps.labels.utils import is_labels_feature_enabled
 from apps.mobile_app.auth import MobileAppAuthTokenAuthentication
 from apps.user_management.models import Team, User
 from common.api_helpers.exceptions import BadRequest
-from common.api_helpers.filters import (
-    ByTeamModelFieldFilterMixin,
-    DateRangeFilterMixin,
-    ModelFieldFilterMixin,
-    TeamModelMultipleChoiceFilter,
-)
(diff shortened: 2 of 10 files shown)


engine/apps/api/serializers/alert_group.py
```diff
@@ -125,6 +125,7 @@ class AlertGroupListSerializer(
     PREFETCH_RELATED = [
         "dependent_alert_groups",
         "log_records__author",
+        "labels",
```
matiasb, line 128: This reduces the number of queries fetching labels for alert groups in one query instead of one per alert group.

engine/apps/api/views/alert_group.py
```diff
@@ -337,6 +331,16 @@ def get_queryset(self, ignore_filtering_by_available_teams=False):
         if not ignore_filtering_by_available_teams:
             alert_receive_channels_qs = alert_receive_channels_qs.filter(*self.available_teams_lookup_args)
 
+        # Filter by team(s). Since we really filter teams from integrations, this is not an AlertGroup model filter.
+        # This is based on the common.api_helpers.ByTeamModelFieldFilterMixin implementation
```
matiasb, line 335: This is the main change. Removing team from filter above will not add the team condition in the alert groups query. Instead prefer to restrict the team when filtering alert receive channels, which we are already using to filter alert groups per org.

engine/common/api_helpers/filters.py
```diff
@@ -7,6 +7,8 @@
 from apps.user_management.models import Team
 from common.api_helpers.exceptions import BadRequest
 
+NO_TEAM_VALUE = "null"
```
vadimkerr, line 10: 👍 👍

vadimkerr (approved):
LGTM

engine/apps/api/serializers/alert_group.py
Konstantinov-Innokentii: Just double-checking if it should be select relates, since alert_group - labels is a one to many relation

matiasb: In this case there are many labels per alert group, so we need to prefetch them (select related would be if there was a FK from alert group to label but here we have FK from label to alert group).

PR #3416, opened 2023-11-23
Add alert group actions extension point to plugin

sd2k:
# What this PR does

This PR adds an [extension point] to the Grafana plugin, allowing other plugins to add menu items to the dropdown independently of this plugin.

It also moves the 'Declare incident' button into the 'Actions' menu representing the extension point, and displays it conditionally; the menu item is shown if all of the following are true:

1. the Grafana Incident plugin is installed (using the `PluginBridge` as before)
2. the alert group has a valid Grafana Incident URL
3. the alert group hasn't already been linked to an Incident
4. the Grafana Incident plugin hasn't _also_ configured an identical extension link to show in the dropdown.

Ideally the Grafana Incident plugin would soon configure an extension using this extension point, at which point we can remove the custom 'Declare incident' button (and the `PluginBridge`) from the OnCall app.

---

The main use case I have for this is to be able to run a Sift investigation from an OnCall alert group, similar to how it can be done from a Grafana alert instance as of [this PR][pr].

I'll try and get the Grafana ML and Incident plugins running in the same Grafana instance as this to test it out more thoroughly!

Note: this is heavily inspired by, and in some case stolen from, the [Explore toolbar extension point in Grafana](https://github.com/grafana/grafana/blob/main/public/app/features/explore/extensions/ToolbarExtensionPoint.tsx).

Part of https://github.com/grafana/machine-learning/issues/3558.

## Checklist

[extension point]: https://grafana.com/developers/plugin-tools/ui-extensions/create-an-extension-point
[pr]: https://github.com/grafana/grafana/pull/77900

### DIFF
--- a/grafana-plugin/src/components/ExtensionLinkMenu/ExtensionLinkDropdown.tsx
+++ b/grafana-plugin/src/components/ExtensionLinkMenu/ExtensionLinkDropdown.tsx
@@ -25,7 +25,7 @@ export function ExtensionLinkDropdown({
   declareIncidentLink,
   grafanaIncidentId,
 }: Props): ReactElement | null {
-  const [isOpen, setIsOpen] = useState<boolean>(false);
+  const [isOpen, setIsOpen] = useState(false);
   const context = useExtensionPointContext(incident);
   const extensions = useExtensionLinks(context, extensionPointId);
 
--- a/grafana-plugin/src/components/ExtensionLinkMenu/ExtensionLinkDropdown.tsx
+++ b/grafana-plugin/src/components/ExtensionLinkMenu/ExtensionLinkDropdown.tsx
@@ -1,0 +1,82 @@
+import React, { ReactElement, useMemo, useState } from 'react';
+
+// Note: these imports are available in Grafana>=10.0.
+// @ts-expect-error
+import { PluginExtensionLink } from '@grafana/data';
+// @ts-expect-error
+import { getPluginLinkExtensions } from '@grafana/runtime';
+import { Dropdown, ToolbarButton } from '@grafana/ui';
+import { OnCallPluginExtensionPoints } from 'types';
+
+import { Alert } from 'models/alertgroup/alertgroup.types';
+
+import { ExtensionLinkMenu } from './ExtensionLinkMenu';
+
+interface Props {
+  incident: Alert;
+  extensionPointId: OnCallPluginExtensionPoints;
+  declareIncidentLink?: string;
+  grafanaIncidentId: string | null;
+}
+
+export function ExtensionLinkDropdown({
+  incident,
+  extensionPointId,
+  declareIncidentLink,
+  grafanaIncidentId,
+}: Props): ReactElement | null {
+  const [isOpen, setIsOpen] = useState<boolean>(false);
+  const context = useExtensionPointContext(incident);
+  const extensions = useExtensionLinks(context, extensionPointId);
+
+  if (extensions.length === 0) {
+    return null;
+  }
+
+  const menu = (
+    <ExtensionLinkMenu
+      extensions={extensions}
+      declareIncidentLink={declareIncidentLink}
+      grafanaIncidentId={grafanaIncidentId}
+    />
+  );
+
+  return (
+    <Dropdown onVisibleChange={setIsOpen} placement="bottom-start" overlay={menu}>
+      <ToolbarButton aria-label="Actions" variant="canvas" isOpen={isOpen}>
+        Actions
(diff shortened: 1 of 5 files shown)


sd2k:
Not sure if we want a changelog/public docs update since this is really a change for other plugin developers?

sd2k:
FYI I'm working on adding the ML repo to Tilt so that we can actually run our plugins alongside each other in a local dev environment, which will make this a bit easier to test out.

grafana-plugin/src/components/ExtensionLinkMenu/ExtensionLinkMenu.tsx
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    return {
+      uncategorised,
+      categorised,
+    };
+  }, [extensions]);
+}
+
+function truncateTitle(title: string, length: number): string {
```
brojd, line 125: This should ideally live in some common place, e.g. we can create new file `src/utils/string` and put it there

grafana-plugin/src/components/ExtensionLinkMenu/ExtensionLinkMenu.tsx
```diff
@@ -0,0 +1,131 @@
...
+  const declareIncidentExtensionLink = extensions.find(
+    (extension) => extension.pluginId === 'grafana-incident-app' && extension.title === 'Declare incident'
+  );
+  if (
+    // Don't show a custom Declare incident button if the Grafana Incident plugin already configured one.
+    declareIncidentExtensionLink !== undefined ||
+    // Don't show a custom Declare incident button if there's no valid link.
+    declareIncidentLink === undefined ||
+    // Don't show the button if an incident has already been declared from this alert group.
+    grafanaIncidentId !== null
```
brojd, line 56: What if it is `undefined`?

In general maybe those checks could be a bit less narrowed and could cover all falsy values, does it make sense? For example:
```ts
if (
    // Don't show a custom Declare incident button if the Grafana Incident plugin already configured one.
    declareIncidentExtensionLink  ||
    // Don't show a custom Declare incident button if there's no valid link.
    !declareIncidentLink ||
    // Don't show the button if an incident has already been declared from this alert group.
    grafanaIncidentId
  }
```

grafana-plugin/src/components/ExtensionLinkMenu/ExtensionLinkDropdown.tsx
```diff
@@ -0,0 +1,82 @@
...
+  );
+}
+
+function useExtensionPointContext(incident: Alert): PluginExtensionOnCallAlertGroupContext {
+  return { alertGroup: incident };
+}
+
+function useExtensionLinks<T>(context: T, extensionPointId: OnCallPluginExtensionPoints): PluginExtensionLink[] {
+  return useMemo(() => {
+    // getPuginLinkExtensions is available in Grafana>=10.0,
```
brojd, line 59: typo in first word:)

grafana-plugin/src/components/ExtensionLinkMenu/ExtensionLinkDropdown.tsx
```diff
@@ -0,0 +1,82 @@
...
+  grafanaIncidentId: string | null;
+}
+
+export function ExtensionLinkDropdown({
+  incident,
+  extensionPointId,
+  declareIncidentLink,
+  grafanaIncidentId,
+}: Props): ReactElement | null {
+  const [isOpen, setIsOpen] = useState<boolean>(false);
```
brojd, line 28: Type would be inferred from default value so `<boolean>` can be omitted

brojd (approved):

grafana-plugin/src/components/ExtensionLinkMenu/ExtensionLinkMenu.tsx
sd2k: Done!

grafana-plugin/src/components/ExtensionLinkMenu/ExtensionLinkDropdown.tsx
sd2k: Thanks 😅 fixed.

grafana-plugin/src/components/ExtensionLinkMenu/ExtensionLinkMenu.tsx
sd2k: Yep makes sense. I've pushed a fix using your recommendation 🙂

PR #3420, opened 2023-11-23
Trigger distribute task after alert is committed

matiasb:
Fix issue triggering task retries because alert is not yet committed to the DB.
Similar to https://github.com/grafana/oncall/pull/3001.

### DIFF
--- a/engine/apps/alerts/tests/test_alert.py
+++ b/engine/apps/alerts/tests/test_alert.py
@@ -7,22 +7,32 @@ from apps.alerts.tasks import distribute_alert, escalate_alert_group
 
 
 @pytest.mark.django_db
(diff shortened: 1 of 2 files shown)


engine/apps/alerts/tests/test_alert.py
```diff
@@ -7,22 +7,32 @@
...
-    alert = Alert.create(
-        title="the title",
-        message="the message",
-        alert_receive_channel=alert_receive_channel,
-        raw_request_data={},
-        integration_unique_data={},
-        image_url=None,
-        link_to_upstream_details=None,
-    )
+    with django_capture_on_commit_callbacks(execute=True) as callbacks:
```
vadimkerr, line 22: 👍 TIL

vadimkerr (approved):
LGTM

PR #3432, opened 2023-11-27
Adds tojson filter to simplify work with webhooks

Konstantinov-Innokentii:
# What this PR does
  This PR adds new tojson filter. For webhook templates I needed something simpler that tojson_pretty, which produces json with indents. I found had existing json_dumps filter with same functionality, which just call json.dumps but it wasn't documented and included to web editor cheatsheet. I named it to match existing and documented tojson_pretty.

### DIFF
--- a/dev/scripts/generate-fake-data/requirements.txt
+++ b/dev/scripts/generate-fake-data/requirements.txt
@@ -1,3 +1,3 @@
-aiohttp==3.8.6
+aiohttp==3.9.0
 Faker==16.4.0
 tqdm==4.64.1


joeyorlando (approved):
you _may_ need to also "register" this function in `engine/common/jinja_templater/jinja_template_env.py` so that it's available. See [here](https://github.com/grafana/oncall/pull/3426/files#diff-376347930586cfedfb588eea67e590fa07892b3eeff9d4d16082598b8defc4c2R26) for an example.

mderynck:
Isn't the `tojson` filter already present we just need to add it to the MonacoEditor?

Konstantinov-Innokentii:
@mderynck yep, you are right, it's already present, for some reason locally it was unavailable for me.

docs/sources/jinja2-templating/_index.md
```diff
@@ -201,7 +201,8 @@ Built-in functions:
 ### Functions added by Grafana OnCall
 
 - `time` - current time
-- `tojson_pretty` - JSON prettified
+- `tojson` - dumps a structure to JSON
```
vadimkerr, line 204: nit: technically this one is built-in and is not added by OnCall

vadimkerr (approved):

docs/sources/jinja2-templating/_index.md
Konstantinov-Innokentii: I guess still worth to be documented

PR #3446, opened 2023-11-28
improve error handling and notifications for labels

brojd:
https://github.com/grafana/oncall-private/issues/2304

# What this PR does
- introduce @WithGlobalNotifiaction decorator to simplify setting notifications
- use it for alert group labels
- add npm commands to simplify linking gops-labels package 

## Checklist


grafana-plugin/src/models/label/label.ts
```diff
@@ -3,34 +3,34 @@ import { action, observable } from 'mobx';
...
 
   constructor(rootStore: RootStore) {
     super(rootStore);
 
     this.path = '/labels/';
   }
 
-  @action
-  public async loadKeys() {
+  @action.bound
```
teodosii, line 23: what's the difference between @action.bound and changing the method to be an arrow function instead where we no longer miss the this context?

teodosii (approved):
Overall looks good but I wonder if we can replace those binders with arrow functions instead, we could still decorate those

grafana-plugin/src/models/label/label.ts
brojd: Nice question, I encountered the same hesitation during development and here's what I've learnt :D 

The reason is that with arrow fn syntax, we switch from method to property decorator:
image

In property decorators [you don't have access](https://www.typescriptlang.org/docs/handbook/decorators.html#property-decorators) to the third argument of our decorator (descriptor).

In method decorator [you do have](https://www.typescriptlang.org/docs/handbook/decorators.html#method-decorators).

And having in mind that we will switch soon to `makeAutoObservable` when those `action.bound` decorators will simply disappear, it's more handy to have access descriptor when using decorators.

PR #3447, opened 2023-11-28
Update when slack client ratelimit retry handler is enabled

matiasb:


engine/apps/slack/tests/test_slack_client.py
```diff
@@ -167,14 +167,33 @@ def test_slack_client_ratelimit(monkeypatch, error, make_organization_with_slack
...
+            client.api_call("auth.test")
+
+    # no slack built-in retries by default
+    assert len(mock_request.mock_calls) == 1
+
+
+@pytest.mark.parametrize("error", ["ratelimited", "rate_limited", "message_limit_exceeded"])
+@pytest.mark.django_db
+def test_slack_client_ratelimit_enable_retry(monkeypatch, error, make_organization_with_slack_team_identity):
+    monkeypatch.undo()  # undo engine.conftest.mock_slack_api_call
```
joeyorlando, line 184: this only undo's the monkeypatch for this test, right?

joeyorlando (approved):

engine/apps/slack/tests/test_slack_client.py
matiasb: correct!

PR #3448, opened 2023-11-28
Update transaction.on_commit to use partial instead of lambda

matiasb:
For reference, https://adamj.eu/tech/2022/08/22/use-partial-with-djangos-transaction-on-commit/.

I think this may help with this one too: https://github.com/grafana/oncall-private/issues/2318

### DIFF
--- a/engine/apps/alerts/tasks/custom_button_result.py
+++ b/engine/apps/alerts/tasks/custom_button_result.py
@@ -1,5 +1,6 @@
 import json
 import logging
+from functools import partial
 
 from django.conf import settings
 from django.db import transaction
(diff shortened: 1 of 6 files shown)


vadimkerr (approved):
TIL! Should we enable [flake8-bugbear B023](https://adamj.eu/tech/2022/08/22/use-partial-with-djangos-transaction-on-commit/#linting-for-late-binding-bugs-with-flake8-bugbear) to avoid this in the future?

matiasb:
> TIL! Should we enable [flake8-bugbear B023](https://adamj.eu/tech/2022/08/22/use-partial-with-djangos-transaction-on-commit/#linting-for-late-binding-bugs-with-flake8-bugbear) to avoid this in the future?

I was thinking on adding it to our linting process, will prepare some PR :+1:

PR #3477, opened 2023-11-30
Adds escalation step to notify a team

### LINKED ISSUE #3407: "Notify users" escalation should allow sending to a team

Maybe I'm missing something fundamental but it seems OnCall doesn't support a way to escalate to a team?

The use case here would be something like:
 - Existing escalation rules
 - Wait for 30 minutes
 - Escalate to second team (Notify users to a team rather than individual users)

### Product Area

Other

### Anything else to add?

Coupled with this bug/issue https://github.com/grafana/oncall/issues/3081 it makes escalating issues to another team very hard to do without annoying people.

Documentation at https://grafana.com/docs/oncall/latest/escalation-chains-and-routes/ says 
> Notify users - send a notification to a user or a group of users.
however groups of users doesn't seem to mean teams.

This could also imply to:
> Notify users one by one (round robin)

xssfox:
# What this PR does
Adds a "Notify Team" escalation step which will notify all users in the assigned grafana team.

Screenshot 2023-12-01 at 10 35 47 am
Screenshot 2023-12-01 at 10 36 10 am

Fixes #3407 

## Checklist


iskhakov:
Thank you for the contribution! Great work! 
We might use step "Notify Team" to notify team's escalation chains. Could you please consider renaming "Notify team" to something more specific, like "Notify all team members" or "Notify all users from team" in the code.

xssfox:
Thanks @iskhakov 

I've updated the code to mostly be `notify_team_members` to keep the variables not too long and made the frontend changes a bit more clearer

image
image

There's currently a test failing `engine/apps/integrations/tests/test_views.py` however it seems to be unrelated to my changes. I performed a `make build`, I suspect my branch doesn't have the required requirement for `DjangoDbBlocker`

xssfox:
I think this PR is still valid however I'm not going to put effort into fixing up conflicts if there's no intention to merge

engine/apps/alerts/tasks/notify_team_members.py
```diff
@@ -0,0 +1,38 @@
...
+    autoretry_for=(Exception,), retry_backoff=True, max_retries=1 if settings.DEBUG else None
+)
+def notify_team_members_task(
+    team_pk,
+    alert_group_pk,
+    previous_notification_policy_pk=None,
+    reason=None,
+    prevent_posting_to_thread=False,
+    notify_even_acknowledged=False,
+    important=False,
```
matiasb, line 20: I don't see the `important` flag to be set anywhere? I think you need to check the escalation step and set accordingly before queuing the task. Also, maybe re-check what params are used/needed here (I don't think we need them all?)

engine/apps/public_api/serializers/escalation_policies.py
```diff
@@ -280,6 +295,11 @@ def update(self, instance, validated_data):
                     EscalationPolicy.STEP_NOTIFY_MULTIPLE_USERS_IMPORTANT,
                 ]:
                     instance.notify_to_users_queue.clear()
+                if step not in [
+                    EscalationPolicy.STEP_NOTIFY_TEAM_MEMBERS,
+                    EscalationPolicy.STEP_NOTIFY_TEAM_MEMBERS_IMPORTANT
+                ]:
+                    instance.notify_to_team_members = None
```
matiasb, line 302: Would you add some test(s) for the changes here?

engine/apps/alerts/escalation_snapshot/snapshot_classes/escalation_policy_snapshot.py
```diff
@@ -87,6 +90,7 @@ def __init__(
         self.custom_webhook = custom_webhook
         self.notify_schedule = notify_schedule
         self.notify_to_group = notify_to_group
+        self.notify_to_team_members = notify_to_team_members
```
matiasb, line 93: To be sure, have you checked how this works when deserializing an escalation policy snapshot where the `notify_to_team_members` field does not exist? (we should still be able to deserialize snapshots created before the field was around) Maybe we need a test for this scenario to double-check.

matiasb (commented):
Thanks (again) for the contribution, I like the idea. Leaving a few comments I think we need to address before merging the changes. Happy to keep reviewing/discussing updates :+1:

xssfox:
Thanks for the detailed comments, will have a look in the coming days :)

engine/apps/alerts/escalation_snapshot/snapshot_classes/escalation_policy_snapshot.py
xssfox: I've fixed this along with adding a test for it.

engine/apps/alerts/tasks/notify_team_members.py
xssfox: Fixed

engine/apps/public_api/serializers/escalation_policies.py
xssfox: Added in `test_create_escalation_policy_using_notify_team_members`

xssfox:
Thanks @matiasb, I've rebased to latest dev, added some tests and I believe fixed the issues above. I've also tested locally to make sure everything still works and it seems to. Let me know if there are any other changes that need looking at.

matiasb:
Thanks for the updates, I will be taking a look before the end of the week :+1:

matiasb:
Hi @xssfox! I hope you don't mind but it seems I'm not able to push updates to this PR, so forked it [here](https://github.com/grafana/oncall/pull/3908) to implement some additional refactoring and changes. Happy to share any details on the updates, let me know if that looks ok to you. Will get someone else to do a final review, but I think it should be ready to merge with that. Thanks for your contribution, and your patience :+1:

xssfox:
@matiasb Looks good, thanks!

PR #3505, opened 2023-12-05
Add script to run a detached celery beat

matiasb:


joeyorlando:
just for my own curiosity, what benefits do we gain from running this detached? I know you made the changes recently to our scheduled tasks to only use `crontab`, I assume this is related to that?

matiasb:
> just for my own curiosity, what benefits do we gain from running this detached? I know you made the changes recently to our scheduled tasks to only use `crontab`, I assume this is related to that?

The idea was to make sure we don't restart beat every time we restart workers, the beat restarts seemed to be missing some scheduled tasks. However, monitoring after the crontab changes were deployed, it looks like that helped with that too, so it may not be needed for now (that's why I changed this to draft)

joeyorlando:
is this PR still needed?

PR #3518, opened 2023-12-06
Return 404 from mobile app gateway viewset

vadimkerr:
# What this PR does

Changes how the `MOBILE_APP_GATEWAY_ENABLED` feature flag enables/disables the mobile app gateway viewset.

## Checklist

### DIFF
--- a/engine/apps/api/views/labels.py
+++ b/engine/apps/api/views/labels.py
@@ -25,9 +25,9 @@ logger = logging.getLogger(__name__)
 
 class LabelsFeatureFlagViewSet(ViewSet):
     def initial(self, request, *args, **kwargs):
-        super().initial(request, *args, **kwargs)
         if not is_labels_feature_enabled(self.request.auth.organization):
             raise NotFound
+        super().initial(request, *args, **kwargs)
 
 
 class LabelsViewSet(LabelsFeatureFlagViewSet):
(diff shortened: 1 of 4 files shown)


engine/apps/api/views/labels.py
```diff
@@ -25,9 +25,9 @@
 
 class LabelsFeatureFlagViewSet(ViewSet):
     def initial(self, request, *args, **kwargs):
-        super().initial(request, *args, **kwargs)
         if not is_labels_feature_enabled(self.request.auth.organization):
             raise NotFound
+        super().initial(request, *args, **kwargs)
```
vadimkerr, line 30: The `initial` method does a bunch of different things like auth, throttling, etc. If the feature flag is disabled, we want to return 404 right away, so I moved the super method call down

engine/apps/mobile_app/urls.py
```diff
@@ -34,11 +33,10 @@
     optional_slash_path("fcm_relay", FCMRelayView.as_view(), name="fcm_relay"),
 ]
 
-if settings.MOBILE_APP_GATEWAY_ENABLED:
```
vadimkerr, line 37: this check is now [here](https://github.com/grafana/oncall/pull/3518/files#diff-cd8caa0fd8e0fba883880308df6d0dfc5db91323ab90207bf155d97fc810fcf3R104-R105)

PR #3528, opened 2023-12-07
Brojd/polishing labels

brojd:
# What this PR does
- auto adjust column widths in alert group table
- add missing space below drawer buttons 

**https://github.com/grafana/oncall-private/issues/2355**

## Checklist


grafana-plugin/src/pages/incidents/Incidents.module.scss
```diff
@@ -25,6 +25,12 @@
   align-items: center;
 }
 
+.horizontal-scroll-table table td:global(.rc-table-cell) {
+  white-space: nowrap;
+  padding-right: 16px;
+  max-width: 600px;
```
teodosii, line 31: hmm maybe 600px is still low for horizontal scrolling enabled. on bigger displays 600 isn't that much. I'd put it to 1000 instead, wdyt? Also I think this exception should apply only to columns that are expect to have long content, it doesn't make sense to set the max-width to such huge value for columns like team/status/id, where they barely take 100-200px

brojd: done as discussed:)

PR #3542, opened 2023-12-11
Plugin: remove duplicate Declare Incident button

sd2k:
This button is now being added into the 'Actions' menu by the
'ExtensionLinkDropdown' component, so we don't need to add
another one to the page.

See the duplicated button in this screenshot:
image


sd2k:
@grafana/grafana-oncall-frontend this is good to go I think, could we get it merged in?

brojd (approved):

PR #3544, opened 2023-12-11
Don't update alert group metrics when deleting an alert group

joeyorlando:

Fixes https://github.com/grafana/oncall-private/issues/2376

## Checklist

### DIFF
--- a/engine/apps/alerts/models/alert_group.py
+++ b/engine/apps/alerts/models/alert_group.py
@@ -1191,8 +1191,6 @@ class AlertGroup(AlertGroupSlackRenderingMixin, EscalationSnapshotMixin, models.
     def delete_by_user(self, user: User):
         from apps.alerts.models import AlertGroupLogRecord
 
-        initial_state = self.state
-
         self.stop_escalation()
         # prevent creating multiple logs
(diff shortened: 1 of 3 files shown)


engine/apps/alerts/models/alert_group.py
```diff
@@ -1219,13 +1217,10 @@ def delete_by_user(self, user: User):
             force_sync=True,
         )
 
-        dependent_alerts = list(self.dependent_alert_groups.all())
-
         self.hard_delete()
-        # Update alert group state metric cache
-        self._update_metrics(organization_id=user.organization_id, previous_state=initial_state, state=None)
```
joeyorlando, line 1226: an alternative here could be to call `self._update_metrics()` _before_ calling `self.hard_delete()`

Ferril: I don't think it makes sense to call it before, since it updates metrics async and we don't protect ourself of error it should fix.
I'd propose either call it sync here (but we change it to async everywhere by purpose) or expend parameters we pass to `_update_metrics` to allow it to update metrics anyway.

For now I think it's ok to remove metrics update from here, since we recalculate all metrics once a day

Ferril (approved):
LGTM

PR #3546, opened 2023-12-11
Fix telegram retrying task after alert group was deleted

Ferril:
# What this PR does

## Checklist

### DIFF
--- a/engine/apps/telegram/tasks.py
+++ b/engine/apps/telegram/tasks.py
@@ -226,8 +226,19 @@ def on_alert_group_action_triggered_async(log_record_id):
     from .alert_group_representative import AlertGroupTelegramRepresentative
 
     logger.info(f"AlertGroupTelegramRepresentative ACTION SIGNAL, log record {log_record_id}")
-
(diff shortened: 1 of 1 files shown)


engine/apps/telegram/tasks.py
```diff
@@ -226,8 +226,19 @@ def on_alert_group_action_triggered_async(log_record_id):
...
+    except AlertGroupLogRecord.DoesNotExist as e:
+        retries_count = on_alert_group_action_triggered_async.request.retries
+        if retries_count >= 50:
+            logger.error(
+                f"AlertGroupTelegramRepresentative: was not able to get AlertGroupLogRecord, probably alert group "
+                f"was deleted. log record {log_record_id}, retries: {retries_count}"
+            )
+            return
+        else:
+            raise e
```
joeyorlando, line 241: do we need to retry if the alert group record can't be found? retrying it would likely imagine in the same result. wdyt?

Ferril: I think we need to retry in cases if we get db connection problems

PR #3553, opened 2023-12-12
Fix task retries for deleted alert groups

Ferril:
# What this PR does

## Checklist

### DIFF
--- a/engine/apps/alerts/tasks/send_update_log_report_signal.py
+++ b/engine/apps/alerts/tasks/send_update_log_report_signal.py
@@ -7,7 +7,7 @@ from .task_logger import task_logger
 
 
 @shared_dedicated_queue_retry_task(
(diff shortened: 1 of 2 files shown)


engine/apps/alerts/tasks/send_update_log_report_signal.py
```diff
@@ -7,7 +7,7 @@
 
 
 @shared_dedicated_queue_retry_task(
-    autoretry_for=(Exception,), retry_backoff=True, max_retries=1 if settings.DEBUG else None
+    autoretry_for=(Exception,), retry_backoff=True, max_retries=1 if settings.DEBUG else 10
```
Ferril, line 10: it is safe to set maximum retries there since the only place where the task can get exception is getting alert group.

joeyorlando (approved):

PR #3571, opened 2023-12-14
Limit how long acknowledge reminders can run for

mderynck:
# What this PR does
Stops rescheduling of `acknowledge_reminder_task` after 2 weeks. Assumption being if it has been sitting for that long in acknowledged state it is likely to not need more reminders that it is still acknowledged.  Notifications for thread were probably muted a long time ago.

## Checklist


engine/apps/alerts/tasks/acknowledge_reminder.py
```diff
@@ -61,6 +63,11 @@ def acknowledge_reminder_task(alert_group_pk: int, unacknowledge_process_id: str
             (alert_group.pk, unacknowledge_process_id), countdown=unacknowledge_timeout
         )
     else:
+        if alert_group.started_at < timezone.now() - relativedelta(months=1):
```
matiasb, line 66: Wondering if a month is still too much? 2 weeks is not enough? Eventually we could have some backoff strategy?
Also, maybe I would use `timezone.timedelta` instead of `relativedelta` (although I see there is no `month` support there; in any case, 1 month for `relativedelta` is a fixed amount of days, right? since you give no context of the "current" month)

matiasb (approved):
LGTM

engine/apps/alerts/tasks/acknowledge_reminder.py
mderynck: Good points, I'll pull it up into settings and make it a configurable amount of days and we can start it at 14.

PR #3578, opened 2023-12-15
Speed up escalation auditor

Ferril:
# What this PR does
Speed up escalation auditor
- use raw escalation snapshot instead of serialized one

## Checklist

### DIFF
--- a/engine/apps/alerts/tasks/check_escalation_finished.py
+++ b/engine/apps/alerts/tasks/check_escalation_finished.py
@@ -29,26 +29,26 @@ def send_alert_group_escalation_auditor_task_heartbeat() -> None:
 
 
 def audit_alert_group_escalation(alert_group: "AlertGroup") -> None:
-    escalation_snapshot = alert_group.escalation_snapshot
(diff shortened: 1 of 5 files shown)


engine/apps/alerts/tasks/check_escalation_finished.py
```diff
@@ -29,26 +29,26 @@ def send_alert_group_escalation_auditor_task_heartbeat() -> None:
 
 
 def audit_alert_group_escalation(alert_group: "AlertGroup") -> None:
-    escalation_snapshot = alert_group.escalation_snapshot
+    raw_escalation_snapshot: dict = alert_group.raw_escalation_snapshot
```
matiasb, line 32: To be sure, the difference here is we use the raw representation to get the information instead of going through the deserialized version?

Ferril: that's correct, we get information directly from alert group json field without serializing and getting additional objects from database, since we don't need it here.

PR #3583, opened 2023-12-19
Brojd/implement insights

brojd:
# What this PR does
- Use Grafana Scenes to add Insights as a separate page in OnCall
- Add an option to run Prometheus instance via helm so that Prometheus Exporter feature can be used easily without the need of setting up Prometheus separately

https://github.com/grafana/oncall-private/issues/2382

## Checklist


grafana-plugin/src/state/rootBaseStore/index.ts
```diff
@@ -285,7 +285,8 @@ export class RootBaseStore {
     return GRAFANA_LICENSE_OSS;
   }
 
-  isOpenSource(): boolean {
+  @computed
+  get isOpenSource(): boolean {
```
teodosii, line 289: :+1:

teodosii (approved):

maskin25 (approved):

helm/oncall/values.yaml
```diff
@@ -708,3 +708,12 @@ ui:
...
   env: {}
+
+prometheus:
+  enabled: false
+  extraScrapeConfigs: |
+    - job_name: 'oncall-exporter'
+      metrics_path: /metrics/
+      static_configs:
+        - targets:
+          - oncall-dev-engine.default.svc.cluster.local:8080
```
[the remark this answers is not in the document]
brojd: I set it this way so that on by default github actions + local env you just need to set `enabled: true` for prometheus then it starts to scrape from `oncall-dev-engine` correctly. Alternatively we could skip `extraScrapeConfig` here and move it to `.github/helm-values.yml`. Or set some other URL that will work for prod helm chart instead and then override in `.github/helm-values.yml`? 

Could you help me with that? :)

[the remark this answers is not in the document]
brojd: We talked with @joeyorlando and we agreed to go with https://github.com/grafana/oncall/pull/3583/commits/fd02e6a76170529f0576b0b07ebb994f056b370d

.github/helm-values.yml
```diff
@@ -4,6 +4,8 @@ base_url_protocol: http
 env:
   - name: GRAFANA_CLOUD_NOTIFICATIONS_ENABLED
     value: "False"
+  - name: FEATURE_PROMETHEUS_EXPORTER_ENABLED
+    value: "True"
```
iskhakov, line 8: I assume prometheus is required for e2e tests

iskhakov (approved):
looks good to me

.github/helm-values.yml
brojd: Correct

PR #3587, opened 2023-12-19
Handle telegram message to reply to not found on send log task

matiasb:
Similar to https://github.com/grafana/oncall/pull/404


vadimkerr (approved):
👍

PR #3612, opened 2024-01-03
Setup one-at-a-time lock for sync_organization tasks

matiasb:
Related to https://github.com/grafana/support-escalations/issues/8844

Queuing multiple sync_organization tasks for the same org could lead to parallel running of the sync task for the same organization, potentially creating duplicated entries and/or generating multiple unneeded API calls. This prevents running an organization sync while there is a sync for that same org in progress.


joeyorlando (approved):
nice! 👍

Issue #3576, opened 2023-12-15
Timezone issues on Schedules page


brojd:
### What went wrong?

We have multiple bugs on Schedules that are related with handling current timezone:
<img width="1448" alt="image" src="https://github.com/grafana/oncall/assets/12073649/3bfc7e8b-df33-4f79-b5b7-64205ed3a403">

- [x] Rotation form (shift start / shift end times) shows incorrect time that is dependent on DST. Instead it should always consider Current timezone that is agnostic to DST (GMT+1, GMT+2 etc)
- [x] Shift tooltip shows incorrect time affected by DST changes while it should correct time according to GMT+1, GMT+2 etc timezones which are DST agnostic

![image](https://github.com/grafana/oncall/assets/12073649/c0076c25-5eb8-41fa-afd6-7766450d801b)
- [x] Schedule team and timezones -> user avatar tooltip -> in local time section the timezone is dependent on "Current timezone" selected on top while it should show GMT timezone of currently logged in OnCall user always
- [x] Schedule team and timezones -> user avatar tooltip -> "Local time" should be renamed to "Your current time" and "User's time" to "User's local time". Also swap the columns
- [x] "Current timezone: Europe/Bucharest, local time: 12:05" should reflect GMT timezone selected on top ("Current timezone: +2 GMT, local time: 12:05"
- [x] Rename "+1 GMT" to "GMT+1" in all places
- [x] Create e2e test to verify above

### How do we reproduce it?

1. Open Grafana OnCall and go to Schedules (e.g. https://oncalldev.grafana-dev.net/a/grafana-oncall-app/schedules/SWUULKYC26T1W?p=1&team=TGH742BKMFZH4)
2. Create schedule with start time before DST change
3. Select timezone GMT+1 or GMT+2
4. Navigate in schedule to dates before DST change
5. Open Rotation form, shift tooltip, user avatar tooltip and take a look on Current timezone in "Schedule team and timezones"

### Grafana OnCall Version

1.3.80

### Product Area

Schedules

### Other notes

Those fixes will solve https://github.com/grafana/support-escalations/issues/8467

PR #3618, opened 2024-01-05
Brojd/fix timezone issues

brojd:
# What this PR does
- fix multiple issues on schedule related to handling timezones

https://github.com/grafana/oncall/issues/3576

## Checklist


brojd:
To be fixed: calendar change on Schedules page

grafana-plugin/src/containers/RotationForm/ScheduleOverrideForm.tsx
```diff
@@ -172,7 +166,7 @@ const ScheduleOverrideForm: FC<RotationFormProps> = (props) => {
     setErrors({});
 
     store.scheduleStore
-      .updateRotationPreview(scheduleId, shiftId, startMoment, true, params)
+      .updateRotationPreview(scheduleId, shiftId, store.timezoneStore.calendarStartDate, true, params)
```
maskin25, line 169: I thought you **always** do it in the following way `const {timezoneStore: {calendarStartDate}} = useStore();` =)

grafana-plugin/src/containers/UsersTimezones/UsersTimezones.helpers.ts
```diff
@@ -0,0 +1,6 @@
+import { Dayjs } from 'dayjs';
+
+export const calculateTimePassedInDayPercentage = (date: Dayjs) => {
+  const midnight = date.startOf('day');
+  return (date.diff(midnight, 'minutes') / 1_440) * 100;
```
maskin25, line 5: nice, I didn't know about numeric separators...

grafana-plugin/src/models/schedule/schedule.ts
```diff
@@ -506,6 +514,34 @@ export class ScheduleStore extends BaseStore {
...
+      this.updateEvents(this.scheduleId, startMoment, 'rotation'),
+      this.updateEvents(this.scheduleId, startMoment, 'override'),
+      this.updateEvents(this.scheduleId, startMoment, 'final'),
+      this.updateShiftSwaps(this.scheduleId, startMoment),
+    ]);
+  }
+
+  @action.bound
+  setScheduleId(id: string) {
+    this.scheduleId = id;
```
maskin25, line 542: hmm, why do we need to save some scheduleId in the store?

grafana-plugin/src/plugin/dayjs.ts
maskin25, line 1: 👍

grafana-plugin/tsconfig.json
```diff
@@ -6,8 +6,8 @@
     "rootDirs": ["src", "frontend_enterprise/src"],
     "baseUrl": "src",
     "typeRoots": ["./node_modules/@types"],
-    "noUnusedLocals": true,
-    "noUnusedParameters": true,
+    "noUnusedLocals": false,
+    "noUnusedParameters": false,
```
maskin25, line 10: why so?

maskin25 (approved):

grafana-plugin/src/containers/RotationForm/ScheduleOverrideForm.tsx
brojd: I do it when we already have destructuring from `store`, seems that here we don't (e.g. the line above we have `store.scheduleStore`, not `scheduleStore`). This PR is already quite big so to make changes more understandable I decided to not refactor those places and follow whatever is done in respective components:) But in general yes you're right - I think destructuring the store is a good thing, I just didn't want to refactor too much in one PR

grafana-plugin/src/models/schedule/schedule.ts
brojd: `scheduleId` observable was already there in store and it's good, I just added action to be able to update it when a single schedule is opened.

We're still passing scheduleId as a prop and that can be changed in the future so that schedule id is retrieved from the store instead. I just didn't do it in scope of this PR

grafana-plugin/src/containers/RotationForm/ScheduleOverrideForm.tsx
maskin25: ok, sounds good

grafana-plugin/src/models/schedule/schedule.ts
brojd: I removed it in the end

grafana-plugin/tsconfig.json
maskin25: but actually compilation doesn't fail, it just show ts errors but doesn't stop the rebuilding process, at least for me...

grafana-plugin/src/models/schedule/schedule.ts
maskin25: oh, I see, cool! I don't like an idea to store current page params like id in the store, we can get such params by react router

grafana-plugin/tsconfig.json
brojd: But when you stop the process and your try again, then it fails for me

Issue #3621, opened 2024-01-05
Telegram HTTP 500s when receiving message from Telegram in discussion group


joeyorlando:
### What went wrong?

**What happened**:
Recurrent HTTP 500s [logs](https://ops.grafana-ops.net/explore?schemaVersion=1&panes=%7B%22sbX%22:%7B%22datasource%22:%22000000193%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22%7Bcluster%3D~%5C%22prod-%28eu-west-0%7Cus-central-0%29%5C%22,%20namespace%3D%5C%22amixr-prod%5C%22%7D%20%7C%3D%20%5C%221fb36fe98b1ead91d15eea30bd56c5d6%2F13289933768182577349%5C%22%22,%22queryType%22:%22range%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22000000193%22%7D,%22editorMode%22:%22code%22%7D%5D,%22range%22:%7B%22from%22:%22now-7d%22,%22to%22:%22now%22%7D%7D%7D&orgId=1)

### How do we reproduce it?

Not yet sure, need to investigate

### Grafana OnCall Version

v1.3.82

### Product Area

Chatops

### Grafana OnCall Platform?

None

joeyorlando:
This appears to be happening when the official "Telegram" user sends a message to a discussion group, and _not_ via the channel. In this case, the `forward_from_chat` object that we expect [here](https://github.com/grafana/oncall/blob/dev/engine/apps/telegram/updates/update_manager.py#L62-L63) is not present. Working on a patch + some tests.

PR #3622, opened 2024-01-05
Address Telegram HTTP 500s when receiving message from Telegram in discussion group

joeyorlando:

Closes https://github.com/grafana/oncall/issues/3621

## Checklist


engine/apps/telegram/updates/update_manager.py
joeyorlando, line 1: another possibility is to add an "update handler" for messages originating directly in a channel (not a discussion group). For these HTTP 500s that we are seeing, they occur when a message is sent from the official Telegram user directly in a channel, not the channel's discussion group.

Right now we're handling these updates w/ the `apps.telegram.updates.update_handlers.channel_to_group_forward.ChannelToGroupForwardHandler` handler (this handler is determined [here](https://github.com/grafana/oncall/pull/3622/files#diff-440894128e379fb8abd0627e0d193d04dbbd8a26512ef0abcffa76bbf9ee6a46R27-R42)).

For now, what I've done in this PR seems like a simpler approach to arrive at the same solution, but open to thoughts/comments.

engine/conftest.py
```diff
@@ -443,19 +442,6 @@ def _make_slack_message(alert_group=None, organization=None, **kwargs):
...
-        """The client with logged in user"""
-
-        client = APIClient()
-        client.force_login(user)
-
-        return client
-
-    return _client_with_user
-
-
```
joeyorlando, line 458: this doesn't appear to be used anywhere

engine/apps/telegram/tests/test_webhook.py
```diff
@@ -0,0 +1,114 @@
+from unittest.mock import patch
+
+import pytest
+from django.urls import reverse
+from rest_framework import status
+from rest_framework.test import APIClient
+
+url = reverse("telegram:incoming_webhook")
```
matiasb, line 8: nit, I guess I would have this as a local var in each test initial setup

matiasb (approved):
LGTM

PR #3632, opened 2024-01-08
manually retry for `requests.exceptions.Timeout` exceptions when sending outgoing webhooks

joeyorlando:

Fixes https://github.com/grafana/oncall-private/issues/2439

## Checklist


engine/apps/webhooks/tasks/trigger_webhook.py
```diff
@@ -172,7 +189,12 @@ def make_request(webhook, alert_group, data):
 
 
 @shared_dedicated_queue_retry_task(
-    autoretry_for=(Exception,), retry_backoff=True, max_retries=1 if settings.DEBUG else 3
+    autoretry_for=(Exception,),
+    # This allows to exclude some exceptions that match autoretry_for but for which you don’t want a retry.
+    # https://docs.celeryq.dev/en/stable/userguide/tasks.html#Task.dont_autoretry_for
+    dont_autoretry_for=(requests.exceptions.Timeout,),
+    retry_backoff=True,
+    max_retries=1 if settings.DEBUG else 3,
```
joeyorlando, line 197: main change (rest is just adding some type hints)

engine/apps/webhooks/tasks/trigger_webhook.py
```diff
@@ -52,15 +66,14 @@ def send_webhook_event(trigger_type, alert_group_id, organization_id=None, user_
     ).exclude(is_webhook_enabled=False)
 
     for webhook in webhooks_qs:
-        print(webhook.name)
```
joeyorlando, line 55: presumably we don't need this?

engine/apps/webhooks/tasks/trigger_webhook.py
```diff
@@ -172,7 +189,12 @@ def make_request(webhook, alert_group, data):
 
 
 @shared_dedicated_queue_retry_task(
-    autoretry_for=(Exception,), retry_backoff=True, max_retries=1 if settings.DEBUG else 3
+    autoretry_for=(Exception,),
+    # This allows to exclude some exceptions that match autoretry_for but for which you don’t want a retry.
+    # https://docs.celeryq.dev/en/stable/userguide/tasks.html#Task.dont_autoretry_for
+    dont_autoretry_for=(requests.exceptions.Timeout,),
```
joeyorlando, line 195: Catching this error will catch both `requests.exceptions.ConnectTimeout` and `requests.exceptions.ReadTimeout` errors ([source code](https://requests.readthedocs.io/en/latest/_modules/requests/exceptions/)):
- `requests.exceptions.ConnectTimeout`: "The request timed out while trying to connect to the remote server"
- `requests.exceptions.ReadTimeout`: "The server did not send any data in the allotted amount of time"

both of these seem out of our control

engine/apps/webhooks/tasks/trigger_webhook.py
joeyorlando: An alternative here would be to _not_ use `dont_autoretry_for` and instead manually retry the task if `isinstance(exception, requests.exceptions.Timeout)` (up to a max of 3 after which point we simply `return`). This would allow to still retry for these tasks but without having to raise an exception (and avoiding triggering our `AmixrRetriedFailedTasks` alert)

mderynck: I think this would be preferred to have some limited retry to cover customer network hiccups

engine/apps/webhooks/tasks/trigger_webhook.py
```diff
@@ -244,5 +265,24 @@ def execute_webhook(webhook_pk, alert_group_id, user_id, escalation_policy_id):
             escalation_error_code=error_code,
         )
 
-    if exception:
+    if isinstance(exception, EXECUTE_WEBHOOK_EXCEPTIONS_TO_MANUALLY_RETRY):
+        msg_details = (
```
mderynck, line 269: _nit_ label_with_underscore=value gives nice default behavior for logfmt in loki.

mderynck (approved):

PR #3638, opened 2024-01-09
Add task logging personal notifications triggered/completed counts

matiasb:
Related to https://github.com/grafana/oncall-private/issues/2347


engine/apps/alerts/tasks/check_escalation_finished.py
```diff
@@ -115,6 +115,55 @@ def check_personal_notifications_task(alert_group_id) -> None:
...
+        Q(type=UserNotificationPolicyLogRecord.TYPE_PERSONAL_NOTIFICATION_FAILED)
+        | Q(type=UserNotificationPolicyLogRecord.TYPE_PERSONAL_NOTIFICATION_SUCCESS),
+        notification_step=UserNotificationPolicy.Step.NOTIFY,
+        created_at__gt=since,
+        created_at__lte=now,
+        alert_group__in=alert_groups,
+    ).count()
+
+    task_logger.info(f"Personal notifications triggered: {triggered}")
+    task_logger.info(f"Personal notifications completed: {completed}")
```
mderynck, line 164: Would these be better as a single log line? If we are querying logs over wider time ranges to produce metrics it will be easier to have them paired rather than using the task id to correlate.  Also `notifications_triggered={triggered} notifications_completed={completed}` has nicer natural behavior with logfmt in loki.

PR #3643, opened 2024-01-10
Adds new templates cheatsheats

Konstantinov-Innokentii:


grafana-plugin/src/components/CheatSheet/CheatSheet.config.ts
```diff
@@ -172,3 +178,138 @@ export const slackMessageTemplateCheatSheet: CheatSheetInterface = {
     },
   ],
 };
+
+export const alertGroupDynamicLabelCheatSheet: CheatSheetInterface = {
+  name: 'Dynamic Label cheatsheet',
+  description: 'Dynamic Label template is used to extract value for a specified key from the alert payload.',
```
Matvey-Kuk, line 184: How is it visible in the UI? Does it make sense to duplicate the name? Maybe:
```
Dynamic Label template is used to extract value for a specified key from the alert payload.
``` 

->

```
Used to extract dynamic value from the alert payload.
```

grafana-plugin/src/components/CheatSheet/CheatSheet.config.ts
```diff
@@ -172,3 +178,138 @@ export const slackMessageTemplateCheatSheet: CheatSheetInterface = {
...
+      name: 'Additional jinja2 variables',
+      listItems: [{ listItemName: 'payload - payload of the first alert in the group' }],
+    },
+  ],
+};
+
+export const alertGroupMultiLabelExtractionCheatSheet: CheatSheetInterface = {
+  name: 'Multi-label extraction cheatsheet',
+  description:
+    'Multi-label extraction template allows extracting and modifying multiple labels from an alert payload. The Jinja template must result in string, representing valid JSON dictionary. See Examples for getting familiar with the idea',
```
Matvey-Kuk, line 221: Same question here

grafana-plugin/src/components/CheatSheet/CheatSheet.config.ts
```diff
@@ -172,3 +178,138 @@ export const slackMessageTemplateCheatSheet: CheatSheetInterface = {
...
+          listItemName: 'Extract labels from different payload fields',
+          codeExample: `{%- set labels = {} -%}
+{# add several labels #}
+{%- set labels = dict(labels, **payload.commonLabels) -%}
+{# add one label #}
+{%- set labels = dict(labels, **{"status": payload.status}) -%}
+{# add label not from payload #}
+{%- set labels = dict(labels, **{"service": "oncall"}) -%}
+{# dump labels dict to json string, so OnCall can parse it #}
+{{ labels | tojson }}
```
Matvey-Kuk, line 240: I like this example!

Matvey-Kuk (approved):

engine/common/api_helpers/mixins.py
```diff
@@ -351,9 +354,19 @@ def get_attr_template(self, attr, alert_receive_channel, render_for=None):
                 return Response({"preview": e.fallback_message}, status.HTTP_200_OK)
         else:
             templated_attr = None
-        response = {"preview": templated_attr}
+        response = {"preview": templated_attr, "is_valid_json_dict": self.is_valid_json_dict(templated_attr)}
```
vadimkerr, line 357: Could you please update the [response schema](https://github.com/grafana/oncall/pull/3643/files#diff-ed4d02ac8c4dc7e83cc8c73c852e084b30a82f0425d1149f39768d41ac44f9fcR300)?

engine/common/api_helpers/mixins.py
```diff
@@ -351,9 +354,19 @@ def get_attr_template(self, attr, alert_receive_channel, render_for=None):
...
         return Response(response, status=status.HTTP_200_OK)
 
+    def is_valid_json_dict(self, json_str):
+        try:
+            json_object = json.loads(json_str)
+            if isinstance(json_object, dict):
+                return True
+            else:
+                return False
```
vadimkerr, line 366:
```suggestion
            return isinstance(json.loads(json_str), dict)
```

engine/common/api_helpers/mixins.py
```diff
@@ -351,9 +354,19 @@ def get_attr_template(self, attr, alert_receive_channel, render_for=None):
                 return Response({"preview": e.fallback_message}, status.HTTP_200_OK)
         else:
             templated_attr = None
-        response = {"preview": templated_attr}
+        response = {"preview": templated_attr, "is_valid_json_dict": self.is_valid_json_dict(templated_attr)}
         return Response(response, status=status.HTTP_200_OK)
 
+    def is_valid_json_dict(self, json_str):
```
vadimkerr, line 360: nit: maybe a better name for this would be `is_valid_json_object`? https://en.wikipedia.org/wiki/JSON#Data_types

vadimkerr (changes requested):

grafana-plugin/src/components/CheatSheet/CheatSheet.config.ts
Konstantinov-Innokentii: It's looks like this: I think It's ok to duplicate, since name is a cheetsheet's name, not a template name
image

engine/common/api_helpers/mixins.py
Konstantinov-Innokentii: Right, thanks

engine/common/api_helpers/mixins.py
Konstantinov-Innokentii: looks better

engine/common/api_helpers/mixins.py
Konstantinov-Innokentii: Updated

vadimkerr (approved):

teodosii (approved):

PR #3668, opened 2024-01-11
AG columns polishing

teodosii:
# What this PR does

https://github.com/grafana/oncall-private/issues/2378

## Checklist


grafana-plugin/src/containers/ColumnsSelectorWrapper/ColumnsSelectorWrapper.tsx
```diff
@@ -125,7 +131,10 @@ const ColumnsSelectorWrapper: React.FC<ColumnsSelectorWrapperProps> = observer((
   }
 
   async function onColumnRemovalClick(): Promise<void> {
-    const columns = store.alertGroupStore.columns.filter((col) => col.id !== columnToBeRemoved.id);
+    const columns = store.alertGroupStore.columns.filter(
+      (col) =>
+        col.id !== columnToBeRemoved.id || (col.id === columnToBeRemoved.id && col.type !== columnToBeRemoved.type)
```
brojd, line 136: Can't we achieve the same with `({ id, type }) => id !== columnToBeRemoved.id && type !== columnToBeRemoved.type`? 🤔

grafana-plugin/src/containers/ColumnsSelector/ColumnsSelector.tsx
```diff
@@ -193,7 +210,10 @@ export const ColumnsSelector: React.FC<ColumnsSelectorProps> = observer(
             tooltipPlacement="top"
             tooltip={'Reset table to default columns'}
             disabled={isResetLoading || isDefaultColumnOrder}
-            onClick={WrapAutoLoadingState(onReset, ActionKey.RESET_COLUMNS_FROM_ALERT_GROUP)}
+            onClick={WrapAutoLoadingState(
+              WrapWithGlobalNotification(onReset, { success: 'Columns list has been reset' }),
```
brojd, line 214: I don't think using `WrapAutoLoadingState` and `WrapWithGlobalNotification` in components directly is a good thing. I can see why we're doing it here: because in components we have functions that wrap multiple store methods and we want loading state / global notification to work with both them at once. 

However, such pattern is a cause of making our containers/components huge. Ideally, we should move towards making our containers and components smaller and less responsible by moving such things into the store layer and encapsulating as much frontend business logic as possible. In this case we could have a dedicated action in MobX store that performs 2 other actions and this wrapper action would be decorated with `@AutoLoadingState` and `@WithGlobalNotification`. Then in component you simply trigger it. And the same could be done with 2 other places in PR where we use `WrapAutoLoadingState(
              WrapWithGlobalNotification(...))`
              
I'm happy to discuss this topic if you want :)

grafana-plugin/src/utils/decorators.ts
```diff
@@ -34,6 +34,23 @@ type GlobalNotificationConfig = {
   failureType?: 'error' | 'warning';
 };
 
+export function WrapWithGlobalNotification(
```
brojd, line 37: As mentioned before, let's consider doing it only in the store layer, then maybe we don't need this util

grafana-plugin/src/containers/ColumnsSelectorWrapper/ColumnsSelectorWrapper.tsx
teodosii: No because it's not the same condition check. Yours assumes that the id always needs to be different, which is not the case.

grafana-plugin/src/containers/ColumnsSelector/ColumnsSelector.tsx
teodosii: I see your point but this would be only possible if I'd move the reset out of the class to the store, but we don't really divide the logic this way in other parts of the app. Reset function handles these 2 calls

```jsx
await alertGroupStore.resetTableSettings();
await alertGroupStore.fetchTableSettings();
```

These together are *only* used in this component, I'd kinda dislike seeing it in the store just so that I can wrap it with the decorators...

brojd: But also `resetTableSettings` is only used by this component so why not to invoke `fetchTableSettings` within `resetTableSettings` in store?

The idea is to flatten our components and make our view layer simple and dumb:)

grafana-plugin/src/containers/ColumnsSelector/ColumnsSelector.tsx
```diff
@@ -206,12 +226,19 @@ export const ColumnsSelector: React.FC<ColumnsSelectorProps> = observer(
       </div>
     );
 
+    function mapColumns(columns: AlertGroupColumn[]) {
```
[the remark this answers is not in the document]
brojd: This function accepts a list of columns and returns new list of columns while each column will have appended id. From the type itself it's easy to understand that it accepts a list of columns and returns also list of columns. What types will not tell you is what's the logic inside this map. And here we append id. In functional programming it's a common way of naming things like that. Then for example if you need to compose multiple map functions it's easy to understand what you're doing, for example:
`appendColumnId(appendColumnName(filterOutRemovedColumns(columns)))`

`mapColumnsToDnd` is ok

maskin25 (approved):

PR #3698, opened 2024-01-16
Update alert group personal notifications checker to check sent SMS

matiasb:
Sent SMS messages are considered completed for our purpose here (ie. do not wait for Twilio delivered confirmation).

### DIFF
--- a/engine/apps/alerts/tasks/check_escalation_finished.py
+++ b/engine/apps/alerts/tasks/check_escalation_finished.py
@@ -7,6 +7,8 @@ from django.db.models import Avg, F, Max, Q
 from django.utils import timezone
 
 from apps.alerts.tasks.task_logger import task_logger
+from apps.phone_notifications.models import SMSRecord
+from apps.twilioapp.models import TwilioSMSstatuses
 from common.custom_celery_tasks.log_exception_on_failure_task import shared_log_exception_on_failure_task
 from common.database import get_random_readonly_database_key_if_present_otherwise_default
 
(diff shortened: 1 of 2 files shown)


engine/apps/alerts/tasks/check_escalation_finished.py
```diff
@@ -106,10 +108,17 @@ def check_alert_group_personal_notifications_task(alert_group_id) -> None:
...
+    sent_but_not_delivered_sms = SMSRecord.objects.filter(
+        represents_alert_group_id=alert_group_id,
+        twilioapp_twiliosmss__status=TwilioSMSstatuses.SENT,
+    ).count()
+
     base_msg = f"Alert group {alert_group_id}"
+    completed += sent_but_not_delivered_sms
     delta = triggered - completed
     if delta > 0:
-        # TODO: when success notifications are setup for every backend, raise exception here
```
joeyorlando, line 112: do we want to raise an exception here still?

joeyorlando (approved):

engine/apps/alerts/tasks/check_escalation_finished.py
matiasb: Good question :+1: . Not now, we changed the approach. The original idea was for this to be part of the alert group auditor checks, so raising an exception would make the escalation checker heartbeat fail, but we switched this to be a separated thing to avoid overloading the alert group auditor checks and mix escalation with personal notifications (personal notifications check run on separated tasks).

Issue #3673, opened 2024-01-12
`/escalate-ops` Slack command is sometimes timing out for ops.grafana-ops.net OnCall


iainlane:
### What went wrong?

**What happened**:
- When I type `/escalate-ops` in Slack I get an `operation_timeout` error:

![image](https://github.com/grafana/oncall/assets/321014/885ef649-13b9-4138-b03a-4871826e41cf)

**What did you expect to happen**:
- I expect to see the UI and be able to initiate a page

### How do we reproduce it?

1. On the Grafana Labs internal Slack, type `/escalate-ops` and press enter
2. After about 3 seconds it times out

This always happens for me, but others report it works for them sometimes.

### Grafana OnCall Version

r309-v1.3.85

### Product Area

Chatops

### Grafana OnCall Platform?

I use Grafana Cloud

iainlane:
[Here's one of my requests showing it's a timeout because we take too long to respond](https://admin-ops-us-east-0.grafana-ops.net/grafana/explore?schemaVersion=1&panes=%7B%22ux1%22:%7B%22datasource%22:%22loki-ops%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22%7Bcluster%3D%5C%22ops-us-east-0%5C%22,%20namespace%3D%5C%22amixr-staging%5C%22%7D%20%7C%20logfmt%20%7C%20google_trace_id%3D%607fee1f7f0581401cb91593bd65c35a6b%2F14022802583067776224%60%22,%22queryType%22:%22range%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22loki-ops%22%7D,%22editorMode%22:%22code%22%7D%5D,%22range%22:%7B%22from%22:%22now-6h%22,%22to%22:%22now%22%7D%7D%7D&orgId=1)

matiasb:
Hi @iainlane, we just deployed the update linked above which should help with the timeout you mention. Let us know if you can take a look and report back if you notice any changes. Thanks!

PR #3699, opened 2024-01-16
Update schedule on-call cache on scheduled refresh tasks

matiasb:
Related to https://github.com/grafana/oncall/issues/3673
Keep cache up to date on every schedule refresh task run (which should keep cache populated every time), helping on any call using cached information (particularly the direct paging slack dialog building).


joeyorlando (approved):
nice 👍

PR #3703, opened 2024-01-17
Update and refactor users API team filter

matiasb:
This should hopefully fix the lint issue [here](https://drone.grafana.net/grafana/oncall/3361/1/7)

### DIFF
--- a/engine/apps/api/views/user.py
+++ b/engine/apps/api/views/user.py
@@ -138,7 +138,7 @@ class UserFilter(ByTeamModelFieldFilterMixin, filters.FilterSet):
     # TODO: remove "roles" in next version
     roles = filters.MultipleChoiceFilter(field_name="role", choices=LegacyAccessControlRole.choices())
     permission = filters.ChoiceFilter(method="filter_by_permission", choices=ALL_PERMISSION_CHOICES)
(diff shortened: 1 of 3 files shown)


joeyorlando (approved):

engine/apps/api/views/user.py
```diff
@@ -138,7 +138,7 @@ class UserFilter(ByTeamModelFieldFilterMixin, filters.FilterSet):
     # TODO: remove "roles" in next version
     roles = filters.MultipleChoiceFilter(field_name="role", choices=LegacyAccessControlRole.choices())
     permission = filters.ChoiceFilter(method="filter_by_permission", choices=ALL_PERMISSION_CHOICES)
-    team = TeamModelMultipleChoiceFilter(field_name="teams", null_label=None, null_value=None)
+    team = TeamModelMultipleChoiceFilter(field_name="teams")
```
vadimkerr, line 141: I set `null_label` and `null_value` to be None in https://github.com/grafana/oncall/pull/3666 so that we don't allow `null` team here, because it's not clear what users are on the `null` team. Do we want to support this?

matiasb: As discussed, we are going to keep the filter for now (it allows filtering users with no assigned team) because other internal endpoints do not restrict this option atm (also, clients are not using this no-team functionality). We may want to decide a unified approach later.

PR #3721, opened 2024-01-19
Chatops api v3

Konstantinov-Innokentii:
This PR makes OnCall compatible with chatops-proxy v3.  When CHATOPS_V3 is enabled, oncall will use new api client to register tenants and slack installations. Also I added v3 routes for slack and telegram, so it's possible to test new chatops proxy.

Currently two versions of chatops-proxy api are deployed, but they are not compatible. They are doing same thing, using different db model and tables. Once only v3 version will be left in prod, I'll remove CHATOPS_V3 env var, all leftovers of previous api client and v3 slack and telegram routes.


engine/apps/slack/views.py
```diff
@@ -578,7 +578,10 @@ def post(self, request):
             if slack_team_identity is not None:
                 clean_slack_integration_leftovers.apply_async((organization.pk,))
                 if settings.FEATURE_MULTIREGION_ENABLED:
-                    delete_slack_connector(str(organization.uuid))
+                    if settings.CHATOPS_V3:
+                        unlink_slack_team(str(organization.uuid), slack_team_identity.slack_id)
+                    else:
+                        delete_slack_connector(str(organization.uuid))
```
vadimkerr, line 584: maybe it'd be better to move this `if settings.CHATOPS_V3` logic to `delete_slack_connector` instead? the same goes for `check_slack_installation_possible`, `create_slack_connector`, `create_oncall_connector`, `delete_oncall_connector`, etc. 
This way a caller wouldn't need to know if it's chatops v3 or not, wdyt?

engine/common/oncall_gateway/client.py
```diff
@@ -0,0 +1,155 @@
...
+        # Since chatops-proxy is internal service messages shouldn't be exposed to the user
+        self.msg = msg
+
+    def __str__(self):
+        return f"LabelsRepoAPIException: status={self.status} url={self.url} method={self.method} error={self.msg}"
+
+
+class ChatopsProxyAPIClient:
+    def __init__(self, url: str, token: str):
+        self.base_url = url
```
vadimkerr, line 52: nit: looks like `self.base_url` is only used in the constructor, consider removing it

engine/common/oncall_gateway/client.py
```diff
@@ -0,0 +1,155 @@
...
+        self.api_base_url = urljoin(self.base_url, "api/v3")
+        self.api_token = token
+
+    # OnCall Tenant
+    def register_tenant(
+        self, service_tenant_id: str, cluster_slug: str, service_type: str
+    ) -> tuple[Optional[Tenant], requests.models.Response]:
```
vadimkerr, line 59:
```suggestion
    ) -> tuple[Tenant, requests.models.Response]:
```

engine/common/oncall_gateway/client.py
```diff
@@ -0,0 +1,155 @@
...
+        url = f"{self.api_base_url}/tenants/register"
+        d = {
+            "tenant": {
+                "service_tenant_id": service_tenant_id,
+                "cluster_slug": cluster_slug,
+                "service_type": service_type,
+            }
+        }
```
vadimkerr, line 67: nit: maybe refactor this to a separate method? there are 3 usages below that are exactly the same

engine/common/oncall_gateway/tasks.py
```diff
@@ -98,3 +99,102 @@ def delete_slack_connector_async_v2(**kwargs):
     except Exception as e:
         task_logger.error(f"Failed to delete SlackConnectorV2 oncall_org_id={oncall_org_id} exc={e}")
         raise e
+
+
+# New tasks to use once chatops v3 is landed
+@shared_dedicated_queue_retry_task(
+    autoretry_for=(Exception,),
+    retry_backoff=True,
+    max_retries=100,
```
vadimkerr, line 108: nit: isn't 100 too much considering it uses exponential backoff? maybe have like 10?

engine/common/oncall_gateway/tasks.py
```diff
@@ -98,3 +99,102 @@ def delete_slack_connector_async_v2(**kwargs):
...
+def unregister_oncall_tenant_async(**kwargs):
+    service_tenant_id = kwargs.get("service_tenant_id")
+    cluster_slug = kwargs.get("cluster_slug")
+    service_type = kwargs.get("service_type")
+
+    client = ChatopsProxyAPIClient(settings.ONCALL_GATEWAY_URL, settings.ONCALL_GATEWAY_API_TOKEN)
+    try:
+        client.unregister_tenant(service_tenant_id, cluster_slug, service_type)
+    except Exception as e:
+        task_logger.error(f"Failed to delete OnCallConnector: {e} service_tenant_id={service_tenant_id}")
```
vadimkerr, line 151:
```suggestion
        task_logger.error(f"Failed to delete OnCallConnector: {e} service_tenant_id={service_tenant_id}")
        raise e
```

engine/common/oncall_gateway/client.py
```diff
@@ -0,0 +1,155 @@
...
+
+class ChatopsProxyAPIClient:
+    def __init__(self, url: str, token: str):
+        self.base_url = url
+        self.api_base_url = urljoin(self.base_url, "api/v3")
+        self.api_token = token
+
+    # OnCall Tenant
+    def register_tenant(
+        self, service_tenant_id: str, cluster_slug: str, service_type: str
```
vadimkerr, line 58: do we plan to add new service types? if not, maybe it makes sense to just use `SERVICE_TYPE_ONCALL`?

engine/common/oncall_gateway/utils.py
```diff
@@ -59,3 +64,92 @@ def create_slack_connector(oncall_org_id: str, slack_id: str, backend: str):
...
+
+
+def unlink_slack_team(service_tenant_id: str, slack_team_id: str):
+    unlink_slack_team_async.apply_async(
+        kwargs={
+            "service_tenant_id": service_tenant_id,
+            "slack_team_id": slack_team_id,
+            "service_type": SERVICE_TYPE_ONCALL,
+        }
+    )
```
vadimkerr, line 155: nit: having a function that just calls a celery task seems a bit sus to me, maybe it makes sense to call the task directly in caller code?

engine/engine/urls.py
```diff
@@ -63,6 +63,7 @@
         path("slack/", include("apps.slack.urls")),
     ]
 
+
```
vadimkerr, line 66: nit

```suggestion
```

vadimkerr (approved):
The changes look good to me! Suggestion: consider adding some simple unit tests to cover the new functionalities.

engine/common/oncall_gateway/utils.py
Konstantinov-Innokentii: The idea is that callers just use utils from the gateway package, without knowing anything about sync/async calls

engine/common/oncall_gateway/client.py
Konstantinov-Innokentii: I made client not aware of exact service type, since it's just a client to call the API, which supports several service types. Service type passed in utils function, which acts as a "logic" layer. So idea is to not to leak any logic (like assumptions about service type) in the client.

engine/common/oncall_gateway/tasks.py
Konstantinov-Innokentii: I choose 100 retries initially to be sure, that we will get the alert about retrying task in the monitoring, since it might affect user's chatops experience. Probably it's better to have some counter metric, but I don't want to change that in  this PR to limit the amount of changes.

engine/common/oncall_gateway/client.py
Konstantinov-Innokentii: Two of them are identical, third has a "slack_team_id" field added. I prefer to keep it simple here for now.

engine/apps/slack/views.py
Konstantinov-Innokentii: Implemented, thanks!

joeyorlando (commented):
can we add some tests for this new functionality?

marsn:
This change appears to break integration with Slack(tested with helm). 
As a workaround setting ONCALL_GATEWAY_URL fixes the issue.
`2024-01-30 17:53:06 source=engine:app google_trace_id=none logger=apps.social_auth.middlewares SocialAuthAuthCanceledExceptionMiddleware.process_exception: Invalid URL 'api/v1/slack_team_connectors/check_installation_possible?slack_team_id=T023JQA1TEY&oncall_org_id=a2734e61-e4cf-425c-94fb-b6cb02d31ca8&backend=None': No scheme supplied. Perhaps you meant https://api/v1/slack_team_connectors/check_installation_possible?slack_team_id=T023JQA1TEY&oncall_org_id=a2734e61-e4cf-425c-94fb-b6cb02d31ca8&backend=None?`

PR #3727, opened 2024-01-22
Show extra fields for servicenow

teodosii:
# What this PR does

https://github.com/grafana/oncall-private/issues/2459


brojd (approved):

grafana-plugin/src/components/GForm/GForm.tsx
```diff
@@ -193,6 +194,14 @@ class GForm extends React.Component<GFormProps, {}> {
               return null; // don't render the field
             }
 
+            if (formItem.type === FormItemType.PlainLabel) {
```
maskin25, line 197: let's maybe use `FormItemType.Other` and provide custom renderer which render just `formItem.label` and nothing else

maskin25: I think we don't need to extend `FormItemType` with  `FormItemType.PlainLabel`, we can make do with existing GForm features

teodosii: it's overhead really, adds just too much code. We will get rid of these anyway, the forms don't work as expected given that there's too many abstraction layers.

Issue #3729, opened 2024-01-22
`AlertGroup matching query does not exist.` in `apps.alerts.tasks.send_update_log_report_signal.send_update_log_report_signal`


joeyorlando:
### What went wrong?

**What happened**:

[logs](https://ops.grafana-ops.net/explore?schemaVersion=1&panes=%7B%22jaq%22:%7B%22datasource%22:%22000000193%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22%7Bcluster%3D~%5C%22%28prod-eu-west-0%7Cprod-us-central-0%7Cops-us-east-0%29%5C%22,%20namespace%3D~%5C%22amixr-%28prod%7Cstaging%29%5C%22,%20pod%3D%5C%22amixr-engine-celery-retry-cccc6bfd7-2zq8s%5C%22%7D%22,%22queryType%22:%22range%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22000000193%22%7D,%22editorMode%22:%22code%22%7D%5D,%22range%22:%7B%22from%22:%221705946457994%22,%22to%22:%221705946472764%22%7D%7D%7D&orgId=1)

```
2024-01-22 13:01:06.018 | apps.alerts.models.alert_group.AlertGroup.DoesNotExist: AlertGroup matching query does not exist. |  
-- | -- | --
  |   | 2024-01-22 13:01:06.018 | raise self.model.DoesNotExist( |  
  |   | 2024-01-22 13:01:06.018 | File "/usr/local/lib/python3.11/site-packages/django/db/models/query.py", line 637, in get |  
  |   | 2024-01-22 13:01:06.018 | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ |  
  |   | 2024-01-22 13:01:06.018 | return getattr(self.get_queryset(), name)(*args, **kwargs) |  
  |   | 2024-01-22 13:01:06.018 | File "/usr/local/lib/python3.11/site-packages/django/db/models/manager.py", line 87, in manager_method |  
  |   | 2024-01-22 13:01:06.018 | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ |  
  |   | 2024-01-22 13:01:06.018 | alert_group = AlertGroup.objects.get(id=alert_group_pk) |  
[...]

PR #3731, opened 2024-01-23
Add transaction on_commit before signals for alert group actions

mderynck:
# What this PR does
Add transactions around log record creation and check transaction on_commit before sending signals passing DB id of alert group log records.  In cases for delete we can then assume any missing IDs on tasks are from intentionally deleted alert groups and we can stop tasks from retrying endlessly.

## Checklist


joeyorlando:
this _may_ also close https://github.com/grafana/oncall/issues/3729

iskhakov (commented):
I like the idea, I see that it worked in the past, but do we have a transaction blocks in these places?

engine/apps/alerts/models/alert_group.py
```diff
@@ -648,11 +653,7 @@ def acknowledge_by_user(self, user: User, action_source: typing.Optional[ActionS
             f"log record {log_record.pk} with type '{log_record.get_type_display()}', action source: {action_source}"
         )
 
-        alert_group_action_triggered_signal.send(
-            sender=self.acknowledge_by_user,
-            log_record=log_record.pk,
-            action_source=action_source,
-        )
+        transaction.on_commit(partial(send_alert_group_signal.delay, log_record.pk))
```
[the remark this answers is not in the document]
mderynck: From the docs:
`If you call on_commit() while there isn’t an open transaction, the callback will be executed immediately.`

engine/apps/telegram/tasks.py
```diff
@@ -239,18 +239,14 @@ def on_alert_group_action_triggered_async(log_record_id):
...
-                f"was deleted. log record {log_record_id}, retries: {retries_count}"
-            )
-            return
-        else:
-            raise e
+    except AlertGroupLogRecord.DoesNotExist:
+        logger.warning(
+            f"AlertGroupTelegramRepresentative: log record {log_record_id} never created or has been deleted"
+        )
+        return
```
matiasb, line 249: Nice

engine/apps/alerts/tasks/delete_alert_group.py
```diff
@@ -25,9 +26,36 @@ def delete_alert_group(alert_group_pk, user_pk):
...
+)
+def send_alert_group_signal_for_delete(alert_group_pk: int, log_record_pk: int) -> None:
     try:
-        alert_group.delete_by_user(user)
+        alert_group_action_triggered_signal.send(
+            sender=None,
+            log_record=log_record_pk,
+            action_source=None,
+            force_sync=True,
+        )
```
matiasb, line 42: So, we trigger the action triggered signal and queue the cleanup task after that. Makes sense :+1:

engine/apps/alerts/tasks/delete_alert_group.py
```diff
@@ -10,7 +11,7 @@
 @shared_dedicated_queue_retry_task(
     autoretry_for=(Exception,), retry_backoff=True, max_retries=1 if settings.DEBUG else None
 )
-def delete_alert_group(alert_group_pk, user_pk):
+def delete_alert_group(alert_group_pk: int, user_pk: int) -> None:
```
matiasb, line 14: Maybe not for this PR, but wondering if this still needs to be a task?

mderynck: I think we can keep it for now, I guess something could go wrong in the stop escalation part of delete and we could retry here.

engine/apps/alerts/models/alert_group.py
```diff
@@ -648,11 +653,7 @@ def acknowledge_by_user(self, user: User, action_source: typing.Optional[ActionS
             f"log record {log_record.pk} with type '{log_record.get_type_display()}', action source: {action_source}"
         )
 
-        alert_group_action_triggered_signal.send(
-            sender=self.acknowledge_by_user,
-            log_record=log_record.pk,
-            action_source=action_source,
```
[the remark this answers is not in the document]
mderynck: I'll double-check this, I think only slack uses action_source from the signal but force_sync was used instead (for the delete case).  In all cases it should be able to be taken from the log record instead of it being passed on the signal, I wanted to avoid changing the signature on send_alert_group_signal/having to add another task with a different signature.

mderynck:
> I like the idea, I see that it worked in the past, but do we have a transaction blocks in these places?

We do not, going to test this on dev to see if it improves the issue.  It comes down to things should have autocommit but we keep finding cases where the objects are missing.

mderynck:
Added back raising exceptions.  In most cases they will not retry and we will get a log message and task failure.  In the delete cases only a log message is written, no reason to raise an exception if it was already deleted.  There are a couple cases for slack and telegram where it is inside of a signal call, here will still retry to not change the behavior at the top level.  If we want to not retry the signal cases we will need to add an intermediate task to isolate the one handler from others.

matiasb (approved):
LGTM

PR #3732, opened 2024-01-23
Minor Fix to Format of REFRESH-INTERVAL in ical_utils.py

ESNewmanium:
# What this PR does
Minor formatting change to the suggested REFRESH-INTERVAL of iCal exports. DURATION units less than 1d must be prefixed by "T". Fixes issue with Atlassian Confluence failing to subscribe to iCal URLs from Grafana OnCall

RFC 2445 explains this a bit more clearly in section 4.3.6 on DURATION. 
https://www.ietf.org/rfc/rfc2445.txt

Obviously I wish Atlassian could just be a little more forgiving in their digestion of the iCal data since clearly Gmail and others have no problem with it, but I doubt I'm going to get much traction with them (we do have a case open though).

I haven't logged one yet but I can if you want. Again, my main issue was with Atlassian Confluence. Kept throwing errors of "The uploaded data does not seem to be iCalendar content" as long as the REFRESH-INTERVAL DURATION was less than 1 day and lacking the "T" character.

## Checklist

### DIFF
--- a/engine/apps/schedules/ical_utils.py
+++ b/engine/apps/schedules/ical_utils.py
@@ -716,7 +716,7 @@ def create_base_icalendar(name: str) -> Calendar:
     cal.add("version", "2.0")
     cal.add("prodid", "//Grafana Labs//Grafana On-Call//")
     # suggested minimum interval for polling for changes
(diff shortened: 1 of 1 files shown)


matiasb:
Thanks for the contribution @ESNewmanium!
Makes sense :+1:

PR #3752, opened 2024-01-25
Integration incoming tab

teodosii:
# What this PR does

## Checklist


grafana-plugin/src/components/IntegrationCollapsibleTreeView/IntegrationCollapsibleTreeView.tsx
```diff
@@ -34,7 +41,12 @@ const IntegrationCollapsibleTreeView: React.FC<IntegrationCollapsibleTreeViewPro
   }, [configElements]);
 
   return (
-    <div className={cx('integrationTree__container')}>
+    <div
+      className={cx(
+        'integrationTree__container',
+        customContainerClassName ? `integrationTree__container--${customContainerClassName}` : ''
```
brojd, line 47: Why do we need `integrationTree__container--` prefix for class provided as a prop? Maybe better just to pass`customContainerClassName` as is and allow adding custom classname while using this component?

grafana-plugin/src/components/Timeline/TimelineItem.tsx
```diff
@@ -22,14 +22,17 @@ const TimelineItem: React.FC<TimelineItemProps> = ({
...
   textColor = '#ffffff',
   number,
 }) => {
   return (
     <li className={cx('item', className)}>
       {!isDisabled && (
-        <div className={cx('dot')} style={{ backgroundColor, color: textColor }}>
+        <div
+          className={cx('dot', backgroundColor.isClassName ? backgroundColor.color : '')}
+          style={{ backgroundColor: backgroundColor.isClassName ? '' : backgroundColor.color, color: textColor }}
```
brojd, line 34: 1. Do we need to apply color as style? Can't we just rely on class?
2. If we really want to keep both, then what do you think about having 2 separate props: `numberClassName: string` and `numberHexColor: string`? `isClassName` is a bit confusing imho and also `backgroundColor.color` doesn't necessarily need to be a class that applies color, it's just any class that is attached to the div

wdyt?

grafana-plugin/src/components/Text/Text.tsx
```diff
@@ -81,8 +83,10 @@ const Text: TextInterface = (props) => {
     setValue(e.target.value);
   }, []);
 
+  const CustomTag = customTag || (`span` as any);
```
brojd, line 86: We should get rid of `as any` by changing the type of `customTag` prop from `string` into union type (like ``span' | 'div' | 'h1' | 'h2'` etc). Then you can also set default prop when destructuring `customTag = 'span', ...rest } = props;`

grafana-plugin/src/containers/EscalationChainSteps/EscalationChainSteps.tsx
```diff
@@ -26,10 +28,19 @@ interface EscalationChainStepsProps {
   offset?: number;
 }
 
+const getStyles = (_theme: GrafanaTheme2) => {
```
brojd, line 31: underscore suggest as if it's unused param, but it is used, can we change it to just `theme`?

grafana-plugin/src/components/Timeline/TimelineItem.tsx
teodosii: This was actually a bugfix and not a needed implementation change. We need both because one is read from emotion and the other is passed as a hex number. I can rename those

grafana-plugin/src/components/IntegrationCollapsibleTreeView/IntegrationCollapsibleTreeView.tsx
teodosii: I guess because I wanted the class to be part of the TreeView component and not from the outside, this way I keep all the display changes private to that scss file. Like I didn't want option to provide a class from the outside because it should be handled by the scss part of the component

grafana-plugin/src/components/Text/Text.tsx
teodosii: Right, just that every typing I tried it was complaining it's not on the IDE's taste :laughing:

grafana-plugin/src/containers/EscalationChainSteps/EscalationChainSteps.tsx
teodosii: Good catch

grafana-plugin/src/components/IntegrationCollapsibleTreeView/IntegrationCollapsibleTreeView.tsx
brojd: With this implementation every time user of IntegrationCollapsibleTreeView component wants to apply styles provided by `.integrationTree__container--timeline-view` class, needs to now exactly that he/she needs to pass `customContainerClassName="timeline-view"`. This is not very handy and is error-prone, any typo in this string will cause a bug. Also when clas`integrationTree__container--timeline-view` is renamed in future, there will be no indication from Typescript that there are places where wrong class name is used.

I suggest one of 2 solutions:
1) Since `customContainerClassName` has only one usage, we can just create cx('timeline-view') class in `ExpandedIntegrationRouteDisplay.module.scss` and add there 
```
{
  margin-left: -19px;
  border: 0 !important;
  width: 4px;
  background-color: var(--secondary-background);
}
```
2) If we believe that it's not a single case and in future we will have more components that will want to apply the same styles, we can change `customContainerClassName: string` prop into `isTimelineView: boolean` in `IntegrationCollapsibleTreeView`. Then apply the class conditionally based on this class:
```
<div
      className={cx(
        'integrationTree__container',
        { `integrationTree__container--timeline-view`: prop.isTimelineView }
      )}
    >
```

grafana-plugin/src/components/IntegrationCollapsibleTreeView/IntegrationCollapsibleTreeView.tsx
```diff
@@ -126,6 +141,22 @@ const IntegrationCollapsibleTreeItem: React.FC<{
...
+    if (item.isTextIcon && elementPosition) {
+      return (
+        <Text type="primary" customTag="h6" className={cx('number-icon')}>
+          {elementPosition}
+        </Text>
+      );
+    }
+
+    if (item.canHoverIcon) {
+      return <IconButton aria-label="" name={getIconName()} onClick={iconOnClickFn} size="lg" />;
```
maskin25, line 154: handleIconClick

grafana-plugin/src/pages/integration/Integration.tsx
```diff
@@ -551,8 +552,10 @@ class Integration extends React.Component<IntegrationProps, IntegrationState> {
           </div>
         ),
       },
-      this.renderRoutesFn(),
-    ].filter((opt) => opt);
+      this.renderRoutesFn() as any,
```
maskin25, line 555: can we get rid of `any` here?

grafana-plugin/src/pages/integration/Integration.tsx
```diff
@@ -551,8 +552,10 @@ class Integration extends React.Component<IntegrationProps, IntegrationState> {
...
+    ];
+
+    return configs.filter((opt) => opt);
```
maskin25, line 558: `configs.filter(Boolean)`

grafana-plugin/src/pages/integrations/Integrations.tsx
```diff
@@ -347,7 +347,7 @@ class Integrations extends React.Component<IntegrationsProps, IntegrationsState>
       >
         <TextEllipsisTooltip placement="top" content={item.verbal_name}>
           <Text type="link" size="medium">
-            <Emoji className={cx('title', TEXT_ELLIPSIS_CLASS)} text={item.verbal_name} />
+            <Emoji className={cx(TEXT_ELLIPSIS_CLASS)} text={item.verbal_name} />
```
maskin25, line 350: AFAIR we use `u-`prefixed classes for such cases like `u-flex-center`

teodosii: TEXT_ELLIPSIS_CLASS was set to `overflow-child` before this PR. Not sure what you meant by this, there's a constant for it just for easier referencing.

grafana-plugin/src/components/Text/Text.tsx
```diff
@@ -26,6 +26,7 @@ interface TextProps extends HTMLAttributes<HTMLElement> {
   editModalTitle?: string;
   maxWidth?: string;
   clickable?: boolean;
+  customTag?: string;
```
brojd, line 29: Can we use union type here as I mentioned in another comment, `'span' | 'h1' | ...`?

grafana-plugin/src/components/Timeline/TimelineItem.tsx
brojd: What about using 2 separate props instead of `backgroundColor` object with `isClassName` boolean as mentioned in point 2 above?

grafana-plugin/src/components/Text/Text.tsx
teodosii: Sorry I missed that part, will add the supported ones

grafana-plugin/src/components/Text/Text.tsx
```diff
@@ -81,8 +83,10 @@ const Text: TextInterface = (props) => {
     setValue(e.target.value);
   }, []);
 
+  const CustomTag = (customTag || `span`) as unknown as React.ComponentType<any>;
```
brojd, line 86: I think now we can get rid of casting here:
```
const CustomTag = customTag
```

And in line 60 apply default prop
```
...,
clickable,
customTag = 'span',
```

brojd (approved):

grafana-plugin/src/assets/style/vars.css
```diff
@@ -46,6 +46,7 @@
   --background-canvas: #f4f5f5;
   --background-primary: #fff;
   --background-secondary: #f4f5f5;
+  --border-medium-color: rgba(36, 41, 46, 0.3);
   --border-medium: 1px solid rgba(36, 41, 46, 0.3);
```
maskin25, line 50: `--border-medium: 1px solid var(--border-medium-color);`

grafana-plugin/src/components/IntegrationCollapsibleTreeView/IntegrationCollapsibleTreeView.module.scss
```diff
@@ -1,46 +1,54 @@
...
-  margin-left: 6px;
+  margin-left: 16px;
   visibility: hidden;
   overflow-y: hidden;
   height: 0;
 
   &--visible {
     visibility: visible;
     height: auto;
+    overflow: unset;
```
maskin25, line 15: `overflow: visible;` ?

grafana-plugin/src/components/IntegrationCollapsibleTreeView/IntegrationCollapsibleTreeView.module.scss
```diff
@@ -1,46 +1,54 @@
...
+  }
 
   &--hidden {
     display: none;
   }
 }
 
+.integrationTree__container--timeline-view .integrationTree__group::before {
+  margin-left: -18px;
+  border: 0 !important;
```
maskin25, line 40: can we get rid of `important` here?

grafana-plugin/src/components/IntegrationCollapsibleTreeView/IntegrationCollapsibleTreeView.tsx
```diff
@@ -103,19 +111,21 @@ const IntegrationCollapsibleTreeView: React.FC<IntegrationCollapsibleTreeViewPro
 
 const IntegrationCollapsibleTreeItem: React.FC<{
   item: IntegrationCollapsibleItem;
+  elementPosition?: number;
   isExpanded: boolean;
   onClick: () => void;
-}> = ({ item, isExpanded, onClick }) => {
-  const iconOnClickFn = !item.isCollapsible ? undefined : onClick;
+}> = ({ item, elementPosition, isExpanded, onClick }) => {
+  const handleIconClick = !item.isCollapsible ? undefined : onClick;
```
maskin25, line 118: more straightforward and easy to read:
`const handleIconClick = item.isCollapsible ? onClick : undefined;`

grafana-plugin/src/components/IntegrationCollapsibleTreeView/IntegrationCollapsibleTreeView.tsx
```diff
@@ -126,6 +136,22 @@ const IntegrationCollapsibleTreeItem: React.FC<{
     </div>
   );
 
+  function renderIcon() {
+    if (item.isTextIcon && elementPosition) {
+      return (
+        <Text type="primary" customTag="h6" className={cx('number-icon')}>
```
maskin25, line 142: <Text.Title level={6} type="primary" className={cx('number-icon')}>

grafana-plugin/src/components/IntegrationCollapsibleTreeView/IntegrationCollapsibleTreeView.tsx
```diff
@@ -126,6 +136,22 @@ const IntegrationCollapsibleTreeItem: React.FC<{
...
+          {elementPosition}
+        </Text>
+      );
+    }
+
+    if (item.canHoverIcon) {
+      return <IconButton aria-label="" name={getIconName()} onClick={handleIconClick} size="lg" />;
```
maskin25, line 149: `aria-label=""` does it make sense?

grafana-plugin/src/components/Text/Text.tsx
```diff
@@ -26,6 +26,7 @@ interface TextProps extends HTMLAttributes<HTMLElement> {
   editModalTitle?: string;
   maxWidth?: string;
   clickable?: boolean;
+  customTag?: 'h6' | 'span';
```
maskin25, line 29: we don't need this prop since we have `Text.Title` component

grafana-plugin/src/containers/IntegrationContainers/ExpandedIntegrationRouteDisplay/ExpandedIntegrationRouteDisplay.tsx
```diff
@@ -183,25 +162,58 @@ const ExpandedIntegrationRouteDisplay: React.FC<ExpandedIntegrationRouteDisplayP
...
+          isHidden: false,
+          isCollapsible: false,
+          isExpanded: false,
+          isTextIcon: true,
+          collapsedView: null,
+          canHoverIcon: false,
+          expandedView: () => (
+            <div className={cx('adjust-element-padding')}>
+              <VerticalGroup spacing="sm">
+                <Text customTag="h6" type="primary">
```
maskin25, line 211: `Text.Title level={6} type="primary"`

grafana-plugin/src/containers/EscalationChainSteps/EscalationChainSteps.tsx
```diff
@@ -80,13 +91,19 @@ const EscalationChainSteps = observer((props: EscalationChainStepsProps) => {
             return null;
           }
 
+          const extraProps: Partial<EscalationPolicyProps> = {};
```
maskin25, line 94: I see a lot of logic inside loop, that not very readable, can we move preparing props for  `EscalationPolicy` outside of loop, e.g create some helper...

PR #3776, opened 2024-01-29
Improve GSelect modelName typing

maskin25:
# What this PR does

Improve GSelect modelName typing

## Checklist

### DIFF
--- a/grafana-plugin/src/utils/types.ts
+++ b/grafana-plugin/src/utils/types.ts
@@ -9,6 +9,6 @@ export interface TableColumn {
   grow?: number;
 }
 
-export type KeysMatching<T extends object, V> = keyof {
+export type PropertiesThatExtendsAnotherClass<T extends object, V> = keyof {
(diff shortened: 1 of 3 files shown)


grafana-plugin/src/utils/types.ts
```diff
@@ -8,3 +8,7 @@ export interface TableColumn {
   className?: string;
   grow?: number;
 }
+
+export type KeysMatching<T extends object, V> = keyof {
+  [K in keyof T as T[K] extends V ? K : never]: any;
+};
```
brojd, line 14: Works cool! I would suggest more explicit naming though + using `unknown` instead of `any`:
sth like
```
export type PropertiesThatExtendsAnotherClass<OriginalObj, AnotherClass> = keyof {
  [Prop in keyof OriginalObj as OriginalObj[Prop] extends AnotherClass ? Prop : never]: unknown;
};
```

maskin25: fixed

brojd (approved):

PR #3759, opened 2024-01-26
Refactor create_alert task

iskhakov:
# What this PR does

This PR simplifies alert group/alert creation, so the alert created and escalation started in the same task. 

## Checklist

- [ ] Unit, integration, and e2e (if applicable) tests updated
- [ ] Documentation added (or `pr:no public docs` PR label added if not required)
- [ ] `CHANGELOG.md` updated (or `pr:no changelog` PR label added if not required)

PR #3824, opened 2024-02-01
fix "AlertGroup not found" exception in `apps.metrics_exporter.tasks.update_metrics_for_alert_group` task

joeyorlando:

[Logs](https://ops.grafana-ops.net/goto/_iQgeLpIR?orgId=1)

Closes https://github.com/grafana/oncall-private/issues/2516

## Checklist


engine/apps/alerts/models/alert_group.py
```diff
@@ -141,7 +142,7 @@ def get_or_create_grouping(self, channel, channel_filter, group_data, received_a
                 web_title_cache=group_data.web_title_cache,
                 received_at=received_at,
             )
-            alert_group_created_signal.send(sender=self.__class__, alert_group=alert_group)
```
joeyorlando, line 144: this signal should be wrapped in a `transaction.on_commit` (to avoid having [this task](https://github.com/grafana/oncall/blob/dev/engine/apps/alerts/models/alert_group.py#L613) being occasionally unable to find the alert group object in the db) **but** if you look at the docstring for this method, it looks like we want to avoid wrapping the `create` call in a transaction?
> Also, "create" is invoked without transaction.atomic to reduce number of ConcurrentUpdateError's which can be
>         raised in AlertGroupQuerySet.create() due to optimistic locking of AlertGroupCounter model.

instead I think the `post_save` signal is probably our best bet here.

engine/apps/metrics_exporter/signals.py
joeyorlando, line 1: this was the only consumer of `alert_group_created_signal`. Also, move the logic in `on_alert_group_created` to `listen_for_alertgroup_model_save`

engine/apps/webhooks/presets/preset_options.py
```diff
@@ -25,6 +25,3 @@ def listen_for_webhook_save(sender: Webhook, instance: Webhook, raw: bool, *args
             WebhookPresetOptions.WEBHOOK_PRESETS[instance.preset].override_parameters_before_save(instance)
         else:
             raise NotImplementedError(f"Webhook references unknown preset implementation {instance.preset}")
-
-
-pre_save.connect(listen_for_webhook_save, Webhook)
```
joeyorlando, line 30: we're already connecting the `listen_for_webhook_save` function to the `pre_save` signal via the `@receiver` decorator. The [docs](https://docs.djangoproject.com/en/5.0/topics/signals/#connecting-receiver-functions) say you should use one or the other:
Screenshot 2024-02-01 at 16 25 13

engine/apps/metrics_exporter/apps.py
joeyorlando, line 1: I believe this was only added ([here](https://github.com/grafana/oncall/pull/3451/files#diff-ec75f8546d9662968ac9da4b2a3ac2b9a2cecb7a124d545c709b009e8b264cd1)) so that the signals would be imported/registered.. no longer needed

engine/apps/alerts/models/alert_group.py
iskhakov: > if you look at the docstring for this method

the comment says about the transaction inside the method `create`. This method is called before the signal, and no matter if it has transaction or not, signal will not be emmited if `create` failed.

This comment refers to the `transaction with select_for_update`, which could lock the table. Instead "optimistic locking" is used, check this file
https://github.com/grafana/oncall/blob/dev/engine/apps/alerts/models/alert_group_counter.py#L14

I'd prefer having the alert_group_created_signal.send called explicitly. We can move this signal to https://github.com/grafana/oncall/blob/e053eb084d84419159c0376be4de3a11a867e5ab/engine/apps/alerts/models/alert.py#L112 and wrap in `transaction.on_commit`, as we do later in the code

joeyorlando: what about this?:
```diff
-- alert_group_created_signal.send(sender=self.__class__, alert_group=alert_group)
++ transaction.on_commit(partial(alert_group_created_signal.send, sender=self.__class__, alert_group=alert_group))
```

I would argue it'd be beneficial to keep the `signal.send` call within the `get_or_create_grouping` method, because even though right now we're only calling `AlertGroupQuerySet.get_or_create_grouping` (within `Alert.create`), if in the future we call this method elsewhere, we'd have to remember to also call `signal.send` alongside the new function invocation (not sure if that makes sense?).

To be fair, I don't quite fully grok Django's `transaction.on_commit`.. In `Alert.create` we're not explicitly opening a transaction, so how does it know to open/close one? The [docs](https://docs.djangoproject.com/en/5.0/topics/db/transactions/#performing-actions-after-commit) say:
> If you call `on_commit()` while there isn’t an open transaction, the callback will be executed immediately.

iskhakov: > To be fair, I don't quite fully grok Django's transaction.on_commit.. In Alert.create we're not explicitly opening a transaction, so how does it know to open/close one? The docs say

Right, there is no transaction yet, so the callback will be called immediately (as there is autocommit). We discussed similar problem in this https://github.com/grafana/oncall/pull/3731#issuecomment-1910899206

I want to add transaction (https://github.com/grafana/oncall/pull/3759/files#r1467196264) once verify it won't break `ConcurrentUpdateError` optimistic locking. 

I found it useful to check if there is an open transaction with
```
from django.db import transaction
print(transaction.get_connection().in_atomic_block)
```

iskhakov: > I would argue it'd be beneficial to keep the signal.send call within the get_or_create_grouping method

That make sense, from the other hand I don't like this pattern as it is really easy to subscibe to that signal and unintentionally slow down alert processing

PR #3830, opened 2024-02-02
Handle ag telegram message does not exist

Ferril:
# What this PR does
Restart `send_link_to_channel_message_or_fallback_to_full_alert_group` task without retry on the first run if alert group telegram message doesn't exist. On the second run raise Exception

https://github.com/grafana/oncall-private/issues/2492
## Checklist

### DIFF
--- a/engine/apps/telegram/tasks.py
+++ b/engine/apps/telegram/tasks.py
@@ -16,6 +16,7 @@ from apps.telegram.decorators import (
     ignore_message_unchanged,
     ignore_reply_to_message_deleted,
 )
+from apps.telegram.exceptions import AlertGroupTelegramMessageDoesNotExist
 from apps.telegram.models import TelegramMessage, TelegramToOrganizationConnector
 from common.custom_celery_tasks import shared_dedicated_queue_retry_task
 from common.utils import OkToRetry
@@ -84,7 +85,7 @@ def edit_message(self, message_pk):
 
 @shared_dedicated_queue_retry_task(bind=True, autoretry_for=(Exception,), retry_backoff=True, max_retries=None)
 def send_link_to_channel_message_or_fallback_to_full_alert_group(
-    self, alert_group_pk, notification_policy_pk, user_connector_pk
+    self, alert_group_pk, notification_policy_pk, user_connector_pk, is_first_message_check=True
 ):
     from apps.telegram.models import TelegramToUserConnector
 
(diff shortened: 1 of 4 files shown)


engine/apps/telegram/tasks.py
```diff
@@ -101,6 +102,16 @@ def send_link_to_channel_message_or_fallback_to_full_alert_group(
         else:
             # seems like the message won't appear in Telegram channel, so send the full alert group to user
             user_connector.send_full_alert_group(alert_group=alert_group, notification_policy=notification_policy)
+    except AlertGroupTelegramMessageDoesNotExist as e:
+        # On the first run, if alert group message doesn't exist in db, restart the task without retry.
+        # The message is expected to appear by the next call, otherwise Exception will be raised.
+        if is_first_message_check:
+            logger.warning(f"{e}. Restarting task")
+            send_link_to_channel_message_or_fallback_to_full_alert_group.apply_async(
+                (alert_group_pk, notification_policy_pk, user_connector_pk, False), countdown=3
```
matiasb, line 111: Curious, did you consider delaying the original `send_link_to_channel_message_or_fallback_to_full_alert_group.delay` call in `telegram/models/connectors/personal.py` instead? Wondering if we just set the `countdown` there would be enough and we don't need to add the extra complexity here (which seems to be a custom/manual retry without the exception :-)). Also, how often do we get this error? (thinking if this failing may indicate some other issue causing the delayed message creation when happening)

matiasb (commented):
Changes look good, but wondering about alternatives.

engine/apps/telegram/tasks.py
Ferril: It happens not often. The case is when the first escalation step is telegram notification, sometimes we don't have alert group telegram message saved in our db by this moment. The issue resolves after one retry.
Your proposal to add a delay to call in `telegram/models/connectors/personal.py` makes sense, I was just thinking about making this more visible, because otherwise if we will call this function from other places in the future we will need to always call this task with a delay, that is only needed to cover this rare case

Ferril: Decided to add a delay to the initial call in `telegram/models/connectors/personal.py` since in this case recalling the task to avoid retry seems to be overcomplication 👍

matiasb (approved):
LGTM

Issue #3831, opened 2024-02-02
`UnicodeEncodeError: 'latin-1' codec can't encode character` in `apps.webhooks.tasks.trigger_webhook.execute_webhook` task


joeyorlando:
### What went wrong?

**What happened**:
[Logs](https://ops.grafana-ops.net/goto/SrFNV8tIR?orgId=1)

**What did you expect to happen**:
No exception should be raised here

### How do we reproduce it?

More investigation is needed to be able to reproduce

### Grafana OnCall Version

v1.3.100

### Product Area

Other

### Grafana OnCall Platform?

None

PR #3832, opened 2024-02-02
patch occasional `UnicodeEncodeError` that occurs with outgoing webhooks

joeyorlando:

Closes https://github.com/grafana/oncall/issues/3831

## Checklist

### DIFF
--- a/engine/apps/webhooks/tests/test_trigger_webhook.py
+++ b/engine/apps/webhooks/tests/test_trigger_webhook.py
@@ -1,6 +1,7 @@
 import json
 from unittest.mock import call, patch
 
+import httpretty
 import pytest
 import requests
 from django.utils import timezone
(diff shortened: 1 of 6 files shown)


engine/apps/webhooks/tests/test_trigger_webhook.py
```diff
@@ -130,51 +131,88 @@ def test_execute_webhook_integration_filter_matching(
     )
 
 
+ALERT_GROUP_PUBLIC_PRIMARY_KEY = "IXJ47FKMYYJ5U"
+
+
+@httpretty.activate(verbose=True, allow_net_connect=False)
```
joeyorlando, line 137: the exception I was trying to fix, and fully test, here happens rather deep inside the `requests` package ([logs](https://ops.grafana-ops.net/goto/SrFNV8tIR?orgId=1)). Rather than trying to figure out what to properly mock there I decided to try out `httpretty` which according to their [docs](https://httpretty.readthedocs.io/en/latest/introduction.html#motivation):

> It can be a bit of a hassle to use something like mock.Mock to stub the requests, this can work well for low-level unit tests but when writing functional or integration tests we should be able to allow the http calls to go through the TCP socket module.
> 
> HTTPretty [monkey patches](http://en.wikipedia.org/wiki/Monkey_patch) Python’s [socket](https://docs.python.org/3/library/socket.html#module-socket) core module with a fake version of the module.
> 
> Because HTTPretty implements a fake the modules [socket](https://docs.python.org/3/library/socket.html#module-socket) and [ssl](https://docs.python.org/3/library/ssl.html#module-ssl) you can use write tests to code against any HTTP library that use those modules.

in other words, we don't have to patch `requests` at all (**note**: I'm `patch`ing `requests` but w/ the `wraps` kwarg which basically sets a "spy" on the package and doesn't mock its actual implementation ([docs](https://docs.python.org/3/library/unittest.mock.html)))

matiasb (approved):
LGTM

Issue #3651, opened 2024-01-10
Migrate from grafana-toolkit to grafana create-plugin


brojd:
### What would you like to see!

grafana-toolkit is deprecated and we have warnings every time we build the frontend that we should switch to create-plugin / publish-plugin. Migration guide: https://grafana.com/developers/plugin-tools/migration-guides/migrate-from-toolkit

Parent issue: [#3650](https://github.com/grafana/oncall/issues/3650)

### Product Area

Other

iskhakov:
Needs help from @mderynck

Issue #3733, opened 2024-01-23
Frontend technical improvements


brojd:
### What would you like to see!

description

### Product Area

Other

PR #3837, opened 2024-02-05
Migrate from grafana-toolkit to grafana plugin tools

maskin25:
# What this PR does

Migrate from grafana-toolkit to grafana plugin tools

https://github.com/grafana/oncall/issues/3651

## Checklist


grafana-plugin/.config/.cprc.json
```diff
@@ -0,0 +1,3 @@
+{
+  "version": "3.3.0"
```
teodosii, line 2: is this scaffolded by `create-plugin` or what's this file for?

grafana-plugin/.nvmrc
```diff
@@ -0,0 +1 @@
+20
```
brojd, line 1: 1. It's pity .nvmrc doesn't support comments but at least maybe we can use "v20" to indicate this s about node version?
2. It would be good to restrict this version in `package.json` using https://docs.npmjs.com/cli/v10/configuring-npm/package-json#engines ? wdyt?
3. Also it this case we stick to Node v20 on local but on on Github Actions and Drone it seems we use 18.16.0. I think we should be consistent here and use the same version. If not in this PR, can we create a task and add it to our [frontend tech backlog](https://github.com/grafana/oncall/issues/3733) pls?

grafana-plugin/package.json
```diff
@@ -7,26 +7,27 @@
...
+    "build": "webpack -c ./webpack.config.ts --env production",
+    "build:dev": "webpack -w -c ./webpack.config.ts --env development",
     "labels:link": "yarn --cwd ../../gops-labels/frontend link && yarn link \"@grafana/labels\" && yarn --cwd ../../gops-labels/frontend watch",
     "labels:unlink": "yarn --cwd ../../gops-labels/frontend unlink",
     "test": "jest --verbose",
     "test:silent": "jest --silent",
     "test:e2e": "yarn playwright test --grep-invert @expensive",
     "test:e2e-expensive": "yarn playwright test --grep @expensive",
+    "e2e": "yarn exec cypress install && yarn exec grafana-e2e run",
+    "e2e:update": "yarn exec cypress install && yarn exec grafana-e2e run --update-screenshots",
```
brojd, line 19: those are not needed I guess?

grafana-plugin/webpack.config.ts
```diff
@@ -0,0 +1,36 @@
...
+    module: {
+      rules: {
+        test: CustomizeRule.Match,
+        use: CustomizeRule.Merge,
+      },
+    },
+  })(baseConfig, customConfig);
+};
+
+export default config;
```
brojd, line 36: Nice!

grafana-plugin/package.json
```diff
@@ -7,26 +7,27 @@
...
-    "ci-build:finish": "grafana-toolkit plugin:ci-build --finish",
-    "ci-package": "grafana-toolkit plugin:ci-package",
-    "ci-report": "grafana-toolkit plugin:ci-report",
+    "dev": "webpack -c ./.config/webpack/webpack.config.ts --env development",
+    "watch": "webpack -w -c ./.config/webpack/webpack.config.ts --env development",
+    "sign": "npx --yes @grafana/sign-plugin@latest",
     "start": "yarn watch",
     "plop": "plop",
+    "test:ci": "jest --passWithNoTests --maxWorkers 4",
+    "typecheck": "tsc --noEmit",
```
teodosii, line 30: do we need these 2?
`test:ci` doesn't seem to be used part of CI
`typecheck` seems more like a job for webpack than for us to run :thinking:

teodosii (approved):

grafana-plugin/.nvmrc
joeyorlando: +1 for keeping the node versions the same on CI as local. A few spots that would need to be changed (this _should_ be all, but you'll want to double check):
- https://github.com/grafana/oncall/blob/dev/grafana-plugin/Dockerfile.dev#L1
- https://github.com/grafana/oncall-private/blob/dev/.drone.yml
- https://github.com/grafana/oncall/blob/dev/.drone.yml
- https://github.com/grafana/oncall/blob/dev/.github/workflows/linting-and-tests.yml
- https://github.com/grafana/oncall/blob/dev/.github/workflows/e2e-tests.yml

grafana-plugin/package.json
maskin25: `typecheck` looks useful for me, let's keep it, and  I've removed `test:ci`

brojd (approved):

.drone.yml
```diff
@@ -380,3 +380,11 @@ name: github_api_token
 ---
 kind: signature
```
mderynck, line 381: Signature should be last thing in the file, also it needs to be updated before it will run.  See instructions [here](https://github.com/grafana/oncall/tree/dev/dev#update-drone-build)

grafana-plugin/src/network/oncall-api/autogenerated-api.types.d.ts
```diff
@@ -1,3 +1,5 @@
+// @ts-nocheck
```
brojd, line 1: We shouldn't edit this file, it's autogenerated so on the next run your change will be lost. Instead we can edit `generate-types` script to do the same on every run. I suggest:

.drone.yml
maskin25: I've moved the signature section to the end and updated the signature according to the instructions

PR #3850, opened 2024-02-07
Allow routing via labels on the Integration page

teodosii:
# What this PR does

Adds label routing on the UI

image

image

Related to https://github.com/grafana/oncall-private/issues/2461


joeyorlando:
@teodosii is this PR still relevant?

grafana-plugin/src/containers/IntegrationContainers/ExpandedIntegrationRouteDisplay/ExpandedIntegrationRouteDisplay.module.scss
```diff
@@ -47,3 +51,42 @@
 .default-route-view {
   min-height: 40px;
 }
+
+.block {
```
teodosii, line 55: Both `Expanded*` & `Collapsed*` components share some of the styles, but this will be further cleaned up when we migrate all `Integration` components to use emotion. Any additional refactoring to the existing *legacy* css/scss files is not worth it.

grafana-plugin/src/components/LabelsTooltipBadge/LabelsTooltipBadge.tsx
```diff
@@ -36,3 +38,32 @@ export const LabelsTooltipBadge: FC<LabelsTooltipBadgeProps> = ({ labels, onClic
...
+        {values.map((label) => (
+          <LabelTag key={label.key.id} label={label.key.name} value={label.value.name} />
+        ))}
+      </HorizontalGroup>
+    );
+  };
+
+  return (
+    <HorizontalGroup spacing="sm">
+      {renderer(labels.slice(0, maxCount))}
```
brojd, line 60: Why do we need renderer function instead of doing this directly?
```
<HorizontalGroup>
  {labels.slice(0, maxCount).map((label) => (
          <LabelTag key={label.key.id} label={label.key.name} value={label.value.name} />
        ))}
</HorizontalGroup>
```

teodosii: Because we reuse it

teodosii: image

grafana-plugin/src/containers/IntegrationContainers/ExpandedIntegrationRouteDisplay/ExpandedIntegrationRouteDisplay.tsx
```diff
@@ -63,6 +70,22 @@ interface ExpandedIntegrationRouteDisplayState {
   routeIdForDeletion: string;
 }
 
+enum ROUTING_OPTION {
```
brojd, line 73: We have similar places in codebase but it would be good to use PascalCase for enum name and its keys just as [official docs](https://www.typescriptlang.org/docs/handbook/enums.html) is doing. Maybe we could bring some simple eslint rule for that

grafana-plugin/src/containers/IntegrationContainers/ExpandedIntegrationRouteDisplay/ExpandedIntegrationRouteDisplay.tsx
```diff
@@ -143,41 +189,105 @@ export const ExpandedIntegrationRouteDisplay: React.FC<ExpandedIntegrationRouteD
...
+
+                            if (option === ROUTING_OPTION.LABELS) {
+                              await alertReceiveChannelStore.saveChannelFilter(channelFilterId, {
+                                filtering_labels: channelFilter.filtering_labels || [],
+                                filtering_term_type: FilteringTermType.labels,
+                              });
+                            }
+
+                            setRoutingOption(option);
+                          }}
```
brojd, line 217: Can we extract this part to the const above return? like `onRoutingOptionChange` or similar

grafana-plugin/src/containers/IntegrationContainers/ExpandedIntegrationRouteDisplay/ExpandedIntegrationRouteDisplay.tsx
```diff
@@ -143,41 +189,105 @@ export const ExpandedIntegrationRouteDisplay: React.FC<ExpandedIntegrationRouteD
...
+
+                              if (!areLabelsValid(labels)) {
+                                return;
+                              }
+
+                              await alertReceiveChannelStore.saveChannelFilter(channelFilterId, {
+                                filtering_labels: labels,
+                                filtering_term_type: FilteringTermType.labels,
+                              });
+                            }}
```
brojd, line 237: Same here, can we extract this to the function / const?

grafana-plugin/src/containers/IntegrationContainers/ExpandedIntegrationRouteDisplay/ExpandedIntegrationRouteDisplay.tsx
```diff
@@ -312,15 +423,38 @@ export const ExpandedIntegrationRouteDisplay: React.FC<ExpandedIntegrationRouteD
...
+                      templates[LabelTemplateOptions.AlertGroupDynamicLabel.key];
+
+                    setLabels([...newLabels]);
+
+                    setCustomLabelIndexToShowTemplateEditor(undefined);
+                  }}
+                />
+              )}
+              <IntegrationCollapsibleTreeView
+                configElements={getTreeViewElements() as any}
```
brojd, line 453: Can we remove `as any`? Seems it won't even throw any TS error now

grafana-plugin/src/containers/IntegrationContainers/ExpandedIntegrationRouteDisplay/ExpandedIntegrationRouteDisplay.tsx
```diff
@@ -337,6 +471,20 @@ export const ExpandedIntegrationRouteDisplay: React.FC<ExpandedIntegrationRouteD
       </>
     );
 
+    function shouldShowLabelAlert() {
+      if (!labels || !labels.length) {
```
brojd, line 475: How about `if (!labels?.length) {`?

grafana-plugin/src/models/label/label.ts
```diff
@@ -26,9 +27,12 @@ export class LabelStore extends BaseStore {
...
 
   @action.bound
-  async loadValuesForKey(key: ApiSchemas['LabelKey']['id'], search = '') {
+  async loadValuesForKey(
+    key: ApiSchemas['LabelKey']['id'],
+    search = ''
+  ): Promise<components['schemas']['LabelOption']> {
     if (!key) {
-      return [];
+      return {} as components['schemas']['LabelOption'];
```
brojd, line 35: Why we need to cast it here? I can see that after removing casting TS throws the errors but I think for a valid reason. Can we adjust the return type of this function and adapt the places where it's used?

grafana-plugin/src/components/LabelsTooltipBadge/LabelsTooltipBadge.tsx
brojd: Ah, missed it! all good

grafana-plugin/src/containers/IntegrationContainers/ExpandedIntegrationRouteDisplay/ExpandedIntegrationRouteDisplay.tsx
teodosii: We didn't have any pattern for this really, or at least I don't remember we agreed on naming. Changed all enums to pascal case

grafana-plugin/src/models/label/label.ts
teodosii: I just added the types, didn't really touch this part. It seems it accepts both the given interface and empty values, I assume this is to prevent NPE errors. The cast is really harmless

grafana-plugin/src/components/LabelsTooltipBadge/LabelsTooltipBadge.tsx
```diff
@@ -36,3 +38,32 @@ export const LabelsTooltipBadge: FC<LabelsTooltipBadgeProps> = ({ labels, onClic
...
     />
   ) : null;
+
+interface LabelBadgesProps {
+  labels: Array<components['schemas']['LabelPair']>;
+  maxCount?: number;
+}
+
+export const LabelBadges: React.FC<LabelBadgesProps> = ({ labels = [], maxCount = 3 }) => {
+  const renderer = (values: Array<components['schemas']['LabelPair']>) => {
```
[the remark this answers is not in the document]
teodosii: Good remark

vadimkerr (approved):
backend changes LGTM

grafana-plugin/src/models/label/label.ts
brojd: As discussed I believe we should get rid of this condition if it's not necessary.

Everything else looks good 👍 approving PR

brojd (approved):

Issue #3858, opened 2024-02-08
Incorrect error when trying to create an integration with the existing name


iskhakov:
### What went wrong?

**What happened**:
- When try to create the integration with the existing name I get misleading errors "There was an issue creating the integration. Please try again." and "Something went wrong, please try again later.". Even if the integration name exists in the other team.
<img width="1237" alt="Screenshot 2024-02-08 at 8 15 29 PM" src="https://github.com/grafana/oncall/assets/2262529/9d121297-556b-4a9b-801f-3851f2d938ee">

**What did you expect to happen**:
- Proper error is displayed, e.g. "Integration with the name XXXX already exists in this organization, please choose other name
- Consider to enforce unique names team wide, not org wide

### How do we reproduce it?

1. Open Grafana OnCall and go to Integrations
2. Try to create a new integration with the same name

### Grafana OnCall Version

drone-7324

### Product Area

Alert Flow & Configuration

### Grafana OnCall Platform?

I use Grafana Cloud

PR #3863, opened 2024-02-08
Restrict integration name uniqueness per team

matiasb:
Related to https://github.com/grafana/oncall/issues/3858

We still need to improve error messages in the UI, [to be discussed](https://raintank-corp.slack.com/archives/C04JCU51NF8/p1707486122947369).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Allow mobile app to access escalation options endpoints @imtoori ([#3847](https://github.com/grafana/oncall/pull/3847))
 - Enable templating for alert escalation mobile app push notifications by @joeyorlando ([#3845](https://github.com/grafana/oncall/pull/3845))
 - Change email notification template to not wrap user template @mderynck ([#3862](https://github.com/grafana/oncall/pull/3862))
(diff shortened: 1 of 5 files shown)


mderynck:
A bit of an edge case, as we don't soft delete teams can we can get into a bad state if the following happens:
1. An organization has 2 integrations with the same name but are on different teams
2. Through sync both teams gets deleted
3. now have 2 with same name and no team if they then try to edit in some other way (templates) will they get an error and be unable to?

matiasb:
> A bit of an edge case, as we don't soft delete teams can we can get into a bad state if the following happens:
> 
>     1. An organization has 2 integrations with the same name but are on different teams
> 
>     2. Through sync both teams gets deleted
> 
>     3. now have 2 with same name and no team if they then try to edit in some other way (templates) will they get an error and be unable to?

Hm.. That's a good point. At first there is no limitation at the model level for both integrations to coexist, but we would get an error if trying to update settings (because there will be multiple returned). I can update the PR to handle that case, and we can then force the user to change the team before saving? (as far as I can tell, any other integration interaction like updating templates shouldn't be affected). Or maybe we could even consider no restricting names in any way?

mderynck:
If there is some hint that the user needs to change the name in an error message that should be enough for them to resolve it on their own.

mderynck (approved):

PR #3886, opened 2024-02-13
Handle 400 status codes from Chatops Proxy

Konstantinov-Innokentii:


joeyorlando (commented):
maybe add some tests here?

PR #3892, opened 2024-02-13
Use pip-tools to handle Python deps

matiasb:
[pip-tools](https://github.com/jazzband/pip-tools) helps making builds deterministic, controlling deps (and indirect deps) upgrades (and versions consistency) avoiding unexpected (and potentially breaking) changes.

We keep our direct deps in `requirements.in` from which we generate the `requirements.txt` (where *all* deps are pinned). We also constrain dev (and enterprise) deps based on base requirements. 

Check how to [update deps](https://github.com/jazzband/pip-tools?tab=readme-ov-file#updating-requirements).


mderynck:
Is the change for pip also needed in drone.yml in (Lint  + Unit Test Backend) and snyk.yml?

matiasb:
> Is the change for pip also needed in drone.yml in (Lint + Unit Test Backend) and snyk.yml?

Not needed as long as we don't update/change `requirements.txt` in any way (`requirements.txt` will now include every dep and subdeps pinned; `pip install` keeps working, main diff is `pip-sync` clears the env in the process too, ensuring only the listed deps are installed as a final result). The main change in our flow would be to add/change/update deps using pip tools (`requirements.in` + `pip-compile`) to ensure we keep everything pinned and no unexpected upgrades happen behind the scenes (I was expecting to discuss this a bit in our call today (we discussed introducing this in the past too), but there was no time; maybe next week we can sync, backend-side).

mderynck (approved):

joeyorlando (approved):

joeyorlando:
@matiasb does this change how we install deps or `pip install -r requirements.txt` is the recommended way?

PR #3899, opened 2024-02-14
Quick fix for labels error on labels cache update

Ferril:
# What this PR does
Check response from labels repo before updating labels cache and update cache only if there are no errors in the response.
https://github.com/grafana/oncall-private/issues/2532
## Checklist


engine/apps/labels/client.py
```diff
@@ -100,7 +100,7 @@ def _check_response(self, response: requests.models.Response):
 
         if 400 <= response.status_code < 500:
             error_data = response.json()
-            message = error_data.get("message", None)
```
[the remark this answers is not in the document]
Konstantinov-Innokentii: I propose to do something like:
```
if 400 <= response.status_code < 500:
    try:
        error_data = response.json()
        message = error_data.get("error", response.reason)
    except JSONDecodeError:
        message = response.reason
elif 500 <= response.status_code < 600:
    message = response.reason
```
to raise exception even if error data not JSON or doesn't contain error message

engine/apps/labels/tests/test_labels_cache.py
```diff
@@ -39,6 +40,14 @@ def test_update_labels_cache_for_key(make_organization, make_label_key_and_value
     assert label_value2.name == old_value2_name
 
 
+@pytest.mark.django_db
+def test_update_labels_cache_error(make_organization, make_label_key_and_value, make_label_value):
```
Konstantinov-Innokentii, line 44: This test should be also removed, once we will process all tasks with wrong args. right? It's not an expected case, when we do not catch exception.

Ferril: makes sense :+1:

Konstantinov-Innokentii (approved):

Issue #3868, opened 2024-02-09
Schedules "Add Rotation" form doesn't fit on all screens


teodosii:
### What went wrong?

**What happened**:
- The "Add Rotation" form does not fit if the screen's vertical space is not big enough, leading to issues hiding some portion of the form content

**What did you expect to happen**:
- The form should be vertically scrollable if its content doesn't fully fit into the current window

### How do we reproduce it?

1. Open schedules
2. Open Add Rotation
3. Add masking, add a few users (on a smaller resolution screen)

### Grafana OnCall Version

all

### Product Area

Schedules

### Grafana OnCall Platform?

None

joeyorlando:
Closing as done in #3904

PR #3904, opened 2024-02-15
Fixed schedule rotations form

teodosii:
# What this PR does

- Improves the calculation of dragging area limits
- Fixes some minor margin/padding inconsistencies for which horizontal scrollbar was appended

https://github.com/grafana/oncall/issues/3868

## Checklist


grafana-plugin/src/containers/RotationForm/RotationForm.module.css
```diff
@@ -97,3 +96,21 @@
   margin-bottom: 10px;
   width: 100%;
 }
+
+@media (max-height: 768px) {
```
teodosii, line 100: Kind of a must because the scrollable container should scale differently based on window height (e. g. 45vh too less for anything higher than 1024px)

grafana-plugin/src/containers/RotationForm/RotationForm.tsx
```diff
@@ -421,6 +422,13 @@ export const RotationForm = observer((props: RotationFormProps) => {
     }
   }, [store.timezoneStore.selectedTimezoneOffset]);
 
+  useEffect(() => {
+    window.addEventListener('resize', onResize);
```
teodosii, line 426: hide modal on resize so that it's not accidentally moved into corners where it's not visible

grafana-plugin/src/containers/RotationForm/RotationForm.tsx
```diff
@@ -699,18 +707,38 @@ export const RotationForm = observer((props: RotationFormProps) => {
     </>
   );
 
+  function onResize() {
+    onHide();
```
brojd, line 711: why not call `onHide` directly without creating proxy fn as it does only one thing?

grafana-plugin/src/containers/Rotation/Rotation.module.css
```diff
@@ -4,6 +4,11 @@
   overflow-x: hidden;
 }
 
+.X {
```
brojd, line 7: is this used anywhere?

brojd (approved):

grafana-plugin/src/containers/Rotation/Rotation.module.css
teodosii: No, just debugging, didn't see it.

grafana-plugin/src/containers/RotationForm/RotationForm.tsx
teodosii: Hmm, the passed prop could change reference and you no longer remove the correct associated handler, that was my thinking back then

brojd: Ok, we can leave like that, I missed it's passed as a prop 👍

Issue #3894, opened 2024-02-14
Filtering bug 


devinrcheevers:
### What went wrong?

**What happened**:
-

1. Have alerts in acknowledged state and try to filter the same in the alert group list. 
2. Clicking firing and resolved works fine. 
3. When i click acknowledged, it continues to show the previous list instead of filtering down to only the acknowledged ones. 
4. Only when i try to update the list of items shown, it'll reset the page and show the acknowledged alerts. 

Uploading bugonackfilter.mov…

**What did you expect to happen**:

Do filter to list of acknowledged alerts  
-

### How do we reproduce it?

1. Have alerts in acknowledged state and try to filter the same in the alert group list. 
2. Clicking firing and resolved works fine. 
3. When i click acknowledged, it continues to show the previous list instead of filtering down to only the acknowledged ones. 
4. Only when i try to update the list of items shown, it'll reset the page and show the acknowledged alerts. 

### Grafana OnCall Version

Cloud

### Product Area

Other

### Grafana OnCall Platform?

I use Grafana Cloud

mackgorski:
@devinrcheevers can you upload the video. I think it did not get through.

mackgorski:
Indeed filtering by `status` works for `firing`, `silenced`, `resolved`.
My assumption was that `acknowledged` will always show regardless of other status as those can still have `acknowledged` on record. But it is not the case as I have tested it as shown in the video. 

`acknowledged` status filtering does not work. @iskhakov I think we should prioritize this. Sending this to the dev channel for someone to check.

Video recorded on `sedemo`
https://github.com/grafana/oncall/assets/14937684/b733b2e2-5f8d-481c-9560-b5196b77f230

brojd:
On this page frontend is doing polling calling GET `/alertgroups` every few secs. Every 2nd call lacks `status` query string. So the results on the page are jumping from filtered ones and not filtered ones depending on which response is returned later (there's a race condition because the one with `status` query string is much slower than the one without it). 

Uploading Screen Recording 2024-02-15 at 20.08.12.mov…

PR #3905, opened 2024-02-16
fix issues related to alert groups table

brojd:
# What this PR does
- get rid of duplicated GET requests during polling
- ignore results of older requests
- fix Refresh btn that doesn't disappear
- show correct data in alert groups table
- show placeholder instead of table when filters are updated

https://github.com/grafana/oncall/issues/3894

## Checklist


grafana-plugin/src/pages/incidents/Incidents.tsx
```diff
@@ -66,9 +68,10 @@ interface IncidentsPageState {
   showAddAlertGroupForm: boolean;
   isSelectorColumnMenuOpen: boolean;
   isHorizontalScrolling: boolean;
+  isFirstIncidentsFetchDone: boolean;
 }
 
-const POLLING_NUM_SECONDS = 15;
+const POLLING_NUM_SECONDS = 5;
```
teodosii, line 74: any reason why this is set to 5 instead of 15? might run into issues with 5 if you're performing any selections on the alert groups and once new data comes in might revert any current state.

teodosii (approved):

grafana-plugin/src/components/GTable/GTable.tsx
```diff
@@ -140,19 +140,25 @@ export const GTable = <RT extends DefaultRecordType = DefaultRecordType>(props:
...
-        columns={columns}
-        data={data}
-        showHeader={showHeader}
-        {...restProps}
-      />
-      {pagination && (
-        <div className={cx('pagination')}>
-          <Pagination hideWhenSinglePage currentPage={page} numberOfPages={numberOfPages} onNavigate={onNavigate} />
-        </div>
+      {loading ? (
```
maskin25, line 143: do we really need to display just loading placeholder during loading stage? In general we are trying to avoid such...

grafana-plugin/src/models/loader/loader.ts
```diff
@@ -13,13 +13,15 @@ class LoaderStoreClass {
   }
 
   @action.bound
-  setLoadingAction(actionKey: string, isLoading: boolean) {
-    this.items[actionKey] = isLoading;
+  setLoadingAction(actionKey: string | string[], isLoading: boolean) {
+    if (typeof actionKey === 'string') {
```
maskin25, line 17: `Array.isArray(actionKey)`

grafana-plugin/src/pages/incidents/Incidents.tsx
maskin25: agree, 5 secs looks to small

brojd: This is only "try every 5 secs". If previous call is pending, it's cancelled immediately and it does nothing. In general I prevented race conditions that led to incorrect alert groups state. For small stacks, when there are not many alert groups, why user should wait 15 secs to see the change applied by another user?

grafana-plugin/src/components/GTable/GTable.tsx
brojd: I'm going to agree on the solution with Raphael. Still I think we need some indicator because for example on ops filtering by `acknowledgment` might take even 10-15secs. Without any indication, user might think that it's just not working and until receives response might refresh the page

grafana-plugin/src/models/loader/loader.ts
brojd: Yes, it will work the same but could you explain the benefit, pls?

grafana-plugin/src/pages/incidents/Incidents.tsx
brojd: brought back 15secs

grafana-plugin/src/components/GTable/GTable.tsx
maskin25: we should display some indicator, but shouldn't hide the whole table

grafana-plugin/src/models/loader/loader.ts
maskin25: my proposal sounds like "if actionKey is an array then iterate, otherwise use it as is", yours sounds like "if it's a string then use  it as is otherwise iterate", I think my approach is more safe

brojd: We're basically deciding here between:
1. My version:
```
if (conditionForSimplerScenario) {
  // cover simple scenario
} else {
  // cover more complex scenario
}
```
2. Your proposal
```
if (conditionForMoreComplexScenario) {
  // cover more complex scenario
} else {
  // cover simple scenario
}
```

I honestly prefer the first one in general, it's similar when you return value from React component. Sometimes if for example you have no data you return `null` (simple scenario). It's more readable to have it before you write some complex JSX (more complex scenario). But anyway it's so minor that let's not spend time discussing it now, I'll just use your suggestion :)

PR #3910, opened 2024-02-17
Repeat Escalation from specific stage

### LINKED ISSUE #3760: Repeat escalations from {number of step of escalation}

We want to add `Repeat escalation from {number of step of escalation}` in Escalation Step type. 
Our case is as follows:
1. Notify users on the schedule every 1 minutes 5 times.
2. (if anybody ack the incident)Notify manager every 1 minutes 5 times.
3. (if manager didn't ack the incident)Notify higher manager or CTO ...

for the number 1, it is easy to setup in escalation chain because we just need to jump up to the beginning. but for the number 2, it is little bit tricky because where we need to jump is not the beggining of the steps. Only choice is to add 10 steps (notify, sleep 1m, notify, sleep 1m, and so on) which is not intuitive enough.

### Product Area

Alert Flow & Configuration

### Anything else to add?

_No response_

ravishankar15:
# What this PR does
Adds the option for the users to run escalation chain from a specific stage five times max

For reviewers,
1. The `run_from_stage` param is one indexed in the UI and API but zero indexed in the DB
2. We are not touching the escalation counter in the serialised object so if the run_from_stage points to another repeat stage it won't repeat rather just moves on from the pointed step.

Fixes #3760

## Checklist


ravishankar15:
> hi there @ravishankar15 👋
> 
> thanks for the contribution, interesting idea! I see that this PR is mostly concerned with being able to create escalation chains, via the internal + public API, that now include this new escalation policy step.
> 
> To actually modify the escalation behaviour you'll need to poke around in `apps.alerts.escalation_snapshot.snapshot_classes.escalation_policy_snapshot`.
> 
> If you're able to get a working POC of this, we'd be willing to review a PR and help you get things across the finish line 🙌

Thanks yup ! Working on the same  🙌

ravishankar15:
@joeyorlando I have implemented the required changes can you take a look ?

donghoon-lee-mrt:
@joeyorlando I also really look forward this feature. Please take a look this PR when you have time. Thanks!

joeyorlando:
@ravishankar15 thanks for all of your recent ideas/contributions 🙏 the core team discussed your contribution and we don't have any immediate plans to add this to the product, but we might possibly implement something similar in the near future. I'm going to close this PR for now, but thanks again for your effort!

Issue #1044, opened 2022-12-26
Email notifications not work


dezzzm:
Oncall v1.1.8. Installed via helm chart.
Email notifications not work.

env example:
```
- name: FEATURE_EMAIL_INTEGRATION_ENABLED
      value: "True"
    - name: EMAIL_HOST
      value: smtp.test.com
    - name: EMAIL_PORT
      value: "465"
    - name: EMAIL_HOST_USER
      value: test@test.ru
    - name: EMAIL_HOST_PASSWORD
      valueFrom:
        secretKeyRef:
          key: smtp-password
          name: oncall-smtp
          optional: true
    - name: EMAIL_USE_TLS
      value: "True"
    - name: EMAIL_FROM_ADDRESS
      value: test@test.ru
```

Logs:
```
2022-12-26 16:37:57 | oncallgrafana/oncall:v1.1.8oncall-celery-6947cd64f7-mrg7s10.111.1.205celeryDeployment/oncall-celerystderr | 2022-12-26 09:37:57,144 source=engine:celery worker=ForkPoolWorker-3 task_id=87f5721a-15d2-471f-b238-a08cb270c89a task_name=apps.email.tasks.notify_user_async name=celery.app.trace level=INFO Task apps.email.tasks.notify_user_async[87f5721a-15d2-471f-b238-a08cb270c89a] retry: Retry in 1s: SMTPServerDisconnected('Connection unexpectedly closed: timed out')
```

MaxDiOrio:
This is probably your SMTP server closing the connection due to configuration issues.  Does your SMTP server have IP address restrictions in place?  Double check your port, 465 is a legacy port that isn't used much anymore.  587 is usually used for ssl/tls.

dezzzm:
We tried 4 different mail servers. Tried port 587. The problem persists.

Matvey-Kuk:
@dezzzm could you please try to enter the pod with engine and try sending SMTP e-mail using one of those answers: https://stackoverflow.com/questions/9998710/is-it-possible-to-send-mails-by-bash-script-via-smtp ? I want to make sure it's not a connectivity problem.

WoodyWoodsta:
I simply cannot get email to work my side as well. I've tried all combinations of ports, enabling and disabling SSL (`EMAIL_USE_SSL="False"`) and TLS. Nothing appears to work!

Any suggestions in 2024?

PR #3911, opened 2024-02-18
Allow setting email app to use SSL instead of TLS

WoodyWoodsta:
# What this PR does
Adds flexibility of the method of encryption in the SMTP email app. Some email servers are configured to use port 465 (intrinsic TLS) which requires `EMAIL_USE_SSL` instead of `EMAIL_USE_TLS`.

Fixes https://github.com/grafana/oncall/issues/1044

## Checklist


engine/apps/base/models/live_setting.py
```diff
@@ -79,6 +80,7 @@ class LiveSetting(models.Model):
         "EMAIL_HOST_USER": "SMTP server user",
         "EMAIL_HOST_PASSWORD": "SMTP server password",
         "EMAIL_USE_TLS": "SMTP enable/disable TLS",
+        "EMAIL_USE_SSL": "SMTP enable/disable SSL. Should be used mutually excelusively with EMAIL_USE_TLS.",
```
joeyorlando, line 83:
```suggestion
        "EMAIL_USE_SSL": "SMTP enable/disable SSL. Should be used mutually exclusively with EMAIL_USE_TLS.",
```

engine/apps/base/models/live_setting.py
joeyorlando, line 1: can you add a check to [`apps.base.utils.LiveSettingValidator`](https://github.com/grafana/oncall/blob/dev/engine/apps/base/utils.py#L36) to validate that `EMAIL_USE_TLS` and `EMAIL_USE_SSL` are actually mutually exclusive?

helm/oncall/templates/_env.tpl
joeyorlando, line 1: similar to my other comment, I would suggest some validation on the mutual-exclusivity of `Values.oncall.smtp.tls` and `Values.oncall.smtp.ssl` (a quick search brought up [this](https://stackoverflow.com/a/53516703)).

Also, please add a helm unit test for this.

joeyorlando (commented):
@WoodyWoodsta thank you for your contribution! 🙏 

A few small comments. Once those are added I'd be happy to get this approved and merged in :)

engine/apps/base/models/live_setting.py
WoodyWoodsta: Ah, I didn't realise that Django enforced the exclusivity. We should definitely do the same. I'll add the checks.

helm/oncall/templates/_env.tpl
WoodyWoodsta: https://github.com/grafana/oncall/pull/3911/commits/0f5a1b81485e1a68c29f899a3a1215e06398a39a

engine/apps/base/models/live_setting.py
WoodyWoodsta: https://github.com/grafana/oncall/pull/3911/commits/0f5a1b81485e1a68c29f899a3a1215e06398a39a

engine/apps/base/models/live_setting.py
WoodyWoodsta: https://github.com/grafana/oncall/pull/3911/commits/0f5a1b81485e1a68c29f899a3a1215e06398a39a

engine/apps/base/models/live_setting.py
joeyorlando: hi 👋 the live setting validation was close but wasn't working 100%. I fixed it in https://github.com/grafana/oncall/pull/3911/commits/406d914f874ce638d7c710cd430493a4d3ac31e0

joeyorlando (approved):

engine/apps/base/models/live_setting.py
WoodyWoodsta: Thanks @joeyorlando - python isn't something I write on a daily basis :D

Issue #3408, opened 2023-11-22
Webhooks table improvements


raphael-batte:
Update table UX by using such design direction.

![Image](https://github.com/grafana/oncall/assets/102970268/18e99a70-3628-494c-a5b6-0b5f47570d68)

Possible directions:
- [x]  Add link to webhook name to open settings
- [x] Add status badge for disabled webhooks
- [x] Remove blue link color from URL
- [x] Rename last run > Last event
- [x] Update last event structure as Timestamp badge+Status badge+Button to open event details

PR #3918, opened 2024-02-19
Snow outgoing webhooks tab

brojd:
# What this PR does
- implement visual part of ServiceNow Outgoing Tab
- reuse created components on outgoing webhooks page
- make yarn:lint:fix to remove unused imports
- fix live reload during development
- remove unused babel dependencies

- https://github.com/grafana/oncall/issues/3408
- partially https://github.com/grafana/oncall-private/issues/2462

## Checklist


grafana-plugin/.eslintrc.js
```diff
@@ -37,7 +37,9 @@ module.exports = {
       },
     ],
     'no-console': ['warn', { allow: ['warn', 'error'] }],
-    'no-unused-vars': [
+    'no-unused-vars': 'off',
+    'unused-imports/no-unused-imports': ['warn'],
+    'unused-imports/no-unused-vars': [
```
teodosii, line 42: is this one a superset of  `no-unused-vars`?

grafana-plugin/e2e-tests/integrations/maintenanceMode.test.ts
```diff
@@ -40,7 +40,7 @@ test.describe('maintenance mode works', () => {
   const enableMaintenanceMode = async (page: Page, mode: MaintenanceModeType): Promise<void> => {
     await _openIntegrationSettingsPopup(page, true);
     // open the maintenance mode settings drawer + fill in the maintenance details
-    await page.getByTestId('integration-start-maintenance').click();
+    await page.getByText('Start Maintenance').click();
```
teodosii, line 43: hmm, why query by text instead of unique id?

grafana-plugin/src/components/IntegrationLogo/IntegrationLogoWithTitle.tsx
```diff
@@ -0,0 +1,20 @@
+import React, { FC } from 'react';
+
+import { HorizontalGroup } from '@grafana/ui';
+
+import { Text } from 'components/Text/Text';
+
+import { IntegrationLogo, IntegrationLogoProps } from './IntegrationLogo';
```
teodosii, line 7: Not related to this PR, but what if we start sticking to just absolute paths instead? Might be a good start with new PRs

grafana-plugin/src/components/Integrations/IntegrationTag.tsx
```diff
@@ -0,0 +1,37 @@
...
+    <Tag className={styles.tag}>
+      <Text type="primary" size="small" className={styles.radius}>
+        {children}
+      </Text>
+    </Tag>
+  );
+};
+
+export const getStyles = (theme: GrafanaTheme2) => ({
+  tag: css({
```
teodosii, line 27: Could you please move these to a separate file so that we'll have the same pattern across the project? I know it's just 2 classes but it enforces consistency this way for future migration

grafana-plugin/src/utils/hooks.tsx
```diff
@@ -74,6 +78,56 @@ export function useDebouncedCallback<A extends any[]>(callback: (...args: A) =>
   };
 }
 
+export const useDrawer = <DrawerKey extends string, DrawerData = unknown>(initialDrawerData?: DrawerData) => {
+  const [openedDrawer, setOpenedDrawer] = useState<DrawerKey>(LocationHelper.getQueryParam('openedDrawerKey'));
```
teodosii, line 82: a bit biased but maybe `openedDrawerKey` instead?

grafana-plugin/src/components/IntegrationLogo/IntegrationLogoWithTitle.tsx
brojd: It's the eslint that should take care of that and currently our setup is:
```
'rulesdir/no-relative-import-paths': ['error', { allowSameFolder: true }],
```

So basically it allows relative path as soon as you import stuff from the same folder. So it allows: `./Dir/AnotherDir` or `./file.ts` but doesn't allow `../AnotherDir/file.ts`. And tbh that's the config I find useful and I've always used. Because if you're working on deeply nested directory like `src/dir/dir2/dir3/dir4/dir5/myFile` and you want to import `myFile2` from your current directory I think it's easier and more readable to have `./myFile2` rather than `src/dir/dir2/dir3/dir4/dir5/myFile2`

teodosii (approved):

grafana-plugin/e2e-tests/integrations/maintenanceMode.test.ts
brojd: test id has been removed as part of my changes and I decided to not recreate it because selecting by text is also a good practice in e2e tests world

grafana-plugin/src/components/HamburgerContextMenu/HamburgerContextMenu.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { isUserActionAllowed, UserAction } from 'utils/authorization/authorization';
+
+interface HamburgerContextMenuProps {
+  items: Array<
+    { onClick?: () => void; label: ReactNode; requiredPermission?: UserAction; hidden?: boolean } | 'divider'
+  >;
+  hamburgerIconClassName?: string;
+}
+
+export const HamburgerContextMenu: FC<HamburgerContextMenuProps> = ({ items, hamburgerIconClassName }) => {
```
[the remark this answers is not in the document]
brojd: Yeah, good idea, let me rename the other one to `HamburgerMenuIcon`

grafana-plugin/src/components/IntegrationLogo/IntegrationLogoWithTitle.tsx
teodosii: Gotcha, might be a good topic to discuss in one of the frontend meetings. Didn't know it's defined as an eslint rule, I thought that's the default behavior on prettier formatting.

grafana-plugin/src/components/Integrations/IntegrationTag.tsx
brojd: I would like to talk about it and reconsider the "styles always in separate file" rule. The thing is if you have small component, it can be cumbersome to every time be forced to create another file for example just for 1-2 classes, add import etc. https://github.com/grafana/grafana/blob/main/contribute/style-guides/styling.md Here you can see that styles are in the same file. 

I'd like to propose that we extract styles to separate file if:
- the component file itself is big enough (which we should avoid)
- you share styles between different components placed in different files

What do you think?

grafana-plugin/src/utils/hooks.tsx
brojd: Sure, we can, I'll do it

grafana-plugin/.eslintrc.js
brojd: Nope, it's an alternative and to make it work properly you need to disable the original `no-unused-vars`. Docs here: https://www.npmjs.com/package/eslint-plugin-unused-imports#usage

PR #3941, opened 2024-02-21
Add task to delete empty deleted integrations from the database

mderynck:
# What this PR does
Add task which will cleanup deleted integrations from the database if they don't have any alert groups.  This is to help address an issue where queries are slowing down due to having a large numbers of ids that do not contribute to the result.

This will be connected as part of sync organization task once it tests out ok.

## Checklist


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+- Add manually run task to cleanup unused integrations @mderynck ([#3941](https://github.com/grafana/oncall/pull/3941))
```
iskhakov, line 12: Not sure if that's have to be a manual task. I'd prefer it to b automatic, otherwise it is not clear who will own it, and we would need to add some docs about that process

engine/apps/grafana_plugin/tasks/sync.py
```diff
@@ -139,3 +141,26 @@ def sync_team_members_for_organization_async(organization_pk):
...
+def cleanup_empty_deleted_integrations(organization_pk, dry_run=True):
+    try:
+        organization = Organization.objects.get(pk=organization_pk)
+    except Organization.DoesNotExist:
+        logger.info(f"Organization {organization_pk} was not found")
+        return
+
+    integrations_qs = (
+        AlertReceiveChannel.objects_with_deleted.filter(deleted_at__isnull=False, organization=organization)
+        .annotate(num_alert_groups=Count("alert_groups"))
```
vadimkerr, line 156: nit: consider using `Exists` instead of `Count`, it could be faster for orgs that have a lot of alert groups: `from django.db.models import Exists`

or this (same logic but the query is simpler):

```python3
AlertReceiveChannel.objects.filter(organization=organization, deleted_at__isnull=True, alert_groups=None).distinct()
```

engine/apps/grafana_plugin/tasks/sync.py
```diff
@@ -139,3 +141,26 @@ def sync_team_members_for_organization_async(organization_pk):
 
     grafana_api_client = GrafanaAPIClient(api_url=organization.grafana_url, api_token=organization.api_token)
     sync_team_members(grafana_api_client, organization)
+
+
+@shared_dedicated_queue_retry_task(autoretry_for=(Exception,), max_retries=1)
+def cleanup_empty_deleted_integrations(organization_pk, dry_run=True):
```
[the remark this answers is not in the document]
mderynck: See added comment re:sync_organization

CHANGELOG.md
mderynck: I would like to release this task to run manually and then in a follow-up PR connect it to run as part of sync_organization to cover everything. So it will only be manual for a short time.

iskhakov (approved):
I don't see any dangerous `on_delete=models.CASCADE` relations. Looks good to me

PR #3946, opened 2024-02-22
Migrate webhooks integration_filter to use a m2m field instead

matiasb:
Related to https://github.com/grafana/oncall-private/issues/2541


mderynck (approved):
I should probably add a comment to WebhookPresets that `controlled_fields` applies to serializer fields and not model fields for clarity.

engine/apps/webhooks/models/webhook.py
```diff
@@ -152,7 +152,9 @@ class Webhook(models.Model):
     http_method = models.CharField(max_length=32, default="POST", null=True)
     trigger_type = models.IntegerField(choices=TRIGGER_TYPES, default=TRIGGER_ESCALATION_STEP, null=True)
     is_webhook_enabled = models.BooleanField(null=True, default=True)
+    # NOTE: integration_filter is deprecated (to be removed), use filtered_integrations instead
     integration_filter = models.JSONField(default=None, null=True, blank=True)
+    filtered_integrations = models.ManyToManyField("alerts.AlertReceiveChannel", related_name="webhooks")
```
vadimkerr, line 157: naming suggestion: maybe just `integrations`?

vadimkerr (approved):
LGTM

engine/apps/webhooks/models/webhook.py
matiasb: I thought about that, but it seemed to me that having `filtered_integrations = []` is clearer that we don't filter by integration (and then this applies to all) when the webhook is triggered than `integrations = []`? Still not super happy with the name, so I'm open to suggestions :-)

vadimkerr: makes sense 👍

PR #3982, opened 2024-02-29
Add trace_id to log lines

Konstantinov-Innokentii:
# What this PR does
This PR set up tracing to propagates trace_id to log lines.
There are two different tracers: local one in manage.py - it's used when app is started via manage.py runserver. In this case spans will be just written in console. Second traces is confugured in wsgi.py. It will be used when app is runned via uwsgi and it will export traces via grpc. Feature is hidden behind the feature flag.


iskhakov (approved):
Good job!

PR #3983, opened 2024-02-29
Add snow integration settings and connections

Ferril:
# What this PR does
Adds:
- support for additional integration settings 
- ability to connect one integration to another via integral API
- internal API endpoint to show connected integrations and update connection settings
https://github.com/grafana/oncall-private/issues/2540
## Checklist


engine/apps/alerts/constants.py
```diff
@@ -22,3 +23,27 @@ class AlertGroupState(str, Enum):
     ACKNOWLEDGED = "acknowledged"
     RESOLVED = "resolved"
     SILENCED = "silenced"
+
+
+# ServiceNow Integration
```
matiasb, line 28: Wondering if these constants could be defined somewhere else? They seem very specific for ServiceNow

engine/apps/alerts/models/alert_receive_channel.py
```diff
@@ -305,6 +306,8 @@ class AlertReceiveChannel(IntegrationOptionsMixin, MaintainableObject):
     alert_group_labels_template: str | None = models.TextField(null=True, default=None)
     """Stores a Jinja2 template for "advanced label templating" for alert group labels."""
 
+    additional_settings: typing.Optional[ServiceNowSettings] = models.JSONField(null=True, default=None)
```
matiasb, line 309: Similar to previous comment, can't we try to define a more general settings type so eventually we can use this for any other integration? Settings should match the integration type, right? (ie. if this is a servicenow integration, here you could have `ServiceNowSettings` but not if the type was different)

matiasb:
Changes split in new PRs, and merged. Closing this one.

PR #3988, opened 2024-03-01
fe-o11-team/upgrade-faro-and-adjust-config

codecapitano:
# What this PR does
* Upgrade Faro to v1.4.0
* Adjust Faro config to the new version
* Set some sane defaults

* It uses `getWebInstrumentations()` to ensure that if new instruments are added they automatically included without touching the code. This ensures compatibility with new frontend-obervabillity features.
* Remove the old `session` object which is not available in the config anymore.
* Enable persistent sessions
  - With the new session manager Faro persists session across pageloads and, if `persistent` session are enabled accross windows as well.
* Adding `beforeSend` hook to filter out events which do not belong to the grafana-oncall-ap. Background: a Faro instance can not auto scope events to a subtree of the DOM and will always capture events from the whole window. It's advisable to filter out unrelated events.

## Checklist


grafana-plugin/src/utils/faro.ts
```diff
@@ -49,19 +50,25 @@ class BaseFaroHelper {
...
-        session: (window as any).__PRELOADED_STATE__?.faro?.session,
         app: {
           name: appName,
           version: plugin?.version,
         },
+        sessionTracking: {
+          persistent: true,
+        },
+        beforeSend: (event) => {
+          if ((event.meta.page?.url ?? '').includes('grafana-oncall-app')) {
```
codecapitano, line 66: Note: It's better to receive the string from the name in package.json but I didn't want to touch too many things.

teodosii (approved):
Looks good, thanks for the contribution

PR #3997, opened 2024-03-04
Use uwsgi instrumentation and log request headers

Konstantinov-Innokentii:
# What this PR does
Use uwsgi instead of TracingMiddleware to not to mess up with span attrs. Also it's an attempt to fix different trace IDs between cortex-gw, oncall and labels plugin. Last, but not least - add a middleware to log request headers if OTEL is enabled. It's needed to debug tracing headers

### DIFF
--- a/engine/tox.ini
+++ b/engine/tox.ini
@@ -11,6 +11,6 @@ banned-modules =
 # https://pytest-django.readthedocs.io/en/latest/configuring_django.html#order-of-choosing-settings
 # https://pytest-django.readthedocs.io/en/latest/database.html
 # dist=load = "load balance by sending any pending test to any available environment"
-addopts = -n auto --dist=load --no-migrations --color=yes --showlocals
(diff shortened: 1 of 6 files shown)


engine/tox.ini
```diff
@@ -11,6 +11,6 @@ banned-modules =
 # https://pytest-django.readthedocs.io/en/latest/configuring_django.html#order-of-choosing-settings
 # https://pytest-django.readthedocs.io/en/latest/database.html
 # dist=load = "load balance by sending any pending test to any available environment"
-addopts = -n auto --dist=load --no-migrations --color=yes --showlocals
```
iskhakov, line 14: is that intentional change?

iskhakov (approved):

engine/tox.ini
Konstantinov-Innokentii: No, nice catch

PR #4005, opened 2024-03-04
Switch to uv Python package installer/resolver

matiasb:
[uv](https://github.com/astral-sh/uv) is an extremely fast Python package installer and resolver, written in Rust, and designed as a drop-in replacement for pip and pip-tools workflows (see [post](https://astral.sh/blog/uv))


joeyorlando (approved):
nice 👍 LGTM. Probably not worthwhile updating our [single `pip` usage in `.drone.yml`](https://github.com/grafana/oncall/blob/dev/.drone.yml#L72) because we will be migrating from Drone -> GitHub Actions next quarter (same [here](https://github.com/grafana/oncall-private/blob/dev/.drone.yml#L197-L198) for oncall-private)

PR #4027, opened 2024-03-06
Add support for integration additional settings via custom serializer

matiasb:
Related to https://github.com/grafana/oncall-private/issues/2540


engine/apps/alerts/models/alert_receive_channel.py
```diff
@@ -305,6 +305,8 @@ class AlertReceiveChannel(IntegrationOptionsMixin, MaintainableObject):
     alert_group_labels_template: str | None = models.TextField(null=True, default=None)
     """Stores a Jinja2 template for "advanced label templating" for alert group labels."""
 
+    additional_settings: typing.Optional[typing.Dict] = models.JSONField(null=True, default=None)
```
vadimkerr, line 308: _nit_, maybe just a preference thing:
```suggestion
    additional_settings: dict | None = models.JSONField(null=True, default=None)
```

engine/apps/api/serializers/alert_receive_channel.py
```diff
@@ -306,6 +316,27 @@ class Meta:
...
+        if self.instance and self.instance.additional_settings:
+            settings_serializer_cls = _additional_settings_serializer_from_type(self.instance.config.slug)
+            if settings_serializer_cls:
+                additional_settings_data = data.get("additional_settings")
+                settings_serializer = settings_serializer_cls(self.instance, data=additional_settings_data)
+                settings_serializer.is_valid()
+                if settings_serializer.errors:
+                    raise ValidationError({"additional_settings": settings_serializer.errors})
+                data["additional_settings"] = settings_serializer.to_internal_value(additional_settings_data)
+        return super().to_internal_value(data)
```
vadimkerr, line 329: nit: maybe invert these ifs to reduce the number of indentations/nested blocks?

engine/apps/api/serializers/alert_receive_channel.py
```diff
@@ -306,6 +316,27 @@ class Meta:
...
+
+    def to_representation(self, instance):
+        result = super().to_representation(instance)
+        if instance.additional_settings and instance.additional_settings.get("password"):
```
vadimkerr, line 333: Why do we rely on `instance.additional_settings.get("password")` here? Maybe add a comment here?

vadimkerr (approved):
LGTM 👍 Do we want to update the OpenAPI schema in this PR or it's better to open a separate PR?

matiasb:
> LGTM 👍 Do we want to update the OpenAPI schema in this PR or it's better to open a separate PR?

Maybe in a separate PR, after we merge the other side of the changes?

engine/apps/api/serializers/alert_receive_channel.py
matiasb: This was a left-over from the original PR! It shouldn't be here, updating.

PR #4030, opened 2024-03-07
Backend support for "connected" integrations

vadimkerr:
# What this PR does

Adds a Django model and internal API for connected integrations. Based on https://github.com/grafana/oncall/pull/3983

Related to https://github.com/grafana/oncall-private/issues/2540

## Checklist


engine/apps/api/serializers/alert_receive_channel_connection.py
```diff
@@ -0,0 +1,36 @@
...
+from apps.api.serializers.alert_receive_channel import FastAlertReceiveChannelSerializer
+
+
+class AlertReceiveChannelSourceChannelSerializer(serializers.ModelSerializer):
+    alert_receive_channel = FastAlertReceiveChannelSerializer(source="source_alert_receive_channel", read_only=True)
+    backsync = serializers.BooleanField()
+
+    class Meta:
+        model = AlertReceiveChannelConnection
+        fields = ["alert_receive_channel", "backsync"]
```
matiasb, line 13: Should `backsync` field be only on the connected side? (it only makes sense for a (source + connected) combination, doesn't it?)

engine/apps/api/tests/test_alert_receive_channel.py
```diff
@@ -1850,3 +1850,185 @@ def test_alert_receive_channel_webhooks_delete(
...
+    source_alert_receive_channel = make_alert_receive_channel(organization)
+    connected_alert_receive_channel_1 = make_alert_receive_channel(organization)
+    connected_alert_receive_channel_2 = make_alert_receive_channel(organization)
+
+    make_alert_receive_channel_connection(source_alert_receive_channel, connected_alert_receive_channel_1)
+    make_alert_receive_channel_connection(source_alert_receive_channel, connected_alert_receive_channel_2)
+
+    client = APIClient()
+    url = reverse(
+        "api-internal:alert_receive_channel-connected-alert-receive-channels-put",
```
matiasb, line 2021: s/-put/-delete?

engine/apps/api/tests/test_alert_receive_channel.py
```diff
@@ -1850,3 +1850,185 @@ def test_alert_receive_channel_webhooks_delete(
...
+    make_user_auth_headers,
+):
+    organization, user, token = make_organization_and_user_with_plugin_token()
+    source_alert_receive_channel = make_alert_receive_channel(organization)
+    alert_receive_channel_to_connect_1 = make_alert_receive_channel(organization)
+    alert_receive_channel_to_connect_2 = make_alert_receive_channel(organization)
+
+    client = APIClient()
+    url = reverse(
+        "api-internal:alert_receive_channel-connected-alert-receive-channels-get",
```
matiasb, line 1926: s/-get/-post? (is the reverse for post also available? I see it's the same endpoint, but that could change in the future?)

vadimkerr: Unfortunately "post" is not reversible. Getting `django.urls.exceptions.NoReverseMatch` when trying to do so.

engine/apps/api/tests/test_alert_receive_channel.py
vadimkerr: same as https://github.com/grafana/oncall/pull/4030#discussion_r1516358997, delete is not reversible

engine/apps/api/serializers/alert_receive_channel_connection.py
vadimkerr: hmm, I think `backsync` makes sense for both the "connected" and "source" fields. the first one is "what integrations are connected to this integration?" and the second is "what integrations this integration is connected to?". so `backsync` applies to both, right?

vadimkerr: here's an example: https://github.com/grafana/oncall/blob/48038438c88b173e3fbabad704d02e485a72d73e/engine/apps/api/tests/test_alert_receive_channel.py#L1856-L1911

the backsync value is the same for both, the only thing that changes is the direction of this relationship. does this make sense?

matiasb (approved):
LGTM, thanks!

PR #4050, opened 2024-03-12
Add test_connection endpoint for alert receive channel backsync setup

matiasb:


Ferril (approved):

engine/apps/api/views/alert_receive_channel.py
```diff
@@ -271,6 +273,25 @@ def send_demo_alert(self, request, pk):
...
+        data = request.data
+        # clear name while testing connection (to avoid name already used validation error)
+        data["verbal_name"] = None
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        instance = AlertReceiveChannel(**serializer.validated_data)
+        test_connection_func = getattr(instance.config, "test_connection", None)
+        if test_connection_func:
+            try:
+                test_connection_func(instance)
```
vadimkerr, line 289: maybe make `test_connection` return a boolean so exceptions are handled in `test_connection` itself? IMHO it would be better if internal API doesn't have to know about those implementation-specific details

PR #4076, opened 2024-03-18
SNOW external ID

vadimkerr:
# What this PR does

- Adds a new model `AlertGroupExternalID` to keep track of incident IDs in external systems
- Adds calls to integration config specific functions on alert group creation and webhook response

Related to https://github.com/grafana/oncall-private/issues/2541

## Checklist


grafana-plugin/src/pages/integration/OutgoingTab/OutgoingWebhookFormFields.tsx
```diff
@@ -101,7 +100,6 @@ export const OutgoingWebhookFormFields: FC<OutgoingWebhookFormFieldsProps> = ({
         <Input
           {...register('url', {
             required: 'URL is required',
-            pattern: { value: VALID_URL_PATTERN, message: 'URL is invalid' },
```
vadimkerr, line 104: this can be a Jinja2 template

engine/apps/alerts/models/alert_receive_channel_connection.py
```diff
@@ -18,3 +18,19 @@ class AlertReceiveChannelConnection(models.Model):
...
+    """
+
+    source_alert_receive_channel = models.ForeignKey(
+        "AlertReceiveChannel", on_delete=models.CASCADE, related_name="external_ids"
+    )
+    alert_group = models.ForeignKey("AlertGroup", on_delete=models.CASCADE, related_name="external_ids")
+    value = models.CharField(max_length=512)
+
+    class Meta:
+        unique_together = ("source_alert_receive_channel", "alert_group")
```
matiasb, line 36: Should we have an index for the value? (I guess we will be querying for the alert group using the external ID and the source channel integration type?)

engine/apps/webhooks/tasks/trigger_webhook.py
```diff
@@ -122,6 +122,13 @@ def _build_payload(
 
     data = serialize_event(event, alert_group, user, webhook, responses_data)
 
+    # TODO: find a better way to filter, additional_settings__isnull is a workaround
+    source_alert_receive_channel = webhook.filtered_integrations.filter(additional_settings__isnull=False).first()
+    if source_alert_receive_channel and hasattr(source_alert_receive_channel.config, "additional_template_context"):
+        data.update(
+            source_alert_receive_channel.config.additional_template_context(alert_group, source_alert_receive_channel)
+        )
```
matiasb, line 130: Would it make sense to update the `serialize_event` helper to add the external_ids (and the corresponding source integration settings) in the context of the alert group, using the external ID model? (ie. does it make sense to do [this](https://github.com/grafana/oncall-private/pull/2581/files#diff-3a474aa4e2ce1b1a0b2c6aaa9483665b3c3eb5b0e5ea10349a743d74bbfcb9d4R200) for all external IDs related to the alert group?)

Also, it will be nice to add some test for this.

engine/apps/webhooks/models/webhook.py
```diff
@@ -337,3 +339,16 @@ class WebhookResponse(models.Model):
...
+def webhook_response_post_save(sender, instance, created, *args, **kwargs):
+    if not created:
+        return
+
+    # TODO: find a better way to filter, additional_settings__isnull is a workaround
+    source_alert_receive_channel = instance.webhook.filtered_integrations.filter(
+        additional_settings__isnull=False
+    ).first()
+    if source_alert_receive_channel and hasattr(source_alert_receive_channel.config, "on_webhook_response_created"):
+        source_alert_receive_channel.config.on_webhook_response_created(instance, source_alert_receive_channel)
```
matiasb, line 354: Would it possible to add some test for this?

engine/apps/webhooks/models/webhook.py
```diff
@@ -337,3 +339,16 @@ class WebhookResponse(models.Model):
...
         if self.content:
             return json.loads(self.content)
+
+
+@receiver(post_save, sender=WebhookResponse)
+def webhook_response_post_save(sender, instance, created, *args, **kwargs):
+    if not created:
+        return
+
+    # TODO: find a better way to filter, additional_settings__isnull is a workaround
```
matiasb, line 349: At this point we should have the alert group for which the webhook was triggered, right? and then, the integration too. Can we filter the source channel using the connection model? Could there be more than one source channel connected?

engine/apps/alerts/models/alert_receive_channel_connection.py
vadimkerr: done in https://github.com/grafana/oncall/pull/4076/commits/5433ab60744048aef71c9625f7c10c37479fd362

engine/apps/webhooks/models/webhook.py
vadimkerr: done in https://github.com/grafana/oncall/pull/4076/commits/34baeba4adbe7f199dce60a7a48e39e92d93fad8

engine/apps/webhooks/models/webhook.py
vadimkerr: changed it to
```python
source_alert_receive_channel = webhook.filtered_integrations.filter(
    connected_alert_receive_channels__isnull=False
).first()
```

which is definitely better than checking by `additional_settings`. I think it'll be possible to have multiple source channels and we'll need to address it, I left a TODO for a future PR.

engine/apps/alerts/models/alert_receive_channel_connection.py
```diff
@@ -30,7 +30,7 @@ class AlertGroupExternalID(models.Model):
         "AlertReceiveChannel", on_delete=models.CASCADE, related_name="external_ids"
     )
     alert_group = models.ForeignKey("AlertGroup", on_delete=models.CASCADE, related_name="external_ids")
-    value = models.CharField(max_length=512)
+    value = models.CharField(max_length=512, db_index=True)
```
matiasb, line 33: FYI, [docs](https://docs.djangoproject.com/en/4.2/ref/models/fields/#db-index) suggest using `Meta.indexes` instead

engine/apps/webhooks/utils.py
```diff
@@ -184,4 +185,19 @@ def serialize_event(event, alert_group, user, webhook, responses=None):
         data["webhook"] = {"id": webhook.public_primary_key, "name": webhook.name, "labels": get_labels_dict(webhook)}
         data["integration"]["labels"] = get_labels_dict(alert_group.channel)
         data["alert_group"]["labels"] = get_alert_group_labels_dict(alert_group)
+
+    # Add additional webhook data if the integration has it
+    source_alert_receive_channel = webhook.filtered_integrations.filter(
+        connected_alert_receive_channels__isnull=False
+    ).first()  # TODO: is it possible to have more than one?
+    if source_alert_receive_channel and hasattr(source_alert_receive_channel.config, "additional_webhook_data"):
+        data.update(source_alert_receive_channel.config.additional_webhook_data(source_alert_receive_channel))
```
matiasb, line 194: Should we add this and the external ID under some key? (if we eventually have more than one external ID?)

matiasb (approved):
LGTM

engine/apps/webhooks/utils.py
vadimkerr: the idea is that `external ID` and `settings` are per "source" integration, so if there are multiple external IDs (e.g. 2 SNOW integrations) they would be referenced in two different sets of webhooks. not sure we want to expose all the different external IDs at this point, wdyt?

engine/apps/alerts/models/alert_receive_channel_connection.py
vadimkerr: https://github.com/grafana/oncall/pull/4076/commits/7e124342cf64bbd4e106a753cd7b0a9bbb906605

engine/apps/webhooks/utils.py
matiasb: hm.. right, I see, makes sense :+1:

brojd (approved):

PR #4078, opened 2024-03-18
Show servicenow indicator for AGs + complete your servicenow config

teodosii:
# What this PR does

- Adds indicator on the AG for servicenow integrations
- Adds "Complete your servicenow configuration" modal on the integration page

## Checklist


grafana-plugin/src/models/alert_receive_channel/alert_receive_channel.ts
```diff
@@ -105,23 +105,60 @@ export class AlertReceiveChannelStore {
...
+    skipErrorHandling?: boolean;
+  }): Promise<void> {
+    const statusList = await onCallApi({ skipErrorHandling }).GET('/alert_receive_channels/{id}/status_options/', {
+      params: { path: { id } },
+    });
 
-    return Promise.resolve();
+    runInAction(() => {
+      // @ts-ignore // looks like wrong schema
+      this.serviceNowStatusList = statusList.data;
```
teodosii, line 121: Ignore this for now, looks like the schema got changed but the endpoint returns as expected on this line

grafana-plugin/src/components/IntegrationInputField/IntegrationInputField.tsx
```diff
@@ -47,7 +51,19 @@ export const IntegrationInputField: React.FC<IntegrationInputFieldProps> = ({
   );
 
   function renderInputField() {
-    return <Input className={inputClassName} value={isInputMasked ? value?.replace(/./g, '*') : value} disabled />;
+    let baseClassname = inputClassName;
+    if (isButtonHeight) {
+      baseClassname = 'button-input-height';
+    }
```
brojd, line 57: It seems we make use of it in one case only. This component already supports `inputClassName` prop so how about we just use existing `inputClassName` to apply the height without introducing another prop?

grafana-plugin/src/containers/ServiceNowConfigDrawer/CompleteServiceNowConfigModal.tsx
```diff
@@ -0,0 +1,134 @@
...
+                Proceed
+              </Button>
+            </HorizontalGroup>
+          </div>
+        </form>
+      </FormProvider>
+    </Modal>
+  );
+
+  async function onDismiss() {
```
brojd, line 71: Can we remove this function and use `onDismiss={onFormAcknowledge` instead? Both x btn and Close should do exactly same thing so they can both use the same handler I suppose?

grafana-plugin/src/containers/ServiceNowConfigDrawer/CompleteServiceNowConfigModal.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+    try {
+      await alertReceiveChannelStore.update({ id, data });
+      openNotification('You successfully completed your ServiceNow configuration');
+      onHide();
+    } finally {
+      setIsFormActionsDisabled(false);
+    }
+  }
+};
```
brojd, line 128: Instead of doing `try...catch...` and managing `isFormActionsDisabled` local state in this component, we can make use of existing `ActionKey.UPDATE_INTEGRATION` loading state and based on that disable the buttons.
```
const isLoading = useIsLoading(ActionKey.UPDATE_INTEGRATION);
...

              <Button variant="secondary" onClick={onFormAcknowledge} disabled={isLoading}>
                Close
              </Button>
              <Button variant="primary" type="submit" disabled={isLoading}>
                Proceed
              </Button>
            </HorizontalGroup>
```

Then remove`useState` usage

grafana-plugin/src/containers/ServiceNowConfigDrawer/ServiceNowAuthSection.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { getCommonServiceNowConfigStyles } from './ServiceNow.styles';
+import { ServiceNowFormFields } from './ServiceNowStatusSection';
+
+export const ServiceNowAuthSection: React.FC = observer(() => {
+  const { getValues } = useFormContext<ServiceNowFormFields | IntegrationFormFields>();
+  const { alertReceiveChannelStore } = useStore();
+
+  const currentIntegration = useCurrentIntegration();
+  const [isAuthTestRunning, setIsAuthTestRunning] = useState(false);
+  const [authTestResult, setAuthTestResult] = useState(undefined);
```
brojd, line 25: it seems that the state here is always `boolean`. How about we make it more type safe and we use `useState(false)`? Then TS will infer the type from the default value. Without it `authTestResult` has `any` type

grafana-plugin/src/containers/ServiceNowConfigDrawer/ServiceNowStatusSection.tsx
```diff
@@ -0,0 +1,205 @@
...
+  const [, forceUpdate] = useReducer((x) => x + 1, 0);
+
+  const { alertReceiveChannelStore } = useStore();
+  const currentIntegration = useCurrentIntegration();
+  const { id } = currentIntegration;
+
+  useEffect(() => {
+    (async () => {
+      await alertReceiveChannelStore.fetchServiceNowStatusList({ id });
+      forceUpdate();
```
brojd, line 37: Hmm it's weird we need this since `serviceNowStatusList` is observable and is updated within `fetchServiceNowStatusList`, maybe we can sync today so we quickly take a look together if we can fix it a different way?

grafana-plugin/src/containers/ServiceNowConfigDrawer/ServiceNowStatusSection.tsx
```diff
@@ -0,0 +1,205 @@
...
+      .map((arr) => arr[1]);
+    const statusList = (alertReceiveChannelStore.serviceNowStatusList || []).map(([name, id]) => ({ id, name }));
+
+    return statusList
+      .filter((status) => values.indexOf(status.id) === -1 || stateMapping?.[currentAction]?.[0] === status.name)
+      .map((status) => ({
+        value: status.id,
+        label: status.name,
+      }));
+  }
```
brojd, line 204: It's not easy to quickly understand how this code is mapping and filtering data. Maybe for things like that we could extract it out of component to helper and have simple unit test so that it's clear right away what's the input and output?

grafana-plugin/src/containers/ServiceNowConfigDrawer/ServiceNowTokenSection.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+import { getCommonServiceNowConfigStyles } from './ServiceNow.styles';
+
+
+export const ServiceNowTokenSection: React.FC = observer(() => {
+  const styles = useStyles2(getStyles);
+  const { id } = useCurrentIntegration();
+  const { alertReceiveChannelStore } = useStore();
+  const [isExistingToken, setIsExistingToken] = useState(undefined);
+  const [currentToken, setCurrentToken] = useState(undefined);
```
brojd, line 23: If we do it this way (`useState(undefined)`), then the state gets `any` type which we want to avoid and at some point ditch from the whole codebase.

So we should either give it some other default value like boolean / string or explicitly provide a type like `useState<string | undefined>(undefined)

grafana-plugin/src/containers/ServiceNowConfigDrawer/ServiceNowTokenSection.tsx
```diff
@@ -0,0 +1,92 @@
...
+  const isLoading = useIsLoading(ActionKey.UPDATE_SERVICENOW_TOKEN);
+
+  useEffect(() => {
+    (async function () {
+      const hasToken = await alertReceiveChannelStore.hasServiceNowToken({ id });
```
brojd, line 28: I suggest that we:
1. Rename `hasServiceNowToken` into `checkIfServiceNowHasToken`
2. Create small boolean observable called `serviceNowHasToken` in `alertReceiveChannelStore` that `checkIfServiceNowHasToken` will update upon receiving response from the backend
3. Remove `isExistingToken` local state and use new `serviceNowHasToken` from the store directly

I already have another place that can reuse it, we have a duplication now

grafana-plugin/src/containers/ServiceNowConfigDrawer/ServiceNowTokenSection.tsx
```diff
@@ -0,0 +1,92 @@
...
+          <Button variant="secondary" onClick={onTokenGenerate} disabled={isLoading}>
+            {isExistingToken ? 'Regenerate' : 'Generate'}
+          </Button>
+        </div>
+      </RenderConditionally>
+    </VerticalGroup>
+  );
+
+  async function onTokenGenerate() {
+    const res = await alertReceiveChannelStore.generateServiceNowToken({ id });
```
brojd, line 80: Similar to the one above - right now `generateServiceNowToken` MobX method doesn't have anything to do with the MobX - it doesn't update any global state and is just making http request, then returning data. That's a pattern we try to avoid. So we have 2 options here:
1. Create a simple observable called `serviceNowToken` that `generateServiceNowToken` will update upon receiving response from the backend. Then use it directly in this component and remove `currentToken` local state
2. Keep using `currentToken` local state but move `generateServiceNowToken` from `alert_receive_channel.ts` into `alert_receive_channel.helpers.ts` just as we did for bunch of other similar methods

grafana-plugin/src/models/alert_receive_channel/alert_receive_channel.ts
```diff
@@ -105,23 +105,60 @@ export class AlertReceiveChannelStore {
...
+    });
+  }
+
+  async hasServiceNowToken({ id }: { id: ApiSchemas['AlertReceiveChannel']['id'] }) {
```
brojd, line 125: it should either update some observable instead of returning data from this store or should be moved to alert_receive_channel.helpers.ts (and actually we already have it there named `checkIfTokenExists`)

grafana-plugin/src/models/alert_receive_channel/alert_receive_channel.ts
```diff
@@ -105,23 +105,60 @@ export class AlertReceiveChannelStore {
...
+      const response = await onCallApi({ skipErrorHandling: true }).GET('/alert_receive_channels/{id}/api_token/', {
+        params: { path: { id } },
+      });
+      return response?.response.status === 200;
+    } catch (ex) {
+      return false;
+    }
+  }
+
+  async testServiceNowAuthentication({ data }: { data: OmitReadonlyMembers<ApiSchemas['AlertReceiveChannelUpdate']> }) {
```
brojd, line 136: it should either update some observable instead of returning data from this store or should be moved to alert_receive_channel.helpers.ts

grafana-plugin/src/models/alert_receive_channel/alert_receive_channel.ts
```diff
@@ -105,23 +105,60 @@ export class AlertReceiveChannelStore {
...
+      });
+      return result?.response.status === 200;
+    } catch (ex) {
+      return false;
+    }
+  }
+
+  @AutoLoadingState(ActionKey.UPDATE_SERVICENOW_TOKEN)
+  @WithGlobalNotification({ failure: 'There was an error generating the token. Please try again' })
+  async generateServiceNowToken({
```
brojd, line 150: it should either update some observable instead of returning data from this store or should be moved to alert_receive_channel.helpers.ts

grafana-plugin/src/containers/ServiceNowConfigDrawer/CompleteServiceNowConfigModal.tsx
teodosii: The integration update seems too generic for me and I don't want to couple it to the loading state of this modal.

grafana-plugin/src/containers/ServiceNowConfigDrawer/ServiceNowAuthSection.tsx
teodosii: Undefined serves its role here because I'm actually checking for undefined and not for false initially `!isAuthTestRunning && authTestResult !== undefined`

grafana-plugin/src/containers/ServiceNowConfigDrawer/ServiceNowStatusSection.tsx
teodosii: I added `SelectableValue` so that it's more obvious it's mapping the data as required by `Select` component. I see what you mean by how would a helper help in this case?
- No other component really need an additional helper so that we justify abstracting logic to a helper and be able to even reuse that logic
- The filtering part is only a few lines, it just filters out the already selected values. Maybe a comment would make it more suggestive for you?

grafana-plugin/src/containers/ServiceNowConfigDrawer/ServiceNowTokenSection.tsx
teodosii: Done, moved to helpers & renamed it. I'm a bit hesitant into adding more data to the alert receive channel store because it's not the kind of data that is largely used by other `alert_receive_channel` consumers... my preference would be to have it in the component because store-wise and being consumed by just 1 component seems kinda meh (maybe there will be 1 or 2 more but the next point still stands). Add to that the store shouldn't know about ServiceNow stuff... Imagine we will add another integrations, it would be nightmare to have so maintain such small data for each integration in the store

grafana-plugin/src/models/alert_receive_channel/alert_receive_channel.ts
teodosii: My bad, totally forgotten about this pattern we recently adopted. Done

grafana-plugin/src/containers/ServiceNowConfigDrawer/ServiceNowStatusSection.tsx
teodosii: The `forceUpdate` is to make sure we call the function that returns the list of possible options for each dropdown. Maybe it's not needed here, I can't totally remember why I added here, but it would be needed in other parts because otherwise that logic isn't being triggered.

grafana-plugin/src/containers/ServiceNowConfigDrawer/ServiceNowStatusSection.tsx
brojd: It would be great if we start flattening our components by extracting things to helpers and store, leveraging pure functions more. In many places our components has grown to enormous sizes and most probably they also started with sth simple. But then, bit after bit, you add one small feature, another etc. The problem here is not a single bit, but the sum of all of them after some time :D And helpers defined outside components or custom hooks, even if used in a single places and not shared outside, are still a great way to make components cleaner.  Another benefit is that you don't create such function on every render.

So - it's a tiny thing in a single PR but there's a larger picture behind that that I'm trying to advocate for :D

grafana-plugin/src/containers/ServiceNowConfigDrawer/ServiceNowTokenSection.tsx
brojd: Got it, helper is ok. As mentioned in previous comment, I'm just trying to point our attention to a bigger picture which is - it would be awesome if we try to make our view layer - components - simpler, flatter etc. There are different ways to achieve that - custom hooks, helpers, better components composition etc

I can already see in your PRs that you pay attention to it and thank you for that! Just giving a bit more suggestions, I hope that's ok

grafana-plugin/src/containers/ServiceNowConfigDrawer/ServiceNowStatusSection.tsx
brojd: it probably has sth to do with react-hook-form Controller

PR #4080, opened 2024-03-18
Fix alert group table default filter being ignored, add 1 week window as default time range

mderynck:
# What this PR does
- Default filters were being ignored on the alerts group page (Only show acknowledged and firing). This was because the object contained an empty array for the team field so it was never considered empty, now all fields are checked.
- Added a 1 week window for `started_at` as a default.  This should give more useful behavior in larger instances so we won't have long queries counting all alert groups since the beginning.

Closes https://github.com/grafana/oncall-private/issues/2520

## Checklist


mderynck:
TODO: Still looking into e2e test issue.

grafana-plugin/src/pages/incidents/Incidents.tsx
```diff
@@ -304,6 +304,8 @@ class _IncidentsPage extends React.Component<IncidentsPageProps, IncidentsPageSt
 
   renderIncidentFilters() {
     const { query, store } = this.props;
+    const defaultStart = moment().subtract(7, 'days');
+    const defaultEnd = moment().add(1, 'days');
```
mderynck, line 308: Using this instead of the 'now-7' options since the +1 day option is not supported.  The +1 day is to make sure we keep loading new alert groups that arrive after the table has loaded

brojd (approved):

PR #4089, opened 2024-03-20
Update alert group state by backsync

Ferril:
# What this PR does
Adds method to update alert group state by backsync
Should be merged with https://github.com/grafana/oncall-private/pull/2606

## Checklist


engine/apps/alerts/models/alert_group.py
```diff
@@ -612,6 +613,115 @@ def _update_metrics(self, organization_id, previous_state, state):
...
+
+        @dataclass
+        class StateData:
+            remove_state: StateUpdateData
+            set_state: StateUpdateData
+
+        state_update_mapping = {
+            AlertGroupState.ACKNOWLEDGED: StateData(
+                set_state=StateUpdateData(
+                    method=self.acknowledge,
```
matiasb, line 638: hm.. I'm probably missing something but to be sure, you don't use the "_by_user" methods because...? passing `user=None, action_source=backsync` seems should work? and it will handle the metrics update, start escalation/ack reminder, log records, signal and dependent alert groups? maybe requiring some tuning? (e.g. I see we sometimes get org from user instead of the channel)

Ferril: I think using method "_by_user" without user looks a bit hacky (e.g. we would need to check if there is a user in the "_by_user" method to set `acknowledged_by=AlertGroup.USER` or `AlertGroup.SOURCE`) 🤔 
In our case methods "_by_source" with some tuning (adding action_source and some additional checks, related to logs) logically look better, but not every state action has this method.

engine/apps/alerts/models/alert_group.py
```diff
@@ -625,28 +651,34 @@ def acknowledge_by_user(self, user: User, action_source: typing.Optional[ActionS
...
-        self.acknowledge(acknowledged_by_user=user, acknowledged_by=AlertGroup.USER)
+        self.acknowledge(acknowledged_by_user=user, acknowledged_by=acknowledged_by)
         # Update alert group state and response time metrics cache
-        self._update_metrics(organization_id=user.organization_id, previous_state=initial_state, state=self.state)
+        self._update_metrics(organization_id=organization_id, previous_state=initial_state, state=self.state)
 
         self.stop_escalation()
-        self.start_ack_reminder_if_needed()
+        if user:  # ack reminder works only for actions performed by user
+            self.start_ack_reminder_if_needed()
```
matiasb, line 674: :+1:

engine/apps/alerts/models/alert_group.py
```diff
@@ -612,11 +612,37 @@ def _update_metrics(self, organization_id, previous_state, state):
...
+            elif self.state == AlertGroupState.RESOLVED:
+                self.un_resolve_by_user_or_backsync(action_source=ActionSource.BACKSYNC)
+            elif self.state == AlertGroupState.SILENCED:
+                self.un_silence_by_user_or_backsync(action_source=ActionSource.BACKSYNC)
+        elif new_state == AlertGroupState.ACKNOWLEDGED:
+            self.acknowledge_by_user_or_backsync(action_source=ActionSource.BACKSYNC)
+        elif new_state == AlertGroupState.RESOLVED:
+            self.resolve_by_user_or_backsync(action_source=ActionSource.BACKSYNC)
+        elif new_state == AlertGroupState.SILENCED:
+            self.silence_by_user_or_backsync(action_source=ActionSource.BACKSYNC)
```
matiasb, line 631: +1, I think it is clearer now what happen when we get an update (and it is nice to reuse the existing methods too)

matiasb (approved):
Thanks for the updates!

vadimkerr (approved):

PR #4104, opened 2024-03-22
Google Calendar Out of Office events - autogenerated shift swap requests

joeyorlando:
# What this PR does

Closes https://github.com/grafana/oncall-private/issues/2590

## Checklist

- [x] Documentation added (or `pr:no public docs` PR label added if not required) - will be done in https://github.com/grafana/oncall-private/issues/2591
    show up in the autogenerated release notes. - will be done in https://github.com/grafana/oncall-private/issues/2591


engine/apps/api/views/shift_swap.py
```diff
@@ -33,13 +33,6 @@ class BaseShiftSwapViewSet(ModelViewSet):
     serializer_class = ShiftSwapRequestSerializer
     pagination_class = FiftyPageSizePaginator
 
-    def _do_create(self, beneficiary: User, serializer: BaseSerializer[ShiftSwapRequest]) -> None:
-        shift_swap_request = serializer.save(beneficiary=beneficiary)
-
-        write_resource_insight_log(instance=shift_swap_request, author=self.request.user, event=EntityEvent.CREATED)
-
-        create_shift_swap_request_message.apply_async((shift_swap_request.pk,))
-
```
joeyorlando, line 42: since both the internal/public API views + now the Google Calendar celery task all have this logic of creating the `ShiftSwapRequest` object + submitting a `create_shift_swap_request_message` task, I decided to move this logic to a `ShiftSwapRequest` `post_save` django signal ([here](https://github.com/grafana/oncall/pull/4104/files#diff-f7ab2fa3cbeb4b36eab0b7904b2c0dd89a780cd73421d32490526538eb1ca68dR264-R272))

engine/apps/google/client.py
```diff
@@ -37,31 +49,28 @@ def fetch_out_of_office_events(self) -> typing.List[GoogleCalendarEvent]:
...
+        )
+
         events_result = (
             self.service.events()
             .list(
                 calendarId="primary",
-                timeMin=now,
-                # timeMax= TODO: should we only fetch out of office events for next X amount of time?
+                timeMin=time_min,
+                timeMax=time_max,
```
joeyorlando, line 68: only consider out of office calendar events for next 30 days (will be sure to document this in forthcoming documentation PR)

engine/apps/google/tasks.py
joeyorlando, line 1: main focus of this PR

engine/apps/schedules/models/on_call_schedule.py
```diff
@@ -542,10 +542,23 @@ def refresh_ical_final_schedule(self):
         self.save(update_fields=["cached_ical_final_schedule"])
 
     def shifts_for_user(
-        self, user: User, datetime_start: datetime.datetime, days: int = 7
+        self,
+        user: User,
+        datetime_start: datetime.datetime,
+        datetime_end: typing.Optional[datetime.datetime] = None,
```
joeyorlando, line 548: added optional `datetime_end` kwarg here which is used in the new celery task

engine/apps/google/client.py
```diff
@@ -2,16 +2,28 @@
...
 logger.setLevel(logging.DEBUG)
 
 
+class GoogleCalendarEvent:
+    def __init__(self, event: GoogleCalendarEventType):
+        self.raw_event = event
+        self._start_time = utils.datetime_strptime(event["start"]["dateTime"])
+        self._end_time = utils.datetime_strptime(event["end"]["dateTime"])
+
+        self.start_time_utc = self._start_time.astimezone(pytz.UTC)
```
matiasb, line 23: FYI, if possible we should try to move to Python [zoneinfo](https://docs.python.org/3/library/zoneinfo.html#module-zoneinfo) while limiting/deprecating the use of [pytz](https://github.com/stub42/pytz/blob/master/src/README.rst#issues--limitations) (which django is also deprecating since [4.0](https://docs.djangoproject.com/en/5.0/releases/4.0/#zoneinfo-default-timezone-implementation))

engine/apps/google/tasks.py
```diff
@@ -15,9 +16,69 @@ def sync_out_of_office_calendar_events_for_user(google_oauth2_user_pk: int) -> N
...
+    user_id = user.pk
+
+    logger.info(f"Syncing out of office Google Calendar events for user {user_id}")
+
+    user_google_calendar_settings = user.google_calendar_settings
+    specific_oncall_schedules_to_sync = user_google_calendar_settings["specific_oncall_schedules_to_sync"]
+
+    users_schedules = OnCallSchedule.objects.related_to_user(user)
+    if specific_oncall_schedules_to_sync:
+        users_schedules = users_schedules.filter(public_primary_key__in=specific_oncall_schedules_to_sync)
```
matiasb, line 29: I see the usage of the `specific_oncall_schedules_to_sync` now. At first read I thought it would sync those calendars with on-call shifts, but now I see the idea is to list the schedules for which you want to automatically request the shifts. Wondering if this could be renamed, or at least add some comment/note in the field definition to make it clearer what means to have an schedule ID listed there?

engine/apps/google/tasks.py
```diff
@@ -15,9 +16,69 @@ def sync_out_of_office_calendar_events_for_user(google_oauth2_user_pk: int) -> N
...
+                    f"Found {len(upcoming_shifts)} upcoming shift(s) for user {user_id} "
+                    f"during the out of office event {event_id}"
+                )
+
+                shift_swap_request_exists = ShiftSwapRequest.objects.filter(
+                    beneficiary=user,
+                    schedule=schedule,
+                    swap_start=start_time_utc,
+                    swap_end=end_time_utc,
+                ).exists()
```
matiasb, line 59: I understand we will not remove swap requests if the OoO events are removed from the Google Calendar, right? (probably complex to detect that if we don't keep a tracking relating swap requests to google events). Probably something we should explicitly mention in docs.

matiasb (approved):

engine/apps/google/client.py
joeyorlando: good call, thanks for pointing that out 👍

engine/apps/google/tasks.py
joeyorlando: that's a good point. After looking at this a second time, I agree, the naming isn't optimal. Maybe something like `oncall_schedules_to_consider_for_shift_swaps`? (a bit long, but more clear of its intention).

I'll also update the UI to be more descriptive on this field 👍

engine/apps/google/tasks.py
joeyorlando: Correct.

I was thinking about persisting some information about calendar events, or at least storing the calendar event ID that was used to generate the shift swap, but decided to keep it simple for now. There's also the caveat where if the OOO is updated, it might do some funky things (like generate an overlapping second shift swap for you 🤔).

My hope is that this case isn't that common, but if it pops up in the future, we can augment the logic here

Issue #4119, opened 2024-03-27
Allow the ability to set custom values for STACK_SLUG in Self Hosted Settings


elburnetto-intapp:
### What would you like to see!

Within the Grafana OnCall Mobile App, if you create an OSS OnCall Stack and link it, you can see the Stack name of 'self_hosted_stack'. If a user had several OSS Stacks (i.e. one for Dev, one for Production), it makes it difficult to differentiate between the two.

It'd be awesome to allow for a custom variable to be set for STACK_SLUG, which is used in the OnCall Engine.

### Product Area

Mobile App, Other

Souldiv:
Hey @joeyorlando @elburnetto-intapp after setting the ENV vars in the dockerfile it still shows up as self_hosted_stack?

PR #4121, opened 2024-03-27
Allow custom values for Self Hosted Clusters (Slugs, Titles, Region)

elburnetto-intapp:
# What this PR does
Allows for environment variables to be set on Grafana OnCall Engine for Self Hosted users, giving them the ability to set values for Stack Slug, Org Slug/Title, Region & Cluster Slugs.

This will mean then using the Grafana OnCall App, when adding multiple OSS Stacks, you can set the correct value of 'stack_slug' so you can differentiate between the stacks in the App. 

Closes [#4119](https://github.com/grafana/oncall/issues/4119)

## Checklist


joeyorlando (approved):
LGTM! Thanks for the contribution ✌️

Issue #2666, opened 2023-07-27
Adopt EmotionJS usage in OnCall to allow better integration with grafana themes


teodosii:
### What went wrong?

Here's one of the slack mentions of issues we're facing right now as we hardcode grafana theme variables in oncall instead of fetching them through EmotionJS.

https://raintank-corp.slack.com/archives/C04JCU51NF8/p1690465362079079

Right now oncall knows how to match theme based on the vars defined, so that results in either light or dark, but it fails for all other. This was already discussed and the solution would be to migrate our styling approach to match grafana's use of Emotion library.

### How do we reproduce it?

Change theme to "System Preference" (and that would need to be light) and then navigate to alert groups or schedules. The theme vars are not being picked up and dark theme shows instead of light.

### Goal

- Migrate existing CSS/SCSS codebase to using Emotion. This will help us with issues regarding themes and OnCall will no longer be hardcoding themes variables inside, but instead rely on what Grafana provides.

In order to allow more visible tracking on this, I suggest we split it more granular so that's easier to work on it (suggested steps below):
- Add the initial emotion setup (package dependency, setup, demo usage)
- Start migrating easier components' style such as OnCall built-in components (Text etc)
- Later on start migrating pages one such as AlertGroups/Integrations which would require more attention and will also take more time

PR #4123, opened 2024-03-27
Migration to emotionjs

teodosii:
# What this PR does

Migrate existing CSS/SCSS to emotion

Closes https://github.com/grafana/oncall/issues/2666


grafana-plugin/src/components/ScheduleQuality/ScheduleQuality.styles.ts
```diff
@@ -0,0 +1,43 @@
...
+      line-height: 16px;
+    `,
+
+    tag: css`
+      font-size: 12px;
+      padding: 5px 10px;
+
+      &--danger {
+        // TODO: emotionjs
+        background-color: var(--tag-background-danger);
```
teodosii, line 23: This needs to be sorted out with @raphael-batte as these tag colors don't really exist in grafana and we need to get rid of the whole custom vars file

grafana-plugin/src/components/Avatar/Avatar.styles.ts
```diff
@@ -0,0 +1,32 @@
+import { css } from '@emotion/css';
+import { GrafanaTheme2 } from '@grafana/data';
+
+export const getAvatarStyles = (_theme: GrafanaTheme2) => {
```
brojd, line 4: I guess we don't need those unused `_theme` params?

grafana-plugin/src/components/Policy/NotificationPolicy.tsx
```diff
@@ -47,21 +43,29 @@ export interface NotificationPolicyProps {
 }
 
 export class NotificationPolicy extends React.Component<NotificationPolicyProps, any> {
+  private styles: ReturnType<typeof getStyles>;
+
+  constructor(props: NotificationPolicyProps) {
+    super(props);
+
+    this.styles = getStyles(props.theme);
```
brojd, line 51: Have you tested if this part works when the theme is changed? Constructor will not be called unless the page reload happens afaik so it might stick to outdated theme

grafana-plugin/src/components/CheatSheet/CheatSheet.styles.ts
```diff
@@ -0,0 +1,40 @@
...
+      min-width: min-content;
+      overflow-y: scroll;
+    `,
+
+    cheatsheetInnerContainer: css`
+      padding-right: 16px;
+      overflow-y: scroll;
+      height: 100%;
+    
+      > div {
```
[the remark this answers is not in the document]
teodosii: right

grafana-plugin/src/components/Policy/NotificationPolicy.tsx
teodosii: That's a good point, let me try that

grafana-plugin/src/components/Avatar/Avatar.styles.ts
brojd: I'm ok with either way, up to you and Maxim

grafana-plugin/src/components/Policy/NotificationPolicy.tsx
teodosii: Updated those 2 usecases, they're now part of componentDidUpdate lifecycle

grafana-plugin/src/components/Policy/EscalationPolicy.tsx
```diff
@@ -55,10 +52,26 @@ export interface EscalationPolicyProps extends ElementSortableProps {
...
+    this.styles = getEscalationPolicyStyles(props.theme);
+  }
+
+  componentDidUpdate(prevProps: Readonly<EscalationPolicyProps>): void {
+    if (prevProps.theme !== this.props.theme) {
+      // fetch new styles whenever the theme changes
+      this.styles = getEscalationPolicyStyles(this.props.theme);
+      this.forceUpdate();
+    }
+  }
```
brojd, line 73: I think just as in functional components we could just define it in render like 
```
render() {
  const styles = getEscalationPolicyStyles(props.theme);
  ...
}
```

brojd (approved):

grafana-plugin/src/components/Avatar/Avatar.styles.ts
maskin25: > I'm ok with either way
I think so too, there is no strong opinion from my side

grafana-plugin/src/components/Integrations/IntegrationBlock.tsx
```diff
@@ -24,22 +24,24 @@ export const IntegrationBlock: React.FC<IntegrationBlockProps> = ({
...
-    <div className={cx('integrationBlock', className)}>
+    <div className={cx(className)}>
       {heading && (
         <Block
           bordered
           shadowed
-          className={cx('integrationBlock__heading', {
-            'integrationBlock__heading--noBorderBottom': !noContent,
+          className={cx(styles.integrationBlockHeading, {
+            [bem(styles.integrationBlockHeading, 'noBorderBottom')]: !noContent,
```
maskin25, line 36:
```
 className={cx(styles.integrationBlockHeading, {
            [bem(styles.integrationBlockHeading, 'noBorderBottom')]: !noContent
})}
```
looks ease o make a typos, how does grafana core deal with nested selectors? I'm pretty sure it does not use bem.

And in general styling looks overcomplicated, to style something we need too much stuff to import

```
import { cx } from '@emotion/css';
import { useStyles2 } from '@grafana/ui';
import { bem } from 'utils/utils';
import { getIntegrationBlockStyles } from './IntegrationBlock.styles';
```

Can we remove at least one thing? the most suitable candidate seems to be a bem.

grafana-plugin/src/components/Policy/EscalationPolicy.tsx
maskin25: agree, ideally styles should be just in render body and nowhere else

teodosii: Too many render functions, that means 1 line for each subrender where we're using the styles... not worth it, it takes 3 lines with cDU lifecycle.

grafana-plugin/src/components/Integrations/IntegrationBlock.tsx
teodosii: Grafana doesn't use BEM pattern. Why remove any of it? I don't understand your reasons...

`cx` is for  composition
`useStyles2` is mandatory ...
`bem` is too build line blocks following BEM pattern so that you don't need to repeat the block, but instead rely on a helper that does that. BEM is very useful because it clearly expresses what the element modifier is supposed to do.
`getIntegrationBlockStyles` is outside the file so that we don't have too much code in the main file

I strongly recommend reading on https://getbem.com/

grafana-plugin/src/components/CheatSheet/CheatSheet.tsx
```diff
@@ -1,39 +1,41 @@
...
-import { HorizontalGroup, IconButton, VerticalGroup } from '@grafana/ui';
-import cn from 'classnames/bind';
+import { HorizontalGroup, IconButton, VerticalGroup, useStyles2 } from '@grafana/ui';
 import CopyToClipboard from 'react-copy-to-clipboard';
 
 import { Block } from 'components/GBlock/Block';
 import { Text } from 'components/Text/Text';
-import { openNotification } from 'utils/utils';
+import { bem, openNotification } from 'utils/utils';
+import { getUtilStyles } from 'utils/utils.styles';
```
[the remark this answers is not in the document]
teodosii: That's a good idea

PR #4128, opened 2024-03-27
Use additional integration settings in templates

matiasb:


engine/common/api_helpers/mixins.py
```diff
@@ -350,8 +350,16 @@ def get_attr_template(self, attr, alert_receive_channel, render_for=None):
             templated_attr = getattr(templated_alert, attr_name)
 
         elif attr_name in BEHAVIOUR_TEMPLATE_NAMES:
+            # are we sure this will always be an integration?
+            channel = self.get_object()
+            extra_context_func = getattr(channel.config, "additional_webhook_data", None)
```
matiasb, line 355: maybe we should rename `additional_webhook_data` to something more like `additional_context`, `integration_extra_context`, or sth else?

PR #4132, opened 2024-03-28
Fix validation of integration name when team  is not present in request data

Konstantinov-Innokentii:
This PR fixes validation of integration name when team is not present in request data. Also it slightly  improves code structure of this validation.


engine/apps/api/serializers/alert_receive_channel.py
```diff
@@ -376,7 +380,24 @@ def validate(self, data):
...
+        otherwise, it returns the team from the existing instance.
+        It's needed to validate name correctly even if team is not present in request data (It's not required).
+
+        If the serializer is used to create a new object, it returns the team from the request data.
+        """
+        if self.instance:
+            if "team" in validated_data:
+                team = validated_data.get("team", None)
+            else:
+                team = self.instance.team
```
matiasb, line 397: Could this if block be simplified to
`team = validated_data.get("team", self.instance.team)`
? (same below)

matiasb (approved):

engine/apps/api/serializers/alert_receive_channel.py
Konstantinov-Innokentii: Hm, it seems so. For some reason I though that if team is in validated data, but it's none (if integration is un-attached from team for example) it will use second argument of get, but it works!

engine/apps/public_api/serializers/integrations.py
```diff
@@ -147,27 +147,50 @@ def update(self, *args, **kwargs):
...
+        """
+        if self.instance:
+            if "team" in validated_data:
+                team = validated_data.get("team", None)
+            else:
+                team = self.instance.team
+        else:
+            team = validated_data.get("team")
+        return team
+
```
joeyorlando, line 193: can we make a `BaseIntegrationSerializer` in `api.serializers` that both the internal and public API serializers inherit from? (to deduplicate this function).

There's also a lot of other duplication in these serializers that we could deduplicate (at a future time 🙂)

Konstantinov-Innokentii: I overcomplicated this logic, it will be just a one-liner :) I'm not sure about having unifying serializer between different APIs since it will violate single responsibility principle.

engine/apps/api/serializers/alert_receive_channel.py
Konstantinov-Innokentii: It seems it could be simplified only to   
`team = validated_data.get("team", self.instance.team) if self.instance else validated_data.get("team")`
becase validate is used both for create when self.instance is None and update when self.instance is object being updated.

PR #4137, opened 2024-03-28
Improve performance for rate-limited, banned and deleted integrations

mderynck:
# What this PR does
- Remove BanAlertConsumptionBasedOnSettingsMiddleware under high traffic scenarios this causes too much DB load and the same effect can be achieved by soft-deleting the integration
- Change lookups for integrations so that non-existent and deleted integrations are also cached to reduce DB load

Closes https://github.com/grafana/oncall-private/issues/2608

## Checklist


engine/apps/integrations/mixins/alert_channel_defining_mixin.py
```diff
@@ -67,18 +78,26 @@ def dispatch(self, *args, **kwargs):
                         alert_receive_channel = obj.object
```
matiasb, line 78: hm.. Not to change anything now (and unrelated to your changes), but I think this json deserialization + loop above this line could be improved.

Issue #4112, opened 2024-03-26
Remove duplicated  /users/<id> HTTP requests


brojd:
We have duplicated requests to /users/<id> that can be reduced, more context:
https://raintank-corp.slack.com/archives/C04JCU51NF8/p1711461446278769

PR #4147, opened 2024-04-03
Google Calendar integration improvements

joeyorlando:
# What this PR does

- UI enhancements
- Fix bug when going through the Google OAuth2 disconnection flow. We should send the `refresh_token` in the `revoke` HTTP request, rather than the `access_token` (`access_token` expires frequently and can result in Google sending back HTTP 400.. `refresh_token` is long lived and can also be sent in the revoke flow)

## Checklist

### DIFF
--- a/grafana-plugin/src/containers/UserSettings/parts/tabs/GoogleCalendar/GoogleCalendar.tsx
+++ b/grafana-plugin/src/containers/UserSettings/parts/tabs/GoogleCalendar/GoogleCalendar.tsx
@@ -1,7 +1,7 @@
 import React, { useEffect, useState } from 'react';
 
 import { css } from '@emotion/css';
-import { Button, HorizontalGroup, InlineSwitch, VerticalGroup, useStyles2 } from '@grafana/ui';
+import { Button, HorizontalGroup, Switch, VerticalGroup, useStyles2 } from '@grafana/ui';
 import { observer } from 'mobx-react';
 
 import { Block } from 'components/GBlock/Block';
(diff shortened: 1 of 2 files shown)


grafana-plugin/src/containers/UserSettings/parts/tabs/GoogleCalendar/GoogleCalendar.tsx
```diff
@@ -21,11 +21,16 @@ const GoogleCalendar: React.FC<{ id: ApiSchemas['User']['pk'] }> = observer(({ i
...
-  const [googleCalendarSettings, setGoogleCalendarSettings] = useState(user?.google_calendar_settings);
-  const [showSchedulesDropdown, setShowSchedulesDropdown] = useState(
-    user.google_calendar_settings?.oncall_schedules_to_consider_for_shift_swaps?.length > 0
-  );
+  const [googleCalendarSettings, setGoogleCalendarSettings] =
+    useState<ApiSchemas['User']['google_calendar_settings']>();
+  const [showSchedulesDropdown, setShowSchedulesDropdown] = useState<boolean>();
+
+  useEffect(() => {
+    userStore.fetchItemById({ userPk: id }).then((user) => {
```
maskin25, line 29: note for @teodosii  and @brojd: we need to fetch user data directly from the component here because user object in the store may not contain `google_calendar_settings` data, because user data retrieved from `GET /users` endpoint does not contain `google_calendar_settings` object (this endpont uses other data serializer than `GET /user/<pk>`), let's discuss this problem together once Dominik is back

teodosii (approved):

mderynck (approved):

grafana-plugin/src/containers/UserSettings/parts/tabs/GoogleCalendar/GoogleCalendar.tsx
brojd: There are probably more information exposed by `GET /user/<pk>` that `GET /users` lacks and that might be needed now or in the future in different parts of UserSettings (in different tabs / sections). GoogleCalendar is nested within one of the tabs of UserSettings. If any other tab needs information from `GET /user/<pk>`, with this approach we would need to replicate HTTP requests (which leads to things like [this one](https://github.com/grafana/oncall/issues/4112) that @joeyorlando brought up for a good reason). It would make sense to me to move this `useEffect` up in the components' structure - to UserSettings and then stop using local state for googleCalendarSettings as I mentioned in the previous PR.

PR #4149, opened 2024-04-03
Check for backsync updates before triggering connected webhooks

matiasb:
Avoid triggering a webhook if it is from a connected integration and the triggering event was caused by a backsync update.

Related to https://github.com/grafana/oncall-private/issues/2615


vadimkerr (approved):
LGTM

PR #4153, opened 2024-04-04
Allow getting details from connected integration webhooks

matiasb:
Related to https://github.com/grafana/oncall-private/issues/2615

### DIFF
--- a/engine/apps/slack/tests/test_scenario_steps/test_alert_group_actions.py
+++ b/engine/apps/slack/tests/test_scenario_steps/test_alert_group_actions.py
@@ -271,6 +271,7 @@ def test_get_alert_group_from_slack_message_in_db_no_alert_group(
         step.get_alert_group(slack_team_identity, payload)
 
 
+@pytest.mark.xdist_group(name="group1")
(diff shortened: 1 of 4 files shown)


engine/apps/slack/tests/test_scenario_steps/test_alert_group_actions.py
```diff
@@ -271,6 +271,7 @@ def test_get_alert_group_from_slack_message_in_db_no_alert_group(
         step.get_alert_group(slack_team_identity, payload)
 
 
+@pytest.mark.xdist_group(name="group1")
```
matiasb, line 274: This is to make sure all these tests reusing the same organization ID, alert group ID, etc. are run in the same worker, sequentially. Otherwise, a parallel run can make them [fail](https://github.com/grafana/oncall/actions/runs/8554723710/job/23441071757?pr=4153#step:5:648). See `loadgroup` [param](https://pytest-xdist.readthedocs.io/en/latest/distribution.html) below.

vadimkerr (approved):

PR #4163, opened 2024-04-04
Improve performance for deleted integration lookups

mderynck:
# What this PR does
- Refactor alert receive channel lookup so it is easier to follow
- Remove the additional lookup that was taking place for alert receive channels that belong to a deleted organization, these can be treated as deleted for usage purposes even though the alert receive channel itself does not have `deleted_at` populated
- Organizations that have been moved will still need to be looked up everytime.  This is not optimized in favor of not maintaining a cache of Organizations.  These are not frequent requests and can be optimized later if necessary.

## Checklist

### DIFF
--- a/engine/apps/integrations/mixins/alert_channel_defining_mixin.py
+++ b/engine/apps/integrations/mixins/alert_channel_defining_mixin.py
@@ -34,7 +34,7 @@ class AlertChannelDefiningMixin(object):
         token = str(kwargs["alert_channel_key"])
         logger.info(f"AlertChannelDefiningMixin started token={token}")
         start = perf_counter()
-        alert_receive_channel, status = self.get_alert_receive_channel_from_cache(token)
+        alert_receive_channel, status = self.get_alert_receive_channel_from_short_term_cache(token)
 
         if not alert_receive_channel or status == CHANNEL_DOES_NOT_EXIST_PLACEHOLDER:
             logger.info(f"Integration {token} does not exist")
@@ -47,7 +47,9 @@ class AlertChannelDefiningMixin(object):
         logger.info(f"AlertChannelDefiningMixin finished in {finish - start}")
         return super(AlertChannelDefiningMixin, self).dispatch(*args, **kwargs)
 
-    def get_alert_receive_channel_from_cache(self, token: str) -> tuple[Optional[AlertReceiveChannel], Optional[str]]:
+    def get_alert_receive_channel_from_short_term_cache(
+        self, token: str
+    ) -> tuple[Optional[AlertReceiveChannel], Optional[str]]:
         # Trying to define from short-term cache
         cache_key_short_term = self.CACHE_KEY_SHORT_TERM + "_" + token
         cached_alert_receive_channel_raw = cache.get(cache_key_short_term)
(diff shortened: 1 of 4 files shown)


engine/apps/integrations/mixins/alert_channel_defining_mixin.py
```diff
@@ -29,83 +31,92 @@ class AlertChannelDefiningMixin(object):
...
+            return self.get_alert_receive_channel_from_db_fallback(token), False
 
-        request = args[0]
-        request.alert_receive_channel = alert_receive_channel
-        finish = perf_counter()
-        logger.info(f"AlertChannelDefiningMixin finished in {finish - start}")
-        return super(AlertChannelDefiningMixin, self).dispatch(*args, **kwargs)
+    def get_alert_receive_channel_from_db_fallback(self, token: str) -> Optional[AlertReceiveChannel]:
+        if cache.get(self.CACHE_KEY_DB_FALLBACK):
+            for obj in serializers.deserialize("json", cache.get(self.CACHE_KEY_DB_FALLBACK)):
```
joeyorlando, line 112: making two `cache.get` calls for the same key here. You could probably do something like:
```python3
if (objects := cache.get(self.CACHE_KEY_DB_FALLBACK)) is not None:
  for obj in serializers.deserialize("json", objects):
    ...
```

engine/apps/integrations/mixins/alert_channel_defining_mixin.py
```diff
@@ -29,83 +31,92 @@ class AlertChannelDefiningMixin(object):
...
+        except OperationalError:
+            logger.info("Cannot connect to database, using cache to consume alerts!")
+            return self.get_alert_receive_channel_from_db_fallback(token), False
 
-        request = args[0]
-        request.alert_receive_channel = alert_receive_channel
-        finish = perf_counter()
-        logger.info(f"AlertChannelDefiningMixin finished in {finish - start}")
-        return super(AlertChannelDefiningMixin, self).dispatch(*args, **kwargs)
+    def get_alert_receive_channel_from_db_fallback(self, token: str) -> Optional[AlertReceiveChannel]:
```
joeyorlando, line 110: _nit_ should this method be renamed? `_from_db_fallback` implies to me it's fetching from a database, but it's fetching from cache

engine/apps/integrations/mixins/alert_channel_defining_mixin.py
```diff
@@ -29,83 +31,92 @@ class AlertChannelDefiningMixin(object):
...
-                raise
-        else:
-            if not alert_receive_channel:
-                logger.info(f"Integration {kwargs['alert_channel_key']} does not exist")
-                raise PermissionDenied(INTEGRATION_PERMISSION_DENIED_MESSAGE)
-            if alert_receive_channel.organization.is_moved:
+    def get_alert_receive_channel_from_cache(self, token: str) -> tuple[Optional[AlertReceiveChannel], Optional[str]]:
+        # Trying to define from short-term cache
+        cache_key_short_term = self.CACHE_KEY_SHORT_TERM + "_" + token
+        cached_alert_receive_channel_raw = cache.get(cache_key_short_term)
```
joeyorlando, line 53: curious why we use two different cache keys (this + `cache.get(self.CACHE_KEY_DB_FALLBACK)` down in `get_alert_receive_channel_from_db`)

mderynck: There are two usages of the cache, 1 which is short lived to optimize regular request traffic by caching individual channels and the other caches everything for if the DB goes down.  The keys are different because the fallback cache is initially loaded by StartupProbeView which does not have a specific channel to operate on.  The fallback channels never get expired (timeout=None) so they should always be available the other key is used to track expiry and update the channels by replacing the value.

joeyorlando: that makes sense, thanks for the breakdown 👍

joeyorlando (approved):

matiasb (approved):

PR #4168, opened 2024-04-05
Optimize GET /users internal api endpoint

Ferril:
# What this PR does
Speed up `GET /users` internal api endpoint by reducing number of calls to database 

Related to slow schedules page issue - https://github.com/grafana/oncall-private/issues/1552
## Checklist

### DIFF
--- a/engine/apps/api/serializers/user.py
+++ b/engine/apps/api/serializers/user.py
@@ -88,7 +88,14 @@ class ListUserSerializer(DynamicFieldsModelSerializer, EagerLoadingMixin):
     cloud_connection_status = serializers.SerializerMethodField()
     working_hours = WorkingHoursSerializer(required=False)
 
-    SELECT_RELATED = ["telegram_verification_code", "telegram_connection", "organization", "slack_user_identity"]
+    SELECT_RELATED = [
+        "telegram_verification_code",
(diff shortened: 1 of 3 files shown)


engine/apps/api/serializers/user.py
```diff
@@ -159,7 +166,13 @@ def get_notification_chain_verbal(self, obj: User) -> NotificationChainVerbal:
         return {"default": " - ".join(default), "important": " - ".join(important)}
 
     def get_cloud_connection_status(self, obj: User) -> CloudSyncStatus | None:
-        if settings.IS_OPEN_SOURCE and live_settings.GRAFANA_CLOUD_NOTIFICATIONS_ENABLED:
+        is_open_source_with_cloud_notifications = self.context.get("is_open_source_with_cloud_notifications", None)
```
joeyorlando, line 169: under what circumstances would `context["is_open_source_with_cloud_notifications"]` not be set?

Ferril: I'm not sure 🤔 It will be set for /user and /users endpoints, but I wouldn't completely rely on this, because we reuse this serizalizer so many times, so it's easy to forget to set the context somewhere

PR #4169, opened 2024-04-05
Optimize GET /schedules and /current_user_events internal api endpoints

Ferril:
# What this PR does
Speed up `GET /schedules` and `GET /current_user_events` internal api endpoints by reducing number of calls to database 

Related to https://github.com/grafana/oncall-private/issues/1552

## Checklist

### DIFF
--- a/engine/apps/api/views/schedule.py
+++ b/engine/apps/api/views/schedule.py
@@ -143,14 +143,10 @@ class ScheduleView(
         if self.action == "list":
             # listing page, only get oncall users for current page schedules, prefetch shift swap requests
             current_schedules = self.filter_queryset(self.get_queryset(annotate=False)).prefetch_related(
-                Prefetch(
-                    "shift_swap_requests",
+                self.prefetch_shift_swaps(
                     queryset=ShiftSwapRequest.objects.filter(
                         swap_start__lte=events_datetime, swap_end__gte=events_datetime
                     )
-                    .select_related("benefactor", "beneficiary")
-                    .order_by("created_at"),
-                    to_attr=PREFETCHED_SHIFT_SWAPS,
                 )
             )
             current_schedules = self.paginate_queryset(current_schedules)
@@ -158,6 +154,14 @@ class ScheduleView(
             # if this is a particular schedule detail, only consider it as current
             current_schedules = [self.get_object(annotate=False)]
         return get_oncall_users_for_multiple_schedules(current_schedules, events_datetime)
+
+    @staticmethod
+    def prefetch_shift_swaps(queryset):
+        return Prefetch(
+            "shift_swap_requests",
+            queryset=queryset.select_related("benefactor", "beneficiary").order_by("created_at"),
+            to_attr=PREFETCHED_SHIFT_SWAPS,
+        )
 
     def get_serializer_context(self):
         context = super().get_serializer_context()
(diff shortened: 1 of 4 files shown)


engine/apps/api/views/schedule.py
```diff
@@ -186,8 +199,9 @@ def get_queryset(self, ignore_filtering_by_available_teams=False, annotate=True)
         )
         if not ignore_filtering_by_available_teams:
             queryset = queryset.filter(*self.available_teams_lookup_args).distinct()
-        if not is_short_request or annotate:
-            queryset = self._annotate_queryset(queryset)
+        if not is_short_request:
+            if annotate:
```
joeyorlando, line 203: could this be simplified to `if not is_short_request and annote`?

engine/apps/api/views/schedule.py
```diff
@@ -138,13 +139,25 @@ def oncall_users(self):
...
+                Prefetch(
+                    "shift_swap_requests",
+                    queryset=ShiftSwapRequest.objects.filter(
+                        swap_start__lte=events_datetime, swap_end__gte=events_datetime
+                    )
+                    .select_related("benefactor", "beneficiary")
+                    .order_by("created_at"),
+                    to_attr=PREFETCHED_SHIFT_SWAPS,
+                )
+            )
```
joeyorlando, line 155: _nit_ consider extracting out the `Prefetch` construction logic into a staticmethod on the view class to avoid the duplication below

engine/apps/schedules/ical_utils.py
```diff
@@ -358,14 +358,14 @@ def list_users_to_notify_from_ical_for_period(
     schedule: "OnCallSchedule",
     start_datetime: datetime.datetime,
     end_datetime: datetime.datetime,
-) -> typing.List["User"]:
+) -> typing.Sequence["User"]:
     users_found_in_ical: typing.Sequence["User"] = []
```
joeyorlando, line 362: it seems like `users_found_in_ical` defined here is never used, since the variable is overwritten down on line 368?

engine/apps/schedules/ical_utils.py
```diff
@@ -358,14 +358,14 @@ def list_users_to_notify_from_ical_for_period(
...
     events = schedule.final_events(start_datetime, end_datetime)
-    usernames = []
+    usernames: typing.List[str] = []
     for event in events:
         usernames += [u["email"] for u in event.get("users", [])]
 
-    users_found_in_ical = users_in_ical(usernames, schedule.organization)
+    users_found_in_ical = memoized_users_in_ical(tuple(usernames), schedule.organization)
     return users_found_in_ical
```
joeyorlando, line 369: why cast `usernames` to a tuple?
```suggestion
    return memoized_users_in_ical(tuple(usernames), schedule.organization)
```

engine/apps/api/views/schedule.py
Ferril: in some cases we need to do `setup_eager_loading`, but don't need to do annotation

engine/apps/schedules/ical_utils.py
Ferril: yeah 🤔 I'll remove this

engine/apps/schedules/ical_utils.py
Ferril: because `memoized_users_in_ical` doesn't work with list, it raises `TypeError: unhashable type: 'list'`

engine/apps/api/views/schedule.py
joeyorlando: ignore my comment 😄  I think I glanced over this too quickly the first time

joeyorlando (approved):
LGTM! 👍

Issue #4142, opened 2024-04-02
Get rid of GForm


maskin25:

joeyorlando:
@maskin25 if this is the direction we want to head in, we can probably close https://github.com/grafana/oncall/issues/3193, wdyt?

maskin25:
yes, agree, we decided to get rid of GForm and gave preference to the direct use of react-hook-form, so I'm closing https://github.com/grafana/oncall/issues/3193

PR #4171, opened 2024-04-05
Replace GForm with react-hook-form

maskin25:
# What this PR does

Replace GForm with react-hook-form

Closes https://github.com/grafana/oncall/issues/4142

## Checklist


grafana-plugin/src/components/ManualAlertGroup/ManualAlertGroup.tsx
```diff
@@ -1,82 +1,113 @@
...
+          </form>
+        </FormProvider>
       </VerticalGroup>
     </Drawer>
   );
 });
+
+export const getStyles = () => ({
+  form: css`
+    width: 100%;
```
teodosii, line 111: there's a util style for `width: 100%`, it's called `width100` located in `utils.styles.ts`, no need to repeat classes with width: 100

grafana-plugin/src/models/alert_receive_channel/alert_receive_channel.ts
```diff
@@ -149,6 +149,7 @@ export class AlertReceiveChannelStore {
 
     this.fetchCounters();
 
+    // @ts-ignore for some reason ts doesn't see `results` as list of ApiSchemas['AlertReceiveChannel']
```
teodosii, line 152: Let's not use ts-ignore unless we're missing backend types. You could easily get around by casting it to `ApiSchemas['AlertReceiveChannel'][]`. Didn't look closely, maybe it's related to the readonly fields in which case you'd need to rely on `OmitReadonlyMembers`

grafana-plugin/src/components/ManualAlertGroup/ManualAlertGroup.tsx
```diff
@@ -1,82 +1,113 @@
...
+
+  const {
+    handleSubmit,
+    control,
+    formState: { errors },
+  } = formMethods;
+
   const hasSelectedEitherATeamOrAUser = selectedTeamResponder !== null || selectedUserResponders.length > 0;
-  const formIsSubmittable = hasSelectedEitherATeamOrAUser && formIsValid;
+  const formIsSubmittable = hasSelectedEitherATeamOrAUser;
```
teodosii, line 47: These 2 in here can be 1 line instead and named directly to `canSubmitForm` or your naming

grafana-plugin/src/models/alert_receive_channel/alert_receive_channel.ts
maskin25: ok, I can do `return results as Array<ApiSchemas['AlertReceiveChannel']>;`  seems `OmitReadonlyMembers` is not a point in my case, @brojd any ideas?

teodosii (approved):

grafana-plugin/src/models/alert_receive_channel/alert_receive_channel.ts
teodosii: The response you actually get from oncall helper is `AlertReceiveChannelPolymorphic` and  that's why it sees it as a missmatch. It's totally fine to just cast it to `AlertReceiveChannel`

grafana-plugin/src/containers/MaintenanceForm/MaintenanceForm.tsx
```diff
@@ -49,34 +54,151 @@ export const MaintenanceForm = observer((props: MaintenanceFormProps) => {
...
+                        },
+                        {
+                          value: 43200,
+                          label: '12 hours',
+                        },
+                        {
+                          value: 86400,
+                          label: '24 hours',
+                        },
+                      ]}
```
brojd, line 174: Maybe we could have a const out of component with those static options (and the ones above)

grafana-plugin/src/containers/OutgoingWebhookForm/OutgoingWebhookForm.tsx
```diff
@@ -59,165 +50,139 @@ export const WebhookTabs = {
...
         }
-        outgoingWebhookStore.setLabelsFormErrors(undefined);
         onHide();
         onUpdate();
-      } catch (err) {
-        if (err.response?.data?.labels) {
-          outgoingWebhookStore.setLabelsFormErrors(err.response.data.labels);
-        }
+      } catch (error) {
+        /*  Object.keys(error.response.data).forEach((key) => {
```
brojd, line 128: Do we need this comment? If yes, could you put one line more with explanation? If not, let's remove it

grafana-plugin/src/containers/OutgoingWebhookForm/OutgoingWebhookForm.tsx
```diff
@@ -370,46 +275,110 @@ export const OutgoingWebhookForm = observer((props: OutgoingWebhookFormProps) =>
...
-  }
-});
+      </div>
+    </Drawer>
+  );
+};
+
+interface EditWebhookTabsProps {
+  id: OutgoingWebhookFormProps['id'];
+  data: any;
```
brojd, line 292: We have couple of places in this component where we use `data: any`. I suggest to improve all those places. I guess we could use `ReturnType<typeof prepareDataForEdit>;` in each?

grafana-plugin/src/containers/OutgoingWebhookForm/OutgoingWebhookForm.tsx
```diff
@@ -370,46 +275,110 @@ export const OutgoingWebhookForm = observer((props: OutgoingWebhookFormProps) =>
...
+  onTemplateEditClick: (params: TemplateParams) => void;
+  preset: OutgoingWebhookPreset;
+}
+
+const EditWebhookTabs = (props: EditWebhookTabsProps) => {
+  const { id, data, action, onHide, onUpdate, onDelete, onSubmit, onTemplateEditClick, preset } = props;
+
+  const history = useHistory();
+
+  const [activeTab, setActiveTab] = useState<string>(
```
brojd, line 307: It's minor but I think the `string` type will be inferred from the default value so we don't need to provide it

grafana-plugin/src/containers/OutgoingWebhookForm/OutgoingWebhookForm.types.ts
```diff
@@ -15,4 +15,11 @@ export const WebhookFormFieldName = {
   ForwardAll: 'forward_all',
   Data: 'data',
 } as const;
+
 export type WebhookFormFieldName = (typeof WebhookFormFieldName)[keyof typeof WebhookFormFieldName];
+
+export interface TemplateParams {
+  name: WebhookFormFieldName;
+  value: any;
```
brojd, line 23: Why is it `any`? It seems we always invoke it with the `string`, can we use `string` instead of `any` in this case?

grafana-plugin/src/models/alert_receive_channel/alert_receive_channel.ts
brojd: @maskin25 let's go with `return results as Array<ApiSchemas['AlertReceiveChannel']>;` , we need to clarify with backend under which conditions it's possible to get from this endpoint `components['schemas']['FilterAlertReceiveChannel'];` and improve the types. For now this casting is ok and better than ts-ignore

grafana-plugin/src/containers/MaintenanceForm/MaintenanceForm.tsx
maskin25: yes, we can, but it doesn't make sense a lot because there no others consumers of those consts

grafana-plugin/src/containers/OutgoingWebhookForm/OutgoingWebhookForm.tsx
maskin25: Ive uncommented, it injects errors into form that come from BE

PR #4174, opened 2024-04-05
add PUT/PATCH alert group internal API endpoints

joeyorlando:
# What this PR does

- Add PUT/PATCH alert group internal API endpoints. Will be consumed by the Grafana Incident API to update `grafana_incident_id`.
  - **TODO**: figure out how we will handle authnz here (we're currently talking with Identity Access team to possibly use the Cloud Auth API)
- Update the `grafana_incident_id` column to be a JSON column (so that we can support storing >= 1 Incident ID for a given alert group)
- Allow filtering alert groups by `grafana_incident_id`
- Update `AlertGroup.declare_incident_link` to include ID

## Checklist


engine/apps/api/serializers/alert_group.py
```diff
@@ -242,3 +242,8 @@ def get_limited_alerts(self, obj: "AlertGroup"):
 
     def get_paged_users(self, obj: "AlertGroup") -> typing.List[PagedUser]:
         return obj.get_paged_users()
+
+
+class AlertGroupUpdateSerializer(AlertGroupSerializer):
+    class Meta(AlertGroupSerializer.Meta):
+        read_only_fields = [field for field in AlertGroupSerializer.Meta.fields if field not in ["grafana_incident_id"]]
```
matiasb, line 249: Curious, could we instead set `fields = ["grafana_incident_id"]` and limit the serializer/update to only work for that field?

robert-northmind (commented):
LGTM,
but.. 😄 I am no python or django developer but I think it looks good. 
thanks @joeyorlando  for driving this.

joeyorlando:
blocked by https://github.com/grafana/incident/issues/5266

PR #4175, opened 2024-04-07
Allow webhook modification by API for advanced webhook

### LINKED ISSUE #3958: Relax webhook modification via API limitation

Currently if a webhook is created in the UI it cannot be modified through the API, you instead get the message `{"non_field_errors":["Preset webhooks must be modified through web UI"]}` The reason for this is presets apply logic and limitations better communicated through the UI.

This limitation should be removed so that webhooks created with the `Advanced` preset can be modified by the API since this preset does not apply custom logic and exposes all available fields.

### Product Area

Alert Flow & Configuration

### Anything else to add?

_No response_

ravishankar15:
# What this PR does

Enables the API to perform updates on the advanced webhooks created via the UI

Closes #3958 

## Checklist

### DIFF
--- a/engine/apps/public_api/serializers/webhooks.py
+++ b/engine/apps/public_api/serializers/webhooks.py
@@ -162,6 +162,8 @@ class WebhookCreateSerializer(serializers.ModelSerializer):
         if (
             self.instance
             and self.instance.preset
+            and WebhookPresetOptions.ADVANCED_PRESET_META_DATA
+            and WebhookPresetOptions.ADVANCED_PRESET_META_DATA.id
             and self.instance.preset != WebhookPresetOptions.ADVANCED_PRESET_META_DATA.id
         ):
             raise serializers.ValidationError(PRESET_VALIDATION_MESSAGE)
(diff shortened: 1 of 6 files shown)


ravishankar15:
The `preset` param is not allowed in the API. This made sense because the preset for the web hook created from API is set to null. Now that we are allowing the `advanced_webhook` preset to be updated from the API the current state of the PR would retain this behaviour i.e The preset param will not be allowed in the API.

**Reason**
If we have to allow the `preset` param in the API the possible value would be `null` or `advanced_webhook`.

**Not related to this PR**
I feel we can merge the `null` or `advanced_webhook` behaviour and handle like,
1. The web hook created from API will always be of type `advanced_webhook`.
2. The preset param from the API can only have `advanced_webhook`
3. Can make the `preset` Enum on the database level.
4. Run migration to replace null value with `advanced_webhook`.
5. Remove any `null` value handling for the preset param

engine/apps/public_api/serializers/webhooks.py
```diff
@@ -158,7 +159,11 @@ def validate_preset(self, preset):
         raise serializers.ValidationError(PRESET_VALIDATION_MESSAGE)
 
     def validate(self, data):
-        if self.instance and self.instance.preset:
+        if (
+            self.instance
+            and self.instance.preset
+            and self.instance.preset != WebhookPresetOptions.ADVANCED_PRESET_META_DATA.id
```
matiasb, line 165: nit, but just in case, if for some reason the `ADVANCED_WEBHOOK_PRESET` setting is not set (e.g. to disable them), this would explode, right? (there is no `.id`)

matiasb (approved):
Makes sense to me, minor comment added.

mderynck (approved):

matiasb:
Thanks for your contribution @ravishankar15! :+1:

PR #4204, opened 2024-04-10
Add endpoint for organization config checks

matiasb:
Related to https://github.com/grafana/oncall-private/issues/2563

### DIFF
--- a/engine/apps/api/serializers/organization.py
+++ b/engine/apps/api/serializers/organization.py
@@ -2,6 +2,7 @@ from dataclasses import asdict
 
 from rest_framework import serializers
 
+from apps.base.messaging import get_messaging_backend_from_id
 from apps.base.models import LiveSetting
 from apps.phone_notifications.phone_provider import get_phone_provider
 from apps.slack.models import SlackTeamIdentity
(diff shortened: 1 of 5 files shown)


engine/apps/api/serializers/organization.py
```diff
@@ -107,3 +108,30 @@ class FastOrganizationSerializer(serializers.ModelSerializer):
...
+            or obj.telegram_channel.exists()  # telegram is connected
+            or (msteams_backend and msteams_backend.is_configured_for_organization(obj))  # msteams is connected
+        )
+
+    def get_is_integration_chatops_connected(self, obj):
+        return (
+            obj.alert_receive_channels.filter(channel_filters__notify_in_slack=True).exists()
+            or obj.alert_receive_channels.filter(channel_filters__notify_in_telegram=True).exists()
+            or obj.alert_receive_channels.filter(channel_filters__notification_backends__MSTEAMS__enabled=True).exists()
+        )
```
joeyorlando, line 137: I wonder if there's any way we can share the various "chatops connected" logic with this django query?

https://github.com/grafana/oncall/blob/dev/engine/apps/user_management/models/organization.py#L316-L345

engine/apps/base/messaging.py
```diff
@@ -43,6 +43,9 @@ def unlink_user(self, user):
     def is_enabled_for_organization(organization):
         return True
 
+    def is_configured_for_organization(self, organization):
```
joeyorlando, line 46: is this expected to be overridden in classes that subclass `BaseMessagingBackend`?

joeyorlando (approved):

engine/apps/base/messaging.py
joeyorlando: ahh nvm, looking at [the other PR](https://github.com/grafana/oncall-private/pull/2634/files#diff-5308bafce8207e7b8e5b7049a10a9748055e408407efcb5215cf42ecaf64271aR17-R18), I understand now 🙂

engine/apps/api/serializers/organization.py
matiasb: I see your point, but it seems tricky? I think they have similar but slightly different conditions. Here we aren't really interested in knowing which integration, with less specific restrictions, and we "short-circuit" as soon as we have a satisfying condition (and I would keep the other query as "one query to get them all" for now).

Issue #4122, opened 2024-03-27
Better Slack link generation


imtoori:
### What went wrong?

**What happened**:
- the slack link under `alertGroup.permalinks.slack` uses an old version of slack links

**What did you expect to happen**:
- we should use this new format https://api.slack.com/reference/deep-linking#app_or_bot

### How do we reproduce it?

get an alert group from the backend and verify the slack link has not this format https://api.slack.com/reference/deep-linking#app_or_bot

### Grafana OnCall Version

v1.3.113

### Product Area

Alert Flow & Configuration

### Grafana OnCall Platform?

None

### Anything else to add?

This is important for the mobile app because with the current format we cannot open the slack app, so every slack links defaults to the browser.

PR #4205, opened 2024-04-10
Add Slack deep link entry to alert group permalinks

matiasb:
Related to https://github.com/grafana/oncall/issues/4122

### DIFF
--- a/engine/apps/alerts/models/alert_group.py
+++ b/engine/apps/alerts/models/alert_group.py
@@ -505,6 +505,10 @@ class AlertGroup(AlertGroupSlackRenderingMixin, EscalationSnapshotMixin, models.
         return None if self.slack_message is None else self.slack_message.permalink
 
     @property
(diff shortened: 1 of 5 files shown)


engine/apps/alerts/models/alert_group.py
```diff
@@ -527,6 +531,7 @@ def telegram_permalink(self) -> typing.Optional[str]:
     def permalinks(self) -> Permalinks:
```
joeyorlando, line 531: does this affect the output of `yarn generate-types`? (wondering if the `Permalinks` type needs to be updated here?)

joeyorlando (approved):

engine/apps/alerts/models/alert_group.py
matiasb: good point! updating and pushing :+1:

maskin25 (approved):

brojd (approved):

PR #4212, opened 2024-04-12
add support for logging in to mobile using deep links

imtoori:
# What this PR does

Optimize mobile app connection layout for smaller screens and add LinkLoginButton component

https://github.com/grafana/oncall-private/issues/2528

## Checklist

### DIFF
--- a/grafana-plugin/src/containers/MobileAppConnection/MobileAppConnection.tsx
+++ b/grafana-plugin/src/containers/MobileAppConnection/MobileAppConnection.tsx
@@ -20,6 +20,7 @@ import styles from './MobileAppConnection.module.scss';
 import { DisconnectButton } from './parts/DisconnectButton/DisconnectButton';
 import { DownloadIcons } from './parts/DownloadIcons/DownloadIcons';
 import { QRCode } from './parts/QRCode/QRCode';
+import { LinkLoginButton } from './parts/LinkLoginButton/LinkLoginButton';
 
 const cx = cn.bind(styles);
 
@@ -143,6 +144,7 @@ export const MobileAppConnection = observer(({ userPk }: Props) => {
   }
 
   let content: React.ReactNode = null;
+  const QRCodeDataParsed = QRCodeValue && getParsedQRCodeValue();
 
   if (fetchingQRCode || disconnectingMobileApp || !userPk || !basicDataLoaded) {
     content = <LoadingPlaceholder text="Loading..." />;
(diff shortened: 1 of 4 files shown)


robert-northmind (approved):
LGTM

grafana-plugin/src/containers/MobileAppConnection/MobileAppConnection.tsx
```diff
@@ -192,12 +192,19 @@ export const MobileAppConnection = observer(({ userPk }: Props) => {
     );
   }
 
+  const isMobile = window.matchMedia('(max-width: 768px)').matches;
```
brojd, line 195: can we put it in some global place like `src/utils/utils`?

imtoori: done ✅

grafana-plugin/src/containers/MobileAppConnection/MobileAppConnection.tsx
```diff
@@ -192,12 +192,19 @@ export const MobileAppConnection = observer(({ userPk }: Props) => {
...
+
   return (
     <VerticalGroup>
       <div className={cx('container')}>
         <Block shadowed bordered withBackground className={cx('container__box')}>
           <DownloadIcons />
         </Block>
+        {QRCodeValue && isMobile && (
```
brojd, line 201: maybe better to have a check of `QRCodeDataParsed && isMobile` here instead? `QRCodeDataParsed` has `any` type 🙃 if it's `undefined` (which can happen if `JSON.parse` in `getParsedQRCodeValue` fails), then we will get "Cannot read property ... of undefined" error here

imtoori: I took inspiration from line 169, maybe we should change it there too?

brojd: There situation is a bit different because we have first part that depends just on `QRCodeValue` and then second nested part that depends on `QRCodeDataParsed`. Second part already have a check for existence of `QRCodeDataParsed`.
image

So I think we should be good to not touch this place and only use `QRCodeDataParsed && isMobile` condition in your new change

imtoori: ah ok cool, updated

PR #4227, opened 2024-04-15
Prepare insight metrics structure for adding service_name label

Ferril:
# What this PR does
Prepare insight metrics for adding `service_name` label.
This PR updates metrics cache structure, supporting both old and new version of cache.
`service_name` label can be added with additional PR when all metric cache is updated.

https://github.com/grafana/oncall-private/issues/2610

## Checklist


engine/apps/metrics_exporter/helpers.py
```diff
@@ -253,22 +261,29 @@ def metrics_update_alert_groups_state_cache(states_diff, organization_id):
...
-            else:
-                integration_alert_groups[previous_state] = 0
-        for new_state, counter in integration_states_diff["new_states"].items():
-            integration_alert_groups[new_state] += counter
+        for service_name, service_state_diff in service_data.items():
+            if "services" in integration_alert_groups:
+                states_to_update = integration_alert_groups["services"].setdefault(
+                    service_name, get_default_states_dict()
+                )
+            else:  # support previous version of metrics cache. This clause can be removed later
```
iskhakov, line 273: Could you please add more context, which version and when should be removed

engine/apps/metrics_exporter/helpers.py
```diff
@@ -243,7 +251,7 @@ def metrics_bulk_update_team_label_cache(teams_updated_data, organization_id):
     cache.set(metric_alert_groups_response_time_key, metric_alert_groups_response_time, timeout=metrics_cache_timeout)
 
 
-def metrics_update_alert_groups_state_cache(states_diff, organization_id):
+def metrics_update_alert_groups_state_cache(states_diff: dict, organization_id: int):
     """Update alert groups state metric cache for each integration in states_diff dict."""
```
iskhakov, line 255: the comment seems outdated. Could you please add more info and examples

engine/apps/metrics_exporter/tasks.py
```diff
@@ -111,38 +114,72 @@ def calculate_and_cache_metrics(organization_id, force=False):
...
-        # get response time
-        all_response_time = integration.alert_groups.filter(
-            started_at__gte=response_time_period,
-            response_time__isnull=False,
-        ).values_list("response_time", flat=True)
-
-        all_response_time_seconds = [int(response_time.total_seconds()) for response_time in all_response_time]
-
-        metric_alert_group_response_time[integration.id] = {
+            # count alert groups with `service_name` label group by label value
```
iskhakov, line 130: Have we tried these queries on production db? Do we need to apply any optimizations?

engine/apps/metrics_exporter/constants.py
```diff
@@ -25,6 +29,7 @@ class AlertGroupsResponseTimeMetricsDict(typing.TypedDict):
     slug: str
     id: int
     response_time: list
```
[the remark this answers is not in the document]
Ferril: yes, thx 👍

engine/apps/metrics_exporter/helpers.py
```diff
@@ -253,23 +282,40 @@ def metrics_update_alert_groups_state_cache(states_diff, organization_id):
...
+                states_to_update = integration_alert_groups["services"].setdefault(
+                    service_name, get_default_states_dict()
+                )
             else:
-                integration_alert_groups[previous_state] = 0
-        for new_state, counter in integration_states_diff["new_states"].items():
-            integration_alert_groups[new_state] += counter
+                # support version of metrics cache without service name. This clause can be removed when all metrics
+                # cache is updated on prod (~2 days after release)
+                states_to_update = integration_alert_groups
```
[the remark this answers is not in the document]
Ferril: yes, it should happen within 24 hours for each organization

engine/apps/metrics_exporter/tasks.py
Ferril: > +1 to optimize as possible (using the index improved things?)

the slowest call from my tests is ~10s now. It uses readonly db, so I don't think this would be a problem 🤔 

> n any case, this shouldn't affect the response time in the scrape endpoint right? (there we are still getting the data from cache and returning it, correct?)

Correct, we get all data from the cache and don't do any calculations there

engine/apps/metrics_exporter/metrics_cache_manager.py
```diff
@@ -40,56 +40,59 @@ def get_default_states_diff_dict():
         return default_dict
 
     @staticmethod
-    def update_integration_states_diff(metrics_dict, integration_id, previous_state=None, new_state=None):
-        metrics_dict.setdefault(integration_id, MetricsCacheManager.get_default_states_diff_dict())
+    def update_integration_states_diff(metrics_dict, integration_id, service_name, previous_state=None, new_state=None):
+        state_per_service = {service_name: MetricsCacheManager.get_default_states_diff_dict()}
+        metrics_dict.setdefault(integration_id, state_per_service)
```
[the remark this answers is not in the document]
Ferril: Makes sense. It was used for a bulk update cache, and the incoming data there was grouped by integration. We don't use a bulk update anymore and the metrics_dict always contains only one integration with information about one alert group, so there is no difference there, but since we have it, it's better that it works correctly in different cases 👍

iskhakov (approved):

PR #4230, opened 2024-04-15
Handle alert group deleted when task is already queued

mderynck:
# What this PR does
- Since send_alert_create_signal is inside transaction on_commit we can conclude that if it does not exist it was intentionally deleted before the task could run and the task can exit instead of retrying
- Improve logging when send_alert_create_signal is called so both alert and alert group are in the same line so you don't need to search the logs as much
- Improve logging on public api delete alert group so we can know what the alert group belonged to and the responsible user/org
- Remove distribute_alerts (Stopped using a while back, code should be safe to remove now, no tasks running in system)

Closes https://github.com/grafana/oncall-private/issues/2640

## Checklist

### DIFF
--- a/engine/apps/alerts/tasks/distribute_alert.py
+++ b/engine/apps/alerts/tasks/distribute_alert.py
@@ -1,46 +1,9 @@
 from django.conf import settings
 
-from apps.alerts.constants import TASK_DELAY_SECONDS
-from apps.alerts.signals import alert_create_signal, alert_group_escalation_snapshot_built
+from apps.alerts.signals import alert_create_signal
 from common.custom_celery_tasks import shared_dedicated_queue_retry_task
 
 from .task_logger import task_logger
-
-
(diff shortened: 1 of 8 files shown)


engine/apps/alerts/tasks/distribute_alert.py
```diff
@@ -1,56 +1,23 @@
...
 
 
-@shared_dedicated_queue_retry_task(
-    autoretry_for=(Exception,), retry_backoff=True, max_retries=1 if settings.DEBUG else None, default_retry_delay=60
-)
-def distribute_alert(alert_id):
-    """
-    We need this task to make task processing async and to make sure the task is delivered.
-    This task is not used anymore, but we keep it for the tasks in the queue to be processed.
-    TODO: remove this task after all the tasks in the queue are processed.
```
mderynck, line 17: This should be safe to remove now

vadimkerr (approved):
LGTM

PR #4254, opened 2024-04-18
SNOW final tweaks

teodosii:
- Show `external_url` for an AG that has its integration connected to ServiceNow
- Addressed some of @raphael-batte 's remarks
- Prevent SNOW creation if connection to servicenow instance fails
- Skip error handling that showed 500 or 401: Unauthorized on some calls to test_connection and instead rely purely on displaying `OK` or `Error`


grafana-plugin/src/containers/IntegrationForm/IntegrationForm.tsx
```diff
@@ -414,11 +419,19 @@ export const IntegrationForm = observer(
         labels: labels ? [...labels] : undefined,
       };
 
-      if (formData.integration !== 'servicenow') {
+      const isServiceNow = formData.integration === 'servicenow';
+      const isCreate = id === 'new';
```
maskin25, line 423: `nit` I think it's better to create vars if they are consumed more than once only or if they contans complex logic behind (a large set of consecutive OR and AND statements), in this case adding a `isCreate` var looks redundant

grafana-plugin/src/pages/incident/Incident.tsx
```diff
@@ -278,7 +278,7 @@ class _IncidentPage extends React.Component<IncidentPageProps, IncidentPageState
     const showLinkTo = !incident.dependent_alert_groups.length && !incident.root_alert_group && !incident.resolved;
     const integrationNameWithEmojies = <Emoji text={incident.alert_receive_channel.verbal_name} />;
     const sourceLink = incident?.render_for_web?.source_link;
-    const isServiceNow = incident?.alert_receive_channel?.integration === 'servicenow';
+    const isServiceNow = Boolean(incident?.external_urls?.find((el) => el.integration_type === 'servicenow'));
```
maskin25, line 281: `nit` you use "servicenow" string several times, let's maybe extract this string to the const like `INTEGRATION_SERVICENOW='servicenow'` or something

maskin25 (approved):

grafana-plugin/src/containers/IntegrationForm/IntegrationForm.tsx
teodosii: IMO this healps with readability, much easier to read `isCreate` than `id==='new'`, we have some other places like this as well

`isNew` for example on IntegrationForm

grafana-plugin/src/pages/incident/Incident.tsx
teodosii: Done

grafana-plugin/src/containers/IntegrationForm/IntegrationForm.tsx
teodosii: Done, didn't notice we already had `isNew`

Issue #3313, opened 2023-11-09
Cannot search items within Type of integration dropdown


brojd:
### What went wrong?

**What happened**:
- Within filters of integrations when I want to search for specific Type of integration (e.g. Alertmanager, Webhook), the search doesn't narrow down the list of options

https://github.com/grafana/oncall/assets/12073649/051e784b-8e36-4c27-b996-72ec39e18395

**What did you expect to happen**:
- after typing a valid part of any of the integration Type, list of dropdown's options should be narrowed down accordingly

### How do we reproduce it?

1. Open Grafana OnCall and go to Integrations
2. Click on Search or filter results
3. Select "Type"
4. Type "Alert"
5. I should see that list is narrowed down to Alertmanager but instead I can still see all the options

### Grafana OnCall Version

1.3.55

### Product Area

Alert Flow & Configuration

### Grafana OnCall Platform?

None

### Anything else to add?

@joeyorlando made initial investigation and seems that frontend is trying to use `search` param in the HTTP request while it is not supported on the backend. It seems the filtering should be done client side

bndrv:
+1
I'm not sure if my case matches this task, but it seems so. Searching by group alert names doesn't work. That is, nothing happens, and the queue doesn't change. Also, searching doesn't work when trying to attach a group of alerts to another group. The search brings up a list of other alert groups, but the list doesn't change based on the character input (number or part of the alert group name), making it impossible to find the desired alert.

joeyorlando:
@bndrv the issue you are mentioning is different. It is being tracked by https://github.com/grafana/oncall/issues/2232

teodosii:
Fixed by @matiasb [here](https://github.com/grafana/oncall/pull/4255)

PR #4255, opened 2024-04-18
Enable search for integration type options filter internal endpoint

matiasb:


Ferril (approved):

joeyorlando:
@matiasb @teodosii I think this also closes #3313, wdyt?

PR #4261, opened 2024-04-20
Modify alert grouping behaviour for demo alerts

### LINKED ISSUE #3129: Override send demo alert grouping behavior

Currently when you Send demo alert it ignores the grouping template.  This behaviour is useful since often when sending a demo alert we are more interested in testing notifications and escalation chains so if the demo alert was grouped to an already existing group it would be confusing.  

However in the case you actually want to test the grouping behaviour you are not able to do so through the web UI (it always generates a unique grouping).  You can work around this by having the UI generate a curl command, if you use the curl command it will not override the grouping behaviour (which is confusing in a different way).

It would be nice if we exposed a consistent way to enable/disable grouping when sending demo alerts.  Possibly this could be a specific key/value in the payload we document instead of the behaviour being fixed in code.

### Product Area

Alert Flow & Configuration

### Anything else to add?

_No response_

ravishankar15:
# What this PR does

1. Exposes a boolean param `group_alerts` for the send_demo_alert endpoint
2. Adds radio button on the frontend for toggling the grouping behaviour

Screenshot 2024-04-20 at 5 44 50 PM

Closes https://github.com/grafana/oncall/issues/3129

## Checklist


nxxqxxj:
Will this be merged soon? This is very helpful while using demo alerts to test template grouping. Thanks.

PR #4270, opened 2024-04-23
Update docker-compose.yml stop exposing ports publicly

fhackenberger:
You don't want ports exposed publicly, bypassing many firewall set-ups, just by trying out this docker-compose.yml. Better only bind them to 127.0.0.1

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -21,7 +21,7 @@ services:
     image: grafana/oncall
     restart: always
     ports:
-      - "8080:8080"
+      - "127.0.0.1:8080:8080"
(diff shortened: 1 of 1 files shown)


matiasb:
Thanks for the contribution. This docker-compose file is mainly for hobby setups/testing, and in some cases it is still useful to have the ports accessible in the network. Maybe we could update the readme or docs to mention this instead?

fhackenberger:
Imho especially Hobbyist should have a security by default set-up as they might be less aware of the consequences. It would be easy to add a commented out BIND_TO=0.0.0.0 into the .env, so they can easily re-enable it if they need the service on the network. Would you accept the PR, if I add this?

PR #4279, opened 2024-04-25
Refactor how RBAC enabled/disabled status is determined for Grafana Cloud stacks

joeyorlando:
# What this PR does

In cloud we are currently (somewhat) improperly determining whether or not a Grafana stack had the `accessControlOnCall` feature flag enabled. At first things worked fine. We would enable this feature toggle via the Grafana Admin UI, and then the OnCall backend would read this value from GCOM's `GET /instance/<stack_id>` endpoint (via `config.feature_toggles`), and everything worked as expected.

There was a recent change made in `grafana/deployment_tools` to set this feature flag to True for all stacks. However, for some reason, the GCOM endpoint above doesn't return the `accessControlOnCall` feature toggle value in `config.feature_toggles` if it is set in this manner (it only returns the value if it is set via the Grafana Admin UI).

So what we should instead be doing is such instead of asking GCOM for this feature toggle, infer whether RBAC is enabled on the stack by doing a `HEAD /api/access-control/users/permissions/search` (this endpoint _is only_ available on a Grafana stack if `accessControlOnCall` is enabled).

**Few caveats to this ☝️**
1. we first have to make sure that the cloud stack is in an `active` state (ie. not paused). This is because, no matter if the `accessControlOnCall` is enabled or not, if the stack is in a `paused` state it will ALWAYS return `HTTP 200` which can be misleading and lead to bugs (this feels like a bug on the Grafana API, will follow up with core grafana team)
2. Once we roll out this change we will effectively **actually** be enabling RBAC for OnCall for all orgs. The Identity Access team would prefer a progressive rollout, which is why I decided to introduce the concept of [`settings.CLOUD_RBAC_ROLLOUT_PERCENTAGE`](https://github.com/grafana/oncall/pull/4279/files#diff-3383aef931e41e44d95829ad971641eeb98fe001be2f5da92217446d300ea1b3R918) (see also [`Organization. should_be_considered_for_rbac_permissioning`](https://github.com/grafana/oncall/pull/4279/files#diff-2ca9917f4f56349be39545ee8abd459be5076295d02ca3a7ec545152fcddccdfR348-R362))

Related to https://github.com/grafana/identity-access-team/issues/667

## Checklist

### DIFF
--- a/engine/apps/grafana_plugin/tests/test_grafana_api_client.py
+++ b/engine/apps/grafana_plugin/tests/test_grafana_api_client.py
@@ -9,7 +9,6 @@ API_TOKEN = "dfjkfdjkfd"
 
 
 class TestGetUsersPermissions:
-
     @pytest.mark.parametrize("api_response_data", [None, []])
     @patch("apps.grafana_plugin.helpers.client.GrafanaAPIClient.api_get")
     def test_api_call_returns_none_or_list(self, mocked_grafana_api_client_api_get, api_response_data):
@@ -36,69 +35,67 @@ class TestGetUsersPermissions:
 
 
 class TestGetUsers:
-
-    @pytest.mark.parametrize("rbac_is_enabled,api_get_return_value,get_users_permissions_return_value,expected", [
-        # RBAC is enabled - permissions are returned
-        (
-            True,
-            [
-                {"userId": 1, "foo": "bar"},
-                {"userId": 2, "foo": "baz"},
-            ],
-            {
-                "1": [
-                    {"action": "grafana-oncall-app.alert-groups:read"},
-                    {"action": "grafana-oncall-app.alert-groups:write"},
+    @pytest.mark.parametrize(
+        "rbac_is_enabled,api_get_return_value,get_users_permissions_return_value,expected",
+        [
+            # RBAC is enabled - permissions are returned
+            (
+                True,
+                [
+                    {"userId": 1, "foo": "bar"},
+                    {"userId": 2, "foo": "baz"},
                 ],
-            },
-            [
                 {
-                    "userId": 1,
-                    "foo": "bar",
-                    "permissions": [
+                    "1": [
                         {"action": "grafana-oncall-app.alert-groups:read"},
                         {"action": "grafana-oncall-app.alert-groups:write"},
-                    ]
+                    ],
                 },
-                {
-                    "userId": 2,
-                    "foo": "baz",
-                    "permissions": [],
-                },
-            ],
-        ),
-        # RBAC is enabled - permissions endpoint returns no permissions (ex. HTTP 500)
-        (
-            True,
-            [
-                {"userId": 1, "foo": "bar"},
-                {"userId": 2, "foo": "baz"},
-            ],
-            None,
-            [],
-        ),
-        # RBAC is not enabled - we don't fetch permissions (hence don't care about its response)
-        (
-            False,
-            [
-                {"userId": 1, "foo": "bar"},
-                {"userId": 2, "foo": "baz"},
-            ],
-            None,
-            [
-                {
-                    "userId": 1,
-                    "foo": "bar",
-                    "permissions": []
-                },
-                {
-                    "userId": 2,
-                    "foo": "baz",
-                    "permissions": [],
-                },
-            ],
-        ),
-    ])
+                [
+                    {
+                        "userId": 1,
+                        "foo": "bar",
+                        "permissions": [
+                            {"action": "grafana-oncall-app.alert-groups:read"},
+                            {"action": "grafana-oncall-app.alert-groups:write"},
+                        ],
+                    },
+                    {
+                        "userId": 2,
+                        "foo": "baz",
+                        "permissions": [],
+                    },
+                ],
+            ),
+            # RBAC is enabled - permissions endpoint returns no permissions (ex. HTTP 500)
+            (
+                True,
+                [
+                    {"userId": 1, "foo": "bar"},
+                    {"userId": 2, "foo": "baz"},
+                ],
+                None,
+                [],
+            ),
+            # RBAC is not enabled - we don't fetch permissions (hence don't care about its response)
+            (
+                False,
+                [
+                    {"userId": 1, "foo": "bar"},
+                    {"userId": 2, "foo": "baz"},
+                ],
+                None,
+                [
+                    {"userId": 1, "foo": "bar", "permissions": []},
+                    {
+                        "userId": 2,
+                        "foo": "baz",
+                        "permissions": [],
+                    },
+                ],
+            ),
+        ],
+    )
     @patch("apps.grafana_plugin.helpers.client.GrafanaAPIClient.api_get")
     @patch("apps.grafana_plugin.helpers.client.GrafanaAPIClient.get_users_permissions")
     def test_it_returns_none_if_permissions_call_returns_none(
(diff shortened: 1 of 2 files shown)


engine/apps/grafana_plugin/helpers/client.py
```diff
@@ -356,6 +356,10 @@ def _feature_is_enabled_via_enable_key(
 
     def _feature_toggle_is_enabled(self, instance_info: GCOMInstanceInfo, feature_name: str) -> bool:
         """
+        NOTE: there are some edge-cases to consider if you decide to use this method
+        1. If a feature flag is enabled outside of the Admin UI (ex. via deployment_tools), this endpoint
+        will not return the ACTUAL runtime value that the Grafana stack itself is using/respecting
+
```
joeyorlando, line 362: I'm honestly considering getting rid of this client method (+ all the logic to handle the various feature toggle edge cases 😆)..

We got bit here because the `accessControlOnCall` feature toggle was "enabled" for all stacks in production, but this was done via `grafana/deployment_tools`. **However**, GCOM apparently doesn't return the value of this feature toggle if it is set in this manner (so we were basically considering RBAC as disabled for almost all stacks unless they had this feature flag value explicitly set in the Grafana Admin UI)

engine/apps/grafana_plugin/helpers/client.py
```diff
@@ -409,10 +404,17 @@ def get_instances(self, query: str, page_size=None):
...
+    def _is_stack_in_certain_state(self, stack_id: str, state: str) -> bool:
+        instance_info = self.get_instance_info(stack_id)
+        if not instance_info:
+            return False
+        return instance_info.get("status") == state
+
     def is_stack_deleted(self, stack_id: str) -> bool:
-        url = f"instances?includeDeleted=true&id={stack_id}"
-        instance_infos, _ = self.api_get(url)
-        return instance_infos["items"] and instance_infos["items"][0].get("status") == self.STACK_STATUS_DELETED
```
joeyorlando, line 415: refactoring to use `GET /instances/{stack_id}` like we were already doing in `is_rbac_enabled_for_stack` (`status` attribute is also by the underlying endpoint used there)

engine/apps/user_management/sync.py
```diff
@@ -30,14 +30,25 @@ def _sync_organization(organization: Organization) -> None:
...
         gcom_client = GcomAPIClient(settings.GRAFANA_COM_ADMIN_API_TOKEN)
-        rbac_is_enabled = gcom_client.is_rbac_enabled_for_stack(organization.stack_id)
+
+        # the stack MUST be active for this check.. if it is in any other state
+        # the Grafana API risks returning an HTTP 200 but the actual permissions data that is
+        # synced later on will be empty (and we'd erase all RBAC permissions stored in OnCall)
+        if gcom_client.is_stack_active(stack_id):
+            rbac_is_enabled = grafana_api_client.is_rbac_enabled_for_organization()
+        else:
+            rbac_is_enabled = organization.is_rbac_permissions_enabled
```
joeyorlando, line 51: main change

engine/apps/user_management/tests/test_sync.py
```diff
@@ -399,38 +416,39 @@ def test_sync_organization_is_rbac_permissions_enabled_cloud(
...
-                            ),
-                        ):
-                            with patch.object(
-                                GrafanaAPIClient,
-                                "get_grafana_labels_plugin_settings",
-                                return_value=(
-                                    {"enabled": True, "jsonData": {}},
-                                    None,
-                                ),
-                            ):
```
joeyorlando, line 423: I think I'll also refactor the tests in this file to use a fixture, rather than 9 nested context managers 😆

engine/apps/user_management/tests/test_sync.py
```diff
@@ -20,6 +21,64 @@
...
+            None,
+        )
+        mock_client_instance.get_grafana_labels_plugin_settings.return_value = (
+            {"enabled": True, "jsonData": {}},
+            None,
+        )
+        mock_client_instance.check_token.return_value = (None, {"connected": True})
+        mock_client_instance.is_rbac_enabled_for_organization.return_value = is_rbac_enabled_for_organization
+
+        yield mock_grafana_api_client
```
joeyorlando, line 79: most of the changes in this file are related to this..

we had a lot of duplication around mocking out responses on the `GrafanaAPIClient` (plus several instances where we had like 8 nested context managers 😆 ).. this gets rid of that

engine/apps/user_management/models/organization.py
```diff
@@ -344,6 +345,22 @@ def get_notifiable_direct_paging_integrations(self) -> "RelatedManager['AlertRec
...
+        the permissions search endpoint, this will effectively turn on RBAC for all orgs.. soo instead lets
+        slowly turn it on via the logic here
+        """
+        cloud_rbac_rollout_percentage = settings.CLOUD_RBAC_ROLLOUT_PERCENTAGE
+
+        # if rbac permissions are already enabled for the org, they're "grandfathered" in
+        if not cloud_rbac_rollout_percentage or self.is_rbac_permissions_enabled:
+            return True
+        return self.id <= math.floor(Organization.objects.count() * cloud_rbac_rollout_percentage)
+
```
joeyorlando, line 363: cc: @gamab

joeyorlando: this will eventually be removed once we're comfortable rolling this out to all stacks in cloud (ie. we'll slowly ramp up the value of `CLOUD_RBAC_ROLLOUT_PERCENTAGE`)

gamab (commented):
LGTM. As discussed with you, I left a couple of questions.

Side note: what happens if the search endpoint returns an error (!=200) do we remove the users permissions?

engine/apps/grafana_plugin/helpers/client.py
mderynck: I'm not sure if there are other implications but in principle I agree with getting rid of this.  GCOM only knows about overrides, the "default" state is in hosted-grafana config so that's why we weren't getting it returned here.

engine/apps/grafana_plugin/helpers/client.py
mderynck: is_stack_deleted needs to keep query parameter for includeDeleted=true otherwise it will not find anything

engine/apps/grafana_plugin/helpers/client.py
mderynck: Digging around a bit `accessControlOnCall` is available by asking the grafana instance instead on `/api/frontend/settings`

engine/apps/user_management/sync.py
```diff
@@ -28,16 +28,28 @@ def sync_organization(organization: Organization) -> None:
...
     if settings.LICENSE == settings.CLOUD_LICENSE_NAME:
+        # We cannot simply rely on the HEAD call in cloud because if an instance is not active
+        # the grafana gateway will still return 200 for the HEAD request.
+        stack_id = organization.stack_id
         gcom_client = GcomAPIClient(settings.GRAFANA_COM_ADMIN_API_TOKEN)
-        rbac_is_enabled = gcom_client.is_rbac_enabled_for_stack(organization.stack_id)
+
+        if not organization.should_be_considered_for_rbac_permissioning():
+            rbac_is_enabled = False
+        elif gcom_client.is_stack_active(stack_id):
```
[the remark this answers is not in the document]
joeyorlando: as discussed, this is handled further down (maybe an improvement/refactor for a later date)

engine/apps/grafana_plugin/helpers/client.py
joeyorlando: good point, reverted change to `GcomAPIClient.is_stack_deleted` for now 👍 

(tested `GET /instances/<stackid>` call with `?includeDeleted=true` and it returns the same results as calling `GET /instances` with that query param.. won't include it in this PR as there are other calls to `GcomAPIClient.get_instance_info` and I'm not 100% sure if there would be unintended consequences elsewhere)

mderynck (approved):

PR #4280, opened 2024-04-25
Tag component tweaks to get rid of hardcoded tag variables

teodosii:
# What this PR does

- Removed the usage of `var(--` within the Tag component to help get rid of the vars file once we fully migrate to emotion
- Added few other display tweaks and migrated a few stylesheets to emotion


grafana-plugin/src/styles/utils.styles.ts
```diff
@@ -24,6 +29,34 @@ export const getUtilStyles = (_theme: GrafanaTheme2) => {
   };
 };
 
+export function getLabelBackgroundTextColorObject(
```
teodosii, line 32: This part is actually the core of `Badge` component in grafana but changing `Tag` to be replaced to `Badge` in all our usecases would have taken a lot more time, and instead I preferred this approach where we rely on tinycolor to generate these (same as grafana)

maskin25: is there any helper in `@grafana/ui` package that returns such object? I think it could be there...

maskin25: I found several helpers: `getTagColorsFromName, getTextColorForBackground, getTextColorForAlphaBackground` etc, maybe the can be used here

grafana-plugin/src/styles/utils.styles.ts
```diff
@@ -39,5 +72,6 @@ export const bem = (...args: string[]) =>
 export enum COLORS {
   ALWAYS_GREY = '#ccccdc',
   GRAY_8 = '#595959',
+  GRAY_9 = '#434343',
```
maskin25, line 75: can we get such color values from the `theme`?

grafana-plugin/src/styles/utils.styles.ts
teodosii: Neither of them is closer to what this does :(

brojd (approved):

maskin25 (approved):

PR #4281, opened 2024-04-25
GCal autogenerated shift swap requests - don't recreate if one was previously created and deleted

joeyorlando:
# What this PR does

Addresses two issues:
- addresses an internal feature request ([in Slack](https://raintank-corp.slack.com/archives/C03KS498VGV/p1713550543916289?thread_ts=1713546008.831749&cid=C03KS498VGV)) to not have a new shift swap request auto-generated if one previously was, but than the user decided to delete that SSR
- when disconnecting the GCal integration from your user, I've seen cases where Google will return HTTP 400 as such:
  ```
  {'error': 'invalid_token', 'error_description': 'Token expired or revoked'}
  ```
  I can't seem to find detailed documentation on the revoke endpoint (`GET https://accounts.google.com/o/oauth2/revoke?token=<token>`) to try and better understand the possible `error` values.. but I think our best bet here is to just continue forward w/ `user.finish_google_oauth2_disconnection_flow()` (which deletes the `GoogleOAuth2User` associated with the user and sets `user.google_calendar_settings = None`)

## Checklist


matiasb (approved):
+1!

Issue #4272, opened 2024-04-24
Display human readable  time ranges (Add support for relative time periods)


maskin25:
For time based  AG  filters like Started At, Resolved At we need to display human readable values: Last 24 hours, Last 2 days, Last 7 days instead of absolute time ranges to improve UX

![Image](https://github.com/grafana/oncall/assets/3278022/250e42b4-b613-4efd-b4ca-07915fef566c)

PR #4288, opened 2024-04-26
Display human readable time ranges in AG filters

maskin25:
# What this PR does

Display human readable time ranges in AG filters

Closes https://github.com/grafana/oncall/issues/4272

## Checklist


grafana-plugin/src/containers/RemoteFilters/RemoteFilters.tsx
```diff
@@ -316,19 +317,37 @@ class _RemoteFilters extends Component<RemoteFiltersProps, RemoteFiltersState> {
...
+            value = {
+              from: moment(fromString + 'Z'),
+              to: moment(toString + 'Z'),
+              raw: {
+                from: fromString,
+                to: toString,
+              },
+            };
+          }
+        }
```
brojd, line 346: Can we extract this part to the helper function like `getValueForDateRangeFilterType`, place it in RemoteFilters.helpers and have it unit tested?

Also seems you could create:
```
const raw = {
                from: fromString,
                to: toString,
              }
```

and reuse it in both conditions to avoid duplication

grafana-plugin/src/models/filters/filters.helpers.ts
```diff
@@ -7,3 +11,13 @@ export const getApiPathByPage = (page: string) => {
     }[page] || page
   );
 };
+
+export const normalizeFilters = (filters: FiltersValues, filterOptions: FilterOption[]) => {
+  const normalizeFilters = { ...filters };
```
brojd, line 16: This const doesn't represent a function so it shouldn't have a verb in the name. Also we don't do any normalization here, we just create a shallow copy. You could name this const `filtersCopy`

grafana-plugin/src/models/filters/filters.helpers.ts
```diff
@@ -7,3 +11,13 @@ export const getApiPathByPage = (page: string) => {
     }[page] || page
   );
 };
+
+export const normalizeFilters = (filters: FiltersValues, filterOptions: FilterOption[]) => {
```
brojd, line 15: We don't really normalize the object here, we just conditionally map one of its values. How about the name: "convertRelativeToAbsoluteDateForDateRange"?

grafana-plugin/src/utils/datetime.ts
```diff
@@ -1,68 +1,19 @@
...
-  { from: 'now/fQ', to: 'now/fQ', display: 'This fiscal quarter' },
-  { from: 'now/fy', to: 'now', display: 'This fiscal year so far' },
-  { from: 'now/fy', to: 'now/fy', display: 'This fiscal year' },
-];
-
 export const mapOptionToTimeRange = (option: TimeOption, timeZone?: TimeZone): TimeRange => {
   return rangeUtil.convertRawToRange({ from: option.from, to: option.to }, timeZone);
 };
 
 export function convertRelativeToAbsoluteDate(dateRangeString: string) {
```
brojd, line 8: Let's have a unit test for it pls

grafana-plugin/src/models/filters/filters.helpers.ts
maskin25: I think it should have more generic name because the main idea is to cast internal representation in a format that the backend understands, maybe `castFiltersToBackendFormat`?

brojd: Sounds good!

brojd (approved):

Ferril (approved):

PR #4306, opened 2024-05-02
Set a timeout for mobile app incident proxy requests

matiasb:


mderynck (approved):
It might be nice to have some logging similar to outgoing [API calls](https://github.com/grafana/oncall/blob/713c51ce07f865020f692ccfef9447c2e26c11af/engine/apps/grafana_plugin/helpers/client.py#L150) here so we have a bit more visibility into this.

PR #4307, opened 2024-05-03
Remove schedule events duplicate requests and ignore window.resize

maskin25:
# What this PR does

Remove schedule events duplicate requests

## Checklist


grafana-plugin/src/containers/RotationForm/RotationForm.tsx
```diff
@@ -480,13 +480,6 @@ export const RotationForm = observer((props: RotationFormProps) => {
     }
   }, [store.timezoneStore.selectedTimezoneOffset]);
 
-  useEffect(() => {
-    window.addEventListener('resize', onResize);
-    return () => {
-      window.removeEventListener('resize', onResize);
-    };
-  }, []);
-
```
joeyorlando, line 489: was there maybe a valid reason why we called `onHide` on window resize? 🤔

grafana-plugin/src/containers/Rotations/ScheduleFinal.tsx
```diff
@@ -62,10 +61,6 @@ const _ScheduleFinal: FC<ScheduleFinalProps> = observer(
       onShowOverrideForm('new', shiftStart, shiftEnd);
     };
 
-    useEffect(() => {
-      refreshEvents(scheduleId);
-    }, [selectedTimezoneOffset]);
```
joeyorlando, line 67: so this was refetching events if you changed the timezone dropdown and this will no longer be the case? Is there any negative consequences of that?

grafana-plugin/src/containers/RotationForm/RotationForm.tsx
maskin25: https://github.com/grafana/oncall/assets/3278022/bf01ea8b-d615-4c6a-ac75-c2c97ad214c2

I found the only case when window resizng can be an issue (see video attached), in that case user can just resize window back, it's better than just close the modal on any resize

grafana-plugin/src/containers/Rotations/ScheduleFinal.tsx
maskin25: Since your comment I've added refreshing schedules events on TimezoneSelect onChange callback, it settles things

brojd (approved):

grafana-plugin/src/containers/RotationForm/RotationForm.tsx
teodosii: IMO we should add some calculations if we remove the resize handler. When that happens it's not very clear for the user that the resize triggered that behavior, and **we will run into escalations again**. Either we

- close the modal on resizing (no one on prod environment is expected really to open the devtools)
- **OR** figure out boundaries for the modal on each `resize` event, and eventually, if it doesn't fit, you'll still need to close it, so it's basically the first option with 2 steps

teodosii: Either way I think some implementation needs to be added when the browser is being narrowed, otherwise users will run into scenario where they cannot  close/submit the modal anymore due to vertical/horizontal restrains.

teodosii (changes requested):

teodosii (approved):

PR #4325, opened 2024-05-08
Minor query optimizations

matiasb:
Related to some API latency issues we were noticing.

### DIFF
--- a/engine/apps/public_api/serializers/incidents.py
+++ b/engine/apps/public_api/serializers/incidents.py
@@ -1,8 +1,6 @@
-from django.db.models import Prefetch
 from rest_framework import serializers
 
 from apps.alerts.models import AlertGroup
-from apps.telegram.models.message import TelegramMessage
 from common.api_helpers.custom_fields import UserIdField
 from common.api_helpers.mixins import EagerLoadingMixin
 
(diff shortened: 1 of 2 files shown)


engine/apps/public_api/serializers/incidents.py
```diff
@@ -19,14 +17,6 @@ class IncidentSerializer(EagerLoadingMixin, serializers.ModelSerializer):
...
 
     SELECT_RELATED = ["channel", "channel_filter", "slack_message", "channel__organization"]
-    PREFETCH_RELATED = [
-        "alerts",
-        Prefetch(
-            "telegram_messages",
-            TelegramMessage.objects.filter(chat_id__startswith="-", message_type=TelegramMessage.ALERT_GROUP_MESSAGE),
-            to_attr="prefetched_telegram_messages",
-        ),
-    ]
```
matiasb, line 29: We don't use alerts data now, do not prefetch. Stop prefetching telegram messages for permalinks (internal API is not doing that and it seems it is not helping here).

engine/apps/public_api/serializers/incidents.py
```diff
@@ -50,7 +40,7 @@ def get_title(self, obj):
         return obj.web_title_cache
 
     def get_alerts_count(self, obj):
-        return len(obj.alerts.all())
+        return obj.alerts.count()
```
matiasb, line 43: This count query is more efficient than prefetching alerts (loading them all in mem) and count them later.

mderynck (approved):

PR #4232, opened 2024-04-16
Prepare OnCall for Unified Slack App

Konstantinov-Innokentii:
This PR does a bunch of changes to prepare OnCall for Unified Slack App:
1. Install Slack via Chatops-Proxy. This change contains two parts: getting a Slack install link from chatops-proxy ([code](https://github.com/grafana/oncall/pull/4232/files#diff-437a77d49fc04b92d315651b3df5991000b1ab74cf60aabb21aa77cb2823bf52R46)) and receiving a "slack installed" event from chatops-proxy ([code](https://github.com/grafana/oncall/pull/4232/files#diff-976d106f0962be5c1de5e35582193f68435ed0c17f2defd6bd2857bf6e27f65d)). Also it means that OnCall doesn't need to register slack_links anymore when slack is connected/disconnected. These changes are behind UNIFIED_SLACK_APP_ENABLED flag and should be no-op if flag is not enabled.
2. Get rid of Multiregionatily restrictions - instrument all slack interactions with a ProxyMeta - json data telling chatops-proxy where to route the interaction. Note, that it doesn't apply for "Add to resolution notes" message action - it will be handled differently in following PR. 
3. Move all chatops-proxy related stuff from common/oncall-gateway to apps/chatops-proxy 

Minor changes:
1. Remove usage of **CHATOPS_V3** flag. Chatops v3 is already released (It's a refactoring from previous quarter)

PR #4350, opened 2024-05-16
Support slack install multiple response status codes

teodosii:
# What this PR does

- Removed OTEL from `makeRequest`
- Since `makeRequest` only returned the data of the response, I added a `makeRequestRaw` that returns whole response. At the moment of writing this, only `onCallAPI` returns the full response, but that helper is only for typed endpoints.
- Added checks for slack install for 200, 201 and 500


Konstantinov-Innokentii:
Merged to https://github.com/grafana/oncall/pull/4232

PR #4356, opened 2024-05-16
Make type required when creating an escalation policy step

matiasb:
Related to https://github.com/grafana/oncall-private/issues/2689


engine/apps/user_management/tests/test_organization.py
```diff
@@ -303,8 +303,11 @@ def test_should_be_considered_for_rbac_permissioning(
     NUM_ORGS = 5
     settings.CLOUD_RBAC_ROLLOUT_PERCENTAGE = rollout_percentage
 
+    # make sure there are no other orgs around
+    Organization.objects.all().delete()
```
mderynck, line 307: Can we do this in a different way? Scary if someone runs this test connected to the wrong environment.

iskhakov (approved):

PR #4362, opened 2024-05-20
More emotionjs migration changes

teodosii:
# What this PR does

- Get rid of some of the hardcoded css variables defined in `vars.css`, specifically tags-related, but not only
- Migrated a few more components to emotion


grafana-plugin/src/containers/AddResponders/AddResponders.styles.ts
```diff
@@ -0,0 +1,72 @@
...
+      width: 100%;
+      position: relative;
+    `,
+
+    respondersList: css`
+      list-style-type: none;
+      margin-bottom: 20px;
+      width: 100%;
+
+      & > li .hover-button {
```
teodosii, line 16: @joeyorlando I'm migrating more code to emotion and have come across this part, can't really figure out where's `hover-button` coming from as it's not being defined in any of the add responder styles. Do you happen to know where is that coming from? I know you've been working on this stuff

The only `hover-button` I see is part of the integration display within the routes, but that shouldn't be related to add responders in any way

grafana-plugin/src/containers/AddResponders/AddResponders.tsx
```diff
@@ -31,21 +29,24 @@ type Props = {
...
-        <Icon name="external-link-alt" />
-      </HorizontalGroup>
-    </Text>
-  </a>
-);
+const LearnMoreAboutNotificationPoliciesLink: React.FC = () => {
+  const styles = useStyles2(getAddRespondersStyles);
+  return (
+    <a
+      className={cx(styles.learnMoreLink)}
```
brojd, line 36: This is just an example but this comment applies to plenty other similar places - why do we use `cx(styles.someClassName)` instead of just `styles.someClassName`? `cx` is handy in 2 situations: when you want to add multiple classes and/or apply some class conditionally. In case we set a single class without any condition it doesn't make much sense. 

How about we get rid of them in such situations and keep only if it brings benefit?

grafana-plugin/src/styles/utils.styles.ts
```diff
@@ -26,6 +31,37 @@ export const getUtilStyles = (theme: GrafanaTheme2) => {
...
+
+    overflowChild: css`
+      overflow: hidden;
+      text-overflow: ellipsis;
+      display: -webkit-box;
+      white-space: initial;
+      -webkit-line-clamp: 2;
+      -webkit-box-orient: vertical;
+
+      &--line-1 {
```
maskin25, line 43: it looks to much specific, like not a good candidate for reusing from util styles

grafana-plugin/src/styles/utils.styles.ts
```diff
@@ -26,6 +31,37 @@ export const getUtilStyles = (theme: GrafanaTheme2) => {
...
+      }
+    `,
+
+    wordBreakAll: css`
+      word-break: break-all;
+    `,
+
+    flex: css`
+      display: flex !important;
+      flex-direction: row;
```
maskin25, line 58: `row` is already default direction for flex

grafana-plugin/src/styles/utils.styles.ts
```diff
@@ -26,6 +31,37 @@ export const getUtilStyles = (theme: GrafanaTheme2) => {
...
+        -webkit-line-clamp: 3;
+      }
+    `,
+
+    wordBreakAll: css`
+      word-break: break-all;
+    `,
+
+    flex: css`
+      display: flex !important;
```
maskin25, line 57: can we remove `!important` here? it considers a bad practive

brojd (approved):
great job!

grafana-plugin/src/pages/settings/SettingsPage.tsx
```diff
@@ -134,36 +130,45 @@ interface TabsContentProps {
...
 const TabsContent = (props: TabsContentProps) => {
   const { activeTab } = props;
+  const styles = useStyles2(getStyles);
 
   return (
-    <div className={cx('tabs__content')}>
+    <div className={styles.tabsContent}>
       {activeTab === SettingsPageTab.MainSettings.key && (
-        <div className={cx('tab__page')}>
+        <div>
```
maskin25, line 138: do we really need `div`here and below?

grafana-plugin/src/pages/settings/SettingsPage.tsx
```diff
@@ -41,7 +37,7 @@ class Settings extends React.Component<SettingsPageProps, SettingsPageState> {
   }
 
   render() {
-    return <div className={cx('root')}>{this.renderContent()}</div>;
+    return <div>{this.renderContent()}</div>;
```
maskin25, line 40: do we need `div` here?

grafana-plugin/src/pages/integrations/Integrations.tsx
```diff
@@ -458,14 +471,16 @@ class _IntegrationsPage extends React.Component<IntegrationsProps, IntegrationsS
...
-  renderMaintenance(item: ApiSchemas['AlertReceiveChannel']) {
+  renderMaintenance = (item: ApiSchemas['AlertReceiveChannel']) => {
+    const { theme } = this.props;
     const maintenanceMode = item.maintenance_mode;
+    const utilStyles = getUtilStyles(theme);
 
     if (maintenanceMode === MaintenanceMode.Debug || maintenanceMode === MaintenanceMode.Maintenance) {
       return (
-        <div className={cx('u-flex')}>
+        <div className={cx(utilStyles.flex)}>
```
maskin25, line 483:
```suggestion
        <div className={utilStyles.flex}>
```

grafana-plugin/src/pages/integrations/Integrations.styles.tsx
```diff
@@ -0,0 +1,71 @@
...
+      white-space: nowrap;
+      border-left: 2px solid transparent;
+      cursor: pointer;
+      min-width: 84px;
+      display: flex;
+      gap: 8px;
+      flex-direction: row;
+
+      &:hover {
+        background: ${theme.isLight ? 'rgba(244, 245, 245)' : COLORS.GRAY_9};
```
maskin25, line 59: can we use some constant for the light theme like for the dark?

grafana-plugin/src/containers/AddResponders/__snapshots__/AddResponders.test.tsx.snap
```diff
@@ -90,7 +90,7 @@ exports[`AddResponders should properly display the add responders button when hi
 exports[`AddResponders should render properly in create mode 1`] = `
 <div>
   <div
-    class="body"
+    class="css-1si66qn"
```
maskin25, line 93: are such classes persistent?

grafana-plugin/src/pages/settings/SettingsPage.tsx
teodosii: This is existing structure, I'm not changing how the layout was done

grafana-plugin/src/styles/utils.styles.ts
teodosii: Hmm I believe this is needed if you want to overwrite the display. Why is it considered a bad practice in this case? The utils are not too specific (as in `div.className`) and they need to overwrite some properties

grafana-plugin/src/styles/utils.styles.ts
teodosii: Right, but that's just to make sure you overwrite it, you know what I mean? Even though that's the default flex-direction, when you apply this util class you wanna make sure it applies all styles (and overwrites anything else). This is not new code, this was already being used in other places

grafana-plugin/src/styles/utils.styles.ts
teodosii: Hmm, what  do you suggest then? This is used for clipping in a few other places, I put it in here so that's being reused

grafana-plugin/src/pages/integrations/Integrations.tsx
teodosii: Good point, I need to apply regex for these cases too! :+1:

grafana-plugin/src/containers/AddResponders/__snapshots__/AddResponders.test.tsx.snap
teodosii: Didn't notice any real problem with snapshots so far, this one shows here because it has been changed from old CSS to new classes

grafana-plugin/src/styles/utils.styles.ts
teodosii: https://stackoverflow.com/a/3706876 came across this post on stackoverflow mentioning the usage of !important within utils classes

grafana-plugin/src/styles/utils.styles.ts
maskin25: I think utils is for very very common classes like `u-position-relative`, `width100` etc, if we start put everything being reused to utils.css then the we come back to the early 2000s when all the styles were placed in the single file `style.css`

grafana-plugin/src/pages/settings/SettingsPage.tsx
maskin25: I'm almost sure this div was added just to hold `root` class and nothing else, but you are right is not the scope of your task

grafana-plugin/src/styles/utils.styles.ts
teodosii: Can split utils per category then

grafana-plugin/src/containers/Rotations/Rotations.styles.ts
```diff
@@ -0,0 +1,93 @@
...
+import { GrafanaTheme2 } from '@grafana/data';
+
+export const getRotationsStyles = (theme: GrafanaTheme2) => {
+  return {
+    root: css`
+      background: 1px solid ${theme.colors.background.secondary};
+      border: ${theme.colors.border.weak};
+      display: flex;
+      flex-direction: column;
+      border-radius: ${theme.shape.radius.default};
```
hdhoang, line 11: With grafana 9.4 themes, `shape.radius` is not defined.

Could you return to fixed value, thanks.

Issue #2333, opened 2023-06-26
Longer silence periods


msbannert:
### What would you like to see!

Alert Groups can be "silenced" for a limited period up to 24 hours.
It would be nice to have longer periods selectable in the dropdown (1 week, up to 1 month).
Use case: Suppress an alert group  while dev teams are working on a fix, which usually takes some days or weeks.

### Product Area

Alert Flow & Configuration

chencs:
This is quite hacky, but I worked around this by editing and resending the `silence` API request via my browser's network panel with `{delay: $numSeconds}` in the body.

PR #4375, opened 2024-05-22
Allow custom silence period for alert groups + display how much left is for a silenced AG

teodosii:
# What this PR does

https://github.com/grafana/oncall/issues/2333

- Allow custom input for silence period (either by duration or by setting a future date in the calendar, both inputs are syncedd/editable)
- Show how much time is left for a silenced alert group

Closes https://github.com/grafana/oncall/issues/2333


grafana-plugin/src/pages/incidents/parts/IncidentDropdown.tsx
```diff
@@ -55,6 +58,8 @@ function IncidentStatusTag({
   );
 }
 
+const CUSTOM_VALUE_NUM = -100;
```
teodosii, line 61: Just to differentiate between all other values (-1 is taken for forever, all other positive values are matching values)

brojd (approved):

grafana-plugin/src/components/RenderConditionally/RenderConditionally.tsx
```diff
@@ -2,9 +2,20 @@ import React, { FC, ReactNode } from 'react';
...
-export const RenderConditionally: FC<RenderConditionallyProps> = ({ shouldRender, children, backupChildren = null }) =>
-  shouldRender ? <>{children}</> : <>{backupChildren}</>;
+export const RenderConditionally: FC<RenderConditionallyProps> = ({
+  shouldRender,
+  children,
+  render,
+  backupChildren = null,
+}) => {
+  if (render) {
+    return shouldRender ? <>{render()}</> : null;
```
brojd, line 17: just thinking if in this case we shouldn't use backupChildren instead of null here

grafana-plugin/src/pages/incident/Incident.tsx
```diff
@@ -632,11 +653,15 @@ class _IncidentPage extends React.Component<IncidentPageProps, IncidentPageState
     };
   };
 
-  getSilenceClickHandler = (incidentId: ApiSchemas['AlertGroup']['pk']) => {
+  getSilenceClickHandler = (incident: ApiSchemas['AlertGroup']) => {
     const { store } = this.props;
 
     return (value: number) => {
```
brojd, line 659: can we mark this fn as `async` as it is returning Promises?

grafana-plugin/src/pages/incident/Incident.tsx
```diff
@@ -632,11 +653,15 @@ class _IncidentPage extends React.Component<IncidentPageProps, IncidentPageState
...
-      return store.alertGroupStore.doIncidentAction(incidentId, AlertAction.Silence, value);
+      if (value === -100) {
```
brojd, line 660: Can we use the const that you defined here?

grafana-plugin/src/pages/incident/Incident.tsx
teodosii: Had to add the promise because the other caller was expecting a promise return instead of void

PR #4381, opened 2024-05-22
Fix order filtering when executing notify all/group steps from snapshot

matiasb:
Fixes https://github.com/grafana/oncall-private/issues/2708


vadimkerr (approved):
LGTM

PR #4389, opened 2024-05-23
Drone -> GitHub Actions migration

joeyorlando:
# What this PR does

Related to https://github.com/grafana/oncall-private/issues/2692

This PR simply deduplicates a lot of steps in our `linting-and-tests.yml` GitHub Actions workflow. This will make it much easier in `grafana/oncall-private` to be able to reuse some of these composable building blocks.


.github/workflows/linting-and-tests.yml
```diff
@@ -263,57 +208,45 @@ jobs:
...
+        uses: actions/checkout@v3
+      - name: Setup Python
+        uses: ./.github/actions/setup-python
         with:
-          python-version: "3.11.4"
-          cache: "pip"
-          cache-dependency-path: |
-            engine/requirements.txt
-            engine/requirements-dev.txt
+          python-requirements-paths: "engine/requirements.txt engine/requirements-dev.txt"
```
matiasb, line 216: If this is the default value, is it needed here? (same below)

.github/workflows/linting/backend-migrations.yml
matiasb, line 1: The plan is to move the db migrations lint job here?

matiasb (approved):
LGTM

.github/workflows/linting-and-tests.yml
joeyorlando: good catch, thanks 👍

PR #4392, opened 2024-05-23
Basic working backend plugin proxy

mderynck:
This is a basic working version of the backend proxy to talk to the OnCall backend.

Ignore my hacky UI changes for now, they are just for debugging and I encountered some issues with the useBackendPlugin setting.


grafana-plugin/pkg/plugin/resources.go
```diff
@@ -2,36 +2,133 @@ package plugin
...
+		return
+	}
+
+	err = SetXGrafanaContextHeader(user, proxyReq)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	if proxyMethod == "POST" || proxyMethod == "PUT" {
```
matiasb, line 210: Maybe we should also include `PATCH` here?

engine/apps/auth_token/auth.py
```diff
@@ -149,16 +149,17 @@ def _get_user(request: Request, organization: Organization) -> User:
...
+            if "UserId" in context or "UserID" in context:
+                try:
+                    user_id = context["UserId"]
+                except KeyError:
+                    user_id = context["UserID"]
+                return organization.users.get(user_id=user_id)
+            elif "Login" in context:
+                return organization.users.get(username=context["Login"])
+            else:
+                raise exceptions.AuthenticationFailed("Grafana context must specify a User or UserID.")
```
matiasb, line 162: Would this be equivalent to the above?
```
user_id = context.get("UserId", context.get("UserID"))
if user_id is not None:
    return organization.users.get(user_id=user_id)
elif "Login" in context:
    ...
```

mderynck: I'm not sure if the Login part will stay now that we have ID available from the backend pluigin.

grafana-plugin/pkg/plugin/resources.go
```diff
@@ -2,36 +2,231 @@ package plugin
...
+			return 0, err
+		}
+		return int(id), nil
+	}
+
+	return 0, fmt.Errorf("User %s not found", user.Login)
+}
+
+func (a *App) SetPermissionsHeader(userID int, settings OnCallPluginSettings, req *http.Request) error {
+	permissionsURL, err := url.JoinPath(settings.GrafanaURL, fmt.Sprintf("api/access-control/users/%d/permissions", userID))
```
brojd, line 129: Is it working for you? Locally I needed to switch to 
`permissionsURL, err := url.JoinPath(settings.GrafanaURL, "api/access-control/user/permissions")` endpoint documented [here](https://grafana.com/docs/grafana/latest/developers/http_api/access_control/#list-your-permissions), otherwise I'm getting 404 (userId is correctly populated)

dev/grafana/provisioning/plugins/grafana-oncall-app-provisioning.yaml
```diff
@@ -5,5 +5,5 @@ apps:
     jsonData:
       stackId: 5
       orgId: 100
-      onCallApiUrl: http://oncall-dev-engine:8080
-      useBackendPlugin: false
+      onCallApiUrl: http://host.docker.internal:8080
```
brojd, line 8: Fyi this one is not working by default if you run the project with Tilt

grafana-plugin/pkg/plugin/resources.go
```diff
@@ -2,36 +2,231 @@ package plugin
...
+			IsAnonymous: true,
+		}
+	} else {
+		xGrafanaContext = XGrafanaContextJSONData{
+			ID:          userID,
+			IsAnonymous: false,
+			Name:        user.Name,
+			Login:       user.Login,
+			Email:       user.Email,
+			Role:        user.Role,
```
matiasb, line 62: Should/Can we get information about teams membership of the current user?

grafana-plugin/pkg/plugin/resources.go
mderynck: Yes but I am running enterprise with RBAC locally. We can fallback to use the role when that is not enabled.

dev/grafana/provisioning/plugins/grafana-oncall-app-provisioning.yaml
mderynck: Yeah I'll remove this before it gets merged.

grafana-plugin/pkg/plugin/resources.go
mderynck: We should be able to get it as another API call

dev/grafana/provisioning/plugins/grafana-oncall-app-provisioning.yaml
mderynck: We might want to rethink how this file gets used as it gets loaded everytime the grafana instance is restarted.  This is another cause of confusion in helm and dev setups when it does not agree with config done from within the plugin itself.

grafana-plugin/src/state/plugin/plugin.ts
```diff
@@ -274,30 +276,31 @@ export class PluginState {
...
-      });
-    } catch (e) {
-      return this.getHumanReadableErrorFromGrafanaProvisioningError(
-        e,
-        onCallApiUrl,
-        errorMsgVerb,
-        onCallApiUrlIsConfiguredThroughEnvVar
-      );
-    }
+    // try {
```
mderynck, line 279: Hack to workaround existing code to not overwrite what is configured by backend plugin.

Issue #4371, opened 2024-05-21
Escalation chain step not rendered on integrations page


mderynck:
### What went wrong?

**What happened**:
- Escalation step when viewing on integrations page does not show selected user group

![Screenshot from 2024-05-21 13-33-04](https://github.com/grafana/oncall/assets/28077050/9a600317-68c1-41a5-9b47-b4d7d6458f95)

![Screenshot from 2024-05-21 13-32-53](https://github.com/grafana/oncall/assets/28077050/aa1808db-7642-42a1-b281-7ad8967e527c)

**What did you expect to happen**:
- User group selected for escalation step should be visible on integrations page

### How do we reproduce it?

1. Create an escalation chain with a Notify Slack User Group step and select a user group
2. Go to an integration under the route and select the escalation chain
3. The user group will not be displayed, instead it says to select one, also confusing because it is read only here.

### Grafana OnCall Version

v1.5.0

### Product Area

Alert Flow & Configuration

### Grafana OnCall Platform?

I use Grafana Cloud

PR #4394, opened 2024-05-24
Fixed UserGroup selectors not showing the set value

teodosii:
# What this PR does

Fix for https://github.com/grafana/oncall/issues/4371

- Fixed UserGroups' selectors not showing up the stored value - this behavior was found in Schedule Form and Escalation Chains

Closes https://github.com/grafana/oncall/issues/4371


grafana-plugin/src/models/user_group/user_group.ts
```diff
@@ -46,6 +46,21 @@ export class UserGroupStore extends BaseStore {
...
+  async fetchItemById(id: string) {
+    const response: UserGroup[] = (await makeRequest(`${this.path}`, {
+      params: { id },
+    })) as UserGroup[];
+
+    runInAction(() => {
+      this.items = {
+        ...this.items,
+        // backend returns an array
+        [id]: Array.isArray(response) ? (response.length > 0 ? response[0] : undefined) : response,
```
teodosii, line 59: For some reason, backend returns an array when fetching an individual response instead of an object. Added this so that if backend will decide to change it in the near future, it won't cause a npe on the frontend

brojd (approved):

PR #4396, opened 2024-05-24
Setup team uniqueness at DB level

matiasb:
Avoid potential duplicated teams. Also it should help ensuring `sync_organization` idempotency.


engine/apps/user_management/migrations/0022_alter_team_unique_together.py
```diff
@@ -0,0 +1,60 @@
...
+            team_id=team_id,
+        ).order_by("id").first()
+
+        # migrate resources associated to duplicated entries
+        duplicated_teams = Team.objects.filter(
+            organization_id=organization_id,
+            team_id=team_id,
+        ).exclude(id=first_team.id)
+
+        for team in duplicated_teams:
```
mderynck, line 31: Do we need to handle user.current_team values here?

matiasb: Hm.. I think we are not using `current_team` anymore? I can do the update here in any case (otherwise `current_team` will be reset to `None` for those users with a deleted current_team, which afaict shouldn't break anything either)

mderynck (approved):

PR #4406, opened 2024-05-27
Refactoring plugin auth getting user

matiasb:


engine/apps/grafana_plugin/views/status.py
```diff
@@ -64,14 +65,14 @@ def post(self, request: Request) -> Response:
             )[0].boolean_value
 
         # If user is not present in OnCall database, set token_ok to False, which will trigger reinstall
-        if not request.user:
-            logger.info(f"Status - user not found org={organization.pk} " f"setting token_status to PENDING")
-            token_ok = False
-            organization.api_token_status = Organization.API_TOKEN_STATUS_PENDING
-            organization.save(update_fields=["api_token_status"])
+        # if not request.user:
```
matiasb, line 68: User does not exist but it should be created on-the-fly in a later step.

engine/apps/grafana_plugin/views/status.py
```diff
@@ -64,14 +65,14 @@ def post(self, request: Request) -> Response:
...
+        #     logger.info(f"Status - user not found org={organization.pk} " f"setting token_status to PENDING")
+        #     token_ok = False
+        #     organization.api_token_status = Organization.API_TOKEN_STATUS_PENDING
+        #     organization.save(update_fields=["api_token_status"])
 
-        # Start task to refresh organization data in OnCall database with Grafana
-        plugin_sync_organization_async.apply_async((organization.pk,))
+        # # Start task to refresh organization data in OnCall database with Grafana
+        # plugin_sync_organization_async.apply_async((organization.pk,))
```
matiasb, line 75: Disable sync to check the user is created during auth.

engine/apps/user_management/sync.py
```diff
@@ -132,6 +134,17 @@ def _sync_grafana_incident_plugin(organization: Organization, grafana_api_client
         )
 
 
+def sync_user_from_request(organization: Organization, request: Request) -> User:
```
matiasb, line 137: This may be moved somewhere else, and it should eventually create the user from the information in the headers as provided by the backend plugin.

PR #4420, opened 2024-05-29
Oncall initialization frontend cleanup

brojd:
https://github.com/grafana/oncall-private/issues/2721

## Checklist


mderynck (approved):

grafana-plugin/src/plugin/GrafanaPluginRootPage.tsx
```diff
@@ -54,9 +55,11 @@ export const GrafanaPluginRootPage = (props: AppRootProps) => {
       }}
     >
       {() => (
-        <Provider store={rootStore}>
-          <PluginSetup InitializedComponent={Root} {...props} />
-        </Provider>
+        <RenderConditionally shouldRender={isInitialized}>
+          <Provider store={rootStore}>
```
teodosii, line 59: This still "renders" within React if we use it like this, while being hidden on the UI. Could there be the case that the insides of `Provider` depend on data that we haven't initialized yet? I'm thinking this could be a cause of a potential issue

If so, it'd be better to run `RenderConditionally` with the other prop to lazy render the content

```jsx
<RenderConditionally shouldRender={...} render={() => <Provider... }/>
```

grafana-plugin/src/utils/hooks.tsx
```diff
@@ -138,3 +141,38 @@ export const useOnMount = (callback: () => void) => {
...
+    // trigger users sync
+    try {
+      await makeRequest(`/plugin/status`, {
+        method: 'POST',
+      });
+    } catch (_err) {
+      await install();
+    }
+
+    setIsInitialized(true);
```
teodosii, line 170: What if the install fails? What's the fallback now? Shouldn't we attempt to try it again at least?

brojd: I don't think it makes sense to have second attempt, if it failed once, it would fail again. Anyway, this is kinda temporary solution because we hope to move this part to backend proxy too so that frontend doesn't call install explicitly

teodosii (approved):

PR #4457, opened 2024-06-04
User notifications bundle

Ferril:
# What this PR does
This PR adds two new models: UserNotificationBundle and BundledNotification (proposals for naming are welcome).

`UserNotificationBundle` manages the information about last notification time and scheduled notification task for bundled notifications. It is unique per user + notification_channel + notification importance.

`BundledNotification` contains notification policy and alert group, that triggered the notification. The BundledNotification instance is created in `notify_user_task` for every notification, that should be bundled, and is attached to UserNotificationBundle by ForeignKey connection.

How it works:
If the user was notified recently (within the last two minutes) by the current notification channel, and this channel is bundlable, BundledNotification instance will be created and attached to the UserNotificationBundle instance, and `send_bundled_notification` task will be scheduled to execute in 2 min.
In `send_bundled_notification` task we get all BundledNotification attached to the current UserNotificationBundle instance, check if alert groups are still active and if there is only one notification - perform regular notification by calling `perform_notification` task, otherwise call "notify_by_<channel>_bundle" method for the current notification channel.

PR with method to send notification bundle by SMS - https://github.com/grafana/oncall/pull/4624

**This feature is disabled by default by feature flag. Public docs will be added in a separate PR with enabling this feature.**
related to https://github.com/grafana/oncall-private/issues/2712

## Checklist


engine/apps/alerts/tasks/notify_user.py
```diff
@@ -418,17 +487,119 @@ def perform_notification(log_record_pk):
...
+        existing_notification_policies = set(
+            UserNotificationPolicy.objects.filter(id__in=notification_policy_ids).values_list("id", flat=True)
+        )
+        current_notification_data = []
+        created_lod_records = []
+
+        for entry in user_notification_bundle.notification_data:
+            if entry["alert_group_id"] not in active_alert_groups:
+                task_logger.info(
+                    f"alert_group {entry['alert_group_id']} is not active or doesn't exist, skip notification"
```
mderynck, line 530: 👍 nice! this will come in handy for those users that are deleting groups in the API before we handle them.

engine/apps/alerts/tasks/notify_user.py
```diff
@@ -170,51 +194,97 @@ def notify_user_task(
...
-                update_metrics_for_user.apply_async((user.id,))
+                    user_notification_bundle = UserNotificationBundle.objects.filter(
+                        pk=user_notification_bundle.pk
+                    ).select_for_update()[0]
+                    # check if notification needs to be bundled
+                    if (
+                        notification_policy.notify_by in UserNotificationBundle.NOTIFICATION_CHANNELS_TO_BUNDLE
+                        and user_notification_bundle.notified_recently()
+                    ):
+                        user_notification_bundle.notification_data.append(
```
mderynck, line 209: We might want some upper limit on what can be appended here.  It's an edge case but at some point the bundle has so much in it that it's not useful it may save resources to just discard new entries at beyond that point.

engine/apps/alerts/tasks/notify_user.py
```diff
@@ -170,51 +194,97 @@ def notify_user_task(
...
-                and previous_notification_policy_pk is None
-                and not user.personal_log_records.filter(
-                    type=UserNotificationPolicyLogRecord.TYPE_PERSONAL_NOTIFICATION_TRIGGERED,
-                    alert_group_id=alert_group_pk,
-                ).exists()
-            ):
-                update_metrics_for_user.apply_async((user.id,))
+                    user_notification_bundle = UserNotificationBundle.objects.filter(
+                        pk=user_notification_bundle.pk
+                    ).select_for_update()[0]
```
mderynck, line 203: I have some concerns around both this task and the send_bundled_notification locking the bundle, in a flood scenario there will be a lot of contention for that record and each save gets bigger as we append to notification_data.  We can test it out before we make things more complicated though.

engine/apps/alerts/tasks/notify_user.py
```diff
@@ -418,17 +487,119 @@ def perform_notification(log_record_pk):
...
+        # perform regular notification
+        log_record = created_lod_records[0]
+        task_logger.info(
+            f"there is only one alert group in bundled notification, perform regular notification. "
+            f"alert_group {log_record.alert_group_id}"
+        )
+        transaction.on_commit(partial(create_perform_notification_task, log_record.pk, log_record.alert_group_id))
+    else:
+        task_logger.info(f"perform notification for alert groups with ids: {active_alert_groups}")
+        if user_notification_bundle.notification_channel == UserNotificationPolicy.NotificationChannel.SMS:
```
mderynck, line 587: Will the sms send happen as part of this task or a separate task? I'm curious to see how it works in terms of if that needs to be completed successfully before this transaction commits and clears the notification_data.

engine/apps/alerts/tasks/notify_user.py
Ferril: In a flood scenario send_bundled_notification will be called once in ~2min. It's not so often, I think, probably it wouldn't be a problem, need to test it 🤔 But I agree that we need to add some limitation for notification_data

engine/apps/alerts/tasks/notify_user.py
Ferril: I think it will be done in a separate task using data collected in `current_notification_data`. In this case we don't need to rework retry logic.
I was thinking about making the second part of the current task similar to the `perform_notification` task, but calling methods of notification channels async

engine/apps/alerts/tasks/notify_user.py
```diff
@@ -170,16 +195,57 @@ def notify_user_task(
...
-                        type=UserNotificationPolicyLogRecord.TYPE_PERSONAL_NOTIFICATION_TRIGGERED,
-                        notification_policy=notification_policy,
-                        alert_group=alert_group,
-                        reason=reason,
-                        slack_prevent_posting=prevent_posting_to_thread,
-                        notification_step=notification_policy.step,
-                        notification_channel=notification_policy.notify_by,
-                    )
+                    # todo: feature flag
+                    if notification_policy.notify_by in UserNotificationBundle.NOTIFICATION_CHANNELS_TO_BUNDLE:
```
Konstantinov-Innokentii, line 199: nit: worth to make is_bundlable method for user_notification_policy

engine/apps/alerts/tasks/notify_user.py
```diff
@@ -418,17 +473,138 @@ def perform_notification(log_record_pk):
...
+                acknowledged=False,
+                silenced=False,
+            ).values_list("id", flat=True)
+        )
+        created_lod_records = []
+        skip_notification_ids = []
+        perform_notifications = []
+
+        # update metric
+        existing_logs = set(
```
Konstantinov-Innokentii, line 513: could metrics be updated outside of transaction to make it shorter?

engine/apps/alerts/tasks/notify_user.py
```diff
@@ -170,16 +195,57 @@ def notify_user_task(
...
+                    if notification_policy.notify_by in UserNotificationBundle.NOTIFICATION_CHANNELS_TO_BUNDLE:
+                        user_notification_bundle, _ = UserNotificationBundle.objects.get_or_create(
+                            user=user, important=important, notification_channel=notification_policy.notify_by
+                        )
+                        user_notification_bundle = UserNotificationBundle.objects.filter(
+                            pk=user_notification_bundle.pk
+                        ).select_for_update()[0]
+                        # check if notification needs to be bundled
+                        if user_notification_bundle.notified_recently():
+                            user_notification_bundle.attach_notification(alert_group, notification_policy)
```
Konstantinov-Innokentii, line 208: nit: it's probably append_notification

engine/apps/alerts/tasks/notify_user.py
```diff
@@ -418,17 +468,144 @@ def perform_notification(log_record_pk):
...
+            return
 
-    from apps.base.models import UserNotificationPolicyLogRecord
+        notifications = user_notification_bundle.notifications.filter(bundle_uuid__isnull=True)
+
+        alert_group_ids = [i.alert_group_id for i in notifications]
+        # get active alert groups to notify about
+        active_alert_groups = set(
+            AlertGroup.objects.filter(
+                id__in=alert_group_ids,
```
Konstantinov-Innokentii, line 504: Is it using index for query?

engine/apps/alerts/tasks/notify_user.py
```diff
@@ -418,17 +468,144 @@ def perform_notification(log_record_pk):
...
+            ).values_list("id", flat=True)
+        )
+        created_lod_records = []
+        skip_notification_ids = []
+        perform_notifications = []
 
-    task_logger.debug(f"LOG RECORD PK: {log_record_pk}")
-    task_logger.debug(f"LOG RECORD LAST: {UserNotificationPolicyLogRecord.objects.last()}")
+        # create logs
+        for notification in notifications:
```
Konstantinov-Innokentii, line 515: Maybe it could be once for cycle, since we already iterated through notifications in line 500

engine/apps/alerts/models/user_has_notification.py
```diff
@@ -16,3 +16,7 @@ class UserHasNotification(models.Model):
 
     class Meta:
         unique_together = ("user", "alert_group")
+
+    def update_active_task_id(self, task_id):
+        self.active_notification_policy_id = task_id
```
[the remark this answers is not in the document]
Ferril: It's a part of the old model `UserHasNotification` and wasn't added or changed in this PR. I just moved these lines in separate method to make `notify_user_task` more readable. Makes sense to add the comment though.

engine/apps/alerts/tasks/notify_user.py
```diff
@@ -418,17 +468,144 @@ def perform_notification(log_record_pk):
...
+                f"send_bundled_notification: notification_task_id mismatch. "
+                f"Duplication or non-active notification triggered. "
+                f"Active: {user_notification_bundle.notification_task_id}"
+            )
+            return
 
-    from apps.base.models import UserNotificationPolicyLogRecord
+        notifications = user_notification_bundle.notifications.filter(bundle_uuid__isnull=True)
+
+        alert_group_ids = [i.alert_group_id for i in notifications]
```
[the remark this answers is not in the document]
Ferril: We will limit it during rendering notification message, but log records with type "notification triggered" will be created for all alert groups from notification bundle

engine/apps/alerts/tasks/notify_user.py
```diff
@@ -418,17 +468,144 @@ def perform_notification(log_record_pk):
...
+                # todo:
+                #  notify_by_sms_bundle:
+                #  - call send sms async
+                #  - filter notifications by bundle_uuid
+                #  - send sms
+                #  - create logs
+                #  - delete notifications
+                # phone_backend = PhoneBackend()
+                # phone_backend.notify_by_sms_bundle(user_notification_bundle.user_id, bundle_uuid)
+                pass
```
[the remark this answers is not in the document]
Ferril: I think it makes sense to add the similar logic as in `perform_notification` task

engine/apps/alerts/tasks/notify_user.py
```diff
@@ -170,16 +194,53 @@ def notify_user_task(
...
-                        notification_channel=notification_policy.notify_by,
-                    )
+                    if (
+                        settings.FEATURE_NOTIFICATION_BUNDLE_ENABLED
+                        and notification_policy.notify_by in UserNotificationBundle.NOTIFICATION_CHANNELS_TO_BUNDLE
+                    ):
+                        user_notification_bundle, _ = UserNotificationBundle.objects.get_or_create(
+                            user=user, important=important, notification_channel=notification_policy.notify_by
+                        )
+                        user_notification_bundle = UserNotificationBundle.objects.filter(
```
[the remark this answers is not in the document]
iskhakov: Can we rewrite it in a following way?
```
user_notification_bundle, _ = UserNotificationBundle.objects.select_for_update().get_or_create(
                            user=user, important=important, notification_channel=notification_policy.notify_by
                        )
```

engine/apps/alerts/tasks/notify_user.py
```diff
@@ -420,17 +486,137 @@ def perform_notification(log_record_pk):
...
+# deprecated
+@shared_dedicated_queue_retry_task(
+    autoretry_for=(Exception,), retry_backoff=True, max_retries=0 if settings.DEBUG else None
+)
+def send_user_notification_signal(log_record_pk):
+    # Triggers user_notification_action_triggered_signal
+    # This signal is only connected to UserSlackRepresentative and triggers posting message to Slack about some
+    # FAILED notifications (see NotificationDeliveryStep and ERRORS_TO_SEND_IN_SLACK_CHANNEL).
+    # No need to call it here.
+    pass
```
joeyorlando, line 622: can this be removed?

engine/apps/metrics_exporter/tests/test_update_metrics_cache.py
```diff
@@ -607,6 +608,82 @@ def get_called_arg_index_and_compare_results(cache_was_updated=False):
...
+        assert mock_cache_set_called_args[0].args[1] == expected_result_metric_user_was_notified
+        # create new log record for alert_group_1
+        make_user_notification_policy_log_record(
+            type=UserNotificationPolicyLogRecord.TYPE_PERSONAL_NOTIFICATION_TRIGGERED,
+            author=user,
+            alert_group=alert_group_1,
+        )
+        # check metric was not updated
+        update_metric_if_needed(user, [alert_group_1.id, alert_group_2.id, alert_group_3.id])
+        mock_cache_set.assert_called_once()
```
joeyorlando, line 684: should this test be added in `apps/alerts/tests/test_notify_user.py`?

engine/apps/alerts/tasks/notify_user.py
```diff
@@ -172,16 +215,50 @@ def notify_user_task(
...
-                        reason=reason,
-                        slack_prevent_posting=prevent_posting_to_thread,
-                        notification_step=notification_policy.step,
-                        notification_channel=notification_policy.notify_by,
-                    )
+                    if (
+                        settings.FEATURE_NOTIFICATION_BUNDLE_ENABLED
+                        and UserNotificationBundle.notification_is_bundleable(notification_policy.notify_by)
+                    ):
+                        user_notification_bundle, _ = UserNotificationBundle.objects.select_for_update().get_or_create(
```
joeyorlando, line 222: why `select_for_update()` and then `get_or_create()`?

engine/apps/alerts/tasks/notify_user.py
```diff
@@ -1,23 +1,83 @@
...
+
+    collected_steps = ", ".join(
+        UserNotificationPolicy.NotificationChannel(step_id).label for step_id in collected_steps_ids
+    )
+    reason = ("Reason: " + reason + "\n") if reason is not None else ""
+    reason += ("Further notification plan: " + collected_steps) if len(collected_steps_ids) > 0 else ""
+    return reason
+
+
+def update_metric_if_needed(user, active_alert_group_ids):
```
joeyorlando, line 62: _nit_ consider adding types here

engine/apps/alerts/models/user_notification_bundle.py
```diff
@@ -0,0 +1,72 @@
...
+        return last_notified + timezone.timedelta(seconds=BUNDLED_NOTIFICATION_DELAY_SECONDS)
+
+    def append_notification(self, alert_group, notification_policy):
+        self.notifications.create(
+            alert_group=alert_group, notification_policy=notification_policy, alert_receive_channel=alert_group.channel
+        )
+
+    @classmethod
+    def notification_is_bundleable(cls, notification_channel):
+        return notification_channel in cls.NOTIFICATION_CHANNELS_TO_BUNDLE
```
joeyorlando, line 61: _nit_ consider adding types to these methods

engine/apps/alerts/models/user_notification_bundle.py
```diff
@@ -0,0 +1,72 @@
...
+
+class BundledNotification(models.Model):
+    alert_group = models.ForeignKey("alerts.AlertGroup", on_delete=models.CASCADE)
+    alert_receive_channel = models.ForeignKey("alerts.AlertReceiveChannel", on_delete=models.CASCADE)
+    notification_policy = models.ForeignKey("base.UserNotificationPolicy", on_delete=models.SET_NULL, null=True)
+    notification_bundle = models.ForeignKey(
+        UserNotificationBundle, on_delete=models.CASCADE, related_name="notifications"
+    )
+    created_at = models.DateTimeField(auto_now_add=True)
+    bundle_uuid = models.CharField(max_length=100, null=True, default=None, db_index=True)
```
joeyorlando, line 72: I see this is _primarly_ used right now in as the second arg to [`PhoneBackend.notify_by_sms_bundle_async`](https://github.com/grafana/oncall/pull/4457/files#diff-08e0e52cc6ff4bfe4a07d2931fe3385b62cb9265ffc646675865f63a7535c91bR197), but can we just use `pk` instead of needing also `bundle_uuid`?

engine/apps/alerts/tasks/notify_user.py
```diff
@@ -420,17 +486,137 @@ def perform_notification(log_record_pk):
...
+            )
+            return
+
+        notifications = user_notification_bundle.notifications.filter(bundle_uuid__isnull=True).select_related(
+            "alert_group"
+        )
+
+        log_records_to_create = []
+        skip_notification_ids = []
+        active_alert_group_ids = set()
```
joeyorlando, line 529: _nit_ consider adding types on these lists/set

joeyorlando (approved):
overall LGTM!

engine/apps/alerts/models/user_has_notification.py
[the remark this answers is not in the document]
Ferril: Agree 💯 I'd propose to do it with `notify_user_task` refactoring (https://github.com/grafana/oncall/issues/4629)

engine/apps/alerts/tasks/notify_user.py
```diff
@@ -172,16 +215,50 @@ def notify_user_task(
...
+                        settings.FEATURE_NOTIFICATION_BUNDLE_ENABLED
+                        and UserNotificationBundle.notification_is_bundleable(notification_policy.notify_by)
+                    ):
+                        user_notification_bundle, _ = UserNotificationBundle.objects.select_for_update().get_or_create(
+                            user=user, important=important, notification_channel=notification_policy.notify_by
+                        )
+                        # check if notification needs to be bundled
+                        if user_notification_bundle.notified_recently():
+                            user_notification_bundle.append_notification(alert_group, notification_policy)
+                            # schedule send_bundled_notification task if it hasn't been scheduled or the task eta is
```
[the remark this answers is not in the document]
Ferril: If for some reason the task didn't start (e.g. rabbitmq issue)

engine/apps/alerts/tasks/notify_user.py
Ferril: It can be removed after release, because we want to make sure there are no more such tasks in the celery queue. Otherwise, a celery exception may occur when the scheduled task is not found

engine/apps/metrics_exporter/tests/test_update_metrics_cache.py
Ferril: 🤔 I think it makes sense to keep it here since it's about updating metrics and doesn't affect user notification flow

engine/apps/alerts/tasks/notify_user.py
Ferril: it works the same way as
```
user_notification_bundle, _ = UserNotificationBundle.objects.get_or_create(
    user=user, important=important, notification_channel=notification_policy.notify_by
)
user_notification_bundle = UserNotificationBundle.objects.filter(pk=user_notification_bundle.pk).select_for_update()[0]
```

engine/apps/alerts/models/user_notification_bundle.py
Ferril: The idea is that `bundle_uuid` groups a bunch of `BundledNotification`s, that were processed in the `send_bundled_notification` task (if user got flooded, it can be a hundreds of them). While we are processing these notifications, new notifications can be created and attached to the current `UserNotificationBundle` and they will have their own `bundle_uuid` and will be processed and sent in a separate task.
That means, that we can't use `UserNotificationBundle` pk, because there could be processed and unprocessed notifications attached at the same time, and we want to keep them separately. We don't want to use `BundledNotification` pks, because there could be a long list of them

engine/apps/alerts/tasks/notify_user.py
joeyorlando: makes sense!

joeyorlando: I'll add a note to https://github.com/grafana/oncall/issues/4629 to remove it as part of that task

mderynck (approved):

Issue #4454, opened 2024-06-04
Issues with refreshing the  alert groups


teodosii:
### What went wrong?

https://raintank-corp.slack.com/archives/C0229FD3CE9/p1717169088020959

iskhakov:
should it go to "to be groomed" or straight to "sprint backlog" if urgent?

joeyorlando:
do you mind adding a story point estimate to this?

PR #4463, opened 2024-06-05
Fix alert groups automatic refresh

teodosii:
# What this PR does

- Fixes AG auto-refresh (every 15s)
- Reflect correct data on the UI after doing a bulk update
- Few tweaks regarding UX around bulk updates/loading
- Added some missing typescript support

Closes https://github.com/grafana/oncall/issues/4454


grafana-plugin/src/containers/RemoteFilters/RemoteFilters.tsx
```diff
@@ -48,7 +48,7 @@ interface RemoteFiltersProps extends WithStoreProps {
   grafanaTeamStore: GrafanaTeamStore;
   skipFilterOptionFn?: (filterOption: FilterOption) => boolean;
 }
-interface RemoteFiltersState {
+export interface RemoteFiltersState {
```
teodosii, line 51: Exporting these to have the typing needed in Incidents for now

grafana-plugin/src/models/alertgroup/alertgroup.helpers.ts
```diff
@@ -33,6 +35,26 @@ export class AlertGroupHelper {
     return (await onCallApi().POST('/alertgroups/bulk_action/', { params: {}, body: data })).data;
   }
 
+  @AutoLoadingState(ActionKey.INCIDENTS_BULK_UPDATE)
+  static async updateBulkActionAndRefresh(
+    data: ApiSchemas['AlertGroupBulkActionRequest'],
+    alertGroupStore: AlertGroupStore,
+    onFinally?: () => void
```
teodosii, line 42: This made more sense to be here rather than in the component, it also accepts `AutoLoadingState`

grafana-plugin/src/models/alertgroup/alertgroup.ts
```diff
@@ -79,7 +79,7 @@ export class AlertGroupStore {
     const newAlerts = new Map(
       results.map((alert: ApiSchemas['AlertGroup']) => {
         const oldAlert = this.alerts.get(alert.pk) || {};
-        const mergedAlertData = { ...oldAlert, ...alert, undoAction: alert.undoAction };
+        const mergedAlertData = { ...oldAlert, ...alert };
```
teodosii, line 82: Removed the `undoAction`, it didn't seem to help in any way :thinking:

grafana-plugin/src/models/alertgroup/alertgroup.ts
```diff
@@ -305,8 +294,14 @@ export class AlertGroupStore {
...
       [AlertAction.unResolve]: this.unresolve,
     };
+
     if (actionToMethodMap[action]) {
-      await actionToMethodMap[action](id, delay);
+      try {
+        this.setLiveUpdatesPaused(true);
+        await actionToMethodMap[action](id, delay);
+      } finally {
+        this.setLiveUpdatesPaused(false);
```
teodosii, line 303: Centralizing these 2 calls here rather than in each method

grafana-plugin/src/pages/incidents/Incidents.tsx
```diff
@@ -428,23 +431,23 @@ class _IncidentsPage extends React.Component<IncidentsPageProps, IncidentsPageSt
...
 
     if (!store.alertGroupStore.bulkActions) {
       return null;
     }
 
-    const { results } = AlertGroupHelper.getAlertSearchResult(store.alertGroupStore);
-
     const hasSelected = selectedIncidentIds.length > 0;
-    const isLoading = LoaderHelper.isLoading(store.loaderStore, ActionKey.FETCH_INCIDENTS);
-    const hasInvalidatedAlert = Boolean(
```
teodosii, line 442: I believe this didn't add any value whatsoever, due to this we were showing "Out of date results" which didn't make sense as we didn't invalidate any data whatsoever

grafana-plugin/src/models/alertgroup/alertgroup.helpers.ts
```diff
@@ -33,6 +35,26 @@ export class AlertGroupHelper {
...
+
+      await AlertGroupHelper.bulkAction(data);
+
+      // pull new data
+      await alertGroupStore.fetchAlertGroups();
+    } finally {
+      alertGroupStore.setLiveUpdatesPaused(false);
+
+      onFinally?.();
+    }
```
brojd, line 55: Why do we need try {} finally {}? Afaik it doesn't make much sense if we omit `catch`, we could just call those functions in the same order without try{}finally{} or I'm missing sth?

Also this way I think we don't handle errors if they appear in try {} block, right?

teodosii: My idea was to make sure we're calling `setLiveUpdatesPaused` and the passed handler even if the request fails. I guess we could add a catch clause and show notification?

brojd: In this we can add catch clause and at least rethrow the error so it's handled by our error boundaries. Without it the error will be completely swallowed

brojd:
Can we refetch stats whenever alert status is updated (either with bulk or single update)? Right now user needs to wait for the next polling tick:

https://github.com/grafana/oncall/assets/12073649/8d844b82-7b42-4915-8708-dccba69d595f

brojd (approved):

Issue #4428, opened 2024-05-31
Schedule Issues


teodosii:
### What went wrong?

- Rotation Form: Time is being each time the date gets changed
- UI: Calendar shifts display once you click it (the one next to Period Selection)
- Shifts starting with date in the past do not have preview on the UI
- Changing the timezone doesn't reflect changes in the Rotation Modal
- Rotations by 2 weeks still display only 1 week gap between rotations
- Override Form: Changing the date changes the time as well incrementally
- In existing OnCall schedule we see "Add Rotations to see users" and no users are being displayed on top unless you change period

### How do we reproduce it?

-

### Grafana OnCall Version

all

### Product Area

Schedules

### Grafana OnCall Platform?

None

### Anything else to add?

-

mderynck:
When requesting a shift swap the dialog is populated with the first shift times instead of the shift being swapped

raphael-batte:
- [ ] Part of the shift which continues from last week doesn't have visible username
![image](https://github.com/grafana/oncall/assets/102970268/a45d562a-8e8a-418f-831d-abe766eb04b3)
- [ ] Calendar opens inline instead of overlay 
![image](https://github.com/grafana/oncall/assets/102970268/ccc926d2-0268-4397-ba33-b58e50e2869f)

raphael-batte:
@teodosii @matiasb It seems this is BE issue and now out of scope for FE 
_Rotations by 2 weeks still display only 1 week gap between rotations_

iskhakov:
> When requesting a shift swap the dialog is populated with the first shift times instead of the shift being swapped

Originally reported in https://raintank-corp.slack.com/archives/C0229FD3CE9/p1717106994906299

PR #4481, opened 2024-06-06
Make rotation start and rotation end take timezone into consideration

teodosii:
# What this PR does

- Fixes Rotation Start and Rotation End to take selected offset into consideration
- Fixed issue where week/month period was being restored when the offset was being changed (now it defaults to start of week, midnight, in selected timezone offset)

Related to https://github.com/grafana/oncall/issues/4428


grafana-plugin/src/containers/RotationForm/RotationForm.tsx
```diff
@@ -435,9 +434,10 @@ export const RotationForm = observer((props: RotationFormProps) => {
   useEffect(() => {
     if (shift) {
       setRotationName(getShiftName(shift));
-      const shiftStart = getDateTime(shift.shift_start);
+      const shiftStart = getDateTime(shift.shift_start).utcOffset(store.timezoneStore.selectedTimezoneOffset);
```
teodosii, line 437: This might not be needed, ignore it for now

joeyorlando (commented):
maybe worthwhile adding some unit tests for some of the time related utility functions, wdyt?

brojd:
I can see the issue when you create Rotation that starts before / after DST change:

https://github.com/grafana/oncall/assets/12073649/3d794a36-7bcd-4a2b-8ead-843e4bacdcf1

brojd:
Another issue is that the day jumps to previous day if you date in before DST change:

https://github.com/grafana/oncall/assets/12073649/e5097e78-a191-4e9a-95b4-7769645e6b08

brojd:
I can see the following issues:
1) I create rotation that starts before DST change (27th March) in my system timezone at 2:00 having GMT+2 offset selected (so start time is 00:00 UTC), then created rotation has start time 01:00 UTC:

https://github.com/grafana/oncall/assets/12073649/5e35fe89-80ca-4f14-8c20-b0599693f54c

2) I create rotation that starts before DST change (27th March) in my system timezone having GMT+0 offset selected, then shifts are overlapping and "Limit each shift length" is enabled as soon as the rotation is created:

https://github.com/grafana/oncall/assets/12073649/d8b2a33a-06f1-48a9-8b09-940c5ee2dc4c

3) When I have GMT+0 offset selected, then I navigate in calendar to some date that is after the DST change in my system timezone, then the default value for start/end date of rotation is moved back to 23:00 of the previous day:

https://github.com/grafana/oncall/assets/12073649/2c1f7a63-0723-4099-a1f0-7723c3affd0e

4) The rotation preview misses the last day of the week / month:

https://github.com/grafana/oncall/assets/12073649/6cd906b0-4086-47bc-b41d-90f656286bfd

teodosii:
Last issues reported no2 and no4 are not related to this work.

grafana-plugin/src/containers/RotationForm/RotationForm.tsx
```diff
@@ -82,6 +85,25 @@ interface RotationFormProps {
...
+  if (isNewRotation) {
+    const localMoment = start.utcOffset(utcOffset);
+    const newValue = localMoment
+      .set('date', 1)
+      .set('year', start.year())
+      .set('month', start.month())
+      .set('date', start.date())
+      .set('hour', 0)
+      .set('minute', 0)
+      // .set('minute', 12) // debug
```
brojd, line 98: can be removed?

grafana-plugin/src/containers/RotationForm/RotationForm.tsx
```diff
@@ -82,6 +85,25 @@ interface RotationFormProps {
...
+      .set('second', 0);
+
+    return newValue;
+  }
+
+  return start.utcOffset(utcOffset); // .set('minute', 40); // debug
```
brojd, line 104: can we remove comment?

grafana-plugin/src/containers/RotationForm/parts/DateTimePicker.tsx
```diff
@@ -73,7 +88,7 @@ export const DateTimePicker = observer(
             <DatePickerWithInput
               open
               disabled={disabled}
-              value={getDateForDatePicker(valueInSelectedTimezone)}
```
brojd, line 76: I guess getDateForDatePicker is no longer used? Can we remove it in this case?

teodosii: yes, I have it removed in a stash. It will be removed.

grafana-plugin/src/containers/RotationForm/RotationForm.tsx
teodosii: :+1:

grafana-plugin/src/containers/RotationForm/parts/DateTimePicker.tsx
```diff
@@ -27,40 +26,56 @@ interface DateTimePickerProps {
...
-        .set('minute', valueAsDate.getMinutes())
-        .set('second', valueAsDate.getSeconds());
-
-      onChange(newValue);
+    const { timezoneStore } = useStore();
+    const { selectedTimezoneOffset } = timezoneStore;
+
+    const handleDateChange = (value: Date) => {
+      const newDate = dayjs(value)
+        .utcOffset(selectedTimezoneOffset)
```
brojd, line 34: In all the places when we do `someDayjsDate.utcOffset(timezoneStore.selectedTimezoneOffset)`, can we use `timezoneStore.getDateInSelectedTimezone(someDayjsDate)` instead? It works the same, right? We do it already in other places so we could be consistent

brojd (approved):

grafana-plugin/src/containers/RotationForm/parts/DateTimePicker.tsx
teodosii: I slightly improved this part since the last time you checked. I added a helper `toDateWithTimezoneOffset`

```jsx
export const toDateWithTimezoneOffset = (date: dayjs.Dayjs, timezoneOffset: number) => {
  return date.utcOffset() === timezoneOffset ? date : date.utcOffset(timezoneOffset);
};
```

The idea is that if you ever apply the offset twice on a dayjs object, it might have some weird implications like getting weird format or yielding weird results for `hour()` and `minute()`. So instead of appending `utcOffset` to those calls, I rely on this method that checks for an already applied offset

teodosii:
Tests will be added separately as there's quite many cases to be handled and verified. before dst/during dst/after, switching from one period to another, changing the timezone etc

brojd (approved):

PR #4501, opened 2024-06-11
use self hosted github runners for actions workflows

joeyorlando:
# What this PR does

Use self-hosted GitHub runners as opposed to default GitHub ones

### DIFF
--- a/.github/workflows/build-engine-docker-image-and-publish-to-dockerhub.yml
+++ b/.github/workflows/build-engine-docker-image-and-publish-to-dockerhub.yml
@@ -13,7 +13,7 @@ on:
 jobs:
   build-engine-docker-image-and-publish-to-dockerhub:
     name: Build engine Docker image and publish to Dockerhub
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-x64-large
     # These permissions are needed to assume roles from Github's OIDC.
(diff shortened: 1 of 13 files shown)


.github/workflows/build-engine-docker-image-and-publish-to-dockerhub.yml
```diff
@@ -13,7 +13,7 @@ on:
 jobs:
   build-engine-docker-image-and-publish-to-dockerhub:
     name: Build engine Docker image and publish to Dockerhub
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-x64-large
```
joeyorlando, line 16: trying to see if I can speed up building of our multi-platform Docker image (linux/arm64 + linux/amd64) by just throwing more resources at it.. it seems that [installing the Python deps on linux/arm64 takes ~400s](https://github.com/grafana/oncall/actions/runs/9419197948/job/25948473369#step:4:1579)

alternative here would be to build the different image variants on different hardware and publish the image manifest separately (something like [this](https://github.com/docker/build-push-action/issues/671))

joeyorlando:
closing as self-hosted runners are not yet available to open-source repos

https://github.com/grafana/oncall/actions/runners?tab=self-hosted

PR #4518, opened 2024-06-13
Support message shortcut broadcast

vadimkerr:
# What this PR does

Related to https://github.com/grafana/oncall-gateway/issues/206

## Checklist


engine/apps/slack/scenarios/resolution_note.py
```diff
@@ -82,10 +83,11 @@ def process_scenario(
                 _slack_team_identity=slack_team_identity,
                 channel_id=channel_id,
             )
-        except KeyError:
-            self.open_warning_window(payload, warning_text)
-            return
-        except SlackMessage.DoesNotExist:
+        except (KeyError, SlackMessage.DoesNotExist):
```
Konstantinov-Innokentii, line 86: It's not clear for me, how it works. Why both KeyError and SlackMessage.DoesNotExist indicates that event belongs to another cluster?

engine/apps/slack/scenarios/resolution_note.py
```diff
@@ -82,10 +83,11 @@ def process_scenario(
...
+            if settings.FEATURE_MULTIREGION_ENABLED:
```
Konstantinov-Innokentii, line 87: I propose to check UNIFIED_SLACK_APP_ENABLED flag. MULTIREGION is already enabled.
P.S.
Yes, these flags need some pruning :)

Konstantinov-Innokentii:
Thanks for the PR. What's unclear to me is the case when an organization was migrated. In this case, both clusters will have a Slack message, so we will not receive a Slack message.DoesNotExist error and will attempt to add a resolution note twice - in the old and new cluster?

engine/apps/slack/scenarios/resolution_note.py
vadimkerr: Getting a `KeyError` here means that this is not a thread message (thus can't be added as a resolution note). In this case we just return a 200 response to avoid multiple regions opening the same warning window multiple times.
Getting a `SlackMessage.DoesNotExist` means that there could be another region that can handle this event.

I changed it to two separate checks to make this clearer in https://github.com/grafana/oncall/pull/4518/commits/853a135ccc8e70dcb111e519be3d66ae0543216a

engine/apps/slack/scenarios/resolution_note.py
vadimkerr: done in https://github.com/grafana/oncall/commit/853a135ccc8e70dcb111e519be3d66ae0543216a

vadimkerr:
@Konstantinov-Innokentii good catch! I added a check for `organization.deleted_at` to handle this case: https://github.com/grafana/oncall/commit/853a135ccc8e70dcb111e519be3d66ae0543216a

engine/apps/slack/scenarios/resolution_note.py
```diff
@@ -76,16 +77,27 @@ def process_scenario(
 
         warning_text = "Unable to add this message to resolution note, this command works only in incident threads."
 
+        thread_ts = payload.get("message", {}).get("thread_ts")
```
Konstantinov-Innokentii, line 80: Could you please add a comment that it checks if message is a thread message?

Konstantinov-Innokentii (approved):

engine/apps/slack/scenarios/resolution_note.py
vadimkerr: done in https://github.com/grafana/oncall/pull/4518/commits/caafcac76edf4239dd43dd19ad15d1f99a341a45

PR #4519, opened 2024-06-13
set default grafana_url if GRAFANA_API_URL set

juriku:
# Issue Oncall backend calling grafana api when grafana is self hosted and under certain path

fixes errors like:
```
source=engine:app google_trace_id=none logger=apps.grafana_plugin.helpers.client Error connecting to api instance 404 Client Error: Not Found for url: https://test.com/api/access-control/users/permissions/search?actionPrefix=grafana-oncall-app
```
while GRAFANA_API_URL=https://test.com/grafana

relevant for when external self-hosted grafana has same URL, but under different path, like /grafana

Also fixes issue when self.api_url has path, but no '/' in the end
```
if urlparse(self.api_url).path and not self.api_url.endswith('/'):
            self.api_url += '/'
```
because in case of `urljoin(self.api_url, endpoint)`
Example:
`https://test.com/path`, `api/v1` -> will become `https://test.com/api/v1`
`https://test.com/path/`, `api/v1` -> will become `https://test.com/path/api/v1`

Another small fix if to change from HEAD to GET for api/org, as there is not HEAD, according to documentation
```
logger=apps.grafana_plugin.helpers.client Error connecting to api instance 404 Client Error: Not Found for url: https://test.com/api/org
```


engine/apps/grafana_plugin/helpers/client.py
```diff
@@ -242,7 +244,7 @@ def get_users_permissions(self) -> typing.Optional[UserPermissionsDict]:
         return all_users_permissions
 
     def is_rbac_enabled_for_organization(self) -> bool:
-        _, resp_status = self.api_head(self.USER_PERMISSION_ENDPOINT)
+        _, resp_status = self.api_get(self.USER_PERMISSION_ENDPOINT)
```
joeyorlando, line 247: > Another small fix if to change from HEAD to GET for api/org, as there is not HEAD, according to documentation

I don't believe this change does what you intend it to do. `/api/access-control/users/permissions/search` will in fact exist as a Grafana API endpoint, under certain circumstances (ie. Grafana Enterprise and Grafana Cloud ([docs](https://grafana.com/docs/grafana/latest/administration/roles-and-permissions/access-control/))). It's used [here](https://github.com/grafana/oncall/blob/dev/engine/apps/user_management/sync.py#L50-L52)

juriku: it does fix the error. 
I think it's [this API request](https://grafana.com/docs/grafana/latest/developers/http_api/access_control/#list-permissions-assigned-to-a-user)
Also if you check 10 lines up in file you will see another `self.api_get(self.USER_PERMISSION_ENDPOINT)` , so somebody just forgot to change that.

Issue #96, opened 2022-06-15
Make integration and notifications via mattermost.


BortovEvgeniy:
https://github.com/mattermost/mattermost-server

Thanks!

22itay:
+1

Matvey-Kuk:
👋 Kindly asking anyone interested in Mattermost support to leave some note so we'll be able to see how many people are interested.

Also check [Integration with Discord #186](https://github.com/grafana/oncall/issues/186)

apsharapov:
+

Anon-bit911:
Mattermost and slack integration  most desired for grafana on call after webhook and email

ravenjm:
I'm also intersted in integration with mattermost!

jakubvojacek:
Very much interested!

okycelt:
We're also very interested

drobek94:
We're also interested.

M4t7e:
A Mattermost integration would be awesome 👍

m0sh1x2:
Quite a lot of users are using Mattermost for notifications so it would be great to have a working integration.

ravenjm:
@Matvey-Kuk should we expect feature in the near future?

Matvey-Kuk:
Friends, we really want to deliver Mattermost integration as soon as possible. Since OnCall is an OSS project and the core team already has a huge backlog, we came up with an idea to organise a group of volunteers in order to make it happen.

If you are interested in participating in the working group which will build Mattermost integration, please sign up to our public slack: https://slack.grafana.com/ and join #grafana-oncall channel 👋

We will organise a call and discuss the architecture, approaches and best practices we learned from Slack and MS Teams integration.

Let's build it ☺️

Tagging everyone who liked above: @roock, @22itay, @ravenjm, @floweb, @M4t7e, @aevteev, @azielke, @DmitriyMurzin, @ilya-jetruby, @floweb, @m0sh1x2, @M4t7e, @m0sh1x2, @ravenjm .

Matvey-Kuk:
@matiasb it would be cool if you provide some information about `messaging backends` (our plugin system for ChatOps) here in writing or in a future video call. WDYT?

Matvey-Kuk:
We will discuss Mattermost integration on the next community call: https://github.com/grafana/oncall/discussions/451

xom4ek:
```json
{
    "username": "Grafana OnCall",
    "attachments": [
        {
[...]

PR #4534, opened 2024-06-14
Mattermost OAuth Flow

ravishankar15:
# What this PR does
This PR covers the OAuth installation flow of the mattermost integration

Closes [issue link here]

## Checklist


ravishankar15:
@matiasb Can you check this PR for the backend changes ! if the flow looks good I will make the front end changes to show the installation link to the user.

joeyorlando:
this _seems_ to be the start of https://github.com/grafana/oncall/issues/96

engine/apps/api/views/organization.py
```diff
@@ -108,6 +109,28 @@ def get(self, request):
         return Response(code)
 
 
+class GetMattermostSetupDetails(APIView):
+    authentication_classes = (PluginAuthentication,)
+    permission_classes = (IsAuthenticated, RBACPermission)
+
+    rbac_permissions = {
+        "get": [RBACPermission.Permissions.INTEGRATIONS_WRITE],
```
matiasb, line 117: It would be nice to move this logic to the backend itself, maybe using the existing `GetChannelVerificationCode`? (although it could be confusing because in this case the code would be for a mattermost instance instead of a telegram or msteams channel), or defining a new optional messaging backend step for this. In an ideal world, it should be possible to write a messaging backend without changing any code in the other apps. Will give it some thought.

engine/apps/api/views/organization.py
```diff
@@ -108,6 +109,28 @@ def get(self, request):
...
+    def get(self, request):
+        organization = request.auth.organization
+        user = request.user
+        from apps.mattermost.models import MattermostAuthToken
+
+        with suppress(MattermostAuthToken.DoesNotExist):
+            existing_auth_token = organization.mattermost_auth_token
+            existing_auth_token.delete()
+        _, auth_token = MattermostAuthToken.create_auth_token(user=user, organization=organization)
+        manifest_link = f"{settings.BASE_URL}/mattermost/manifest?auth_token={auth_token}"
```
matiasb, line 129: It may be better to use `create_engine_url` helper to build the URL here.

engine/apps/mattermost/views.py
```diff
@@ -0,0 +1,64 @@
...
+            "expand": {"app": "summary", "acting_user": "summary"},
+            "state": {"auth_token": auth_token},
+        }
+
+    def _build_bindings_callback(self, auth_token: str) -> dict:
+        return {"path": "/mattermost/bindings", "state": {"auth_token": auth_token}}
+
+    def _build_manifest(self, auth_token: str) -> dict:
+        return {
+            "app_id": "Grafana-Oncall",
```
matiasb, line 32: Case for oncall should be OnCall (here and below)

engine/apps/mattermost/views.py
```diff
@@ -0,0 +1,64 @@
...
+            "version": "1.0.0",
+            "display_name": "Grafana Oncall",
+            "description": "Grafana Oncall app for sending and receiving events from mattermost",
+            "homepage_url": "https://grafana.com/docs/oncall/latest/",
+            "requested_permissions": ["act_as_bot"],
+            "requested_locations": ["/in_post", "/post_menu", "/command"],
+            "on_install": self._build_on_install_callback(auth_token=auth_token),
+            "bindings": self._build_bindings_callback(auth_token=auth_token),
+            "http": {"root_url": "http://host.docker.internal:8080"},
```
matiasb, line 41: I guess we will need to update this URL?

engine/apps/mattermost/views.py
```diff
@@ -0,0 +1,64 @@
...
+        view_text = MATTERMOST_CONNECTED_TEXT.format(
+            app_id=app_id, organization_title=request.auth.organization.org_title
+        )
+        # TODO: Create and save bot user and access token and also link org and user with mattermost
+        response = {"type": "ok", "text": view_text}
+        return Response(response, status=status.HTTP_200_OK)
+
+
+class MattermostBindings(APIView):
+    authentication_classes = (MattermostWebhookAuthTokenAuthentication,)
```
matiasb, line 59: I think bindings should authenticate the user behind the request instead (or too), right?

matiasb (commented):
Thanks for start working on this! Looking good, added a few comments/questions.
It will be nice to start adding some initial tests too.

engine/apps/mattermost/views.py
ravishankar15: This call is to return the bindings supported by our app hence it has the token authentication. Once we decide on the bindings like `/alert`  we will create separate paths (or reuse old ones) in which case those paths will have separate authentication.

engine/apps/api/views/organization.py
ravishankar15: I gave some thought. One way it could be is via polymorphic association we could have a single model to generate the tokens for each backend apps with a `type` field but that might be slightly complicated to unify that approach for the existing apps token implementation.

engine/apps/mattermost/views.py
```diff
@@ -0,0 +1,66 @@
...
+            "app_id": "Grafana-OnCall",
+            "version": "1.0.0",
+            "display_name": "Grafana OnCall",
+            "description": "Grafana OnCall app for sending and receiving events from mattermost",
+            "homepage_url": "https://grafana.com/docs/oncall/latest/",
+            "requested_permissions": ["act_as_bot"],
+            "requested_locations": ["/in_post", "/post_menu", "/command"],
+            "on_install": self._build_on_install_callback(auth_token=auth_token),
+            "bindings": self._build_bindings_callback(auth_token=auth_token),
+            "http": {"root_url": live_settings.MATTERMOST_WEBHOOK_HOST},
```
matiasb, line 43: I was just reviewing my notes and draft code, and I think that my original plan was to [use JWT](https://developers.mattermost.com/integrate/apps/structure/manifest/#http) for the app auth/link (see also this [example](https://developers.mattermost.com/integrate/apps/quickstart/quick-start-jwt/), and that's what the `authenticate` function was supposed to do in the WIP I shared; also, that should allow us identify the user making requests to OnCall from Mattermost).

So, I think, we could generate a random token to store in a Mattermost Server model (similar to the one in the gist). This token will be asked to the user during install, and we can then link the Mattermost App with Server (and the respective org).

The flow I'm imagining will be something like:
- Admin user goes to a "setup mattermost" endpoint
  - this endpoint will setup a server instance with a generated token
  - returns a manifest URL (no auth, this is public info) with the `use_jwt` flag enabled
- User install app using the manifest URL; it should be asked for the generated secret
  - manifest includes an `install` URL (this URL could include the org public ID, and it should require a valid and matching token/secret, server-side).
  - when install endpoint is hit, this will complete the (OnCall Mattermost) server setup

And that's it (?). From that moment on, app should be installed and be able to talk to OnCall, and OnCall should have the information for the mattermost app linked to a specific org.

Makes sense? Sorry I didn't try to summarize this before, but I couldn't find the time to understand my notes and test a few things until now. I think that the basics for this are in the shared gist.

Let me know if you think this won't work, or if you prefer me to work on a better PoC following the above.

ravishankar15: Hi ! Correct me if my understanding is right,

1. We have two endpoints,
     - `/manifest`  --> Returns the manifest json
     - `/install` --> The on_install call back
2. Do we have to authenticate the `/manifest` endpoint ? If No then wouldn't that mean this endpoint can be hit by anyone of the internet users ? Like I can hit the server of any customers oncall instance and spam them right ?
3. When the user hits the `Setup Mattermost` We generate the JWT signing secret and create the JWT token with org id in it. The JWT token will be shared in the UI along with the Mattermost json url. When the user installs the app the Mattermost app will send the jwt token in the request header. For authenticating when we receive the jwt token we get the org id from the jwt payload and using the org id we get the signing secret from our DB for that org id and then verify the signature before proceeding Is that right ?


If (3) is the expected behaviour. When we do the link-user like linking the oncall user to Mattermost user ? I believe we will generate a separate token which needs to be sent when the user types /link-user command from Mattermost ? using which we link the user.

matiasb: > Hi ! Correct me if my understanding is right,
> 
>     1. We have two endpoints,
>        
>        * `/manifest`  --> Returns the manifest json
>        * `/install` --> The on_install call back

Right.

>     2. Do we have to authenticate the `/manifest` endpoint ? If No then wouldn't that mean this endpoint can be hit by anyone of the internet users ? Like I can hit the server of any customers oncall instance and spam them right ?

Hm.. Not exactly, because `/install` will require a valid JWT token. Also we should generate the `manifest.json` using org id info. So, when an admin generates the JWT token, they also get a URL to the manifest with a param indicating the org ID, which is used in the rendered `/install` URL. Then, anyone will be able to get the base `manifest.json`, but to make things work you still need a JWT token and (the correspondent) an org ID... we could also require a fresh enough token before allowing installation. Makes sense?

>     3. When the user hits the `Setup Mattermost` We generate the JWT signing secret and create the JWT token with org id in it. The JWT token will be shared in the UI along with the Mattermost json url. When the user installs the app the Mattermost app will send the jwt token in the request header. For authenticating when we receive the jwt token we get the org id from the jwt payload and using the org id we get the signing secret from our DB for that org id and then verify the signature before proceeding Is that right ?

Yeah, something like that is what I was thinking.

> If (3) is the expected behaviour. When we do the link-user like linking the oncall user to Mattermost user ? I believe we will generate a separate token which needs to be sent when the user types /link-user command from Mattermost ? using which we link the user.

That's correct, we will still need some link-user command (for which we could generate a token, similar to telegram or msteams flows) to bind an OnCall account with a Mattermost user.

How does it sound?

ravishankar15: Agreed yeah `/install` will require the JWT which will have the org id but for the `/manifest` endpoint we should be using a token which maps to the org like we have in the current state of the PR right ? But the token will be a one time usage which after manifest is downloaded will not be used or sent in the `state` param of any  callback ?  
> Hm.. Not exactly, because /install will require a valid JWT token. Also we should generate the manifest.json using org id info. So, when an admin generates the JWT token, they also get a URL to the manifest with a param indicating the org ID, which is used in the rendered /install URL. Then, anyone will be able to get the base manifest.json, but to make things work you still need a JWT token and (the correspondent) an org ID... we could also require a fresh enough token before allowing installation. Makes sense?


I think overall I got the context expect the above point mentioned I will start making the changes and will update the PR over this week.

matiasb: > Agreed ! yeah `/install` will require the JWT which will have the org id but for the `/manifest` endpoint we should be using a token which maps to the org like we have in the current state of the PR ? or should we just send the org_id as plain text in the manifest url like `/manifest?org_id=<ORG_ID>` ?

:+1:  I think using the `org_id` as a query param should be ok, doesn't it? But if we are including the org id as part of the secret maybe we can even omit it here.

ravishankar15: Yeah makes sense ! I think I will skip the org id in the manifest and use the org id in the secret ! will start making the changes in this week !

engine/apps/api/serializers/organization.py
```diff
@@ -113,14 +113,23 @@ class Meta:
 class CurrentOrganizationConfigChecksSerializer(serializers.ModelSerializer):
     is_chatops_connected = serializers.SerializerMethodField()
     is_integration_chatops_connected = serializers.SerializerMethodField()
+    mattermost = serializers.SerializerMethodField()
```
matiasb, line 116: I wouldn't change this API for this, since this is being used outside oncall to confirm a minimal setup required steps. If we need mattermost specific info, maybe we need a new endpoint?

engine/apps/social_auth/backends.py
```diff
@@ -201,3 +203,58 @@ class InstallSlackOAuth2V2(SlackOAuth2V2):
...
+
+    def authorization_url(self):
+        return f"{live_settings.MATTERMOST_HOST}/oauth/authorize"
+
+    def access_token_url(self):
+        return f"{live_settings.MATTERMOST_HOST}/oauth/access_token"
+
+    def get_user_details(self, response):
+        """
+        Return user details from Google API account
```
matiasb, line 232: I guess it would be Mattermost here?

engine/apps/auth_token/auth.py
```diff
@@ -264,6 +270,11 @@ class SlackTokenAuthentication(_SocialAuthTokenAuthentication[SlackAuthToken]):
     model = SlackAuthToken
 
 
+class MattermostTokenAuthentication(_SocialAuthTokenAuthentication[MattermostAuthToken]):
```
matiasb, line 273: Not needed right now, but how hard would be to have all mattermost-related model/logic in a separate app?

matiasb (commented):
Looking good, minor comments added.

To be sure, what will the flow look like at a high-level? (how do we get channels information? how each user links their account?) Is there a [bot account](https://developers.mattermost.com/integrate/reference/bot-accounts/) involved?

engine/apps/api/serializers/organization.py
ravishankar15: I was planning on using the existing 
API that we have used for telegram found this note in that flow hence moved the logic 
over here. https://github.com/grafana/oncall/blob/dev/engine/apps/api/serializers/organization.py#L91

engine/apps/auth_token/auth.py
ravishankar15: Its not a hard thing to do. My thought process was since we are using the common OAuth view to complete the OAuth flow. We can keep the auth model alone in the `auth_token` folder along with other utilities like crypto, exceptions etc. The other models and class will be inside the mattermost app.

ravishankar15:
> Looking good, minor comments added.
> 
> To be sure, what will the flow look like at a high-level? (how do we get channels information? how each user links their account?) Is there a [bot account](https://developers.mattermost.com/integrate/reference/bot-accounts/) involved?

1. The user will install the mattermost app using the flow from this PR
2. A bot account will be created along with the install flow using the REST API
3. The mattermost public channels will be fetched using Public channel endpoint https://api.mattermost.com/#tag/channels/operation/GetPublicChannelsForTeam
4. Each user will have a one to one association with the mattermost user model.
5. Each org will have a one to one association with the mattermost team.
6. The users can link their account via a similar OAuth flow to connect their account -> A mattermost user will be created with data and linked with the oncall user model.

In the follow up PR I am thinking,

To create a small client class for making API calls to mattermost along with the data saving portion of OAuth flow for the install

engine/apps/api/serializers/organization.py
matiasb: Hm... I think the comment refers to something else (the comment was added [here](https://github.com/grafana/oncall/pull/1713/files#diff-5e7f252d928ee561501ecdf2c8f730cc0259c58fc49eb4d5e5880b41b4ee2a4bR117)). The API here was added much later and it is used for a 3rd party integration, so it would be better if we can leave that one unchanged (and add a different endpoint or for now just update the one you linked despite the comment there).

matiasb:
> 2\. A bot account will be created along with the install flow using the REST API

And then we should get a personal access token for this bot account to post notifications as the bot, right? Can we just ask for the token and make the users setup their bot account? (and then we "just need to link" Mattermost accounts with OnCall users, and all the API requests are done through/as the bot?)

> 5\. Each org will have a one to one association with the mattermost team.

To be sure, do we need this association? (do we need to know anything about Mattermost teams on our side?)

Besides those questions, plan sounds good to me :+1:

ravishankar15:
> > 2. A bot account will be created along with the install flow using the REST API
> 
> And then we should get a personal access token for this bot account to post notifications as the bot, right? Can we just ask for the token and make the users setup their bot account? (and then we "just need to link" Mattermost accounts with OnCall users, and all the API requests are done through/as the bot?)
> 
> > 5. Each org will have a one to one association with the mattermost team.
> 
> To be sure, do we need this association? (do we need to know anything about Mattermost teams on our side?)
> 
> Besides those questions, plan sounds good to me 👍

Yes we need to get the access token. Asking the user to setup the bot account also works for us but that means an additional step for the user to perform even then we still need to make some calls to fetch the bot user id and some meta data that we might need to post messages ? Plus this way the integration is little more automated and self served

We need the mattermost team id for getting the public channels https://api.mattermost.com/#tag/channels/operation/GetPublicChannelsForTeam but I will check once before implementing if we can skip any of these association

Should I raise the PR for this branch or shall we merge this PR to feature branch and I can raise a follow up PR to the feature branch

ravishankar15:
I have the changes for the config API can you take a look and if the changes are good can we merge this PR to feature branch so that i can raise the follow up PR to the feature branch !

PR #4301, opened 2024-04-30
Replace docs/reference shortcode with ref URIs

jdbaldry:
You can use `ref` URIs in admonitions (or any shortcodes) because they are inline and not subject to the issues noted in the [`admonition` shortcode](https://grafana.com/docs/writers-toolkit/write/shortcodes/#code-shortcode:~:text=to%20core%20understanding.-,WARNING,For%20more%20information%2C%20refer%20to%20Markdown%20Reference%20Links%20in%20Shortcodes.,-Examples).

The `ref` URIs perform the same pattern matching as `docs/reference` but don't require the use of reference-style links and the destinations are ordinary (full) URLs that can include version substitution. Unlike `docs/reference`, the implementation doesn't use `relref` so you don't have to be careful with omitting trailing slashes and the links will follow redirects.

Documentation: https://grafana.com/docs/writers-toolkit/write/links/#link-from-source-content-thats-used-in-multiple-projects

To check the links, refer to the deploy preview in https://github.com/grafana/website/pull/19630.

Signed-off-by: Jack Baldry <jack.baldry@grafana.com>

PR #4549, opened 2024-06-17
move jinja2 templating docs as a subsection of the integrations docs

joeyorlando:
# What this PR does

- moves the contents of [`/docs/oncall/configure/jinja2-templating`](https://grafana.com/docs/oncall/latest/configure/jinja2-templating/) to a subsection of [`/docs/oncall/integrations`](https://grafana.com/docs/oncall/latest/integrations/)
  - as agreed upon on Slack [here](https://raintank-corp.slack.com/archives/C04JCU51NF8/p1718633939361189), the current layout is a bit confusing. I was searching for documentation around how to configure templates for an integration. I saw that we had a main nav section for "Integrations", but then I had to jump around into a doc named "Configure > Jinja2 templating". The naming here is a bit ambigous because we use Jinja2 templates not only for Integration templates but also for other things like Outgoing Webhooks, label templates, etc. I'm proposing we combine it with the index page for "Integrations" (which already contains a lot of info on how to "configure" Integrations).

## Checklist


docs/sources/configure/jinja2-templating/index.md
joeyorlando, line 1: @grafana/docs-gops do we need to do anything with the aliases here? is there a way to add a redirect from this page to `/docs/oncall/integrations#templates`?

.github/CODEOWNERS
```diff
@@ -1,6 +1,6 @@
 * @grafana/grafana-oncall-backend
 /grafana-plugin @grafana/grafana-oncall-frontend
-/docs @grafana/docs-gops @grafana/grafana-oncall
+/docs @grafana/grafana-oncall
```
joeyorlando, line 3: should we maybe add `@grafana/docs-squad` here in lieu of `@grafana/docs-gops`?

joeyorlando:
tested the changes locally via:

```bash
cd docs
make docs
```

All of the links that were modified seem to work as expected 👍

docs/sources/integrations/_index.md
```diff
@@ -18,17 +21,15 @@ An "Integration" is a main entry point for alerts being consumed by Grafana OnCa
...
 escalations.
 
-For more information about the templating used in OnCall, refer to [Jinja2 templating][].
-
 ## Learn Alert Flow Within Integration
 
 1. An Alert is received on an integration's **Unique URL** as an HTTP POST request with a JSON payload (or via
 [Inbound email][], for inbound email integrations)
-1. Routing is determined for the incoming alert, by applying the [Routing Template][]
-1. Alert Grouping is determined based on [Grouping Id Template][]
```
joeyorlando, line 28: also fixed this issue here:
Screenshot 2024-06-17 at 11 44 13

docs/sources/integrations/_index.md
```diff
@@ -53,16 +54,227 @@ Click the **How to connect** link for more information.
...
+   example `{{ True if payload.state == 'OK' else False }}`
+
+> **Pro Tip:** As a best practice, add _Playbooks_, _Useful links_, or _Checklists_ to the
+alert message.
+
+#### How to edit templates
+
+1. Open the **Integration** page for the integration you want to edit
+1. Click the **Edit** button for the Templates Section. Now you can see previews of all
+templates for the Integration
```
joeyorlando, line 184: fixed a small existing issue here:
Screenshot 2024-06-17 at 12 23 10

.github/CODEOWNERS
alyssawada: We wouldn't want to use the @grafana/docs-squad here as our team workflow has these broke into smaller sub-teams. Michael and I did recently discussed updating this as well. But rather to update the @grafana/docs-gops to include a couple more writers and engineers.

joeyorlando, line 3:
```suggestion
/docs @grafana/docs-gops @grafana/grafana-oncall
```

docs/sources/configure/jinja2-templating/index.md
alyssawada: Usually our approach here would be to move the existing file and update the aliases to include the new route, this way the previous aliases redirect to the new location. 

My initial thought is that this content is still worth a stand alone topic. As you mentioned, the Jinja templates are not only for integration templates but also for other things like webhooks, and label templates. 

What if we keep the original topic with a rename, in addition to adding the content you added for to the integrations _index.md page?

joeyorlando: > What if we keep the original topic with a rename, in addition to adding the content you added for to the integrations _index.md page?

Do you mind rephrasing this, I'm not sure I grok 100%. Do you mean to rename Configure > Jinja2 Templates to Configure > Integrations? In the last part of that:

> in addition to adding the content you added for to the integrations _index.md page?

Do you mean to duplicate the content of Configure > Jinja2 Templates into Integrations > _index.md?

alyssawada: Apologies, that wasn't the clearest train of thought. I have two concerns with deleting the topic entirely:
1. I think this information is worth a standalone topic, albeit in a different location with a different title. 
2. Updating and moving the topic would be a more effective way to address the discoverability issue because adding the information into a relatively long page of content introduces a different discoverability issue in itself, IMO.


I have an outline that proposes an updated TOC that addresses this topic in a different way.. I'd be really interested to get your thoughts on. I planned on sharing it out to the team for input and feedback later this week (after I share it with @mackgorski on Wednesday).

If it's alright with you, I'll snag a quick 30 mins of your time to get your initial thoughts, and sync on this templating content?

jdbaldry:
I'd be keen that https://github.com/grafana/oncall/pull/4301/ is reviewed and merged before this to avoid conflicts on my end.

joeyorlando:
after chatting with @alyssawada, we're going to close this out in favour of some upcoming docs restructuring ✌️

Issue #4547, opened 2024-06-17
Override Form: Changing the date changes the time as well incrementally (by the timezone offset)


teodosii:
### What went wrong?

-

### How do we reproduce it?

-

### Grafana OnCall Version

latest

### Product Area

Schedules

### Grafana OnCall Platform?

None

### Anything else to add?

-

PR #4553, opened 2024-06-18
Tweaks to overrides (added draggable bounds, take timezone into consideration)

teodosii:
# What this PR does

Part of https://github.com/grafana/oncall/issues/4428

Closes https://github.com/grafana/oncall/issues/4547


grafana-plugin/e2e-tests/schedules/addRotation.test.ts
```diff
@@ -0,0 +1,39 @@
...
+  await page.getByRole('dialog').click(); // clear focus
+
+  await page.getByTestId('timezone-select').locator('svg').click();
+  await page.getByTestId('timezone-select').getByText('GMT', { exact: true }).click();
+
+  // expect times to go back by -3
+  await expect(startEl.getByTestId('date-time-picker').getByRole('textbox')).toHaveValue('23:00');
+  await expect(endEl.getByTestId('date-time-picker').getByRole('textbox')).toHaveValue('09:00');
+
+  async function changeDatePickerTime(element: Locator, value: string) {
```
teodosii, line 32: In future iterations some parts will be reused, but for now I'll leave it like this

grafana-plugin/src/containers/RotationForm/ScheduleOverrideForm.tsx
```diff
@@ -280,4 +315,26 @@ export const ScheduleOverrideForm: FC<RotationFormProps> = (props) => {
...
     </Modal>
   );
+
+  async function onResize() {
+    setOffsetTop(await calculateOffsetTop());
+
+    setDraggablePosition({ x: 0, y: 0 });
+  }
+
+  async function calculateOffsetTop() {
```
brojd, line 325: I can see that the same function is declared in another component. Since it doesn't depend on anything fro within the component, can we move it to some common util file and avoid duplication?

brojd (approved):
nice job!

grafana-plugin/e2e-tests/schedules/addOverride.test.ts
```diff
@@ -20,3 +20,37 @@ test('default dates in override creation modal are correct', async ({ adminRoleP
...
+  await expect(overrideEndEl.getByTestId('date-time-picker').getByRole('textbox')).toHaveValue('12:00');
+
+  await page.getByRole('dialog').click(); // clear focus
+
+  await page.getByTestId('timezone-select').locator('svg').click();
+  await page.getByTestId('timezone-select').getByText('GMT', { exact: true }).click();
+
+  // expect times to go back by -3
+  await expect(overrideStartEl.getByTestId('date-time-picker').getByRole('textbox')).toHaveValue('23:00');
+  await expect(overrideEndEl.getByTestId('date-time-picker').getByRole('textbox')).toHaveValue('09:00');
```
brojd, line 47: On CI this test can run in different timezones so you can't really predict how many hours will be added / subtracted. I think to make it reliable the test can include the following steps:
1. Set GMT in timezone select
2. Add override
3. Change offset in timezone select (let's say to GMT-3)
4. Assert times in date time picker

teodosii: I thought by default its set to GMT+3, I'll just call the fuction that sets it to Moscow time then as in other places. It changes back to GMT so there's a 3 hour difference, initially I wanted to fill in a value like Hawaii to have a larger gap, but I couldnt figure out how to instruct playwright do it as the component is being rendered in a portal

Issue #4353, opened 2024-05-16
Add filters to users page


raphael-batte:
At the moment it's not possible to filter the list of users by specific criteria but there are some requests to have multiple options to filter users.

Suggested filters:

- Team (list)
- Notification methods (list incudes available for stack options: Slack, mobile app, phone call etc)
- Important policy: toggle y/n
- Warnings: toggle y/n

joeyorlando:
@iskhakov I saw you have #4575 open, do you mind adding an estimate to this issue? 🙏

iskhakov:
> @iskhakov I saw you have #4575 open, do you mind adding an estimate to this issue? 🙏

Done @joeyorlando

PR #4575, opened 2024-06-24
Add filtering by team, is_currently_oncall and search on the user page

iskhakov:
# What this PR does

This PR adds filtering by team and is_currently_oncall on the user page

Closes https://github.com/grafana/oncall/issues/4353

## Checklist


grafana-plugin/src/pages/users/Users.tsx
```diff
@@ -250,6 +230,37 @@ class Users extends React.Component<UsersProps, UsersState> {
...
+    const currentTablePage = filtersStore.currentTablePageNum[PAGE.Users];
+
+    LocationHelper.update({ p: currentTablePage }, 'partial');
+
+    this.setState({ filters }, () => {
+      this.updateUsers();
+    });
+  };
+
+  applyFilters = () => {
```
brojd, line 260: Seems this method is unused, can we remove it? Not related to your changes but there's another unused method in this component - `renderContacts`, we could remove it too

brojd (approved):

engine/apps/api/views/user.py
```diff
@@ -841,6 +842,25 @@ def export_token(self, request, pk) -> Response:
...
+            {
+                "name": "team",
+                "type": "team_select",
+                "href": api_root + "teams/",
+                "global": True,
+            },
+        ]
+
+        if filter_name is not None:
+            filter_options = list(filter(lambda f: filter_name in f["name"], filter_options))
```
vadimkerr, line 860: does the frontend use this search functionality? if not, I think it might make sense to remove this

vadimkerr (approved):

engine/apps/api/views/user.py
```diff
@@ -841,6 +842,25 @@ def export_token(self, request, pk) -> Response:
             return Response(status=status.HTTP_204_NO_CONTENT)
         return Response(status=status.HTTP_405_METHOD_NOT_ALLOWED)
 
+    @action(methods=["get"], detail=False)
```
joeyorlando, line 845: I believe we have openapi docs for the `/users` routes ([code](https://github.com/grafana/oncall/blob/dev/engine/settings/base.py#L326)). do we need to add anything here to document this new endpoint?

joeyorlando (approved):

engine/apps/api/views/user.py
iskhakov: yes, it does

engine/apps/api/views/user.py
iskhakov: good point, added

engine/apps/api/views/user.py
vadimkerr: I can't find any usages in the frontend code 🤔 Am I missing something?

dmitry-tiger:
Hello @iskhakov, can you please give comments why this code was removed at https://github.com/grafana/oncall/pull/4575/commits/9dbcf017fab3c6862cb9f7f1759bb0eb14a856db
As I understand pr should have added is_currently_oncall filter but it was removed from filter list on backend and it in new versions I can't see that filter in place

PR #4600, opened 2024-06-27
make `make cleanup` prune volumes

vadimkerr:


Makefile
```diff
@@ -156,7 +156,8 @@ build:  ## rebuild images (e.g. when changing requirements.txt)
 cleanup: stop  ## this will remove all of the images, containers, volumes, and networks
                ## associated with your local OnCall developer setup
 	$(call echo_deprecation_message)
-	docker system prune --filter label="$(DOCKER_COMPOSE_DEV_LABEL)" --all --volumes
+	docker system prune --filter label="$(DOCKER_COMPOSE_DEV_LABEL)" --all --volumes --force
```
vadimkerr, line 159: Apparently `docker system prune --volumes` only removes anonymous volumes: https://docs.docker.com/reference/cli/docker/system/prune/#options

vadimkerr, line 159: Also adding `--force` to remove the confirmation dialog

iskhakov (approved):

PR #4604, opened 2024-07-01
Sync teams on user creation during first login

matiasb:
Include teams membership details in user header (allowing teams setup during first login user creation).
Add authentication bits for sync endpoint.


grafana-plugin/pkg/plugin/proxy.go
joeyorlando, line 1: super small thing.. should we update `.github/CODOWNERS` to have the backend team be responsible for `grafana-plugin/pkg`?

matiasb: Makes sense :+1:

brojd (approved):

Issue #4412, opened 2024-05-28
Outgoing Webhook "Forward All" field - improve UI/UX/confusion around this field wrt "Data" field


joeyorlando:
### What would you like to see!

When creating an advanced Outgoing Webhook the current UI around the configuration of "Forward All" and "Data" is a bit confusing (as reported by a customer):
![Screenshot 2024-05-28 at 13 37 43](https://github.com/grafana/oncall/assets/9406895/5de6c009-5b31-4314-8f3a-2bbf90a448c5)

When "Forward All" is enabled (which is the default), this essentially overrides whatever the user inputs into the "Data" field. The wording in the form aren't entirely clear that this is the case. We should either improve wording here, or as another possible solution, maybe hide "Data" if "Forward All" is enabled.

### Product Area

Other

raphael-batte:
During the latest UI work, there was a new possible direction for the webhook drawer. We can use this specific part from it:

![Image](https://github.com/grafana/oncall/assets/102970268/2d360d68-6210-4e62-8415-4d6dd49248f7)

or update the entire drawer:

![Image](https://github.com/grafana/oncall/assets/102970268/96cefe78-ac78-40d1-a747-d2fa460ca936)


There is also an issue that is relevant to such work https://github.com/grafana/oncall/issues/3409

PR #4622, opened 2024-07-08
Webhook form - changed Forward all

teodosii:
# What this PR does

- added validation for `Data` field, backend seems to be a bit clumsy on validating that field
- Changed `Fordwar All` / `Customise data` to be more straightforward for the user

Closes https://github.com/grafana/oncall/issues/4412


grafana-plugin/src/pages/outgoing_webhooks/OutgoingWebhooks.tsx
```diff
@@ -415,11 +415,7 @@ class OutgoingWebhooks extends React.Component<OutgoingWebhooksProps, OutgoingWe
   };
 
   handleOutgoingWebhookFormHide = () => {
-    const { history } = this.props;
-
     this.setState({ outgoingWebhookId: undefined, outgoingWebhookAction: undefined });
```
teodosii, line 418: The current behavior was that after hiding the webhook form the user would be redirect to `/outgoing_webhooks` but this way it does a page refresh and it loses the filters. This remains as a TODO to figure out a way to refresh with filters in place, best would be just to discard current id.

brojd (approved):

PR #4624, opened 2024-07-08
Add method to send notification bundle by SMS

Ferril:
# What this PR does
Adds method to render and send notification bundle by sms.

Example of SMS message:
```
Grafana OnCall: Alert groups #1, #2, #3 and 2 more 
from stack: TestOrganization, 
integrations: Grafana Alerting and 1 more.
```

Should be merged with https://github.com/grafana/oncall/pull/4457

https://github.com/grafana/oncall-private/issues/2713

## Checklist


engine/apps/phone_notifications/models/sms.py
```diff
@@ -45,7 +45,7 @@ class Meta:
     notification_policy = models.ForeignKey(
         "base.UserNotificationPolicy", on_delete=models.SET_NULL, null=True, default=None
     )
-
+    represents_bundle_uuid = models.CharField(max_length=100, null=True, default=None, db_index=True)
```
joeyorlando, line 48: can this be an optional foreign key to a `BundledNotification` object instead?

engine/apps/phone_notifications/phone_backend.py
```diff
@@ -147,16 +162,109 @@ def notify_by_sms(self, user, alert_group, notification_policy):
...
+        channels_and_alert_groups_count = notifications.aggregate(
+            channels_count=Count("alert_receive_channel", distinct=True),
+            alert_groups_count=Count("alert_group", distinct=True),
+        )
+        # get 3 unique alert groups from notifications
+        alert_groups_to_render = []
+        for notification in notifications:
+            if notification.alert_group not in alert_groups_to_render:
+                alert_groups_to_render.append(notification.alert_group)
+                if len(alert_groups_to_render) == 3:
```
joeyorlando, line 216: what happens if there's more than 3 alert groups to notify about?

engine/apps/phone_notifications/phone_backend.py
```diff
@@ -147,16 +162,109 @@ def notify_by_sms(self, user, alert_group, notification_policy):
...
+
+        notifications = BundledNotification.objects.filter(bundle_uuid=bundle_uuid).select_related("alert_group")
+
+        if not notifications:
+            logger.info("Notification bundle is empty, related alert groups might have been deleted")
+            return
+        channels_and_alert_groups_count = notifications.aggregate(
+            channels_count=Count("alert_receive_channel", distinct=True),
+            alert_groups_count=Count("alert_group", distinct=True),
+        )
```
joeyorlando, line 210: what is this doing? 🤔

engine/apps/phone_notifications/phone_backend.py
```diff
@@ -147,16 +162,109 @@ def notify_by_sms(self, user, alert_group, notification_policy):
...
+        # get 3 unique alert groups from notifications
+        alert_groups_to_render = []
+        for notification in notifications:
+            if notification.alert_group not in alert_groups_to_render:
```
joeyorlando, line 214: would a `set` work for this?

engine/apps/phone_notifications/models/sms.py
Ferril: it represents a bunch of `BundledNotification`s, so it should be `ManyToMany`, what means +1 db table. But since we already have a unique identifier for `BundledNotification`, we can avoid adding additional many-to-many table and use this identifier instead 🤔

engine/apps/phone_notifications/phone_backend.py
Ferril: the sms message would be like `"Alert group(s) #1, #2, #3 and X more"` where `x=total alert groups to notify about - 3`, but log records will be created for all alert groups.

engine/apps/phone_notifications/phone_backend.py
Ferril: gets number of unique `alert_receive_channel`s and `alert_group`s in notifications to pass it to sms message renderer to add "and X more" to the message if needed.
Example of the rendered sms message with 5 alert groups and 2 integrations:
```
Grafana OnCall: Alert group(s) #1, #2, #3 and 2 more 
from stack: TestOrganization, 
integration(s): Grafana Alerting and 1 more.
```

engine/apps/phone_notifications/phone_backend.py
Ferril: yes, makes sense to use it instead 👍

Ferril: it this case it doesn't work, since it's not subscriptable

engine/apps/phone_notifications/phone_backend.py
```diff
@@ -148,16 +163,109 @@ def notify_by_sms(self, user, alert_group, notification_policy):
...
+
+        from apps.alerts.models import BundledNotification
+        from apps.base.models import UserNotificationPolicy, UserNotificationPolicyLogRecord
+
+        notifications = BundledNotification.objects.filter(bundle_uuid=bundle_uuid).select_related("alert_group")
+
+        if not notifications:
+            logger.info("Notification bundle is empty, related alert groups might have been deleted")
+            return
+        channels_and_alert_groups_count = notifications.aggregate(
```
vadimkerr, line 208: maybe move this to `AlertGroupSMSBundleRenderer`? and change the `AlertGroupBundleBaseRenderer` to something like this:


```python3
class AlertGroupBundleBaseRenderer:
    MAX_ALERT_GROUPS = 3
    MAX_CHANNELS = 1

    def __init__(self, notifications):
        self.notifications = notifications
```

and

```python3
class AlertGroupSMSBundleRenderer(AlertGroupBundleBaseRenderer):
    def render(self) -> str:
        channels_and_alert_groups_count = notifications.aggregate(...)
        ...
```

I think it makes it more clear, otherwise the rendering logic is spread across different places.

Ferril: makes sense 👍

engine/apps/alerts/incident_appearance/renderers/sms_renderer.py
```diff
@@ -24,3 +28,25 @@ def render(self):
...
+        ]
+        numbers_str = ", ".join(f"#{x}" for x in alert_group_inside_organization_numbers)
+
+        other_alert_groups_count = self.alert_groups_count - len(self.alert_groups_to_render)
+        if other_alert_groups_count > 0:
+            numbers_str += f" and {other_alert_groups_count} more"
+
+        channel_names = ", ".join([channel.short_name for channel in self.channels_to_render])
+        if self.channels_count > 1:
+            channel_names += f" and {self.channels_count - len(self.channels_to_render)} more"
```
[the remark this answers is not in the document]
Ferril: good catch, thanks, I'll fix it 👍 we render only one channel for sms, but if this is changed, this part won't work properly

engine/apps/alerts/incident_appearance/renderers/sms_renderer.py
```diff
@@ -24,3 +30,53 @@ def render(self):
             f"integration: {self.alert_group.channel.short_name}, "
             f"alerts registered: {self.alert_group.alerts.count()}."
         )
+
+
+class AlertGroupSMSBundleRenderer(AlertGroupBundleBaseRenderer):
```
mderynck, line 35: While we are here do we want to do some further segment optimization on both bundled and non bundled too? Happy to open an issue and do it as a separate task too.  Since our integration names often contain emojis we are getting around 1/2 the characters per segment [(doc)](https://www.twilio.com/docs/glossary/what-sms-character-limit), [Calculator](https://twiliodeved.github.io/message-segment-calculator/).  Potentially we could be much more aggressive with substrings here to ensure we limit segments.  For example `Grafana OnCall: # alert groups @TestOrganization / Grafana Alerting 🦗`

Ferril: That means, that we pay x2/x3 for SMS with emoji, right? 😨
I think we can exclude emojis from SMS. I'm assuming that integrations created after we disabled auto-naming rarely have emoji in the title
Anyway, I'd propose to open a separate issue for that and continue discussion there

PR #4628, opened 2024-07-08
notify user task patch + small update to user notification rules public API docs

joeyorlando:
# What this PR does

Patches a small bug noticed (locally) by @Ferril 🙏  + updates our user notification rules public API docs to include `notify_by_msteams` as a valid `type` value (cloud only)

## Checklist

- [ ] Unit, integration, and e2e (if applicable) tests updated
- [ ] Documentation added (or `pr:no public docs` PR label added if not required)
- [x] Added the relevant release notes label (see labels prefixed w/ `release:`). These labels dictate how your PR will
    show up in the autogenerated release notes.

Issue #4410, opened 2024-05-28
Add ability to manage user notification policy resources via Terraform and Crossplane


joeyorlando:
### What would you like to see!

As the title mentions, it would be handy to be able to manage a user's personal notification policies via Terraform.

### Product Area

Terraform

joeyorlando:
**NOTE**: once this issue is completed we can _probably_ close #1193

joeyorlando:
This is currently blocked (see conversation [here](https://raintank-corp.slack.com/archives/C04JCU51NF8/p1719520920744319))

joeyorlando:
✅ 
https://registry.terraform.io/providers/grafana/grafana/latest/docs/resources/oncall_user_notification_rule

https://marketplace.upbound.io/providers/grafana/provider-grafana/v0.18.0/resources/oncall.grafana.crossplane.io/UserNotificationRule/v1alpha1

PR #4635, opened 2024-07-08
patch default user notification policy changes

joeyorlando:
# What this PR does

This is a follow-up PR to https://github.com/grafana/oncall/pull/4628. As @Ferril pointed out, there was a slight issue in `apps.alerts.tasks.notify_user.perform_notification` method when using a "fallback"/default user notification policy. This is because the `log_record_pk` arg passed into `perform_notification` will fetch the `UserNotificationPolicyLogRecord` object, but that object will have a `notification_policy` set to `None` (because there's no persistent `UserNotificationPolicy` object to refer to).

Instead we now pass in a second argument to `perform_notification`, `use_default_notification_policy_fallback`. If this is true, simply grab the transient/in-memory `UserNotificationPolicy` and use that inside of this task

Related to https://github.com/grafana/oncall/issues/4410

## Checklist

### DIFF
--- a/engine/apps/alerts/tasks/notify_user.py
+++ b/engine/apps/alerts/tasks/notify_user.py
@@ -121,9 +121,7 @@ def notify_user_task(
             reason = None
 
         def _create_user_notification_policy_log_record(**kwargs):
-            if using_fallback_default_notification_policy_step and "notification_policy" in kwargs:
-                kwargs["notification_policy"] = None
-            return UserNotificationPolicyLogRecord(**kwargs)
+            return UserNotificationPolicyLogRecord(**kwargs, using_fallback_default_notification_policy_step=using_fallback_default_notification_policy_step)
 
         if notification_policy is None:
             stop_escalation = True
@@ -211,14 +209,25 @@ def notify_user_task(
         if not stop_escalation:
             if notification_policy.step != UserNotificationPolicy.Step.WAIT:
 
-                def _create_perform_notification_task(log_record_pk, alert_group_pk):
-                    task = perform_notification.apply_async((log_record_pk,))
+                def _create_perform_notification_task(
+                    log_record_pk, alert_group_pk, use_default_notification_policy_fallback
+                ):
+                    task = perform_notification.apply_async(
+                        (log_record_pk, use_default_notification_policy_fallback)
+                    )
                     task_logger.info(
                         f"Created perform_notification task {task} log_record={log_record_pk} "
                         f"alert_group={alert_group_pk}"
                     )
 
-                transaction.on_commit(partial(_create_perform_notification_task, log_record.pk, alert_group_pk))
+                transaction.on_commit(
+                    partial(
+                        _create_perform_notification_task,
+                        log_record.pk,
+                        alert_group_pk,
+                        using_fallback_default_notification_policy_step,
+                    )
+                )
 
             delay = NEXT_ESCALATION_DELAY
             if countdown is not None:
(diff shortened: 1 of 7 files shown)


grafana-plugin/e2e-tests/utils/userSettings.ts
joeyorlando, line 1: These changes _should_ fix [this](https://github.com/grafana/oncall/actions/runs/9843903564/job/27176150098#step:20:15399) failing e2e test (which started failing after we recently got rid of autogenerating default user notification policy steps)

engine/apps/alerts/tasks/notify_user.py
Ferril, line 1: I found another issue. It seems like there is an endless loop of notification. In the `notify_user_task` we call next `notify_user_task` here (line 241):
```
transaction.on_commit(
    partial(
        notify_user_task.apply_async,
        (user.pk, alert_group.pk, notification_policy.pk, reason),
        ...
  )
```
The next task will start with `previous_notification_policy_pk=None`, notify user again and call next `notify_user_task` with the same parameters -> 🔁 

The regular flow is:
1. call `notify_user_task` with `previous_notification_policy_pk=None`
2. get the first notification policy
3. create log with type `TRIGGERED` and call `perform_notification`
4. call next `notify_user_task` with `previous_notification_policy_pk=<current notification_policy.pk>`
5. in the next `notify_user_task` get notifiction_policy.next()
6. if it's not None: repeat from step 3. Otherwise create notification log with type `FINISHED` and stop notification.

So for one existing notification step the `notify_user_task` is called twice.
For default notification we can add a flag to `notify_user_task` when we call it again OR we can skip steps 4 and 5, create `FINISHED` log and stop notification 🤔 wdyt?

engine/apps/email/tasks.py
```diff
@@ -65,7 +78,7 @@ def notify_user_async(user_pk, alert_group_pk, notification_policy_pk):
 
     emails_left = user.organization.emails_left(user)
     if emails_left <= 0:
-        UserNotificationPolicyLogRecord.objects.create(
+        _create_user_notification_policy_log_record(
```
Ferril, line 81: There are a few more UserNotificationPolicyLogRecord in this task, that need to be updated 👀

joeyorlando: good catch! addressed in https://github.com/grafana/oncall/pull/4635/commits/a70b1dfdf19bc11ae4cacd9ce42688867358e7c2

engine/apps/alerts/tasks/notify_user.py
joeyorlando: great catch and thanks for the really detailed breakdown, this was super helpful in understanding the flow of things 🫶  This should now be addressed in https://github.com/grafana/oncall/pull/4635/commits/716ad63039c862a21175191922345b37a6e0d2fa

**Before**
Screenshot 2024-07-09 at 10 39 03

**After**
Screenshot 2024-07-09 at 10 39 10

engine/apps/alerts/tasks/notify_user.py
```diff
@@ -42,7 +43,6 @@ def notify_user_task(
 
     countdown = 0
     stop_escalation = False
-    log_message = ""
```
joeyorlando, line 45: `log_message` is _only_ ever updated [here](https://github.com/grafana/oncall/pull/4635/files#diff-fa8513da06f09885842f55c35f819b5a14dd5e3b82ed8a5e1c4719bb64f1c447L137), but then it is never referenced/used anywhere

engine/apps/alerts/tasks/notify_user.py
Ferril: LGTM 👍

Ferril (approved):

Issue #1908, opened 2023-05-09
Make Declare incident button visible by default in Slack


raphael-batte:
An important button for **declaring an incident** is invisible by default in Slack.
I think that it would be good to move it to the first row instead of **Attach to** button

Matvey-Kuk:
Last time I tested, it's moving not only "Attach to", but also Snooze which is used pretty frequently.

PR #4642, opened 2024-07-09
Show incident button first when alert group is acknowledged

iskhakov:
# What this PR does

* Move declare incident button to first position when incident is acknowledged
* Move resolved text above buttons, so buttons remain on the same place when clicked
Screenshot 2024-07-09 at 11 52 33 PM
Screenshot 2024-07-09 at 11 52 41 PM

Closes https://github.com/grafana/oncall/issues/1908

## Checklist


engine/apps/alerts/incident_appearance/renderers/slack_renderer.py
```diff
@@ -301,7 +310,7 @@ def _get_buttons_blocks(self):
             buttons.append(resolution_notes_button)
 
             # Declare incident button
-            if self.alert_group.channel.organization.is_grafana_incident_enabled:
+            if self.alert_group.channel.organization.is_grafana_incident_enabled and not self.alert_group.acknowledged:
```
joeyorlando, line 313: shouldn't you still be able to see the Declare Incident button even if the Alert Group is ack'd?

engine/apps/alerts/incident_appearance/renderers/slack_renderer.py
```diff
@@ -194,6 +194,15 @@ def _get_buttons_blocks(self):
                         },
                     )
                 else:
+                    if True or self.alert_group.channel.organization.is_grafana_incident_enabled:
```
joeyorlando, line 197: `if True`

joeyorlando (approved):

engine/apps/alerts/incident_appearance/renderers/slack_renderer.py
iskhakov:
```suggestion
                    if self.alert_group.channel.organization.is_grafana_incident_enabled:
```

engine/apps/alerts/incident_appearance/renderers/slack_renderer.py
iskhakov: if it is ack'd we add the button in the first place

joeyorlando: gotcha 👍 

what do you think about defining the `incident_button` variable higher up so that we don't need to duplicate the definition on lines 198 and 314?

PR #4643, opened 2024-07-09
helm: allow configuration of smtp credentials via existing secret

hendrikheil:
# What this PR does

This PR adds three keys to the helm chart values: `smtp.existingSecret`, `smtp.usernameKey` and `smtp.passwordKey` which allows for secure configuration of the smtp integration.

We noticed this, as we wanted to set all integration keys through secrets. We tried to bypass the Helm chart behavior by defining environment variables through `env` directly. However as we use Argo CD which has to diff the current state with the desired state, it caused issues. Diffing requires environment key names to be unique, in order to properly create a diff, unfortunately. Instead of just not setting any of the environment keys that the chart is currently setting, I opted to just allow setting the smtp credentials through an existing secret, which is how the other integrations handle it too.

I also added the usernameKey to the secret, as we use Postmark for SMTP where the "password" is also the username, so it needs to be in a secret too.

There is no related existing issue yet.

## Checklist


joeyorlando (approved):
LGTM 👍 thanks for the contribution @hendrikheil!

joeyorlando (commented):
do you mind adding a few helm unit tests? (examples in [here](https://github.com/grafana/oncall/tree/dev/helm/oncall/tests))

More instructions on how to run these tests locally [here](https://github.com/grafana/oncall/tree/dev/dev#helm-unit-tests)

PR #4644, opened 2024-07-09
Don't consider email message limit exceeded notification failure

mderynck:
# What this PR does
If an email message notification exceeds the limit and does not get sent it is expected and can be treated like other rate-limit violations, although the notification does not happen it is behaving as expected.  This change is to reduce some noise in the notification failure alerts.

An alternative would be to ignore this specific condition in the queries for `check_alert_group_personal_notifications_task` but this seemed more straight forward.

Closes [issue link here]

## Checklist

### DIFF
--- a/engine/apps/email/tasks.py
+++ b/engine/apps/email/tasks.py
@@ -80,7 +80,7 @@ def notify_user_async(user_pk, alert_group_pk, notification_policy_pk):
     if emails_left <= 0:
         _create_user_notification_policy_log_record(
             author=user,
-            type=UserNotificationPolicyLogRecord.TYPE_PERSONAL_NOTIFICATION_FAILED,
+            type=UserNotificationPolicyLogRecord.TYPE_PERSONAL_NOTIFICATION_FINISHED,
             notification_policy=notification_policy,
             alert_group=alert_group,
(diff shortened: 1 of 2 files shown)


engine/apps/email/tasks.py
```diff
@@ -80,7 +80,7 @@ def _create_user_notification_policy_log_record(**kwargs):
     if emails_left <= 0:
         _create_user_notification_policy_log_record(
             author=user,
-            type=UserNotificationPolicyLogRecord.TYPE_PERSONAL_NOTIFICATION_FAILED,
+            type=UserNotificationPolicyLogRecord.TYPE_PERSONAL_NOTIFICATION_FINISHED,
```
matiasb, line 83: hm.. I think the FINISHED log record indicates the end of the personal policies (it stops escalation for user), so may not be the best option here (or it could be confusing). I think we could maybe add a new type here, or even consider using SUCCESS (updating the rendered log accordingly) since we did our part. Note we would have the same issue with SMS and phone calls, I think? Alternatively, yeah, ignoring FAILED log records for limit exceeded entries would work too.

mderynck:
Needs a bit more investigation before proceeding.

PR #4648, opened 2024-07-10
Update consts.ts

iskhakov:
# What this PR does

Closes [issue link here]

## Checklist

### DIFF
--- a/grafana-plugin/src/utils/consts.ts
+++ b/grafana-plugin/src/utils/consts.ts
@@ -32,7 +32,7 @@ export const PLUGIN_ROOT = '/a/grafana-oncall-app';
 
 // Environment options list for onCallApiUrl
 export const ONCALL_PROD = 'https://oncall-prod-us-central-0.grafana.net/oncall';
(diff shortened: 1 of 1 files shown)


brojd (approved):

joeyorlando (changes requested):
I think this is proof that these shouldn't be hardcoded.. we discussed this a while back and decided to take the lower effort way out at the time.

My opinion would be to inject these env vars at build time, rather than having the app even need to know what prod/ops/dev is.

joeyorlando (commented):
@mderynck also had some thoughts here. Instead of adding these consts at plugin build time, read them from a (new?) endpoint.

Maybe [this one](https://github.com/grafana/oncall/blob/dev/engine/apps/grafana_plugin/views/status.py#L76-L90)?

iskhakov:
@joeyorlando let's do this work within the new issue/pr scope. I need to merge this one as part of https://github.com/grafana/oncall-private/issues/2609

joeyorlando (approved):

PR #4655, opened 2024-07-10
Handle inbound email integration alert getting multiple recipients

matiasb:
Related to https://github.com/grafana/oncall-private/issues/2683
(when using mailgun backend, you can get multiple recipients, keep the first one matching the domain; other backends seem to just return the first one)


vadimkerr (approved):
LGTM

PR #4666, opened 2024-07-11
Add task to trigger Grafana to send sync data

mderynck:
- Adds a new scheduled task (default every 5 mins but configurable) to make requests to initiate Grafana instances to send sync data to OnCall
- Backend plugin sync endpoint can also be used to manually send sync data.
- Backend plugin sync endpoint adds lock so only 1 can sync at a time, currently returns 400 if already in use. Callers are engine and manual users so this should be ok, engine will just log it.
- Removed routes from plugin.json (I don't think we need these anymore)
- Needed to add plugins.app:access permission to engine can call plugin.


brojd (approved):

engine/settings/base.py
```diff
@@ -509,6 +509,7 @@ class BrokerTypes:
 )
 
 CELERY_BEAT_SCHEDULE_FILENAME = os.getenv("CELERY_BEAT_SCHEDULE_FILENAME", "celerybeat-schedule")
+CELERY_BEAT_SYNC_ORGANIZATIONS_INTERVAL_MINUTES = os.getenv("CELERY_BEAT_SYNC_ORGANIZATIONS_INTERVAL_MINUTES", 5)
```
matiasb, line 512: Wondering if we should start with a more conservative interval?

engine/apps/grafana_plugin/tasks/sync_v2.py
```diff
@@ -0,0 +1,42 @@
...
+    for idx, org in enumerate(organization_qs):
+        client = GrafanaAPIClient(api_url=org.grafana_url, api_token=org.api_token)
+        _, status = client.sync()
+        if status["status_code"] != 200:
+            logger.error(
+                f"Failed to request sync stack_slug={org.stack_slug} status_code={status['status_code']} url={status['url']} message={status['message']}"
+            )
+        if idx % orgs_per_second == 0:
+            logger.info(f"Sleep 1s after {idx + 1} organizations processed")
+            sleep(1)
```
matiasb, line 42: Does this mean we will have a ~4min running task every 5mins? Will that keep one worker busy most of the time?

engine/settings/base.py
mderynck: I thought we can give it a shot at this frequency. Since they are spread out the worst case still is an org is at the beginning in one task an then the end of another resulting in 9 mins between sync. Maybe that can be adjusted see next comment.

engine/apps/grafana_plugin/tasks/sync_v2.py
mderynck: Yeah the thinking here was to avoid having the task slam requests as fast as possible. Not sure it is worth doing or we should try to get through them as fast as possible.  Another option could be running a few workers concurrently which handle different hashes of ids to divide the work. wdyt?

matiasb: +1 to spread the updates. I guess we can see how this behaves and have a plan in case we need to tune settings if needed? Also we may need to keep track of the timing of sync requests to know how this could affect latency too? Eventually if we could skip no-op syncs would be great (and/or apply "diffs" as we also discussed).

matiasb (approved):

PR #4687, opened 2024-07-17
Label error handling tweaks

teodosii:
# What this PR does

These changes make it so that we no longer show an error notification when you first add an empty pair `key/value` due to the undefined key. Other than that, the handling remains the same, with some refactorings as we don't need the try catch clauses.

Would be nice in the future to find a way not to duplicate the usage of those 2 functions that fetch the key and the value and share them somehow 🤔


grafana-plugin/src/containers/RouteLabelsDisplay/RouteLabelsDisplay.tsx
```diff
@@ -32,28 +32,15 @@ export const RouteLabelsDisplay: React.FC<RouteLabelsDisplayProps> = ({ labels,
...
-      openErrorNotification('There was an error processing your request. Please try again');
-    }
-
+    const result = await labelsStore.loadKeys(search);
     return splitToGroups(result);
   };
 
   const onLoadValuesForKey = async (key: string, search?: string) => {
-    let result = undefined;
+    if (!key) {return [];}
```
brojd, line 40: looks like wrong formatting

brojd (approved):

grafana-plugin/src/containers/RouteLabelsDisplay/RouteLabelsDisplay.tsx
teodosii: Yeah, `yarn run lint:fix` seems not to complain on this one... I don't really understand why `pre-commit` yields different results compared to `lint:fix`, they look the same

Issue #4031, opened 2024-03-07
App plugin: Update to react-router v6


leventebalogh:
Starting from Grafana 10, plugins can start using the v6 of react-router. Overall, react-router v6 aims to simplify route configuration and provide a more flexible and intuitive API for developers.

### Why?
We are going to deprecate using react-router v5 in plugins from Grafana 11, and we would like to make sure our internal app plugins are using the v6 version by the time we finish migrating Grafana core itself.

### How?
We have created a migration guide for plugins (please see the links below), and also don't hesitate to reach out in `#grafana-plugins-platform` in Slack in case you have any questions.

- [How to migrate a Grafana app plugin to react-router v6](https://grafana.com/developers/plugin-tools/migration-guides/update-from-grafana-versions/migrate-10_0_x-to-10_1_x#update-to-react-router-v6)
- [Official React Router v5 to v6 migration guide](https://reactrouter.com/en/main/upgrading/v5)
- [Community topic](https://community.grafana.com/t/migrating-app-plugins-to-use-react-router-v6/115410)

joeyorlando:
added this to "To Be Groomed" because it appears we are currently using a deprecated version of react router (Grafana 11 has been out for ~2 months as of the writing of this comment)

teodosii:
Blocked until https://github.com/grafana/scenes/issues/608 is sorted out - Right now all other pages in OnCall work with v6, except for Insights which uses scenes

teodosii:
Opened a PR https://github.com/grafana/oncall/pull/4703 for it - it changes OnCall so that we rely on the `react-router-compat` version but its using v6 syntax. Can't fully migrate to v6 as grafana/scenes still relies on v5 and that seems to break rendering on the Insights page

PR #4703, opened 2024-07-18
Migrate react-router to v6

teodosii:
# What this PR does

- Migrate react-router from v5 to v6

Closes https://github.com/grafana/oncall/issues/4031


teodosii:
Blocked until scenes migrates to react-router v6

grafana-plugin/.config/webpack/webpack.config.ts
```diff
@@ -46,11 +50,11 @@ const config = async (env): Promise<Configuration> => {
       '@grafana/slate-react',
       'react',
       'react-dom',
+      'react-router-dom',
```
teodosii, line 53: This seems to be needed even with the compat version, otherwise Insights won't render due to their dependency on v5.

grafana-plugin/.config/webpack/webpack.config.ts
```diff
@@ -199,7 +240,7 @@ const config = async (env): Promise<Configuration> => {
 
   if (isWSL()) {
     baseConfig.watchOptions = {
-      poll: 3000,
+      // poll: 3000,
```
teodosii, line 243: on WSL this isn't really needed. I use WSL and this breaks reloading actually, doesn't seem to be configurable :(

grafana-plugin/.eslintrc.js
```diff
@@ -8,6 +8,10 @@ module.exports = {
     'import/internal-regex':
       '^assets|^components|^containers|^contexts|^icons|^models|^network|^pages|^services|^state|^utils|^plugin',
   },
+  parserOptions: {
```
brojd, line 11: is it needed here since we have it in overrides already? Seems it's a duplication in this case?

grafana-plugin/src/pages/escalation-chains/EscalationChains.tsx
```diff
@@ -480,4 +495,6 @@ class _EscalationChainsPage extends React.Component<EscalationChainsPageProps, E
   };
 }
 
-export const EscalationChainsPage = withRouter(withMobXProviderContext(withTheme2(_EscalationChainsPage)));
+export const EscalationChainsPage = withRouter<RouteProps, Omit<EscalationChainsPageProps, 'store' | 'meta' | 'theme'>>(
+  withMobXProviderContext(withTheme2(_EscalationChainsPage))
+); // as unknown as React.ComponentClass<Omit<EscalationChainsPageProps, 'store'>>;
```
brojd, line 500: can we remove comment?

grafana-plugin/src/pages/incident/incident.styles.ts
```diff
@@ -0,0 +1,220 @@
+import { css } from '@emotion/css';
```
brojd, line 1: nit: seems our convention for file naming is `Incident.styles.ts` (uppercase)

brojd (commented):
great job! let me just make sure tomorrow if this will work ok in the IRM monorepo setup (incident uses v5.3.3)

grafana-plugin/.eslintrc.js
teodosii: I'd say not as I undo-ed the other packages' updates. But I believe once you bump grafana/eslint, it is needed.

grafana-plugin/src/pages/incident/incident.styles.ts
teodosii: Didn't notice it, good catch

brojd (approved):

PR #4714, opened 2024-07-22
Fix slack app redirect

Konstantinov-Innokentii:


engine/apps/chatops_proxy/utils.py
```diff
@@ -29,7 +30,7 @@ def get_installation_link_from_chatops_proxy(user) -> typing.Optional[str]:
         link, _ = client.get_slack_oauth_link(
             org.stack_id,
             user.user_id,
-            org.web_link,
+            urljoin(org.web_link, "settings?p=1&tab=TeamsSettings&chatOpsTab=Slack"),
```
vadimkerr, line 33: Is this `p=1` param required here?

```suggestion
            urljoin(org.web_link, "settings?tab=TeamsSettings&chatOpsTab=Slack"),
```

vadimkerr (approved):

engine/apps/chatops_proxy/utils.py
Konstantinov-Innokentii: Removed, for some reason it was in my browser address bar

PR #4716, opened 2024-07-23
[OnCall Initialization] Backend plugin call caching

mderynck:
- Manual sync endpoint
- Add cache to settings to avoid 2 calls on every proxy call to get other plugin settings
- Add cache to all users to avoid getting all users on every proxy call and so it can be reused
- Add cache for user objects to avoid repeated calls for teams and permissions for every proxy call
- All caches currently expire after 60s.
- Temporary add call counters for debugging (Remove after testing)


grafana-plugin/pkg/plugin/settings.go
```diff
@@ -92,11 +101,10 @@ func (a *App) OnCallSettingsFromContext(ctx context.Context) (*OnCallPluginSetti
 
 	if settings.License == "" {
 		cloudRe := regexp.MustCompile(CLOUD_VERSION_PATTERN)
-		cloudDevRe := regexp.MustCompile(CLOUD_DEV_VERSION_PATTERN)
 		ossRe := regexp.MustCompile(OSS_VERSION_PATTERN)
-		if version == DEV_PATTERN || ossRe.MatchString(version) {
+		if ossRe.MatchString(pluginContext.PluginVersion) {
```
matiasb, line 105: Should we still use the `version` value here? (added to support previous Grafana versions where version wasn't available in context)

grafana-plugin/pkg/plugin/debug.go
```diff
@@ -7,6 +7,17 @@ import (
...
 
+type OnCallDebugStats struct {
+	SettingsCallCount           int32 `json:"settingsCallCount"`
+	AllUsersCallCount           int32 `json:"allUsersCallCount"`
+	PermissionsCallCount        int32 `json:"permissionsCallCount"`
+	AllPermissionsCallCount     int32 `json:"allPermissionsCallCount"`
+	TeamForUserCallCount        int32 `json:"teamForUserCallCount"`
+	AllTeamsCallCount           int32 `json:"allTeamsCallCount"`
+	TeamMembersForTeamCallCount int32 `json:"teamMembersForTeamCallCount"`
+	CheckHealthCallCount        int32 `json:"checkHealthCallCount"`
```
matiasb, line 18: Curious, could we assume there is a health call per each plugin request? (trying to think how we map these counts to API calls)

grafana-plugin/pkg/plugin/settings.go
```diff
@@ -128,22 +136,88 @@ func (a *App) OnCallSettingsFromContext(ctx context.Context) (*OnCallPluginSetti
...
+		log.DefaultLogger.Error("getting incident plugin settings", "error", err)
+	}
+	labelsPluginSettings, err := a.GetOtherPluginSettings(settings, LABELS_PLUGIN_ID)
 	if err != nil {
-		return &settings, err
+		log.DefaultLogger.Error("getting incident plugin settings", "error", err)
 	}
 
-	return &settings, nil
+	o := make(map[string]map[string]interface{})
```
matiasb, line 177: Maybe rename `o` to something more meaningful?

matiasb (approved):
LGTM, minor comments added

grafana-plugin/pkg/plugin/settings.go
mderynck: Good catch! probably why the tests are failing

mderynck: Fixed by @brojd

grafana-plugin/pkg/plugin/debug.go
mderynck: No we only do the check health call on status (initial load) and install since it is a waste for other calls since it will be contacting the engine anyways and should be already working.

brojd (approved):

PR #4686, opened 2024-07-17
Move slack log report from message thread to button

iskhakov:
# What this PR does

Continuation of [Add slack button to show log report #4641](https://github.com/grafana/oncall/pull/4641)

## Which issue(s) this PR closes

Closes https://github.com/grafana/oncall/issues/3849

## Checklist

- [x] Unit, integration, and e2e (if applicable) tests updated
- [x] Documentation added (or `pr:no public docs` PR label added if not required)
- [x] Added the relevant release notes label (see labels prefixed w/ `release:`). These labels dictate how your PR will
    show up in the autogenerated release notes.

PR #4641, opened 2024-07-09
Add slack button to show log report

iskhakov:
# What this PR does

Add a button to show an alert group log report in Slack. After deployed we can remove it from the thread as requested in https://github.com/grafana/oncall/issues/3849

Screenshot 2024-07-09 at 11 29 48 PM
Screenshot 2024-07-09 at 11 29 57 PM

## Which issue(s) this PR closes

Closes [issue link here]

## Checklist

- [ ] Unit, integration, and e2e (if applicable) tests updated
- [ ] Documentation added (or `pr:no public docs` PR label added if not required)
- [ ] Added the relevant release notes label (see labels prefixed w/ `release:`). These labels dictate how your PR will
    show up in the autogenerated release notes.

PR #4750, opened 2024-07-29
Stop posting new slack log reports

iskhakov:
Stop posting a new message, as it is available from the button (https://github.com/grafana/oncall/pull/4641)
This is an intermediate step, so we will only update posted messages but not post new ones
Once the majority of messages are updated, we can remove this step (https://github.com/grafana/oncall/pull/4686)


vadimkerr (approved):
lgtm

PR #4765, opened 2024-07-31
Feature: Use ui extension hooks where available

leventebalogh:
**What this PR does / why we need it:**
This PR updates usage of plugin extensions APIs to take advantage of the new hooks API where available. In older versions we fallback to the currently used hook. This prevents an issue where due to the reactive registry the older APIs don't receive the full list of extensions. It also paves the way for frontend performance improvements in Grafana core.

**Which issue(s) this PR fixes:**
Related: https://github.com/grafana/grafana-community-team/issues/174

**Special notes for your reviewer:**
We would really appreciate some assistance in testing this PR in both the latest version of Grafana 11 and the minimum supported Grafana version.


brojd:
It looks good in general but I want to test it properly just after hackathon + we need to disable deprecation/deprecation eslint rule for now (we've already created a dedicated task to replace deprecated components and then we will enable the eslint rule back). Does it sound ok

brojd (approved):

PR #4779, opened 2024-08-06
Bump kind to match ops-devenv version

Konstantinov-Innokentii:


joeyorlando (approved):

joeyorlando (commented):
consider bumping the version here as well

https://github.com/grafana/oncall/blob/dev/helm/kind.yml#L5

PR #4813, opened 2024-08-13
Deleted organization fixes

vadimkerr:
# What this PR does

Makes it so that:
* Shift swap requests are not created from Google Calendar for deleted orgs
* On-call shift mobile app notifications are not sent to users from deleted orgs

## Checklist


joeyorlando (approved):
LGTM 👍 

_nit_/question: I wonder if it'd be at all useful to move the `organization__deleted_at__isnull=True` logic to the `QuerySet.filter` method for `OnCallSchedule` and `GoogleOAuth2User`. Such that by default you never see these objects ([similar to what we do for the `User` model](https://github.com/grafana/oncall/blob/dev/engine/apps/user_management/models/user.py#L82-L86))

vadimkerr:
@joeyorlando yes that could also work but in other places we filter out deleted instances inside of Celery tasks so I wanted to keep things consistent

PR #4830, opened 2024-08-15
Speed up internal api endpoints

Ferril:
# What this PR does
Reduces number of calls to db for `/schedules`, `/alertgroups` and `/users` endpoints.
Fixes the issue when there was an additional call to db to get organization url to build user avatar full link.

Related to [issue link here]

## Checklist


engine/apps/api/serializers/user.py
```diff
@@ -324,6 +329,10 @@ class Meta:
             "avatar_full",
         ]
 
+    def get_avatar_full(self, obj):
```
matiasb, line 332: Is there any way to reuse the implementation above? (to avoid the duplication)

engine/apps/user_management/models/user.py
```diff
@@ -195,9 +195,12 @@ def google_oauth2_token_is_missing_scopes(self) -> bool:
...
         return not google_utils.user_granted_all_required_scopes(self.google_oauth2_user.oauth_scope)
 
-    @property
-    def avatar_full_url(self):
-        return urljoin(self.organization.grafana_url, self.avatar_url)
+    def avatar_full_url(self, organization: "Organization"):
+        """
+        Use arg `organization` instead of `self.organization` to avoid multiple requests to db when getting avatar for
+        users list
+        """
```
matiasb, line 202: May not be possible or simple, but ensuring we `select_related`  the organization for users instead should also work, right?

matiasb (approved):

engine/apps/user_management/models/user.py
Ferril: It was my first thought 🙂 But we get avatar link in many places (e.g. log.author, alert_group.resolved_by_user, swap_request.beneficiary, etc.). I think: 1. it will make db queries more complex, when it's not necessary (organization is the same for every user and we already have it in memory) 2. it will be easy to forget to add `select_related` in every new place where we will need user info

engine/apps/api/serializers/user.py
Ferril: We have so many user serializers that inherit from each other, so I think it would be better to reuse this since it's just a couple of lines to avoid additional overcomplication 🤔 wdyt?

matiasb: Sounds ok, I guess eventually we may want to define a mixin class that it can be inherited to provide the functionality if there is more shared bits.

PR #4864, opened 2024-08-20
Add stack_id to Tenant dataclass

Konstantinov-Innokentii:
Fix retrying chatops_proxy tasks


vadimkerr:
also added `stack_slug` (as per https://github.com/grafana/oncall-gateway/blob/4b92eebe36990a43b4b3be013157973cf988e737/internal/api/tenants.go#L15-L21)

Issue #2911, opened 2023-08-30
Jinja template result max length too short


12doge-LEO:
### What would you like to see!

Hello, when I tried to send an image by base64, I found the length limits of Jinja template is not long enough.
Currently the length is set to 50,000 like:
```
JINJA_TEMPLATE_MAX_LENGTH = 50000
JINJA_RESULT_TITLE_MAX_LENGTH = 500
JINJA_RESULT_MAX_LENGTH = 50000

```

Can you please change this config to an editable param?

By the way, is there any better solution to send an image via grafana on-call? In my case, image-url cannot be used in this case because our user prefer to direct send an image by base64 or file 

Looking forward to your reply, thanks!

### Product Area

Alert Flow & Configuration

Hridoy-31:
Hi there!

I'd like to express my interest in working on this issue.

If you have any specific guidelines, suggestions, or additional context to provide, please feel free to share them. I'm eager to collaborate and make a positive contribution to this project.

Looking forward to your response!

buddhdev-harsh:
Hello @Matvey-Kuk ,

I am interested in working on this issue.

I think we can solve this issue in one of the 3 ways,

1.  we can remove the length for Jinja template, not sure how that will be consumed once removed (the user will be able to upload big files)
2.  we can keep this range and if the template > 50000 then prompt the user for editable params and allow user to upload the file again it can cause the same issue as above but then we can put another limit, which will make 50000 > template_size < 70000. To prevent the above case.
3.  if we consider the second case then it would be efficient to increase the limit directly to 70000, just to prevent double checking if it >50000.

I would appreciate your feedback on these solutions and how you would like them implemented for this feature.

Riksus:
@Matvey-Kuk @buddhdev-harsh  please add the ability to change from via env

Matvey-Kuk:
@Riksus I stepped back from working on OnCall, forwarding this to @joeyorlando <3

Riksus:
https://github.com/grafana/oncall/blob/7112be811681264db541b89fa94523d940d95525/engine/apps/alerts/models/channel_filter.py#L86
for this too

joeyorlando:
@Riksus we would review a PR if you're wiling to put one together 🙂

PR #4865, opened 2024-08-20
Make jinja template limits configurable by env var

Riksus:
# What this PR does
Fixing variable for jinja templates to env
Related to [issue](https://github.com/grafana/oncall/issues/2911)


engine/apps/alerts/models/channel_filter.py
```diff
@@ -83,7 +84,7 @@ class ChannelFilter(OrderedModel):
 
     created_at = models.DateTimeField(auto_now_add=True)
 
-    FILTERING_TERM_MAX_LENGTH = 1024
+    FILTERING_TERM_MAX_LENGTH = os.getenv("FILTERING_TERM_MAX_LENGTH", 1024)
```
mderynck, line 87: This should not be configurable by env var.  If it changes from one state to another after the database has been initialized then a database migration is required.  The field should be changed to a TextField instead of fixed length CharField, FILTERING_TERM_MAX_LENGTH should be removed and references to it should reference the common JINJA_TEMPLATE_MAX_LENGTH in settings.

Riksus: Does this apply only to FILTERING_TERM_MAX_LENGTH or to other changes too?

Riksus: JINJA_TEMPLATE_MAX_LENGTH = os.getenv("JINJA_TEMPLATE_MAX_LENGTH", 50000)
JINJA_RESULT_TITLE_MAX_LENGTH = os.getenv("JINJA_RESULT_TITLE_MAX_LENGTH", 500)
JINJA_RESULT_MAX_LENGTH = os.getenv("JINJA_RESULT_MAX_LENGTH", 50000)
tit's possible edit vars  on top withiout migration?

Riksus: @mderynck

mderynck: Yes the other ones should be ok to change without needing a database migration.

Riksus: @mderynck i cancel commit with FILTERING_TERM_MAX_LENGTH, pls add other changes

mderynck (approved):

Riksus:
Thanks!

mderynck:
Thank you for your contribution!

PR #4897, opened 2024-08-22
Fixed deprecated imports of H/VGroup in favor of Stack

teodosii:
# What this PR does

Closes https://github.com/grafana/irm/issues/10


grafana-plugin/.eslintrc.js
```diff
@@ -12,7 +12,7 @@ module.exports = {
     {
       files: ['src/**/*.{ts,tsx}'],
       rules: {
-        'deprecation/deprecation': 'off',
+        'deprecation/deprecation': 'warn',
```
brojd, line 15: isn't it warn by default? If so we can just remove  whole `rules` from here

teodosii: First time I attempt this by removing the rule completely didn't trigger any warnings/errors 🤔

.github/workflows/linting-and-tests.yml
```diff
@@ -242,8 +242,7 @@ jobs:
     strategy:
       matrix:
         grafana_version:
-          - 10.1.7
-          - 10.3.3
+          - 11.1.3
```
joeyorlando, line 245: nice 🔥 

should we also still test against `10.x` versions? wdyt?

joeyorlando (approved):

.github/workflows/linting-and-tests.yml
teodosii: Not really possible anymore after adding these changes, afaik these UI changes are only available in grafana 11.

teodosii: Checked&tested, lowest we can do is `10.2.5` which I have it set in the e2e yaml file

grafana-plugin/e2e-tests/outgoingWebhooks/advancedWebhook.test.ts
```diff
@@ -22,12 +22,15 @@ test('create advanced webhook and check it is displayed on the list correctly',
...
   await webhooksFormDivs.locator('[name=name]').fill(WEBHOOK_NAME);
 
-  // Select team
-  await page.getByLabel('New Outgoing Webhook').getByRole('img').nth(1).click(); // Open team dropdown
-  await page.getByLabel('Select options menu').getByText('No team').click(); // Select "No team"
+  // Open team dropdown
+  await page.getByTestId('team-selector').locator('div').filter({ hasText: 'Choose (Optional)' }).nth(1).click();
+  // Set No Team
+  await page.getByLabel('Select options menu').getByText('No team').click();
+  // await page.getByTestId('data-testid Select option').getByText('No team').click();
```
brojd, line 29: Can we remove all those comments with the old ways?

brojd (approved):

PR #4947, opened 2024-08-28
Check for incident enabled before making mobile app request

matiasb:
Related to some issues we have noticed in our logs ([example](https://ops.grafana-ops.net/goto/8sa28TqIR?orgId=1)).
Also make sure unexpected responses are logged too.

### DIFF
--- a/engine/apps/mobile_app/views.py
+++ b/engine/apps/mobile_app/views.py
@@ -197,6 +197,9 @@ class MobileAppGatewayView(APIView):
         downstream_path = kwargs["downstream_path"]
         method = request.method
         user = request.user
+        org = user.organization
+        if not org.is_grafana_incident_enabled:
+            raise NotFound(f"Incident is not enabled for organization {org.pk}")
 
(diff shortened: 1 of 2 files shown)


mderynck (approved):

engine/apps/mobile_app/views.py
```diff
@@ -197,6 +197,9 @@ def _proxy_request(self, request: Request, *args, **kwargs) -> Response:
         downstream_path = kwargs["downstream_path"]
         method = request.method
         user = request.user
+        org = user.organization
+        if not org.is_grafana_incident_enabled:
```
joeyorlando, line 201: _nit_ this is probably fine for now. Eventually might make sense to move this to a separate method, something like `can_proxy_request_to_downstream_backend`, which has different conditional checks based on the the value of `kwargs["downstream_backend"]` (ie. `incident`, `slo`, etc). For now we only support Incident so this is okay.. just thought I'd note.

joeyorlando (approved):

PR #4951, opened 2024-08-28
Compress sync data

mderynck:
# What this PR does
- Compresses sync data being sent to engine
- Minor fix to log messages when JSON parse errors occur

Related to [issue link here]

## Checklist

### DIFF
--- a/engine/apps/grafana_plugin/views/sync_v2.py
+++ b/engine/apps/grafana_plugin/views/sync_v2.py
@@ -29,7 +29,7 @@ class SyncV2View(APIView):
 
     def do_sync(self, request: Request) -> Organization:
         if request.headers.get("Content-Encoding") == "gzip":
-            gzip_data = gzip.GzipFile(fileobj=io.BytesIO(request.body)).read()
+            gzip_data = gzip.GzipFile(fileobj=request).read()
             decoded_data = gzip_data.decode("utf-8")
             data = json.loads(decoded_data)
         else:
--- a/engine/apps/grafana_plugin/views/sync_v2.py
+++ b/engine/apps/grafana_plugin/views/sync_v2.py
@@ -1,5 +1,4 @@
 import gzip
-import io
 import json
 import logging
 from dataclasses import asdict, is_dataclass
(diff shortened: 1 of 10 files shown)


engine/apps/grafana_plugin/views/sync_v2.py
```diff
@@ -25,7 +28,14 @@ class SyncV2View(APIView):
     authentication_classes = (BasePluginAuthentication,)
 
     def do_sync(self, request: Request) -> Organization:
-        serializer = SyncDataSerializer(data=request.data)
+        if request.headers.get("Content-Encoding") == "gzip":
```
mderynck, line 31: Trying this route rather than middleware since we only want to deal with compression on here.

engine/apps/grafana_plugin/views/sync_v2.py
```diff
@@ -25,7 +28,14 @@ class SyncV2View(APIView):
...
+            gzip_data = gzip.GzipFile(fileobj=io.BytesIO(request.body)).read()
```
matiasb, line 32: Curious, can't you pass the `request` directly as the fileobj? (I think the `request` object implements file-like interface?)

grafana-plugin/pkg/plugin/resources.go
```diff
@@ -132,6 +132,7 @@ func (a *App) registerRoutes(mux *http.ServeMux) {
 	//mux.HandleFunc("/debug/settings", a.handleDebugSettings)
 	//mux.HandleFunc("/debug/permissions", a.handleDebugPermissions)
 	//mux.HandleFunc("/debug/stats", a.handleDebugStats)
+	//mux.HandleFunc("/debug/unlock", a.handleDebugUnlock)
```
matiasb, line 135: Any reason to add the handler but keep the route commented?

matiasb (approved):
LGTM

grafana-plugin/pkg/plugin/resources.go
mderynck: Concerned someone will automate API to defeat the whole purpose of having the timeout lock.  On the other hand there are valid use cases when you added some new data and don't want to wait  5 mins.  Maybe it should
always be available under plugin instead?

Currently the endpoint implementation is the minimum for testing, no error handling for when things aren't locked etc.

matiasb: :+1: right; maybe, we can leave it like this for now and revisit later

PR #4955, opened 2024-08-29
Backend Changes for Connecting Mattermost Channels

ravishankar15:
I have created two endpoints,

1. /channels: Create  --> To create/add a new channel
2. /channels/<id>: destroy --> To delete the channel

The create end point will take `channel_name` and `team_name` param as input using which we get the channel information. Ref(https://api.mattermost.com/#tag/channels/operation/GetChannelByNameForTeamName) The team_name is needed because a single bot/user can be a part of multiple teams.

As follow up,
1. I will create tests for this PR changes as a separate commit in this PR. While the team is taking look at the implementation
2. I will raise a separate PR for UI changes

# What this PR does

Related to [issue link here]

## Checklist


engine/apps/mattermost/models/mattermost_channel.py
```diff
@@ -0,0 +1,38 @@
...
+
+    public_primary_key = models.CharField(
+        max_length=20,
+        validators=[MinLengthValidator(settings.PUBLIC_PRIMARY_KEY_MIN_LENGTH + 1)],
+        unique=True,
+        default=generate_public_primary_key_for_slack_channel,
+    )
+
+    channel_id = models.CharField(max_length=100, default=None)
+    channel_name = models.CharField(max_length=100, default=None)
```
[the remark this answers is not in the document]
ravishankar15: Yup storing display_name can be helpful for displaying in UI as channel_name is the kebab case of the display_name added the same to the model

engine/apps/mattermost/serializers.py
```diff
@@ -0,0 +1,48 @@
...
+            "channel_name",
+        ]
+        validators = [
+            UniqueTogetherValidator(queryset=MattermostChannel.objects.all(), fields=["organization", "channel_id"])
+        ]
+
+    def create(self, validated_data):
+        return MattermostChannel.objects.create(**validated_data)
+
+    def to_internal_value(self, data):
```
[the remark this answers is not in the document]
ravishankar15: Yeah my thought process was,

1. With bot token getting list of channels we need to store all of those channels and display the same in UI plus we have to manage the sync of those channels similar to slack implementation but that seems like a overkill at this point. We can have each channel to be integrated manually by the admin as this is just a one time activity and not expected to happen or change often
2. Yeah slash commands is also possible but again the admin has to provide the token for slash command to us using which we can verify the incoming slash command. 

I feel (2) to be a better option than (1). I don't feel much of a feasibility for the admin between (2) and current implementation since both the flows requires relatively similar manual effort. Let me know if you think otherwise implementing with slash commands is not that complicated [Ref](https://developers.mattermost.com/integrate/slash-commands/custom/).

engine/apps/mattermost/models/mattermost_channel.py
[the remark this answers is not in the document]
ravishankar15: I gave a thought about storing the team model but at this point I don't see much of use case for that, All post API are tied to channel or user not to team. The team info is mainly useful on some of the channel APIs for which i don't see a use case. Let me know if you thing otherwise.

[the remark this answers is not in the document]
matiasb: Ack, I was checking since it seems `team_name` is needed to register the channel, it may make sense to store something around team to make sure we can disambiguate later if needed? (at least `team_id`?)

engine/apps/mattermost/serializers.py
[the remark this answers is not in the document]
matiasb: Right. If having a slash command does not imply a lot of changes, it sounds better, since that will be more consistent with the current telegram and msteams experience.

ravishankar15:
@matiasb 
In mattermost,
1. A single user and bot can belong to multiple team
2. A single channel belong to single team

For creating the slash commands we need the team_id [API Reference](https://api.mattermost.com/#tag/commands/operation/CreateCommand) 

With single bot_access_token having access to multiple teams we would need the user to input the team so that we can create the slash command on that team ? or We can create slash commands on all the teams the bot has access to.

How about we do something like this,
1. When the user connects the integration in Oncall we use the bot access token to create slash command `/oncall` in all the teams the bot has access too.
2. We can then implement `/oncall link-team` and `/oncall link-channel` which we can ask the user to do which creates the team and channel in our oncall. Before linking any channel the corresponding team needs to be linked.

Couple of catches,
1. A single Oncall org can have many mattermost team integrated.
2. The user to team link will be many to many relationship.

Does this sound good ?

matiasb:
Hm.. I see your point. I thought we could have a "global" command to register a channel by just calling that command in the context of the channel (I guess that may be possible if going with a plugin implementation).

> How about we do something like this,
> 
>     1. When the user connects the integration in Oncall we use the bot access token to create slash command `/oncall` in all the teams the bot has access too.

This would mean that adding a team later, there will be no slash command for it, right?

>     2. We can then implement `/oncall link-team` and `/oncall link-channel` which we can ask the user to do which creates the team and channel in our oncall. Before linking any channel the corresponding team needs to be linked.

It seems a bit too involved for users. Maybe your original plan works better as a first (and simpler) implementation? (ie. manually registering the channels OnCall-side via web UI providing team/channel, validating the information through the API/bot token)

ravishankar15:
```
This would mean that adding a team later, there will be no slash command for it, right?
```
There will a manual step from the user to do `/oncall link-team` so that we can add the team to oncall. 

I was also feeling the same way getting the team_name and channel_name and go with that flow. Also to confirm we are aligned to store only the team_id to the channel model as per this [comment](https://github.com/grafana/oncall/pull/4955#discussion_r1741126216)


While checking these I realised, There is a small edge case on that,
If we have two teams, Team A and Team B both can have a same channel name. When we are displaying the channel names to the user it might be confusing, There can be couple of ways to handle this, 

1. We can append the team_name to display but that would mean we need to store the team information like team_id and team_name in a model and team_name needs to be fetched with a separate API call with team_id.
2. We can append the last or first 4 to 5 chars of the team_id to the channel name while displaying.
I would prefer (2) for now Thoughts ?

If we are aligned with the existing PR implementation I will write the tests for the same and complete the same within this week.

matiasb:
> I was also feeling the same way getting the team_name and channel_name and go with that flow. Also to confirm we are aligned to store only the team_id to the channel model as per this [comment](https://github.com/grafana/oncall/pull/4955#discussion_r1741126216)

+1

> While checking these I realised, There is a small edge case on that, If we have two teams, Team A and Team B both can have a same channel name. When we are displaying the channel names to the user it might be confusing, There can be couple of ways to handle this,
> 
>     1. We can append the team_name to display but that would mean we need to store the team information like team_id and team_name in a model and team_name needs to be fetched with a separate API call with team_id.
> 
>     2. We can append the last or first 4 to 5 chars of the team_id to the channel name while displaying.
>        I would prefer (2) for now Thoughts ?
> 

Right, we will need something to identify the correct channel. Let's go with (2) for now.

> If we are aligned with the existing PR implementation I will write the tests for the same and complete the same within this week.

Sounds good :+1: Thanks!

engine/apps/mattermost/models/channel.py
```diff
@@ -0,0 +1,72 @@
...
+        unique=True,
+        default=generate_public_primary_key_for_mattermost_channel,
+    )
+
+    mattermost_team_id = models.CharField(max_length=100, default=None)
+    channel_id = models.CharField(max_length=100, default=None)
+    channel_name = models.CharField(max_length=100, default=None)
+    display_name = models.CharField(max_length=100, default=None)
+    is_default_channel = models.BooleanField(null=True, default=False)
+    datetime = models.DateTimeField(auto_now_add=True)
```
matiasb, line 42: `created_at`? I think `datetime` is not clear enough, and it collides with the datetime module.

engine/apps/mattermost/models/channel.py
```diff
@@ -0,0 +1,72 @@
...
+
+    @property
+    def unique_display_name(self) -> str:
+        return f"{self.display_name}-{self.mattermost_team_id[:5]}"
```
matiasb, line 46: if there is no `mattermost_team_id` this will fail, right? (or if there is always a team, should the field be non-nullable?)

engine/apps/mattermost/models/channel.py
```diff
@@ -0,0 +1,72 @@
...
+            old_default_channel = None
+            self.is_default_channel = True
+            self.save(update_fields=["is_default_channel"])
+        else:
+            self.is_default_channel = True
+            with transaction.atomic():
+                old_default_channel.save(update_fields=["is_default_channel"])
+                self.save(update_fields=["is_default_channel"])
+
+        print(f"Model: {self.is_default_channel}")
```
matiasb, line 65: Debugging print?

engine/apps/mattermost/serializers.py
```diff
@@ -0,0 +1,58 @@
...
+            "display_name",
+            "is_default_channel",
+        ]
+
+    def create(self, validated_data):
+        return MattermostChannel.objects.create(**validated_data)
+
+    def to_representation(self, instance):
+        ret = super().to_representation(instance)
+        del ret["mattermost_team_id"]
```
matiasb, line 30: Why is this needed?

engine/apps/mattermost/tests/test_mattermost_channel.py
```diff
@@ -0,0 +1,351 @@
...
+    role,
+    expected_status,
+):
+    client = APIClient()
+    _, user, token = make_organization_and_user_with_plugin_token(role)
+
+    data = make_mattermost_get_channel_by_name_team_name_response()
+    channel_response = requests.Response()
+    channel_response.status_code = status.HTTP_200_OK
+    channel_response._content = json.dumps(data).encode()
```
matiasb, line 139: No need to change but, FYI, we are using [httpretty](https://github.com/grafana/oncall/blob/dev/engine/apps/webhooks/tests/test_trigger_webhook.py#L236) for tests involving http requests

engine/apps/mattermost/serializers.py
ravishankar15: I don't see a need for us to expose the `mattermost_team_id` to UI hence wanted to remove that from payload. Modified the serializers Meta class and included `write_only` flag for neater representation.
Do you feel we should send this in payload ?

engine/apps/mattermost/models/channel.py
ravishankar15: Yup made the `mattermost_team_id` and `channel_id` both non-nullable as they both will be present always and required always.

ravishankar15:
Hey @matiasb I have addressed the comments. If the changes looks good can we merge the changes and also can be rebase our feature branch with new changes from main.

matiasb:
It looks like there is some issues with the API (besides tests not running).

If I tried to register a channel:

```
$ curl -X POST http://oncall:oncall@localhost:3000/api/plugins/grafana-oncall-app/resources/mattermost/channels -d '{"team_name": "general", "channel_name": "alerts"}'

<!doctype html>
<html lang="en">
<head>
  <title>Server Error (500)</title>
</head>
<body>
  <h1>Server Error (500)</h1><p></p>
</body>
</html>
```

While on Django server side:
```
source=engine:app google_trace_id=none logger=django.request Internal Server Error: /api/internal/v1/mattermost/channels
Traceback (most recent call last):
  File "/home/matiasb/grafana/oncall/venv/lib/python3.12/site-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
               ^^^^^^^^^^^^^^^^^^^^^
  File "/home/matiasb/grafana/oncall/venv/lib/python3.12/site-packages/django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/matiasb/grafana/oncall/venv/lib/python3.12/site-packages/django/views/decorators/csrf.py", line 56, in wrapper_view
    return view_func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/matiasb/grafana/oncall/venv/lib/python3.12/site-packages/rest_framework/viewsets.py", line 124, in view
    return self.dispatch(request, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/matiasb/grafana/oncall/venv/lib/python3.12/site-packages/rest_framework/views.py", line 509, in dispatch
    response = self.handle_exception(exc)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/matiasb/grafana/oncall/venv/lib/python3.12/site-packages/rest_framework/views.py", line 469, in handle_exception
    self.raise_uncaught_exception(exc)
  File "/home/matiasb/grafana/oncall/venv/lib/python3.12/site-packages/rest_framework/views.py", line 480, in raise_uncaught_exception
    raise exc
  File "/home/matiasb/grafana/oncall/venv/lib/python3.12/site-packages/rest_framework/views.py", line 506, in dispatch
    response = handler(request, *args, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/matiasb/grafana/oncall/venv/lib/python3.12/site-packages/rest_framework/mixins.py", line 18, in create
    serializer.is_valid(raise_exception=True)
  File "/home/matiasb/grafana/oncall/venv/lib/python3.12/site-packages/rest_framework/serializers.py", line 231, in is_valid
    raise ValidationError(self.errors)
                          ^^^^^^^^^^^
  File "/home/matiasb/grafana/oncall/venv/lib/python3.12/site-packages/rest_framework/serializers.py", line 582, in errors
    return ReturnDict(ret, serializer=self)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/matiasb/grafana/oncall/venv/lib/python3.12/site-packages/rest_framework/utils/serializer_helpers.py", line 19, in __init__
    super().__init__(*args, **kwargs)
ValueError: dictionary update sequence element #0 has length 46; 2 is required
source=engine:app google_trace_id=none logger=root inbound latency=0.037808 status=500 method=POST path=/api/internal/v1/mattermost/channels user_agent=GrafanaOnCall/1.0.0 content-length=50 slow=0 user_id=1 org_id=1 org_slug=self_hosted_org
```

(didn't get to debug the cause, but it seems the request isn't getting to the endpoint?)

ravishankar15:
The issue is in handling the exception of the API. The returned exception should contain dict value hence made changes to use the default BadRequest exception we have defined already. Added a test case as well to catch this scenario. Earlier the test was catching this in client layer only.

For the test not running part, I am not exactly sure why it is failing in pipeline it was running successfully for me in local. However i removed the namespace from the config since it was a duplicate of the app_name. I ran the test and lint in local and it is running successfully.

matiasb:
Hey, it looks like I cannot push to this branch? I updated the base mattermost-chatops branch, fixed the conflicts locally and I'm running the tests without changes and I see them pass. Still debugging the issues with the API call.

ravishankar15:
I think I have a hunch, In pipeline the value of the feature flag could be false `FEATURE_MATTERMOST_INTEGRATION_ENABLED` I have made the value as true hopefully that should resolve the issue

ravishankar15:
Hey, I have rebased the branch with the feature branch !

engine/apps/mattermost/client.py
```diff
@@ -0,0 +1,70 @@
+from dataclasses import dataclass
+from typing import Optional
+
+import requests
+from requests.auth import AuthBase
+from requests.models import PreparedRequest
+
+from apps.base.utils import live_settings
```
matiasb, line 8:
```suggestion
```
We are considering to limit/deprecate the use of `live_settings` (preferring Django settings + env vars)

engine/apps/mattermost/client.py
```diff
@@ -0,0 +1,70 @@
...
+    channel_id: str
+    team_id: str
+    channel_name: str
+    display_name: str
+
+
+class MattermostClient:
+    def __init__(self, token: Optional[str] = None) -> None:
+        self.token = token or live_settings.MATTERMOST_BOT_TOKEN
+        self.base_url = f"{live_settings.MATTERMOST_HOST}/api/v4"
```
matiasb, line 32:
```suggestion
        self.token = token or settings.MATTERMOST_BOT_TOKEN
        self.base_url = f"{settings.MATTERMOST_HOST}/api/v4"
```

engine/apps/mattermost/serializers.py
```diff
@@ -0,0 +1,62 @@
...
+
+        if not channel_name:
+            raise serializers.ValidationError({"channel_name": "This field is required."})
+
+        try:
+            response = MattermostClient().get_channel_by_name_and_team_name(
+                team_name=team_name, channel_name=channel_name
+            )
+        except (MattermostAPIException, MattermostAPITokenInvalid) as ex:
+            raise BadRequest(detail=ex.msg)
```
matiasb, line 53:
```suggestion
        except MattermostAPIException as err:
            raise BadRequest(detail=err.msg)
        except MattermostAPITokenInvalid:
            raise BadRequest(detail="Mattermost API token is invalid.")
```
(`MattermostAPITokenInvalid` does not have a `.msg` attr)

engine/apps/mattermost/tests/test_mattermost_channel.py
```diff
@@ -0,0 +1,381 @@
...
+            "channel_name": second_mattermost_channel.channel_name,
+            "display_name": second_mattermost_channel.unique_display_name,
+            "is_default_channel": second_mattermost_channel.is_default_channel,
+        },
+    ]
+
+    url = reverse("mattermost:channel-list")
+    response = client.get(url, **make_user_auth_headers(user, token))
+    assert response.status_code == status.HTTP_200_OK
+    assert response.json() == expected_payload
```
matiasb, line 283:
```suggestion
    response_data = response.json()
    assert len(response_data) == 2
    for channel_data in expected_payload:
        assert channel_data in response_data
```
(you cannot guarantee an order in the response)

engine/apps/mattermost/tests/test_mattermost_client.py
```diff
@@ -0,0 +1,86 @@
+import json
+from unittest.mock import Mock, patch
+
+import pytest
+import requests
+from rest_framework import status
+
+from apps.base.utils import live_settings
```
matiasb, line 8:
```suggestion
```

engine/apps/mattermost/tests/test_mattermost_client.py
```diff
@@ -0,0 +1,86 @@
...
+from apps.mattermost.client import MattermostAPIException, MattermostAPITokenInvalid, MattermostClient
+
+
+@pytest.mark.django_db
+def test_mattermost_client_initialization():
+    live_settings.MATTERMOST_BOT_TOKEN = None
+    with pytest.raises(MattermostAPITokenInvalid) as exc:
+        MattermostClient()
+        assert type(exc) is MattermostAPITokenInvalid
```
matiasb, line 17:
```suggestion
def test_mattermost_client_initialization(settings):
    settings.MATTERMOST_BOT_TOKEN = None
    with pytest.raises(MattermostAPITokenInvalid):
        MattermostClient()
```

grafana-plugin/src/pages/settings/tabs/ChatOps/ChatOps.tsx
```diff
@@ -150,6 +150,14 @@ const Tabs = (props: TabsProps) => {
...
         </VerticalTab>
       )}
+      {store.hasFeature(AppFeature.Mattermost) && (
+        <VerticalTab id={ChatOpsTab.Mattermost}>
+          <HorizontalGroup>
+            <Icon name="message" />
+            Mattermost
+          </HorizontalGroup>
+        </VerticalTab>
+      )}
```
matiasb, line 160:
```suggestion
      {store.hasFeature(AppFeature.Mattermost) && (
        <VerticalTab id={ChatOpsTab.Mattermost}>
          <Stack>
            <Icon name="message" />
            Mattermost
          </Stack>
        </VerticalTab>
      )}
```
(fixing lint error; also note that this block is duplicated above)

matiasb (commented):
Added a few suggestions. Other than that, things looking good to me :+1: 

Note that after merging changes from upstream, you will also need to update this:
```diff
--- a/grafana-plugin/src/pages/settings/tabs/ChatOps/tabs/MattermostSettings/MattermostSettings.tsx
+++ b/grafana-plugin/src/pages/settings/tabs/ChatOps/tabs/MattermostSettings/MattermostSettings.tsx
@@ -2,6 +2,8 @@ import React, { Component } from 'react';
 
 import { Button, HorizontalGroup, VerticalGroup, Icon } from '@grafana/ui';
 import cn from 'classnames/bind';
+import { DOCS_MATTERMOST_SETUP } from 'helpers/consts';
+import { showApiError } from 'helpers/helpers';
 import { observer } from 'mobx-react';
 
 import { Block } from 'components/GBlock/Block';
@@ -10,8 +12,6 @@ import { Text } from 'components/Text/Text';
 import { AppFeature } from 'state/features';
 import { WithStoreProps } from 'state/types';
 import { withMobXProviderContext } from 'state/withStore';
-import { DOCS_MATTERMOST_SETUP } from 'utils/consts';
-import { showApiError } from 'utils/utils';
 
 import styles from './MattermostSettings.module.css';
```

ravishankar15:
Hey I have made the changes

engine/apps/api/serializers/organization.py
```diff
@@ -117,14 +117,23 @@ class Meta:
 class CurrentOrganizationConfigChecksSerializer(serializers.ModelSerializer):
     is_chatops_connected = serializers.SerializerMethodField()
     is_integration_chatops_connected = serializers.SerializerMethodField()
+    mattermost = serializers.SerializerMethodField()
```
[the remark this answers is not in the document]
ravishankar15: Yeah i did, i think this got added by mistake during one of my rebase possibly. I have removed the change

PR #4968, opened 2024-08-30
fix(helm): Fixed migrate not being consistent in name for hooks

elliotcourant:
# What this PR does

This makes it so that the yaml being created for the migrate job is consistent when `useHook` is `true` or `false`. At the moment if use hook is false, the the name on the yaml and in the metadata match. But if use hook is true then the name on the object itself is different from the one in the metadata.

Didn't open an issue for this since it's minor.

Basically I have an "infra" repo that generates the yaml from several helm charts and version controls the generated yaml. It's an easy way to make sure that a chart update doesn't absolutely clobber something and makes it very easy to see whats actually getting pushed out to the kube cluster before actually applying it.

Because of the way this "metadata.name" field was being generated, it would result in the file being **changed** each time the yaml for the repository was generated which made for some odd diffs when something completely different may have changed.

## Checklist

---

Not aware of any tests for this specifically in this repo that I can see, or documentation around this specific behavior.


elliotcourant:
Just looking for feedback still, not stale

PR #4997, opened 2024-09-06
Bump Redis deps

arslanbekov:
Can we update this dependency? The current version of Redis has CVE.


matiasb (approved):
Thanks for the contribution!

PR #5019, opened 2024-09-12
Fix <a> link escaping in web template

mderynck:
# What this PR does
Removed escaping of html in web template to fix issue where <a> links were not being detected correctly when post processing the web template.

https://github.com/grafana/support-escalations/issues/12412

## Checklist

### DIFF
--- a/engine/apps/alerts/incident_appearance/templaters/web_templater.py
+++ b/engine/apps/alerts/incident_appearance/templaters/web_templater.py
@@ -1,7 +1,7 @@
 import re
 
 from apps.alerts.incident_appearance.templaters.alert_templater import AlertTemplater
-from common.utils import convert_md_to_html, escape_html, url_re, urlize_with_respect_to_a
+from common.utils import convert_md_to_html, url_re, urlize_with_respect_to_a
 
 
 class AlertWebTemplater(AlertTemplater):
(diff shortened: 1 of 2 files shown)


joeyorlando (approved):

engine/apps/alerts/incident_appearance/templaters/web_templater.py
```diff
@@ -15,7 +15,7 @@ def _postformat(self, templated_alert):
         if templated_alert.title:
             templated_alert.title = escape_html(self._slack_format_for_web(templated_alert.title))
         if templated_alert.message:
-            message = escape_html(self._slack_format_for_web(templated_alert.message))
```
matiasb, line 18: Can we just enable HTML links here? (not escaping HTML coming from an alert payload and rendering it could lead to other issues, I guess) I think part of the issue may be the [slack formatter](https://github.com/grafana/oncall/blob/dev/engine/apps/slack/slack_formatter.py#L8) changing links?

mderynck: Thanks! that is probably the correct solution, I'll test and then revert or update things. I'm not sure why the slack formatting is part of the web formatting, it would be nice to separate if slack needs it differently.

Issue #4201, opened 2024-04-10
Migrate remaining stylesheet to emotion (2nd iteration)


teodosii:
A good portion of our styles had been migrated to emotion [here](https://github.com/grafana/oncall/pull/4123), but there's still a good amount of untouched components (like the container components and the main pages) that are left to be migrated entirely to emotion

PR #5022, opened 2024-09-13
Migrate remaining stylesheets to Emotion

teodosii:
Closes https://github.com/grafana/oncall/issues/4201


grafana-plugin/src/components/GBlock/Block.styles.ts
```diff
@@ -16,7 +16,7 @@ export const getBlockStyles = (theme: GrafanaTheme2) => {
       }
 
       &--hover:hover {
-        background: var(--hover-selected);
+        background: ${theme.isLight ? '#f4f5f5' : 'rgba(204, 204, 220, 0.12)'};
```
brojd, line 19: Shouldn't we use here some vars from theme?

grafana-plugin/src/components/UserGroups/UserGroups.styles.ts
```diff
@@ -10,8 +10,8 @@ export const getUserGroupStyles = (theme: GrafanaTheme2) => {
 
     sortable: css`
       z-index: 1062;
-      box-shadow: var(--focused-box-shadow);
-      background: var(--hover-selected-hardcoded) !important;
+      box-shadow: ${theme.isDark ? 'rgb(17 18 23) 0 0 0 2px, rgb(61 113 217) 0 0 0 4px;' : ''};
+      background: ${theme.isDark ? '#34363d' : ''};
```
brojd, line 14: Shouldn't it be some var from theme?

grafana-plugin/src/containers/AddResponders/parts/AddRespondersPopup/AddRespondersPopup.styles.ts
```diff
@@ -0,0 +1,70 @@
...
+             max-height: 500px;
+        overflow: hidden;
+        border: 1px solid ${theme.colors.border.medium};
+        position: absolute;
+        right: 16px;
+        top: 55px;
+        background: ${theme.colors.background.primary}
+        width: 340px;
+        z-index: 10;
+        `,
```
brojd, line 16: seems like some wrong formatting

grafana-plugin/src/components/UserGroups/UserGroups.styles.ts
teodosii: These are custom made for oncall, they're not in core grafana

grafana-plugin/src/components/GBlock/Block.styles.ts
teodosii: This one is a bit odd. If you look for #f4f5f5 in core grafana it will point to `theme.colors.background.canvas` for light, while for dark the other one points to `theme.colors.action.selected` which seems a bit strange if you read it like

`${theme.isLight ? theme.colors.background.canvas : theme.colors.action.selected};`

But I can easily change it, though I must admit it looks a bit odd to either pick canvas background or the color for a selected action 😆

grafana-plugin/src/containers/AddResponders/parts/AddRespondersPopup/AddRespondersPopup.styles.ts
teodosii: Was missing a semi-colon on the `background` property, that fixes it, otherwise prettier wasn't doing any formatting

grafana-plugin/src/components/Modal/Modal.tsx
```diff
@@ -66,7 +66,7 @@ const getStyles = (theme: GrafanaTheme2) => {
       padding: 15px;
       background: ${theme.colors.background.primary};
       border: 1px solid ${theme.colors.border.weak};
-      box-shadow: var(--shadows-z3);
+      box-shadow: ${theme.isLight ? '0 13px 20px 1px rgba(24, 26, 27, 0.18)' : '0 8px 24px rgb(1, 4, 9)'};
```
brojd, line 69: Why do we need this change since it seems `var(--shadows-z3)` holds exactly the same values?

grafana-plugin/src/components/SortableList/SortableList.tsx
```diff
@@ -1,9 +1,9 @@
...
 
-import { SortableContainer } from 'react-sortable-hoc';
+import { SortableContainer, SortableContainerProps } from 'react-sortable-hoc';
 
 import { Timeline } from 'components/Timeline/Timeline';
 
 export const SortableList = SortableContainer(({ className, children }: any) => {
   return <Timeline className={className}>{children}</Timeline>;
-});
+}) as React.ComponentClass<SortableContainerProps & { className: string; children: React.ReactNode[] }>;
```
brojd, line 9: We can pass type as generic and avoid casting
```
export const SortableList = SortableContainer<{ className: string; children: React.ReactNode[] }>(
  ({ className, children }: any) => {
    return <Timeline className={className}>{children}</Timeline>;
  }
);
```

grafana-plugin/src/components/UserGroups/UserGroups.styles.ts
brojd: Can we put them in some global file and import? At least the background color. So it's easier to track what custom colors we have in app and reuse if needed

grafana-plugin/src/containers/MobileAppConnection/parts/DisconnectButton/DisconnectButton.tsx
```diff
@@ -19,7 +15,12 @@ export const DisconnectButton: FC<Props> = ({ onClick }) => (
       variant="destructive"
       onClick={onClick}
       size="md"
-      className={cx('disconnect-button')}
+      className={css`
+        position: absolute;
+        top: 50%;
+        left: 50%;
+        transform: translate(-50%, -50%);
+      `}
```
brojd, line 23: this part is a good candidate for a util that could be in `utils.styles.ts`

grafana-plugin/src/components/GBlock/Block.styles.ts
brojd: I see what you mean. I just want to avoid using raw values directly. Then I don't have any preference whether to add them to Colors enum in utils.styles or use this strange values from theme

grafana-plugin/src/components/SortableList/SortableList.tsx
teodosii: Missed the `any` param, fixed it.

grafana-plugin/src/components/Modal/Modal.tsx
teodosii: `--shadows-z3` doesn't exist

brojd: Hm but I can see it's there, e.g.
image

brojd: We can probably access it with `${theme.shadows.z3}` as well, could you try?

teodosii: Oh right, there's a `shadows` field in there. good catch, I changed it

teodosii: Though not all box-shadows in oncall match the defined shadows in grafana, e.g. SortableContainer

```jsx
box-shadow: ${theme.isDark ? 'rgb(17 18 23) 0 0 0 2px, rgb(61 113 217) 0 0 0 4px;' : ''};
background: ${theme.isDark ? '#34363d' : ''};
```

brojd (approved):

PR #5034, opened 2024-09-18
Mattermost Channel Integration UI Changes

ravishankar15:
# What this PR does
Add UI changes for the channel integration

## Checklist

### DIFF
--- a/engine/apps/mattermost/serializers.py
+++ b/engine/apps/mattermost/serializers.py
@@ -36,19 +36,13 @@ class MattermostChannelSerializer(serializers.ModelSerializer):
         return ret
 
     def to_internal_value(self, data):
-        team_name = data.get("team_name")
-        channel_name = data.get("channel_name")
+        channel_id = data.get("channel_id")
 
-        if not team_name:
-            raise serializers.ValidationError({"team_name": "This field is required."})
-
-        if not channel_name:
-            raise serializers.ValidationError({"channel_name": "This field is required."})
+        if not channel_id:
+            raise serializers.ValidationError({"channel_id": "This field is required."})
 
         try:
-            response = MattermostClient().get_channel_by_name_and_team_name(
-                team_name=team_name, channel_name=channel_name
-            )
+            response = MattermostClient().get_channel_by_id(channel_id=channel_id)
         except MattermostAPIException as ex:
             raise BadRequest(detail=ex.msg)
         except MattermostAPITokenInvalid:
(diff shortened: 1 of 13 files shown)


ravishankar15:
Hey @matiasb  I have added the UI changes for the channel integration, I see a small hiccup,

The input channelName is not same as the name we see in the mattermost UI. The name we see in the mattermost UI is display name. The name value is generated by mattermost from display name [ref](https://github.com/mattermost/mattermost/blob/master/webapp/channels/src/utils/url.tsx#L25). Our API call `get_channel_by_name_and_team_name` expects the name param and not the display name. The name param is not directly visible in the UI. 

When we click on the channel info icon on the top right we can see `Copy Link` button which has the name param in its path. 

Screenshot 2024-09-18 at 10 33 11 AM


On the other hand we have the channel id information on the info view

Should we let the user to get the channel name and keep our inputs as it is or should we just change our input to only channel_id and we can fetch the channel information using that id ?

matiasb:
> Should we let the user to get the channel name and keep our inputs as it is or should we just change our input to only channel_id and we can fetch the channel information using that id ?

Hm.. if we switch to use channel id, then we may not need the team name? ([docs](https://api.mattermost.com/#tag/channels/operation/GetChannel)). Maybe that's simpler? (no need to find out what the channel name is, and there is one less param to register a channel that can be error prone?)

ravishankar15:
Yup right ! I was also inclining in that direction will go ahead and make the changes in the PR

engine/apps/mattermost/serializers.py
```diff
@@ -36,19 +36,13 @@ def to_representation(self, instance):
         return ret
 
     def to_internal_value(self, data):
-        team_name = data.get("team_name")
-        channel_name = data.get("channel_name")
+        channel_id = data.get("channel_id")
```
matiasb, line 39: Since we are now using the mattermost channel ID (without referencing team in any way), wondering if we should display the channel list showing channel display name and channel ID (maybe fully displaying it?), instead of team ID (which user may never need to know, and I think it could be clearer in the UI what this ID is?)

matiasb (commented):
Minor comment added, lgtm (we may need to involve someone from frontend later to review UI changes, but I think that can wait for now).

engine/apps/mattermost/serializers.py
ravishankar15: Yup removed the the unique display name logic and displayed the channel id in the UI directly

matiasb (approved):
Thanks! :+1:

PR #5047, opened 2024-09-19
Add declare incident step and model

Ferril:
Related to https://github.com/grafana/oncall-private/issues/2831

## Checklist


grafana-plugin/src/components/Policy/EscalationPolicy.tsx
```diff
@@ -248,6 +251,32 @@ class _EscalationPolicy extends React.Component<EscalationPolicyProps, any> {
...
+    const { severity } = data;
+
+    return (
+      <WithPermissionControlTooltip key="" userAction={UserActions.EscalationChainsWrite}>
+        <Select
+          menuShouldPortal
+          disabled={isDisabled}
+          placeholder="Severity"
+          className={cx(styles.select, styles.control)}
+          // @ts-ignore
```
brojd, line 267: Seems we don't need it, can we remove?

brojd (approved):

engine/apps/api/views/escalation_policy.py
```diff
@@ -151,3 +161,25 @@ def num_minutes_in_window_options(self, request):
...
+        if organization.is_grafana_labels_enabled:
+            choices = [
+                {
+                    "value": EscalationPolicy.SEVERITY_SET_FROM_LABEL,
+                    "display_name": EscalationPolicy.SEVERITY_SET_FROM_LABEL_DISPLAY_VALUE,
+                }
+            ]
+        incident_client = IncidentAPIClient(organization.grafana_url, organization.api_token)
+        try:
+            severities, _ = incident_client.get_severities()
```
joeyorlando, line 178: _nit_ I suspect these won't change that often, maybe we can cache these for a few hours?

matiasb: This is only called, once, when editing escalation chains, which is not often either. I think caching wouldn't be necessary in this case, but it is something we have a note to take a look after release :+1:

engine/apps/public_api/serializers/escalation_policies.py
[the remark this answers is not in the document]
matiasb: +1

engine/apps/alerts/models/alert_group.py
```diff
@@ -420,8 +422,17 @@ def status(self) -> int:
...
+    # todo: rework using this field to use DeclaredIncident model field instead
     grafana_incident_id = models.CharField(max_length=100, null=True, default=None)
 
+    declared_incident = models.ForeignKey(
+        "alerts.DeclaredIncident",
+        on_delete=models.SET_NULL,
+        null=True,
+        default=None,
+        related_name="attached_alert_groups",
+    )
```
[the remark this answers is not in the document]
matiasb: Mainly as a way to simplify/enforce checks (like having one-active incident per route, or a potential limit of created incidents per org per timeframe) while reducing the incident API calls (and avoiding "complex/big" queries to fetch the information needed; this way we only create incident or query for an incident using the ID).
Eventually we could also store additional metadata in the relation (and as commented in the code, it may make sense to migrate the existing field to use the relation instead).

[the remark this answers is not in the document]
joeyorlando: makes sense! was just curious, thanks for clarifying!

engine/apps/api/views/escalation_policy.py
joeyorlando: > only called, once, when editing escalation chains

ahh gotcha 👍 (agreed, probably not necessary in that case)

Issue #5041, opened 2024-09-18
When creating an oncall-shift via the public API, `rolling_users` is not validated


joeyorlando:
### What went wrong?

**What happened**:

Below is a reproducible example via Terraform: 

```terraform
terraform {
  required_version = ">= 1.5.1"
  required_providers {
    grafana = {
      source  = "grafana/grafana"
      version = ">= 2.0.0"
    }
  }
}


provider "grafana" {
  alias               = "oncall"
  oncall_access_token = "<redacted>"
  oncall_url          = "http://localhost:8080/"
}

data "grafana_oncall_user" "joey" {
  provider = grafana.oncall
  username = "joey"
}

data "grafana_oncall_user" "bob" {
  provider = grafana.oncall
  username = "bob"
}

resource "grafana_oncall_on_call_shift" "example_shift" {
  provider = grafana.oncall
  name       = "Example Shift"
  type       = "rolling_users"
  start      = "2024-09-19T14:00:00"
  duration   = 60 * 60 * 12
  frequency  = "weekly"
  interval   = 1
  by_day     = ["MO", "TU", "WE", "TH", "FR"]
  week_start = "MO"
  rolling_users = [
    ["foo"],
    ["bar"],
    # [data.grafana_oncall_user.bob.id],
    # [data.grafana_oncall_user.joey.id],
  ]
  time_zone = "UTC"
}

resource "grafana_oncall_schedule" "example_schedule" {
  provider = grafana.oncall
  name      = "Example Calendar Schadule"
  type      = "calendar"
  time_zone = "America/New_York"
  shifts = [
    grafana_oncall_on_call_shift.example_shift.id,
  ]
}
```

Then:

```bash
❯ terraform apply
data.grafana_oncall_user.bob: Reading...
data.grafana_oncall_user.joey: Reading...
data.grafana_oncall_user.joey: Read complete after 0s [id=U95J3KBUFXM8J]
data.grafana_oncall_user.bob: Read complete after 0s [id=USSSBK861DG8H]

Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  + create

Terraform will perform the following actions:

  # grafana_oncall_on_call_shift.example_shift will be created
  + resource "grafana_oncall_on_call_shift" "example_shift" {
      + by_day        = [
          + "FR",
          + "MO",
          + "TH",
          + "TU",
          + "WE",
        ]
      + duration      = 43200
      + frequency     = "weekly"
[...]

PR #5050, opened 2024-09-20
fix: Add rolling users validation for oncall shift API

ravishankar15:
# What this PR does
Adds validation for rolling users param in the shift api

Closes [5041](https://github.com/grafana/oncall/issues/5041)

## Checklist

### DIFF
--- a/engine/apps/api/serializers/on_call_shifts.py
+++ b/engine/apps/api/serializers/on_call_shifts.py
@@ -29,7 +29,7 @@ class OnCallShiftSerializer(EagerLoadingMixin, serializers.ModelSerializer):
         allow_null=True,
         required=False,
         child=UsersFilteredByOrganizationField(
-            queryset=User.objects, required=False, allow_null=True
+            queryset=User.objects, db_verification=True, required=False, allow_null=True
         ),  # todo: filter by team?
     )
     updated_shift = serializers.CharField(read_only=True, allow_null=True, source="updated_shift.public_primary_key")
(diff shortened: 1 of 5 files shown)


joeyorlando (approved):
LGTM 👍 thanks for patch @ravishankar15! 🙂

engine/common/api_helpers/custom_fields.py
```diff
@@ -102,7 +103,16 @@ def to_internal_value(self, data):
...
             return None
 
-        return queryset.filter(organization=request.user.organization, public_primary_key__in=data).distinct()
+        users = queryset.filter(organization=request.user.organization, public_primary_key__in=data).distinct()
+        users_ppk = [u.public_primary_key for u in users]
+
+        if not self.db_verification:
+            return users
+
+        for d in data:
```
matiasb, line 112: Would this check be equivalent to do something like
`len(set(data)) == users.count()`
(ie. make sure we have a user per provided ppk)? I guess you could do a set difference to get the missing users too?

engine/apps/api/serializers/on_call_shifts.py
```diff
@@ -29,7 +29,7 @@ class OnCallShiftSerializer(EagerLoadingMixin, serializers.ModelSerializer):
         allow_null=True,
         required=False,
         child=UsersFilteredByOrganizationField(
-            queryset=User.objects, required=False, allow_null=True
+            queryset=User.objects, db_verification=True, required=False, allow_null=True
```
matiasb, line 32: I would suggest a different name for the param, based on my understanding of the check (see also the comment below), eg. `require_all_exist`?

matiasb (commented):
Makes sense to me :+1: 
Minor suggestions added.

engine/common/api_helpers/custom_fields.py
ravishankar15: I have Made the changes accordingly

Issue #4936, opened 2024-08-27
Past users still showing in schedules


matiasb:
### What went wrong?

**What happened**:
- A user that has been removed from a schedule is still shown in the schedule details page, and may still get notifications about shift swaps.

**What did you expect to happen**:
- Once a user has no scheduled shifts in the future should be not considered part of the schedule.

### How do we reproduce it?

Create a schedule with a rotation including a few users. Remove a user (that had been assigned an on-call in the past) from the rotation, save as new rotation. Alternatively, create an override setting up a new user to be on-call (not involved in any rotation). That user will be associated to the schedule since then.

### Grafana OnCall Version

r411-v1.9.16

### Product Area

Schedules

### Grafana OnCall Platform?

None

PR #5053, opened 2024-09-21
fix: Remove past users from schedule details page

ravishankar15:
# What this PR does
Removes users from schedule details page if the user does not have any future events

[#4936](https://github.com/grafana/oncall/issues/4936)

Related to [issue link here]

## Checklist

### DIFF
--- a/engine/apps/schedules/models/shift_swap_request.py
+++ b/engine/apps/schedules/models/shift_swap_request.py
@@ -181,7 +181,7 @@ class ShiftSwapRequest(models.Model):
 
     @property
     def possible_benefactors(self) -> QuerySet["User"]:
-        return self.schedule.related_users().exclude(pk=self.beneficiary_id)
+        return self.schedule.users_in_future_schedule().exclude(pk=self.beneficiary_id)
 
     @property
     def web_link(self) -> str:
--- a/engine/apps/api/tests/test_schedules.py
+++ b/engine/apps/api/tests/test_schedules.py
@@ -1518,6 +1518,7 @@ def test_next_shifts_per_user(
             tomorrow + timezone.timedelta(hours=18),
             user_c.timezone,
         ),
+        user_d.public_primary_key: (None, None, user_d.timezone),
     }
     returned_data = {
         u: (ev.get("start"), ev.get("end"), ev.get("user_timezone")) for u, ev in response.data["users"].items()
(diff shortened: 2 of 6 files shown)


engine/apps/schedules/models/shift_swap_request.py
```diff
@@ -181,7 +181,7 @@ def organization(self) -> "Organization":
 
     @property
     def possible_benefactors(self) -> QuerySet["User"]:
-        return self.schedule.related_users().exclude(pk=self.beneficiary_id)
+        return self.schedule.users_in_future_schedule().exclude(pk=self.beneficiary_id)
```
ravishankar15, line 184: I am taking the users from events in the future considering 30days should we rethink on this thought ?

matiasb:
Hey, in this case I think we really need to change how we determine the schedule [related_users](https://github.com/grafana/oncall/blob/dev/engine/apps/schedules/models/on_call_schedule.py#L344) instead, so we can "ignore" users that have been part of a schedule in the past at every place (e.g. when rendering schedule events, but also when triggering shift swap notifications). This may require changing the way we update the ical representation for each schedule type (or reworking how the related users are calculated), so that we ignore past events.

ravishankar15:
I checked on the things you have mentioned,

The ical representation has events to be the top node the users are appended to the ical summary for us to parse and export. I am not sure if changing the representation can be helpful in this case. I am thing your alternate suggestion could be helpful, We use `recurring-ical-events` library --> `between` method [code](https://github.com/grafana/oncall/blob/dev/engine/apps/schedules/ical_events/adapter/amixr_recurring_ical_events_adapter.py#L36) for fetching the events from calandar. We can have another method in the adapter to get the future events. The library also has a `after` method  [code](https://github.com/niccokunzmann/python-recurring-ical-events/blob/v2.2.3/recurring_ical_events.py#L834). Once we have the future events we can reuse the logics to fetch the users from those events.

Does this sound good ?

Catch: The after method is available in the higher version of recurring-ical-events and we might have to update that library.

matiasb:
Hey, I was taking a look at this (it has been in our backlog/radar for a bit), and prepared https://github.com/grafana/oncall/pull/5101 with the changes I was thinking of. Feel free to check and/or add any suggestions there.

ravishankar15:
Your changes looks more appropriate hence closing this PR

PR #5064, opened 2024-09-24
Fix for user appearing in shift rotation when they should not be

mderynck:
# What this PR does
Fix calculation for interval when building ical events for schedules with end dates and masked days.  Add a test which makes it easier to test different cases that can occur.

Related to https://github.com/grafana/support-escalations/issues/12388

## Checklist


engine/apps/schedules/tests/test_custom_on_call_shift.py
```diff
@@ -1826,3 +1831,69 @@ def test_refresh_schedule(make_organization_and_user, make_schedule, make_on_cal
...
+        "start": start,
+        "rotation_start": start,
+        "until": start + timezone.timedelta(days=total_days),
+        "duration": timezone.timedelta(hours=12),
+        "frequency": frequency,
+        "by_day": day_mask,
+        "schedule": schedule,
+        "interval": interval,
+        "priority_level": 1,
+        "week_start": CustomOnCallShift.MONDAY,
```
mderynck, line 1873: Does this get set to Monday in all cases? It seems to be set in the serializer to Monday when it is None.  DB default is Sunday but without this being set to Monday not all of these test cases pass.

nathG:
Thank you very much, I confirm that this fix works for us and solves the issue we had raised to your Support team.

PR #5087, opened 2024-09-25
fix: address RBAC Admin issue

joeyorlando:
# What this PR does

**NOTE**: should be merged/released after https://github.com/grafana/irm/pull/183 has been rolled out to most stacks (as that frontend update is what will grant that new RBAC "action" to users whom already have the "OnCall Admin" RBAC role assigned)

tldr; from the comment in the `RBACPermission.Permission.ADMIN` comment in `engine/apps/api/permissions.py`:

> NOTE: this is a bit of a hack for now. See https://github.com/grafana/support-escalations/issues/12625
> Basically when it comes to filtering teams that are configured to share their resources with
> "Team members and admins", we have no way of knowing, when a user is ACTUALLY an Admin when RBAC is involed.
>
> Example: Take a user with the basic role of None/Editor/Viewer but with the "OnCall Admin" role assigned.
> Without this RBAC permission, we have no way of knowing that the user is ACTUALLY an "Admin".

Closes https://github.com/grafana/support-escalations/issues/12625

## Checklist

### DIFF
--- a/engine/apps/user_management/models/user.py
+++ b/engine/apps/user_management/models/user.py
@@ -233,9 +233,7 @@ class User(models.Model):
 
     def self_or_has_user_settings_admin_permission(self, user_to_check: "User", organization: "Organization") -> bool:
         has_permission = user_is_authorized(user_to_check, [RBACPermission.Permissions.USER_SETTINGS_ADMIN])
-        return user_to_check.pk == self.pk or (
-            has_permission and organization.pk == user_to_check.organization_id
-        )
+        return user_to_check.pk == self.pk or (has_permission and organization.pk == user_to_check.organization_id)
 
     def get_username_with_slack_verbal(self, mention=False) -> str:
         slack_verbal = None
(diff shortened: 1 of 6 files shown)


engine/apps/user_management/models/user.py
```diff
@@ -172,11 +172,19 @@ def __str__(self):
         return f"{self.pk}: {self.username}"
 
     @property
-    def available_teams(self):
-        if self.role == LegacyAccessControlRole.ADMIN:
```
joeyorlando, line 176: THIS was the crux of the issue

matiasb (approved):

grafana-plugin/src/plugin.json
```diff
@@ -174,9 +174,10 @@
           { "action": "grafana-oncall-app.user-settings:admin" },
 
           { "action": "grafana-oncall-app.other-settings:read" },
-          { "action": "grafana-oncall-app.other-settings:write" }
-        ],
-        "hidden": true
```
joeyorlando, line 179: `hidden` is no longer supported. See https://raw.githubusercontent.com/grafana/grafana/main/docs/sources/developers/plugins/plugin.schema.json

(the Admin role is already _always_ shown in the RBAC role picker in the Grafana UI)

PR #5099, opened 2024-09-30
fix: update internal labels endpoints to work with RBAC

joeyorlando:
# What this PR does

Related to:
- https://github.com/grafana/oncall-private/issues/2943

## Checklist


engine/apps/alerts/models/alert_group.py
```diff
@@ -545,7 +546,7 @@ def web_link(self) -> str:
     @property
     def declare_incident_link(self) -> str:
         """Generate a link for AlertGroup to declare Grafana Incident by click"""
-        incident_link = urljoin(self.channel.organization.grafana_url, "a/grafana-incident-app/incidents/declare/")
+        incident_link = urljoin(self.channel.organization.grafana_url, f"a/{PluginID.INCIDENT}/incidents/declare/")
```
joeyorlando, line 549: I'll finish renaming the hardcoded plugin IDs, to use `PluginID`, in a subsequent PR

engine/apps/api/permissions.py
```diff
@@ -250,6 +247,17 @@ class Permissions:
             Resources.OTHER_SETTINGS, Actions.WRITE, LegacyAccessControlRole.ADMIN
         )
 
+        # NOTE: we don't currently add the label delete permission here because we don't currently use this in OnCall
+        LABEL_CREATE = LegacyAccessControlCompatiblePermission(
+            Resources.LABEL, Actions.CREATE, LegacyAccessControlRole.EDITOR, prefix=PluginID.LABELS
+        )
```
vadimkerr, line 253: I only see the `Labels Reader` and `Labels Writer` roles in the existing Grafana RBAC UI, is this PR introducing a new one for create? maybe it's a good idea to only keep read and write (so create is part of write).

engine/apps/api/permissions.py
```diff
@@ -301,40 +309,6 @@ def has_object_permission(self, request: AuthenticatedRequest, view: ViewSetOrAP
         return True
 
 
-class BasicRolePermission(permissions.BasePermission):
```
vadimkerr, line 304: is it safe to delete this? can it break existing OSS installations that don't have RBAC?

engine/conftest.py
```diff
@@ -405,6 +424,7 @@ def _make_user(role: typing.Optional[LegacyAccessControlRole] = None, **kwargs):
         if permissions is None:
             permissions_to_grant = ROLE_PERMISSION_MAPPING[role] if IS_RBAC_ENABLED else []
             permissions = [GrafanaAPIPermission(action=perm.value) for perm in permissions_to_grant]
+        print("YOOOO", permissions, role)
```
vadimkerr, line 427:
```suggestion
```

engine/apps/api/permissions.py
joeyorlando: the [`user_is_authorized`](https://github.com/grafana/oncall/pull/5099/files#diff-83058af8678d629be3d878a5947537c15c01abf6ef2eb7ba31d08b8febda0853R137-R141) function already handles this scenario (which is used by the `RBACPermission` class) 🙂

engine/apps/api/permissions.py
joeyorlando: tldr; no it's not introducing any new RBAC roles. The `LABEL_CREATE`, `LABEL_READ`, and `LABEL_WRITE` variables are just helpers which translate to the permission strings `grafana-labels-app:label:create`, `grafana-labels-app:label:read`, and `grafana-labels-app:label:write`, respectively.

---

Those are the RBAC _roles_ that're available from Labels. Roles grant >= 1 _permission(s)_ . [These](https://github.com/grafana/gops-labels/blob/main/plugin/src/plugin.json#L30-L54) are the permissions granted by those labels RBAC roles:
```json
"roles": [
    {
      "role": {
        "name": "Labels Writer",
        "description": "Read/write/create/delete Labels",
        "permissions": [
          { "action": "plugins.app:access", "scope": "plugins:id:grafana-labels-app" },
          { "action": "grafana-labels-app.label:read" },
          { "action": "grafana-labels-app.label:write" },
          { "action": "grafana-labels-app.label:create" },
          { "action": "grafana-labels-app.label:delete" }
        ]
      },
      "grants": ["Admin", "Editor"]
    },
    {
      "role": {
        "name": "Labels Reader",
        "description": "Read-only access to Labels",
        "permissions": [
          { "action": "plugins.app:access", "scope": "plugins:id:grafana-labels-app" },
          { "action": "grafana-labels-app.label:read" }
        ]
      },
      "grants": ["Viewer"]
    }
  ]
```

so in OnCall we're basically validating that the user has the necessary _permission_ (action), we don't necessarily care about the role names themselves.

I agree that `:write` and `:create` should be consolidated over in labels but that's the current state of things. If you're curious, see these resources for more context:
- https://github.com/grafana/gops-labels/blob/main/pkg/plugin/auth.go#L42
- https://github.com/grafana/gops-labels/blob/main/pkg/plugin/plugin.go#L70-L74
- https://github.com/grafana/gops-labels/blob/main/pkg/plugin/plugin.go#L84-L96

PR #5112, opened 2024-10-02
Include link information for objects referenced in alert group timeline

matiasb:
Related to https://github.com/grafana/oncall/issues/4537.

Related frontend changes: https://github.com/grafana/oncall/pull/5111/files#diff-98c45c177708c814aa5a8aafad36b0a76e9cf49d7e25dada214ae1ce9ed10699

PR #5123, opened 2024-10-03
Include link information for objects referenced in alert group timeline

matiasb:
Reworked https://github.com/grafana/oncall/pull/5112 (post-revert)


matiasb:
> Lgtm What's changed?

You mean re the reverted PR? There were some changes applied on top of the declared incident related work (which was reverted and still needs rework), so this PR is slightly different (because it is not including refactorings for the declare incident step links)

PR #5137, opened 2024-10-08
update URLs constructed by the backend to support IRM plugin

joeyorlando:
# What this PR does

Introduces a new class, `apps.grafana_plugin.ui_url_builder.UIURLBuilder`, which is responsible for... building UI URLs (😄). The class mainly does two things:
- it will decide if the URL should point to `grafana-oncall-app` or `grafana-irm-app` based on the value of `organization.is_grafana_irm_enabled` (**NOTE**: this value isn't yet being set + defaults to `False`; logic for setting this value will be done in a subsequent PR)
- Adds `enum`s, `OnCallPage` and `IncidentPage` to DRYify hardcoded UI URLs (in case we decide to change these slightly in the near future)

## Checklist


engine/apps/alerts/models/alert_receive_channel.py
```diff
@@ -422,8 +422,9 @@ def emojized_verbal_name(self):
 
     @property
     def new_incidents_web_link(self):
-        return urljoin(
-            self.organization.web_link, f"?page=incidents&integration={self.public_primary_key}&status=0&p=1"
```
joeyorlando, line 426: this previously had a bug in it, the `?page=<page>` URL format is no longer supported, this will now point to the `alert-groups` page w/ the necessary filtering query params

engine/apps/grafana_plugin/ui_url_builder.py
joeyorlando, line 1: main change

engine/apps/oss_installation/cloud_heartbeat.py
```diff
@@ -102,9 +107,16 @@ def send_cloud_heartbeat():
...
+def get_heartbeat_link(
+    organization: "Organization",
+    connector: typing.Optional["CloudConnector"],
+    heartbeat: typing.Optional["CloudHeartbeat"],
+) -> typing.Optional[str]:
     if connector is None:
         return None
     if heartbeat is None:
         return None
-    return urljoin(connector.cloud_url, f"a/grafana-oncall-app/?page=integrations&id={heartbeat.integration_id}")
```
joeyorlando, line 110: this previously had a bug in it, `?page=<page>` URL format is no longer supported

engine/apps/oss_installation/utils.py
```diff
@@ -80,5 +84,7 @@ def cloud_user_identity_status(connector, identity):
         else:
             status = CloudSyncStatus.SYNCED_PHONE_NOT_VERIFIED
 
-        link = urljoin(connector.cloud_url, f"a/grafana-oncall-app/?page=users&p=1&id={identity.cloud_id}")
```
joeyorlando, line 83: this previously had a bug in it, `?page=<page>` URL format is no longer supported

engine/apps/schedules/models/shift_swap_request.py
```diff
@@ -183,11 +179,6 @@ def organization(self) -> "Organization":
     def possible_benefactors(self) -> QuerySet["User"]:
         return self.schedule.related_users().exclude(pk=self.beneficiary_id)
 
-    @property
-    def web_link(self) -> str:
-        # TODO: finish this once we know the proper URL we'll need
-        return f"{self.schedule.web_detail_page_link}"
```
joeyorlando, line 189: this was not referenced anywhere

engine/apps/schedules/models/shift_swap_request.py
```diff
@@ -171,10 +171,6 @@ def slack_channel_id(self) -> str | None:
         """
         return self.schedule.channel
 
-    @property
-    def schedule_slack_url(self) -> str:
-        return f"<{self.schedule.web_detail_page_link}|{self.schedule.name}>"
```
joeyorlando, line 176: this has been replaced w/ `slack_url` [here](https://github.com/grafana/oncall/pull/5137/files#diff-4037b7bbef9fc16d9b541beb3ed46f760916d7cd720847c3123adf7afb5ab4b4R260-R262) (and any references to this updated)

engine/apps/social_auth/middlewares.py
```diff
@@ -1,44 +1,46 @@
...
             # if user canceled authentication, redirect them to the previous page using the same link
             # as we used to redirect after auth/install
-            url_to_redirect = urljoin(request.user.organization.grafana_url, redirect_to)
-            return redirect(url_to_redirect)
+            return redirect(url_builder.build_absolute_plugin_ui_url(redirect_to))
         elif isinstance(exception, exceptions.AuthFailed):
             # if authentication was failed, redirect user to the plugin page using the same link
             # as we used to redirect after auth/install with error flag
-            url_to_redirect = urljoin(
-                request.user.organization.grafana_url, f"{redirect_to}&slack_error={SLACK_AUTH_FAILED}"
```
joeyorlando, line 34: this previously had a bug in it, `f"{redirect_to}&slack_error={SLACK_AUTH_FAILED}` _should've_ been `f"{redirect_to}?slack_error={SLACK_AUTH_FAILED}`, has been updated

engine/apps/user_management/migrations/0023_organization_is_grafana_irm_enabled.py
joeyorlando, line 1: I will create a separate PR which is responsible for setting this flag (likely in the sync flow)

engine/apps/user_management/models/organization.py
```diff
@@ -344,14 +345,15 @@ def get_notifiable_direct_paging_integrations(self) -> "RelatedManager['AlertRec
             .distinct()
         )
 
-    @property
-    def web_link(self):
-        return urljoin(self.grafana_url, "a/grafana-oncall-app/")
```
joeyorlando, line 349: this was not referenced anywhere, only `web_link_with_uuid` is used

joeyorlando:
wrt `is_grafana_irm_enabled`, I will create a separate PR to set this value (likely in the sync flow)

engine/apps/grafana_plugin/ui_url_builder.py
```diff
@@ -0,0 +1,50 @@
...
+if typing.TYPE_CHECKING:
+    from apps.user_management.models import Organization
+
+
+class UIURLBuilder:
+    """
+    If `base_url` is passed into the constructor, it will override using `organization.grafana_url`
+    """
+
+    class OnCallPage(enum.StrEnum):
```
mderynck, line 16: The call signature for build url ends up being a bit long when you reference the enum, I think it would be nicer to use if we define these as functions which call build url instead of passing in str enums.  This way we can also enforce args like id when we want to.
```
Examples:
UIURLBuilder(org).alert_groups()
UIURLBuilder(org).alert_group_detail(id="abc") or ("abc")?
UIURLBuilder(org).settings(extra="?tab=ChatOps&chatOpsTab=Slack")
```

joeyorlando: good point 👍 addressed in https://github.com/grafana/oncall/pull/5137/commits/2570e24981589eff2e16817f0cf4dc59b8076e1e

mderynck (approved):

PR #5142, opened 2024-10-09
feat: add `timedeltaparse` function to the jinja template

kelnage:
# What this PR does
If an OnCall template needs to contain a date relative to a date in the alert response, currently there is no way for the template to add or subtract time from a parsed date.

This PR adds a function that allows a time-window (e.g., 1s, 5m, 6h, 7d, 2w) to be converted into a Python timedelta, which can then be added or subtracted from a datetime. An example usage might be:

```
{% set delta = alert.timeWindow | timedeltaparse %}
{% set time = alert.startsAt | iso8601_to_time - delta | datetimeformat('%s') %}
```

## Checklist

- [X] Unit, integration, and e2e (if applicable) tests updated
- [X] Documentation added (or `pr:no public docs` PR label added if not required)
- [X] Added the relevant release notes label (see labels prefixed w/ `release:`). These labels dictate how your PR will show up in the autogenerated release notes.


joeyorlando (approved):
thanks for the contribution @kelnage!

PR #5172, opened 2024-10-11
Minor optimizations for public API endpoints

matiasb:
- Add/update eager loading for public API serializers
- Use cached final schedule to get currently on-call users


joeyorlando (approved):
nice! LGTM 👍

PR #5173, opened 2024-10-14
Mattermost Alert Flow

ravishankar15:
# What this PR does

Mattermost message screenshot,

Screenshot 2024-10-14 at 1 21 45 PM

Related to [issue link here]

## Checklist


engine/apps/mattermost/alert_rendering.py
```diff
@@ -0,0 +1,128 @@
...
+                    },
+                },
+            }
+
+        token = MattermostEventAuthenticator.create_token(organization=self.alert_group.channel.organization)
+        if not self.alert_group.resolved:
+            if self.alert_group.acknowledged:
+                actions.append(_make_actions("unacknowledge", "Unacknowledge", token))
+            else:
+                actions.append(_make_actions("acknowledge", "Acknonwledge", token))
```
ChristianCiach, line 107: typo: Acknonwledge -> Acknowledge

Also visible in your screenshot :)

engine/apps/mattermost/models/message.py
```diff
@@ -0,0 +1,33 @@
...
+
+    channel_id = models.CharField(max_length=100)
+
+    message_type = models.IntegerField(choices=MATTERMOST_MESSAGE_CHOICES)
+
+    alert_group = models.ForeignKey(
+        "alerts.AlertGroup",
+        on_delete=models.CASCADE,
+        related_name="mattermost_messages",
+    )
```
matiasb, line 25: I think we should have some unique condition (+ use `get_or_create` in the task) between `alert_group`, `channel_id` and `message_type` so some delay task/retries do not create multiple notifications for the same alert group (this happened to me in local testing :-))

engine/apps/mattermost/views.py
```diff
@@ -61,3 +62,29 @@ def perform_destroy(self, instance):
             channel_id=instance.channel_id,
         )
         instance.delete()
+
+
+class MattermostEventView(APIView):
+    def get(self, request, format=None):
```
matiasb, line 68: This event requests will be authenticated using the signing secret and jwt, right?

matiasb (commented):
Looking good, added a few comments/questions :+1:

engine/apps/alerts/models/alert_group.py
```diff
@@ -1989,6 +1990,10 @@ def slack_message(self) -> typing.Optional["SlackMessage"]:
         except AttributeError:
             return self.slack_messages.order_by("created_at").first()
 
+    @property
+    def mattermost_message(self) -> typing.Optional["MattermostMessage"]:
+        return self.mattermost_messages.order_by("created_at").first()
```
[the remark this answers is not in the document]
ravishankar15: One other case we might need this for is posting in thread like we need to pass the post id to which we wanted to thread the message which we might need in the upcoming PR's. I am thinking I will make this change of moving the logic into the task in this PR. if needed in future will add this property.

engine/apps/mattermost/models/message.py
ravishankar15: I am not sure if i understand completely. When we receive the signal for an alert group we enqueue a task which is like a 1:1 mapping for the task and alert group. In case of failures or retry won't it be happening for that alert group only ? 

Also `alert_group, channel_id and message_type` does not uniquely identify the message since for messages in the same thread share a common value for the above combination.

I am not able to see how multiple notifications gets triggered for the same alert group. Can you help me see where i am missing ?

engine/apps/mattermost/utils.py
```diff
@@ -0,0 +1,37 @@
...
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.DEBUG)
+
+
+class MattermostEventAuthenticator:
+    @staticmethod
+    def create_token(organization: typing.Optional["Organization"]):
+        secret = settings.MATTERMOST_SIGNING_SECRET
+        expiration = timezone.now() + datetime.timedelta(days=30)
```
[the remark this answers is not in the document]
ravishankar15: Yes thats correct !

engine/apps/mattermost/views.py
ravishankar15: Yup the secret that is sent as part of the message !

ravishankar15:
Hey @matiasb Please take a look at this PR I have added the tests. Apart from a small clarification on the above comment https://github.com/grafana/oncall/pull/5173#discussion_r1820809559 I have made all the required changes.

engine/apps/mattermost/models/message.py
matiasb: > I am not sure if i understand completely. When we receive the signal for an alert group we enqueue a task which is like a 1:1 mapping for the task and alert group. In case of failures or retry won't it be happening for that alert group only ?

Yeah, but it seems there is some retry condition or update that could enqueue multiple tasks for the same alert group, leading to duplicated notifications (need to take a closer look to check what the context was).

> Also `alert_group, channel_id and message_type` does not uniquely identify the message since for messages in the same thread share a common value for the above combination.

But we don't need to track every message in a thread, right? Only the root (alert group notification) message, which should be only one per alert group?

> I am not able to see how multiple notifications gets triggered for the same alert group. Can you help me see where i am missing ?

As mentioned above, not really sure what the specific conditions triggering the issue were (I was missing some Mattermost secret in my setup first, when set, I got duplicated messages).

engine/apps/alerts/models/alert_group.py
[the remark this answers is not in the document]
matiasb: Ok. My main goal here was to avoid adding mattermost-specific logic in the alert group model (but we can revisit later)

engine/apps/mattermost/models/message.py
ravishankar15: I was not able to reproduce the scenario in my local but I have added the condition to prevent duplicate events and also fixed the linting and test errors.

engine/apps/mattermost/tests/test_utils.py
```diff
@@ -0,0 +1,26 @@
...
+):
+    organization = make_organization()
+    token = MattermostEventAuthenticator.create_token(organization=organization)
+    payload = MattermostEventAuthenticator.verify(token)
+    assert payload["organization_id"] == organization.public_primary_key
+
+
+@pytest.mark.django_db
+def test_jwt_token_validation_failure(
+    make_organization,
```
matiasb, line 20: You should probably use the pytest fixture [settings](https://pytest-django.readthedocs.io/en/latest/configuring_django.html#overriding-individual-settings) for the override here.

engine/apps/mattermost/tests/test_tasks.py
```diff
@@ -0,0 +1,214 @@
...
+    ack_alert_group = make_alert_group(
+        alert_receive_channel, acknowledged_at=timezone.now() + timezone.timedelta(hours=1), acknowledged=True
+    )
+    make_alert(alert_group=ack_alert_group, raw_request_data=alert_receive_channel.config.example_payload)
+    make_mattermost_channel(organization=organization, is_default_channel=True)
+    ack_log_record = make_alert_group_log_record(ack_alert_group, type=AlertGroupLogRecord.TYPE_ACK, author=None)
+    mattermost_message = make_mattermost_message(ack_alert_group, MattermostMessage.ALERT_GROUP_MESSAGE)
+
+    url = "{}/api/v4/posts/{}".format(settings.MATTERMOST_HOST, mattermost_message.post_id)
+    data = make_mattermost_post_response_failure(status_code=data["status_code"])
```
matiasb, line 202: It looks like you only need to parametrize on the status code? (wondering why having a dict) also, you are overriding the data from the parametrize here, which can be confusing

engine/apps/mattermost/tests/test_tasks.py
```diff
@@ -0,0 +1,214 @@
...
+    on_create_alert_async(alert_pk=alert.pk)
+
+    request = httpretty.last_request()
+    assert request.url is None
+
+
+@pytest.mark.django_db
+@httpretty.activate(verbose=True, allow_net_connect=False)
+@pytest.mark.parametrize(
+    "data",
```
matiasb, line 76: (see comment below about data param)

engine/apps/mattermost/tests/test_tasks.py
```diff
@@ -0,0 +1,214 @@
...
+
+    url = "{}/api/v4/posts".format(settings.MATTERMOST_HOST)
+    data = make_mattermost_post_response()
+    mock_response = httpretty.Response(json.dumps(data), status=status.HTTP_200_OK)
+    httpretty.register_uri(httpretty.POST, url, responses=[mock_response])
+
+    on_create_alert_async(alert_pk=alert.pk)
+
+    request = httpretty.last_request()
+    assert request.url is None
```
matiasb, line 70: Curious (didn't check this before), is this the way to check there was no request made to the patched URL?

engine/apps/mattermost/tests/test_representative.py
```diff
@@ -0,0 +1,100 @@
...
+    make_alert_group_log_record,
+):
+    organization = make_organization()
+    alert_receive_channel = make_alert_receive_channel(
+        organization, integration=AlertReceiveChannel.INTEGRATION_GRAFANA
+    )
+    make_mattermost_channel(organization=organization, is_default_channel=True)
+    ack_alert_group = make_alert_group(
+        alert_receive_channel=alert_receive_channel,
+        acknowledged_at=timezone.now() + timezone.timedelta(hours=1),
```
matiasb, line 24: I see this a few times, any particular reason to set the ack_at in the future? (wondering if this could lead to unexpected things later)

engine/apps/mattermost/tests/conftest.py
```diff
@@ -24,3 +26,36 @@ def _make_mattermost_get_user_response():
...
+
+
+@pytest.fixture()
+def make_mattermost_post_response_failure():
+    def _make_mattermost_post_response(**kwargs):
+        return {
+            "status_code": kwargs["status_code"] if "status_code" in kwargs else 400,
+            "id": kwargs["id"] if "id" in kwargs else "itre5wsjnctbz78mkq9z6ci9itue",
+            "message": kwargs["message"] if "message" in kwargs else "API Error",
+            "request_id": kwargs["message"] if "message" in kwargs else "reqe5wsjnctbz78mkq9z6ci9iqer",
```
matiasb, line 50: Should this be "request_id"?

engine/apps/mattermost/models/message.py
matiasb: To be sure, I don't see a unique constraint for the model, should there be one?

matiasb (commented):
LGTM, added a few last comments. Thanks!

engine/apps/mattermost/tests/test_representative.py
ravishankar15: I saw this in couple of other places as well but i think the comment makes sense we can ack the alert at current timestamp

ravishankar15:
Hey @matiasb I have made the changes ! I have added a conditional unique constraint plus we have foreign key reference for alert group which we mostly use for fetching the messages I hope that works !

engine/apps/mattermost/models/message.py
ravishankar15: I think i missed it, I have added a conditional unique constraint plus we have foreign key reference for alert group which we mostly use for fetching the messages.

engine/apps/mattermost/models/message.py
```diff
@@ -0,0 +1,42 @@
...
+
+    created_at = models.DateTimeField(auto_now_add=True)
+
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(
+                fields=["alert_group", "channel_id", "message_type"],
+                condition=models.Q(message_type__in=[0, 1]),
```
matiasb, line 33: Last question, why do we need the condition here? (column is not nullable, and these are the 2 only possible values)

matiasb (approved):
One last question, lgtm :+1:

engine/apps/mattermost/models/message.py
ravishankar15: I think we might add other types for messages that get posted in the thread like from notifications, log message etc ?

matiasb: Ack, I think we can have the uniqueness without a condition for now, but we can revisit later.

PR #5178, opened 2024-10-15
Reschedule rate limited telegram task instead of retry

matiasb:


engine/apps/alerts/tasks/notify_user.py
```diff
@@ -445,10 +448,29 @@ def perform_notification(log_record_pk, use_default_notification_policy_fallback
...
+                )
+                UserNotificationPolicyLogRecord(
+                    author=user,
+                    type=UserNotificationPolicyLogRecord.TYPE_PERSONAL_NOTIFICATION_FAILED,
+                    notification_policy=notification_policy,
+                    reason="Telegram rate limit exceeded",
+                    alert_group=alert_group,
+                    notification_step=notification_policy.step,
+                    notification_channel=notification_channel,
+                    notification_error_code=UserNotificationPolicyLogRecord.ERROR_NOTIFICATION_POSTING_TO_TELEGRAM_IS_DISABLED,
```
Ferril, line 471: I'd suggest creating a new error code or updating the error message for this one since it says that `Telegram is disabled for the route`, what could be confusing for customers (see `render_log_line_action`)

matiasb: Hm.. good point, I now see the reported log (and also that the error code is deprecated). Adding a specific ratelimit error then :+1:

PR #5189, opened 2024-10-17
Add org setting to configure direct paging preferred policy

matiasb:
For context: https://raintank-corp.slack.com/archives/C01DHQ6LH1S/p1729267368387299?thread_ts=1729185737.051889&cid=C01DHQ6LH1S

Next steps:
  - use the setting in frontend (add participants, add responders) and mobile_app
  - expose setting in org settings page


mackgorski:
This is great thanks @matiasb I think we should continue with this. I will create a tasks for proper implementation of this for the next steps and for now let's compel this fro TTD.

joeyorlando (approved):

PR #5211, opened 2024-10-25
Enable RBAC support for public API endpoints

matiasb:

### DIFF
--- a/engine/apps/auth_token/auth.py
+++ b/engine/apps/auth_token/auth.py
@@ -9,7 +9,7 @@ from rest_framework import exceptions
 from rest_framework.authentication import BaseAuthentication, get_authorization_header
 from rest_framework.request import Request
 
-from apps.api.permissions import GrafanaAPIPermissions, LegacyAccessControlRole, RBACPermission, user_is_authorized
+from apps.api.permissions import GrafanaAPIPermissions, LegacyAccessControlRole
 from apps.grafana_plugin.helpers.gcom import check_token
 from apps.grafana_plugin.sync_data import SyncPermission, SyncUser
 from apps.user_management.exceptions import OrganizationDeletedException, OrganizationMovedException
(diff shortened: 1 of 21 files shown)


engine/apps/auth_token/auth.py
```diff
@@ -52,10 +52,8 @@ def authenticate(self, request):
         auth = get_authorization_header(request).decode("utf-8")
         user, auth_token = self.authenticate_credentials(auth)
 
-        if not user.is_active or not user_is_authorized(user, [RBACPermission.Permissions.API_KEYS_WRITE]):
```
joeyorlando, line 55: nice 👍 (this always felt like a hack)

joeyorlando (approved):
this is amazing! 🔥

joeyorlando (commented):
I'm guessing docs + (maybe?) deprecating creation of new API keys (UI + API) will be done in separate PRs?

matiasb:
> I'm guessing docs + (maybe?) deprecating creation of new API keys (UI + API) will be done in separate PRs?

Yeah, there will be more PRs coming (this is one first step from https://github.com/grafana/oncall/pull/5212); in practice this is still not really useful (because only admins can issue tokens for now).

OTOH, not sure about deprecating our API keys, something to discuss probably (with these changes any user could have their own token, with limited perms, which could be still useful to perform user-related actions: ack, resolve, etc., while service accounts are not tied to a user).

joeyorlando:
> in practice this is still not really useful (because only admins can issue tokens for now)

you don't need to be an Admin _per se_. You could have Viewer users that you assign some of [these RBAC roles](https://grafana.com/docs/grafana/latest/administration/roles-and-permissions/access-control/rbac-fixed-basic-role-definitions/) to be able to create service accounts + tokens (but right, I guess this doesn't apply to OSS where RBAC in Grafana isn't available).

PR #5220, opened 2024-10-31
fix: pass redis SSL config to connection pool

tarvip:
# What this PR does

This fix will pass SSL config properly to redis connection pool.
django-redis passes `CONNECTION_POOL_KWARGS` and not `CONNECTION_POOL_CLASS_KWARGS` to connection pool class, related django-redis code can be seen [here](https://github.com/jazzband/django-redis/blob/5.4.0/django_redis/pool.py#L26).

Also, without this fix other settings `max_connections` and `timeout` are also not passed to connection pool.

I had issues with external Redis with in-transit encryption enabled using self-signed certs (Google managed Redis).
After changing this I can properly pass self-signed CA cert to redis client using `REDIS_SSL_CA_CERTS` env variable.

There is no issue created for this problem.

## Checklist

### DIFF
--- a/engine/settings/base.py
+++ b/engine/settings/base.py
@@ -257,7 +257,7 @@ CACHES = {
             "DB": REDIS_DATABASE,
             "PARSER_CLASS": "redis.connection._HiredisParser",
             "CONNECTION_POOL_CLASS": "redis.BlockingConnectionPool",
-            "CONNECTION_POOL_CLASS_KWARGS": REDIS_SSL_CONFIG
+            "CONNECTION_POOL_KWARGS": REDIS_SSL_CONFIG
             | {
                 "max_connections": 50,
                 "timeout": 20,


JNKielmann:
We are running into the same problem trying to use oncall with a Redis on GCP with in-transit encryption enabled.

tarvip:
> We are running into the same problem trying to use oncall with a Redis on GCP with in-transit encryption enabled.

We have Grafana OnCall deployed using a Helm chart. As a workaround, I added the following env variable:
```
env:
  - name: REDIS_URI
    value: "$(REDIS_PROTOCOL)://$(REDIS_USERNAME):$(REDIS_PASSWORD)@$(REDIS_HOST):$(REDIS_PORT)/$(REDIS_DATABASE)?ssl_ca_certs=/mnt/redis-tls/server_ca.pem"
```
GCP cert is mounted to `/mnt/redis-tls/server_ca.pem`.

It would be nice to get this merged. Created this PR almost a month ago, no feedback so far.

PR #5258, opened 2024-11-15
remove add_stack_slug_to_message_title utility from push notification titles

imtoori:
# What this PR does

We noticed that the backend was adding the stack name to the notification title only on Android. 
We thought it makes sense to add the stack name only if the user has more than 1 stack connected, but that's not doable right now since the backend doesn't know how many stacks are connected in the app.
Also we took a look at the analytics for the app and basically 95% of the users have only 1 stack connected.

This pr removes the stack name from the notifications title.
If in the future we think it makes sense to add it conditionally based on the number of stacks we can open another pr, but given the very little amount of users with more than 1 stack I think this is not needed.

## Checklist


vadimkerr (approved):
LGTM

PR #5262, opened 2024-11-18
Merge alert group static labels to integration labels

Ferril:
# What this PR does
- Adds migration to merge static labels to integration labels. On creating new static label in UI saves it as integration label.
- Removes "inheritable" option for integration labels. All integration labels will be inheritable.
This PR should be merged together with frontend changes.

Related to https://github.com/grafana/oncall-private/issues/2973

## Checklist


Konstantinov-Innokentii:
@Ferril Thanks! Frontend changes are going to live in this PR as well, right?

Ferril:
> @Ferril Thanks! Frontend changes are going to live in this PR as well, right?

Not sure how it works now. Better to ask @brojd 😅

engine/apps/api/serializers/alert_receive_channel.py
```diff
@@ -124,15 +125,11 @@ def update(
...
-        inheritable_key_ids = [
-            key_id for key_id, inheritable in alert_group_labels["inheritable"].items() if inheritable
-        ]
-        instance.labels.filter(key_id__in=inheritable_key_ids).update(inheritable=True)
-        instance.labels.filter(~Q(key_id__in=inheritable_key_ids)).update(inheritable=False)
-
         # update DB cache for custom labels
         cls._create_custom_labels(instance.organization, alert_group_labels["custom"])
+        # save static labels as integration labels
+        # todo: it's needed to cover delay between backend and frontend rollout, and can be removed later
```
joeyorlando, line 131: can you elaborate a bit more on this? I think I know what you're referring to, but if we rollout the backend changes, before the frontend changes, would that solve the problem stated here?

engine/apps/api/serializers/alert_receive_channel.py
```diff
@@ -170,18 +167,38 @@ def _create_custom_labels(organization: Organization, labels: AlertGroupCustomLa
         LabelKeyCache.objects.bulk_create(label_keys, ignore_conflicts=True, batch_size=5000)
         LabelValueCache.objects.bulk_create(label_values, ignore_conflicts=True, batch_size=5000)
 
+    @staticmethod
+    def _save_static_labels_as_integration_labels(instance: AlertReceiveChannel, labels: AlertGroupCustomLabelsAPI):
+        labels_associations_to_create = []
+        labels_copy = labels[:]
```
joeyorlando, line 173: why do we need to make a copy of the list + do `labels.remove` further down?

engine/apps/alerts/migrations/0071_migrate_labels.py
```diff
@@ -0,0 +1,59 @@
...
+    AlertReceiveChannel = apps.get_model("alerts", "AlertReceiveChannel")
+
+    logging.info("Start migrating alert group static labels to integration labels")
+
+    labels_associations_to_create = []
+    alert_receive_channels_to_update = []
+
+    alert_receive_channels = AlertReceiveChannel.objects.filter(alert_group_labels_custom__isnull=False)
+    logging.info(f"Found {alert_receive_channels.count()} integrations with custom alert groups labels")
+    for alert_receive_channel in alert_receive_channels:
```
joeyorlando, line 21: do you have the rough number of how many integrations we have in production? (just curious if it's more like 50,000 or 50 million; there's probably not _that_ many, but just want to double check 🙂)

engine/apps/labels/models.py
joeyorlando, line 1: _nit_ I would include the `..._db.py` second migration file in the `engine/apps/labels` directory and just add a comment at the top saying that we'll move it to the `engine/apps/labels/migrations` directory in a future PR/release

joeyorlando (approved):
LGTM 👍  the demo in the [frontend PR](https://github.com/grafana/irm/pull/441#event-15464074572) looks 🔥 (UX seems much simpler)

engine/apps/api/serializers/alert_receive_channel.py
Ferril: that's the point, when backend changes are rolled out, it will still be possible to create static labels using frontend, since frontend has some delay in rollout, so backend will automatically save them as integration labels instead. Does it make sense?

engine/apps/api/serializers/alert_receive_channel.py
Ferril: removing items during the iteration through the original lists causes some issues since the list's length changes. I see two options to avoid this: iterating through the copy, or add items to remove to a separate list and go through it in another iteration. Lmk if I miss something 🙏

engine/apps/api/serializers/alert_receive_channel.py
joeyorlando: ahh okay, yes, I see what you mean, thanks for clarifying 👍

engine/apps/alerts/migrations/0071_migrate_labels.py
joeyorlando: ah okay, in that case this should be very performant, thanks for double checking 🙂

mrbaowei:
It seems that the Grafana Oncall OSS version still cannot use the label feature, is that correct?

joeyorlando:
@mrbaowei correct, this is a Grafana Cloud only feature

PR #5259, opened 2024-11-15
Inbound email improvements

vadimkerr:
# What this PR does

* Allows to use multiple inbound email ESPs at the same time by setting the `INBOUND_EMAIL_ESP` env variable to `amazon_ses,mailgun` for example
* Adds a new ESP `amazon_ses_validated` that performs SNS message vaildation (`django-anymail` doesn't implement it: [comment](https://github.com/anymail/django-anymail/blob/35383c7140289e82b39ada5980077898aa07d18d/anymail/webhooks/amazon_ses.py#L107-L108))

## Which issue(s) this PR closes

Related to https://github.com/grafana/oncall-private/issues/2905

## Checklist

- [ ] Unit, integration, and e2e (if applicable) tests updated
- [ ] Documentation added (or `pr:no public docs` PR label added if not required)
- [ ] Added the relevant release notes label (see labels prefixed w/ `release:`). These labels dictate how your PR will
    show up in the autogenerated release notes.

PR #5263, opened 2024-11-18
Inbound email improvements (continued)

vadimkerr:
follow up to https://github.com/grafana/oncall/pull/5259:

* Auto confirm SNS subsriptions for ESP `amazon_ses_validated`
* Add a couple of tests for SNS message validation (try with wrong SNS topic ARN, try with wrong singature)

### DIFF
--- a/engine/apps/email/inbound.py
+++ b/engine/apps/email/inbound.py
@@ -2,6 +2,7 @@ import logging
 from functools import cached_property
 from typing import Optional, TypedDict
 
+import requests
 from anymail.exceptions import AnymailAPIError, AnymailInvalidAddress, AnymailWebhookValidationFailure
 from anymail.inbound import AnymailInboundMessage
 from anymail.signals import AnymailInboundEvent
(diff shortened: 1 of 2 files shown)


engine/apps/email/inbound.py
```diff
@@ -26,12 +27,14 @@ class AmazonSESValidatedInboundWebhookView(amazon_ses.AmazonSESInboundWebhookVie
 
     def validate_request(self, request):
         """Add SNS message validation to Amazon SES inbound webhook view, which is not implemented in Anymail."""
-
-        super().validate_request(request)
```
vadimkerr, line 30: call to `super().validate_request` is unnecesery here because django-anymail collects these `validate_request` methods from the whole superclass chain and runs them all in `run_validators`: https://github.com/anymail/django-anymail/blob/35383c7140289e82b39ada5980077898aa07d18d/anymail/webhooks/base.py#L28

matiasb (approved):

PR #5266, opened 2024-11-19
Fix issue with DST starting/ending causing overlaps/gaps

### LINKED ISSUE #5247: DST starting/ending causes overlap/gap in schedules
### What went wrong?

When using the API to create a schedule that has a time zone with DST, there will be a one hour gap in the schedule when DST ends and a one hour overlap when DST begins.

Image
Image

### How do we reproduce it?

1) Create a weekly rotating shift
POST {baseUrl}/api/v1/on_call_shifts
```json
{
    "name": "Test API Shift",
    "type": "rolling_users",
    "start": "2024-10-01T09:00:00",
    "duration": 604800,
    "frequency": "weekly",
    "week_start": "MO",
    "rolling_users": [["SomeUserId"]],
    "start_rotation_from_user_index": 0
}
```

2) Create a schedule using the shift created previously. A time zone that has DST must be used.
POST {baseUrl}/api/v1/schedules
```json
{
    "name": "Test API Schedule",
    "type": "calendar",
    "time_zone": "America/Chicago",
    "enable_web_overrides": true,
    "shifts": ["IdOfShiftFromStep1"]
}
```

3) In Grafana, look at the scheduled shifts for when DST ends (Such as November 3, 2024 for America/Chicago). You should see a one hour gap.

4) In Grafana, look at the scheduled shifts for when DST begins (Such as March 9, 2024 for America/Chicago). You should see a one hour overlap.

### Grafana OnCall Version

Cloud (Plugin management says v1.14.1 is installed)

### Product Area

Schedules

### Grafana OnCall Platform?

I use Grafana Cloud

### User's Browser?

_No response_

### Anything else to add?

I found the PR #4103 that says it fixes schedule gaps, but it appears that it only fixed the visual gaps caused by a schedule being in UTC and the user's time zone being non-UTC. In that situation, there is no actual schedule gap because the schedule was in UTC, so the only issue was a visual bug. In this case, I believe there actually is a one hour gap.

zackman0010:
# What this PR does

This older version of recurring_ical_events does not call the pytz .normalize() function, which can cause some invalid datetime objects to return when a DST swap happens. For example: Nov 3, 2024 9:00 AM CDT instead of the correct 8:00 AM CST). By calling tz.normalize on the end date and checking if the time zone information changed, we can detect when DST starts/stops and adjust the end date accordingly.

| | DST stopping on November 3, 2024: | DST starting on March 9, 2024 |
|-|-----------------------------------------|-----------------------------------|
| Before | image | image
| After | image | image |

Closes #5247

## Checklist

### DIFF
--- a/engine/apps/schedules/ical_events/adapter/amixr_recurring_ical_events_adapter.py
+++ b/engine/apps/schedules/ical_events/adapter/amixr_recurring_ical_events_adapter.py
@@ -20,6 +20,15 @@ EXTRA_LOOKUP_DAYS = 16
 
 
 class AmixrRecurringIcalEventsAdapter(IcalService):
+    def _normalize(self, dt: datetime.datetime) -> datetime.datetime:
+        tz = getattr(dt, "tzinfo", None)
+        if tz:
+            normalized = tz.normalize(dt)
+            if normalized.tzinfo != tz:
+                diff = dt.dst() - normalized.dst()
+                dt = normalized + diff
+        return dt
+
     def get_events_from_ical_between(
         self, calendar: Calendar, start_date: datetime.datetime, end_date: datetime.datetime
     ) -> typing.List[Event]:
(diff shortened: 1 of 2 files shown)


matiasb:
Thanks @zackman0010 for the contribution (reporting the bug, and the fix)!
LGTM :+1: just did a minor refactoring and added a test.

PR #5282, opened 2024-11-21
Listen for alert group actions during whole notification call

Konstantinov-Innokentii:
1. Wrap whole message in twiml <Gather>  - that's an actual fix
2. Use twilio helper lib to build twiml queries
3. URLencode twimlquery only once before making a call to reduce code duplication.


engine/apps/twilioapp/gather.py
```diff
@@ -85,5 +85,5 @@ def get_gather_url():
     return create_engine_url(reverse("twilioapp:gather"))
 
 
-def get_gather_message():
+def get_alert_group_gather_instructions():
     return "Press 1 to acknowledge, 2 to resolve, 3 to silence to 30 minutes"
```
joeyorlando, line 89:
```suggestion
    return "Press 1 to acknowledge, 2 to resolve, 3 to silence for 30 minutes"
```

joeyorlando (approved):

Issue #5244, opened 2024-11-07
Migrations failing when upgrading to 1.12.1


WoodyWoodsta:
### What went wrong?

**What happened**:
- Upgrading to 1.12.1 from 1.11.3, the migration job that runs shows the following error:

```
Apply all migrations: admin, alerts, auth, auth_token, base, contenttypes, email, exotel, fcm_django, google, heartbeat, labels, mobile_app, oss_installation, phone_notifications, schedules, sessions, slack, social_django, telegram, twilioapp, user_management, webhooks, zvonok
Running migrations:
source=engine:app google_trace_id=none logger=apps.alerts.migrations.0063_migrate_channelfilter_slack_channel_id Starting migration to populate slack_channel field.
Traceback (most recent call last):
File "/usr/local/lib/python3.12/site-packages/django/db/backends/utils.py", line 87, in _execute
return self.cursor.execute(sql)
^^^^^^^^^^^^^^^^^^^^^^^^
psycopg2.errors.SyntaxError: syntax error at or near "JOIN"
LINE 3: JOIN alerts_alertreceivechannel AS arc ON arc.id = cf.al...
^


The above exception was the direct cause of the following exception:

Traceback (most recent call last):
File "/etc/app/manage.py", line 34, in <module>
execute_from_command_line(sys.argv)
File "/usr/local/lib/python3.12/site-packages/django/core/management/__init__.py", line 442, in execute_from_command_line
utility.execute()
File "/usr/local/lib/python3.12/site-packages/django/core/management/__init__.py", line 436, in execute
self.fetch_command(subcommand).run_from_argv(self.argv)
File "/usr/local/lib/python3.12/site-packages/django/core/management/base.py", line 412, in run_from_argv
self.execute(*args, **cmd_options)
File "/usr/local/lib/python3.12/site-packages/django/core/management/base.py", line 458, in execute
output = self.handle(*args, **options)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/local/lib/python3.12/site-packages/django/core/management/base.py", line 106, in wrapper
res = handle_func(*args, **kwargs)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/local/lib/python3.12/site-packages/django/core/management/commands/migrate.py", line 356, in handle
post_migrate_state = executor.migrate(
^^^^^^^^^^^^^^^^^
[...]

PR #5292, opened 2024-11-22
refactor `SlackMessage.channel_id` (`CHAR` field) to `SlackMessage.channel` (foreign key relationship)

joeyorlando:
# What this PR does

Related to https://github.com/grafana/oncall-private/issues/2947

## Checklist

### DIFF
--- a/engine/apps/slack/models/slack_message.py
+++ b/engine/apps/slack/models/slack_message.py
@@ -18,6 +18,9 @@ from apps.slack.errors import (
 
 if typing.TYPE_CHECKING:
     from apps.alerts.models import AlertGroup
+    from apps.base.models import UserNotificationPolicy
+    from apps.slack.models import SlackChannel
+    from apps.user_management.models import User
 
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.DEBUG)
@@ -25,17 +28,38 @@ logger.setLevel(logging.DEBUG)
 
 class SlackMessage(models.Model):
     alert_group: typing.Optional["AlertGroup"]
+    channel: "SlackChannel"
 
     id = models.CharField(primary_key=True, default=uuid.uuid4, editable=False, max_length=36)
-
     slack_id = models.CharField(max_length=100)
 
-    # TODO: convert this to a foreign key field to SlackChannel
-    channel_id = models.CharField(max_length=100, null=True, default=None)
+    _channel_id = models.CharField(max_length=100, null=True, default=None)
+    """
+    DEPRECATED: this is no longer being referenced/set, drop in a separate PR/release
+    """
+
+    channel = models.ForeignKey(
+        "slack.SlackChannel", on_delete=models.CASCADE, null=True, default=None, related_name="slack_message"
+    )
+    """
+    TODO: once we've migrated the data in `_channel_id` to this field, set `null=False`
+    as we should always have a `channel` associated with a message
+    """
 
     organization = models.ForeignKey(
-        "user_management.Organization", on_delete=models.CASCADE, null=True, default=None, related_name="slack_message"
+        "user_management.Organization",
+        on_delete=models.CASCADE,
+        null=True,
+        default=None,
+        related_name="slack_message",
     )
+    """
+    DEPRECATED: this is no longer being referenced/set
+
+    TODO: do we really need organization here? We now have a foreign key reference to channel, which has a foreign
+    key reference to slack_team_identity, which has a foreign key reference to organization...
+    """
+
     _slack_team_identity = models.ForeignKey(
         "slack.SlackTeamIdentity",
         on_delete=models.PROTECT,
@@ -44,6 +68,12 @@ class SlackMessage(models.Model):
         related_name="slack_message",
         db_column="slack_team_identity",
     )
+    """
+    DEPRECATED: this is no longer being referenced/set
+
+    TODO: drop this field in a separate PR/release
+    since we now have a foreign key reference to channel, channel already has a reference to the team identity
+    """
 
     ack_reminder_message_ts = models.CharField(max_length=100, null=True, default=None)
 
@@ -72,21 +102,12 @@ class SlackMessage(models.Model):
 
     @property
     def slack_team_identity(self):
-        if self._slack_team_identity is None:
-            if self.organization is None:  # strange case when organization is None
-                logger.warning(
-                    f"SlackMessage (pk: {self.pk}) fields _slack_team_identity and organization is None. "
-                    f"It is strange!"
-                )
-                return None
-            self._slack_team_identity = self.organization.slack_team_identity
-            self.save()
-        return self._slack_team_identity
+        return self.channel.slack_team_identity
 
     @property
     def permalink(self) -> typing.Optional[str]:
-        # Don't send request for permalink if there is no slack_team_identity or slack token has been revoked
-        if self.cached_permalink or not self.slack_team_identity or self.slack_team_identity.detected_token_revoked:
+        # Don't send request for permalink if slack token has been revoked
+        if self.cached_permalink or self.slack_team_identity.detected_token_revoked:
             return self.cached_permalink
 
         try:
@@ -105,10 +126,18 @@ class SlackMessage(models.Model):
     def deep_link(self) -> str:
         return f"https://slack.com/app_redirect?channel={self.channel_id}&team={self.slack_team_identity.slack_id}&message={self.slack_id}"
 
-    def send_slack_notification(self, user, alert_group, notification_policy):
+    def send_slack_notification(
+        self,
+        user: "User",
+        alert_group: "AlertGroup",
+        notification_policy: "UserNotificationPolicy",
+    ) -> None:
         from apps.base.models import UserNotificationPolicyLogRecord
 
         slack_message = alert_group.slack_message
+        slack_channel = slack_message.channel
+        channel_id = slack_channel.slack_id
+
         user_verbal = user.get_username_with_slack_verbal(mention=True)
 
         slack_user_identity = user.slack_user_identity
@@ -142,8 +171,8 @@ class SlackMessage(models.Model):
                 },
             }
         ]
+
         sc = SlackClient(self.slack_team_identity, enable_ratelimit_retry=True)
-        channel_id = slack_message.channel_id
 
         try:
             result = sc.chat_postMessage(
@@ -190,12 +219,8 @@ class SlackMessage(models.Model):
             ).save()
             return
         else:
-            alert_group.slack_messages.create(
-                slack_id=result["ts"],
-                organization=self.organization,
-                _slack_team_identity=self.slack_team_identity,
-                channel_id=channel_id,
-            )
+            alert_group.slack_messages.create(slack_id=result["ts"], channel=slack_channel)
+
             # create success record
             UserNotificationPolicyLogRecord.objects.create(
                 author=user,
--- a/engine/apps/alerts/models/alert_group.py
+++ b/engine/apps/alerts/models/alert_group.py
@@ -1983,7 +1983,7 @@ class AlertGroup(AlertGroupSlackRenderingMixin, EscalationSnapshotMixin, models.
         if not self.channel.organization.slack_team_identity:
             return None
         elif self.slack_message:
-            return self.slack_message.channel_id
+            return self.slack_message.channel.slack_id
         elif self.channel_filter:
             return self.channel_filter.slack_channel_id_or_org_default_id
         return None
--- a/engine/apps/alerts/tests/test_alert_group.py
+++ b/engine/apps/alerts/tests/test_alert_group.py
@@ -24,14 +24,15 @@ def test_render_for_phone_call(
     make_alert_receive_channel,
     make_alert_group,
     make_alert,
+    make_slack_channel,
+    make_slack_message,
 ):
-    organization, _ = make_organization_with_slack_team_identity()
+    organization, slack_team_identity = make_organization_with_slack_team_identity()
     alert_receive_channel = make_alert_receive_channel(organization)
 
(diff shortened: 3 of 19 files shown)


engine/apps/slack/models/slack_message.py
```diff
@@ -72,26 +102,21 @@ class Meta:
...
+        return self.channel.slack_team_identity
+
+    @property
+    def channel_slack_id(self) -> str:
+        return self.channel.slack_id
 
     @property
     def permalink(self) -> typing.Optional[str]:
-        # Don't send request for permalink if there is no slack_team_identity or slack token has been revoked
-        if self.cached_permalink or not self.slack_team_identity or self.slack_team_identity.detected_token_revoked:
```
joeyorlando, line 89: I verified (in production) and don't see any records where `slack_team_identity` is null for `SlackMessage` (fwiw, I think this field should eventually be dropped from this model and we should use `slack_message.channel.slack_team_identity` instead; see comments above)

engine/apps/slack/models/slack_message.py
```diff
@@ -72,26 +102,21 @@ class Meta:
 
     @property
     def slack_team_identity(self):
-        if self._slack_team_identity is None:
-            if self.organization is None:  # strange case when organization is None
-                logger.warning(
-                    f"SlackMessage (pk: {self.pk}) fields _slack_team_identity and organization is None. "
-                    f"It is strange!"
-                )
```
joeyorlando, line 80: I wasn't able to find any mentions of this log message for the last 90 days + there's no git history available unfortunately for why this is here. As I am proposing to (eventually) drop `slack_team_identity` from this model, I propose to remove this

engine/apps/slack/migrations/0006_rename_channel_id_slackmessage__channel_id_and_more.py
joeyorlando, line 1: main change

engine/apps/slack/migrations/0007_migrate_slackmessage_channel_id.py
```diff
@@ -0,0 +1,78 @@
...
+            {SlackMessage._meta.db_table} AS sm
+        JOIN {SlackChannel._meta.db_table} AS sc
+            ON sc.slack_id = sm._channel_id
+            AND sc.slack_team_identity_id = sm.slack_team_identity
+        SET
+            sm.channel_id = sc.id
+        WHERE
+            sm._channel_id IS NOT NULL
+            AND sm.slack_team_identity IS NOT NULL;
+        """
```
joeyorlando, line 58: main change. Added queries for sqlite/postgres here to avoid https://github.com/grafana/oncall/issues/5244#issuecomment-2493688544

engine/apps/slack/models/slack_message.py
```diff
@@ -103,12 +128,16 @@ def permalink(self) -> typing.Optional[str]:
 
     @property
     def deep_link(self) -> str:
-        return f"https://slack.com/app_redirect?channel={self.channel_id}&team={self.slack_team_identity.slack_id}&message={self.slack_id}"
+        return f"https://slack.com/app_redirect?channel={self.channel_slack_id}&team={self.slack_team_identity.slack_id}&message={self.slack_id}"
 
-    def send_slack_notification(self, user, alert_group, notification_policy):
+    def send_slack_notification(self, user: "User", notification_policy: "UserNotificationPolicy") -> None:
```
joeyorlando, line 133: given that this is an instance method, `self.alert_group` already is a thing, so I don't _think_ we need to re-pass in `alert_group`?

engine/apps/slack/migrations/0007_migrate_slackmessage_channel_id.py
joeyorlando: TODO: I'm planning on running this SQL statement against a prod clone just to get an idea of performance before merging this PR

engine/apps/schedules/models/shift_swap_request.py
```diff
@@ -164,7 +164,15 @@ def status(self) -> str:
         return self.Statuses.OPEN
 
     @property
-    def slack_channel_id(self) -> str | None:
+    def slack_channel(self) -> typing.Optional["SlackChannel"]:
+        """
+        This is only set if the schedule associated with the shift swap request
+        has a Slack channel configured for it.
+        """
+        return self.schedule.slack_channel
```
joeyorlando, line 172: added for [this](https://github.com/grafana/oncall/pull/5292/files#diff-505c54140ab7b80107fe3845db9554bf9434a10f57fae4335f60eab72d366d91R169)

engine/apps/slack/models/slack_message.py
```diff
@@ -18,24 +18,48 @@
...
+    DEPRECATED: this is no longer being referenced/set, drop in a separate PR/release
+    """
+
+    channel = models.ForeignKey(
+        "slack.SlackChannel", on_delete=models.CASCADE, null=True, default=None, related_name="slack_messages"
+    )
+    """
+    TODO: once we've migrated the data in `_channel_id` to this field, set `null=False`
+    as we should always have a `channel` associated with a message
+    """
```
vadimkerr, line 47: what if Slack channel gets deleted (for example [here](https://github.com/grafana/oncall/blob/14c1b37c87ada3458bbd07024dabc0480f9da3a7/engine/apps/slack/scenarios/slack_channel.py#L56-L59))? will it still be possible to use `SlackMessage.permalink` to redirect users to Slack?

`on_delete=models.CASCADE` this means if alert group was posted to `#alerts` and then `#alerts` is deleted, we will not be able to reference Slack messages from that channel in alert groups (for things like Slack permalink, resolution notes, etc.). I think we don't want that right?

joeyorlando: good question! I was also questioning what the right `on_delete` would be here. My thought was that if you delete a Slack channel.. everything inside the channel vanishes. So permalinks/resolution notes/etc are no longer relevant for those `SlackMessage`s (but maybe I'm overlooking something?)

vadimkerr (commented):
LGTM, just a couple of suggestions about the `SlackTeamIdentity` <-> `Organization` relationship

vadimkerr (approved):

engine/apps/slack/models/slack_message.py
```diff
@@ -72,26 +102,21 @@ class Meta:
...
-                )
-                return None
-            self._slack_team_identity = self.organization.slack_team_identity
-            self.save()
-        return self._slack_team_identity
+        return self.channel.slack_team_identity
+
+    @property
+    def channel_slack_id(self) -> str:
+        return self.channel.slack_id
```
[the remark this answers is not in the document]
joeyorlando: fair point! addressed in https://github.com/grafana/oncall/pull/5292/commits/536b1bc1ec73b553e413bba4a6ed071ab185d8ad

engine/apps/slack/models/slack_message.py
```diff
@@ -18,24 +18,48 @@
...
-        "user_management.Organization", on_delete=models.CASCADE, null=True, default=None, related_name="slack_message"
+        "user_management.Organization",
+        on_delete=models.CASCADE,
+        null=True,
+        default=None,
+        related_name="slack_message",
     )
+    """
+    TODO: do we really need organization here? We now have a foreign key reference to channel, which has a foreign
+    key reference to slack_team_identity, which has a foreign key reference to organization...
```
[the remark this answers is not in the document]
joeyorlando: Hmm right, good call out. I think I slightly misunderstood the relationship, thanks for clarifying 👍
- I assumed that every `SlackChannel` was unique for a given stack's (org) workspace installation.. ex. if you have stack/org A, and stack/org B both installed to workspace X (which has 1 slack channel, Y), you'd have two `slack_team_identity` objects and two `slack_channel` objects, when in reality its only 1 `slack_team_identity` and 1 `slack_channel`

To me it feels duplicative to have a relationship to _both_ `organization` **and** `slack_team_identity` on this table. Given what you said (and my, hopefully correct, understanding) maybe what we could do here instead is, when creating, just set `SlackMessage.organization`, and when we need to reference the `slack_team_identity`, you'd be able to just do `slack_message.organization.slack_team_identity`. This would allow us to (eventually) drop the relationship to `slack_team_identity` on this table. wdyt?

engine/apps/slack/migrations/0006_rename_channel_id_slackmessage__channel_id_and_more.py
```diff
@@ -0,0 +1,61 @@
+# NOTE: the foreign key checks are disabled and re-enabled in this migration
+# see the following resources as to why
+#
+# https://arc.net/l/quote/hgqztayk
+# https://dba.stackexchange.com/a/316683
+#
+# tldr;
+# The INPLACE algorithm is supported when foreign_key_checks is disabled. Otherwise, only the COPY algorithm is supported
+#
+# which means that creating this foreign key constraint, without this, on a large table would take foreverz
```
joeyorlando, line 10: this is worth calling out..

engine/engine/management/commands/batch_migrate_slack_message_channel.py
joeyorlando, line 1: this will be run manually, as [step 2 of the proposed migration plan](https://raintank-corp.slack.com/archives/C06K1MQ07GS/p1732558692071039?thread_ts=1732555465.144099&cid=C06K1MQ07GS), via 👇  (following is from testing this out locally):

```bash
oncall-dev-engine-64985c9d6-qzb8b:/etc/app# python manage.py batch_migrate_slack_message_channel
/usr/local/lib/python3.12/site-packages/telegram/utils/request.py:49: UserWarning: python-telegram-bot is using upstream urllib3. This is allowed but not supported by python-telegram-bot maintainers.
  warnings.warn(
Starting batch update of SlackMessage.channel_id...
Total records to update: 1
Created_at range: 2024-11-25 19:41:10.352158+00:00 to 2024-11-25 19:41:10.352158+00:00
Batch interval: 1 day, 0:00:00
Total batches: 1
Batch 1/1: Processing records from 2024-11-25 19:41:10.352158+00:00 to 2024-11-25 19:41:11.352158+00:00
Batch 1/1: Updated 1 records
Batch update completed successfully. Total records updated: 1
Total time taken: 0.03 seconds
```

engine/apps/slack/tests/test_slack_message.py
```diff
@@ -54,10 +77,63 @@ def test_slack_message_deep_link(
...
+@pytest.mark.django_db
+def test_slack_message_permalink_token_revoked(mock_slack_api_call, slack_message_setup):
+    slack_message = slack_message_setup(cached_permalink=None)
+    slack_message.slack_team_identity.detected_token_revoked = timezone.now()
+    slack_message.slack_team_identity.save()
+
+    assert slack_message.slack_team_identity is not None
+    assert slack_message.permalink is None
+
+    mock_slack_api_call.assert_not_called()
```
joeyorlando, line 139: these tests are consolidated from [`engine/apps/slack/test_slack_message.py`](https://github.com/grafana/oncall/pull/5292/files#diff-cd9b05b5b105ca44555f32c26d78d700008b71869b51b0cfb549c75d5e533ac3)

engine/apps/slack/test_slack_message.py
joeyorlando, line 1: this file has been consolidated into [`engine/apps/slack/tests/test_slack_message.py`](https://github.com/grafana/oncall/pull/5292/files#diff-fc850d58e5dcaa4ab7489755991da6293c25dbe1ef8a501445af7222e9db1005)

engine/apps/slack/models/slack_message.py
```diff
@@ -103,12 +117,28 @@ def permalink(self) -> typing.Optional[str]:
...
+        """
+        NOTE: the reason why we pass in `alert_group` as an argument here, as opposed to just doing
+        `self.alert_group`, is that it "looks like" we may have a race condition occuring between two celery tasks:
+        - one which sends out the initial slack message
+        - one which notifies the user (this method) inside of the above slack message's thread
+
+        Still some more investigation needed to confirm this, but for now, we'll pass in the `alert_group` as an argument
+        """
 
-    def send_slack_notification(self, user, alert_group, notification_policy):
```
joeyorlando, line 108: I tried removing `alert_group` here (since `self.alert_group` is a thing..) but it appears that when notifying a user in the Alert Group message thread, the second `SlackMessage` that gets persisted (the one that represents the thread notification message) has a `null` `alert_group` (and I believe it had `null` for another field such as `organization`?)

engine/apps/slack/models/slack_message.py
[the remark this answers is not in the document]
vadimkerr: yes I think this makes sense 👍 not sure how big of a refactor this would be, but I agree with the idea that having both `organization` and `slack_team_identity` on `SlackMessage` is not optimal

engine/apps/slack/migrations/0006_rename_channel_id_slackmessage__channel_id_and_more.py
```diff
@@ -0,0 +1,61 @@
+# NOTE: the foreign key checks are disabled and re-enabled in this migration
+# see the following resources as to why
+#
+# https://arc.net/l/quote/hgqztayk
```
vadimkerr, line 4: `https://arc.net/l/quote/...` TIL 😎

vadimkerr (approved):

engine/apps/slack/migrations/0006_rename_channel_id_slackmessage__channel_id_and_more.py
joeyorlando: YIL (yesterday-I-learned) as well 😄:

Screenshot 2024-11-26 at 5 22 52 AM

"Copy Link to Highlight" in Arc seems to have its own fancy "URL shortener"

PR #5296, opened 2024-11-26
User Notification and Escalation Chain flow

ravishankar15:
# What this PR does
User notification flow

Related to [issue link here]

## Checklist


engine/apps/mattermost/tasks.py
```diff
@@ -89,3 +93,76 @@ def on_alert_group_action_triggered_async(log_record_id):
...
+        client = MattermostClient()
+        mattermost_post = client.create_post(channel_id=mattermost_channel.channel_id, data=payload)
+    except MattermostAPITokenInvalid:
+        logger.error(f"Mattermost API token is invalid could not create post for alert {alert_group_pk}")
+    except MattermostAPIException as ex:
+        logger.error(f"Mattermost API error {ex}")
+        if ex.status not in [status.HTTP_401_UNAUTHORIZED]:
+            raise ex
+    else:
+        MattermostMessage.create_message(
```
matiasb, line 166: I don't think we need to track this message? (we won't be updating it in any way)
What we will need though is a success `UserNotificationPolicyLogRecord` entry to have a record the notification was sent (and a failed one in case of the exceptions above; we need to make sure there is a `failed` or `success` log record for each `triggered` one)

engine/apps/mattermost/tests/test_tasks.py
```diff
@@ -189,3 +214,290 @@ def test_on_alert_group_action_triggered_async_failure(
...
+    make_user_notification_policy,
+):
+    organization, user = make_organization_and_user()
+
+    alert_receive_channel = make_alert_receive_channel(organization)
+    alert_group = make_alert_group(alert_receive_channel)
+    make_alert(alert_group=alert_group, raw_request_data=alert_receive_channel.config.example_payload)
+    make_mattermost_channel(organization=organization, is_default_channel=True)
+    make_mattermost_message(alert_group, MattermostMessage.ALERT_GROUP_MESSAGE)
+    make_mattermost_user(user=user)
```
matiasb, line 281: You don't need to setup a user in this case, right?

engine/apps/mattermost/tests/test_tasks.py
```diff
@@ -189,3 +214,290 @@ def test_on_alert_group_action_triggered_async_failure(
...
+    make_alert,
+    make_mattermost_channel,
+    make_mattermost_message,
+    make_mattermost_user,
+    make_user_notification_policy,
+):
+    organization, user = make_organization_and_user()
+
+    alert_receive_channel = make_alert_receive_channel(organization)
+    alert_group = make_alert_group(alert_receive_channel)
```
matiasb, line 310: Similarly, you shouldn't need to setup an alert group here?

engine/apps/mattermost/tests/test_tasks.py
```diff
@@ -189,3 +214,290 @@ def test_on_alert_group_action_triggered_async_failure(
...
+    httpretty.register_uri(httpretty.POST, url, responses=[mock_response])
+
+    notify_user_about_alert_async(
+        user_pk=user.pk, alert_group_pk=alert_group.pk, notification_policy_pk=user_notification_policy.pk
+    )
+
+    mattermost_message = alert_group.mattermost_messages.order_by("created_at").last()
+    assert mattermost_message.post_id == data["id"]
+    assert mattermost_message.channel_id == data["channel_id"]
+    assert mattermost_message.message_type == MattermostMessage.USER_NOTIFACTION_MESSAGE
```
matiasb, line 260: This should check for the success log record.

engine/apps/mattermost/tests/test_tasks.py
```diff
@@ -189,3 +214,290 @@ def test_on_alert_group_action_triggered_async_failure(
...
+        step=UserNotificationPolicy.Step.NOTIFY,
+        notify_by=UserNotificationPolicy.NotificationChannel.TESTONLY,
+    )
+
+    with patch("apps.mattermost.client.MattermostClient.create_post") as mock_post_call:
+        notify_user_about_alert_async(
+            user_pk=123, alert_group_pk=alert_group.pk, notification_policy_pk=user_notification_policy.pk
+        )
+
+    mock_post_call.assert_not_called()
```
matiasb, line 293: This should check for the failed log record (similar below).

engine/apps/mattermost/tasks.py
```diff
@@ -89,3 +93,76 @@ def on_alert_group_action_triggered_async(log_record_id):
...
+    payload = {"root_id": mattermost_messsage.post_id, "message": message}
+
+    try:
+        client = MattermostClient()
+        mattermost_post = client.create_post(channel_id=mattermost_channel.channel_id, data=payload)
+    except MattermostAPITokenInvalid:
+        logger.error(f"Mattermost API token is invalid could not create post for alert {alert_group_pk}")
+    except MattermostAPIException as ex:
+        logger.error(f"Mattermost API error {ex}")
+        if ex.status not in [status.HTTP_401_UNAUTHORIZED]:
```
matiasb, line 163: Use `!=` instead?

engine/apps/mattermost/tests/test_tasks.py
```diff
@@ -189,3 +214,290 @@ def test_on_alert_group_action_triggered_async_failure(
...
+    data = make_mattermost_post_response_failure(status_code=status_code)
+    mock_response = httpretty.Response(json.dumps(data), status=status_code)
+    httpretty.register_uri(httpretty.POST, url, status=status_code, responses=[mock_response])
+
+    if status_code != 401:
+        with pytest.raises(MattermostAPIException):
+            notify_user_about_alert_async(
+                user_pk=user.pk, alert_group_pk=alert_group.pk, notification_policy_pk=user_notification_policy.pk
+            )
+    else:
```
matiasb, line 466: What should happen in each case? (401, no retry, log; other, retry, no log?)

engine/apps/mattermost/tests/test_tasks.py
ravishankar15: I think the user is needed for the notification policy so passing the wrong id to the task function

engine/apps/mattermost/tests/test_tasks.py
ravishankar15: Since this is to check if the actual user present we cannot create log records as the user is required for creating the same. Hence I am just adding a log and returning from task.

engine/apps/mattermost/tests/test_tasks.py
ravishankar15: Yup that is the idea

engine/apps/mattermost/tasks.py
```diff
@@ -89,3 +93,91 @@ def on_alert_group_action_triggered_async(log_record_id):
...
+            raise e
+
+    mattermost_channel = MattermostChannel.get_channel_for_alert_group(alert_group=alert_group)
+    if not mattermost_channel:
+        logger.error(f"Mattermost channel not found for user notification {user_pk}")
+        return
+
+    templated_alert = AlertGroupMattermostRenderer(alert_group).alert_renderer.templated_alert
+
+    print("Check identity")
```
matiasb, line 149: Is this a forgotten print?

matiasb (approved):
:+1:

PR #5305, opened 2024-11-28
feat: add service account checks in plugin auth

matiasb:
Related to https://github.com/grafana/irm/pull/459

Allow org sync requests from service account users. Also trigger a sync during public API requests if the org wasn't yet setup.

### DIFF
--- a/engine/apps/auth_token/auth.py
+++ b/engine/apps/auth_token/auth.py
@@ -9,6 +9,7 @@ from rest_framework import exceptions
 from rest_framework.authentication import BaseAuthentication, get_authorization_header
 from rest_framework.request import Request
 
+from apps.auth_token.grafana.grafana_auth_token import setup_organization
 from apps.grafana_plugin.helpers.gcom import check_token
 from apps.grafana_plugin.sync_data import SyncPermission, SyncUser
 from apps.user_management.exceptions import OrganizationDeletedException, OrganizationMovedException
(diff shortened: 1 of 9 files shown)


engine/apps/auth_token/auth.py
```diff
@@ -133,6 +134,14 @@ def _get_user(request: Request, organization: Organization) -> User:
         except KeyError:
             user_id = context["UserID"]
 
+        if context.get("IsServiceAccount", False):
```
mderynck, line 137: This just reminded me we need to clean-up the GrafanaHeadersMixin in mixins.py at some point whenever we remove the old status endpoint.  This way of doing it is better in that it is more tolerant of missing/different fields when dealing with plugin rollout difference between objects.

engine/apps/grafana_plugin/helpers/client.py
```diff
@@ -337,6 +337,9 @@ def create_service_account_token(
     def get_service_account_token_permissions(self) -> APIClientResponse[typing.Dict[str, typing.List[str]]]:
         return self.api_get("api/access-control/user/permissions")
 
+    def setup_organization(self) -> APIClientResponse:
+        return self.api_post(f"api/plugins/{PluginID.ONCALL}/resources/plugin/sync?wait=true")
```
mderynck, line 341: We might also want `force=true` here as well

joeyorlando (commented):
LGTM 👍

engine/apps/auth_token/auth.py
```diff
@@ -357,12 +369,15 @@ def authenticate(self, request):
...
         return self.authenticate_credentials(organization, auth)
 
-    def get_organization(self, request):
+    def get_organization(self, request, auth):
         grafana_url = request.headers.get(X_GRAFANA_URL)
         if grafana_url:
             organization = Organization.objects.filter(grafana_url=grafana_url).first()
             if not organization:
-                raise exceptions.AuthenticationFailed("Invalid Grafana URL.")
+                success = setup_organization(grafana_url, auth)
```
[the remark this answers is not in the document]
matiasb: There is a use case in which you can bootstrap a Grafana stack from scratch (via Terraform), setting up a service account token in the process with which you could hit our API and OnCall may not know about the organization yet, so this should sync the org if the service account token auth passes and we don't have a record for that org yet.

engine/apps/grafana_plugin/helpers/client.py
matiasb: Yeah, I wasn't sure about `force=true`, but maybe we can set it for now

joeyorlando (approved):

PR #5314, opened 2024-11-29
fix: patch slack ID reference issue

joeyorlando:

Fixes https://github.com/grafana/irm/issues/469

## Checklist

### DIFF
--- a/engine/apps/alerts/models/alert_group.py
+++ b/engine/apps/alerts/models/alert_group.py
@@ -1982,9 +1982,7 @@ class AlertGroup(AlertGroupSlackRenderingMixin, EscalationSnapshotMixin, models.
     def slack_channel_id(self) -> str | None:
         channel_filter = self.channel_filter
 
-        if not self.channel.organization.slack_team_identity:
-            return None
-        elif self.slack_message:
+        if self.slack_message:
             # TODO: once _channel_id has been fully migrated to channel, remove _channel_id
             # see https://raintank-corp.slack.com/archives/C06K1MQ07GS/p173255546
             #
--- a/engine/apps/alerts/models/alert_group.py
+++ b/engine/apps/alerts/models/alert_group.py
@@ -2014,6 +2014,8 @@ class AlertGroup(AlertGroupSlackRenderingMixin, EscalationSnapshotMixin, models.
 
     @property
     def slack_channel_id(self) -> str | None:
+        channel_filter = self.channel_filter
+
         if not self.channel.organization.slack_team_identity:
             return None
         elif self.slack_message:
@@ -2022,8 +2024,8 @@ class AlertGroup(AlertGroupSlackRenderingMixin, EscalationSnapshotMixin, models.
             #
             # return self.slack_message.channel.slack_id
             return self.slack_message._channel_id
-        elif self.channel_filter and self.channel_filter.slack_channel_or_org_default:
-            return self.channel_filter.slack_channel_or_org_default.slack_id
+        elif channel_filter and channel_filter.slack_channel_or_org_default:
+            return channel_filter.slack_channel_or_org_default.slack_id
         return None
 
     @property
--- a/engine/apps/alerts/models/alert_group.py
+++ b/engine/apps/alerts/models/alert_group.py
@@ -1978,14 +1978,52 @@ class AlertGroup(AlertGroupSlackRenderingMixin, EscalationSnapshotMixin, models.
     def is_presented_in_slack(self):
         return self.slack_message and self.channel.organization.slack_team_identity
 
+    # TODO: once we've finished migrating slack_message._channel_id to slack_message.channel, we can enable this
+    # and refactor the related code
+    #
+    # @property
+    # def organization(self) -> Organization:
+    #     return self.channel.organization
+
+    # @property
+    # def slack_channel(self) -> typing.Optional["SlackChannel"]:
+    #     """
+    #     If the `AlertGroup` already has a `SlackMessage` associated with it, that means that we already
+    #     know the channel.
+
+    #     Otherwise, infer the Slack channel from the `ChannelFilter` associated with the `AlertGroup`.
+    #     """
+    #     if self.slack_message:
+    #         # NOTE: this isn't available yet because we haven't finished migrating slack_message._channel_id to
+    #         # slack_message.channel
+    #         return self.slack_message.channel
+    #     elif self.channel_filter and self.channel_filter.slack_channel:
+    #         return self.channel_filter.slack_channel
+    #     return self.organization.default_slack_channel
+
+    # @property
+    # def slack_channel_id(self) -> str | None:
+    #     """
+    #     If the `AlertGroup` already has a `SlackMessage` associated with it, that means that we already
+    #     know the channel.
+
+    #     Otherwise, infer the Slack channel ID from the `ChannelFilter` associated with the `AlertGroup`.
+    #     """
+    #     return self.slack_channel.slack_id if self.slack_channel else None
+    # END TODO:
+
     @property
     def slack_channel_id(self) -> str | None:
         if not self.channel.organization.slack_team_identity:
             return None
         elif self.slack_message:
-            return self.slack_message.channel.slack_id
-        elif self.channel_filter:
-            return self.channel_filter.slack_channel_id_or_org_default_id
+            # TODO: once _channel_id has been fully migrated to channel, remove _channel_id
+            # see https://raintank-corp.slack.com/archives/C06K1MQ07GS/p173255546
+            #
+            # return self.slack_message.channel.slack_id
+            return self.slack_message._channel_id
+        elif self.channel_filter and self.channel_filter.slack_channel_or_org_default:
+            return self.channel_filter.slack_channel_or_org_default.slack_id
         return None
 
     @property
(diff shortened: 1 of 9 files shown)


engine/apps/alerts/models/channel_filter.py
```diff
@@ -172,14 +172,8 @@ def slack_channel_slack_id(self) -> typing.Optional[str]:
         return self.slack_channel.slack_id if self.slack_channel else None
 
     @property
-    def slack_channel_id_or_org_default_id(self):
-        organization = self.alert_receive_channel.organization
-
-        if organization.slack_team_identity is None:
-            return None
```
joeyorlando, line 179: I'm not sure this check needs to be here? 🤔

```python
if organization.slack_team_identity is None:
  return None
``` 

`organization.slack_team_identity` and `organization.default_slack_channel` are two distinct things.

Also, I'm _fairly_ certain that when a user uninstalls the Slack integration (ie. `organization.slack_team_identity` is set to `None`) we unset the `default_slack_channel` as well:

https://github.com/grafana/oncall/blob/jorlando/cleanup_slack_id/engine/apps/slack/installation.py#L70

https://github.com/grafana/oncall/blob/jorlando/cleanup_slack_id/engine/apps/slack/tasks.py#L200

---

I noticed we use `if organization.slack_team_identity is not None`, throughout the codebase, to basically mean "does the org have Slack connected?".

I was thinking of adding this to the `Organization` model:

```python
@property
def is_slack_connected(self) -> None:
  return self.slack_team_identity is not None
```

but if we feel that is appropriate, it should _probably_ be refactored in a separate PR

engine/apps/alerts/tests/test_notify_all.py
```diff
@@ -9,24 +9,18 @@
...
     make_alert_group,
 ):
-    organization = make_organization()
-    slack_team_identity = make_slack_team_identity()
+    organization, user, slack_team_identity, _ = make_organization_and_user_with_slack_identities()
     slack_channel = make_slack_channel(slack_team_identity)
-    organization.slack_team_identity = slack_team_identity
-    organization.save()
 
-    user = make_user(organization=organization)
```
joeyorlando, line 29: `make_organization_and_user_with_slack_identities` does the same thing as these 6 lines

engine/apps/alerts/models/channel_filter.py
joeyorlando: (I also refactored the two references to `slack_channel_id_or_org_default_id` to use `slack_channel_or_org_default`)

engine/apps/alerts/models/alert_group.py
```diff
@@ -1978,14 +1978,52 @@ def notify_in_slack_enabled(self):
...
+    # @property
+    # def slack_channel_id(self) -> str | None:
+    #     """
+    #     If the `AlertGroup` already has a `SlackMessage` associated with it, that means that we already
+    #     know the channel.
+
+    #     Otherwise, infer the Slack channel ID from the `ChannelFilter` associated with the `AlertGroup`.
+    #     """
+    #     return self.slack_channel.slack_id if self.slack_channel else None
+    # END TODO:
```
joeyorlando, line 2013: we can't really use this yet because we haven't yet migrated `SlackMessage._channel_id` (char field) to `SlackMessage.channel` (foreign key ref). I'm planning on doing this next week.

Once that is done, this becomes a lot cleaner because:
- `slack_message.channel` will be a `SlackChannel` (`_channel_id` hasn't been migrated to `channel`)
- `channel_filter.slack_channel` already was migrated
- `organization.default_slack_channel` already was migrated

In this case, we can just pass around `SlackChannel`s and not string references to channel IDs 🙂

engine/apps/alerts/models/alert_group.py
```diff
@@ -1978,14 +1978,52 @@ def notify_in_slack_enabled(self):
...
+
     @property
     def slack_channel_id(self) -> str | None:
         if not self.channel.organization.slack_team_identity:
             return None
```
joeyorlando, line 2018: again.. I'm not sure if this check needs to be here?

```python
if not self.channel.organization.slack_team_identity:
            return None
```

engine/apps/slack/scenarios/distribute_alerts.py
```diff
@@ -128,17 +129,29 @@ def process_signal(self, alert: Alert) -> None:
...
+                logger.info(
+                    f"Skipping posting message to Slack for alert_group {alert_group_pk} because we don't know which "
+                    f"Slack channel to post to. channel_filter={channel_filter} "
+                    f"organization.default_slack_channel={organization.default_slack_channel}"
+                )
+
+                alert_group.slack_message_sent = False
+                alert_group.reason_to_skip_escalation = AlertGroup.CHANNEL_NOT_SPECIFIED
+                alert_group.save(update_fields=["slack_message_sent", "reason_to_skip_escalation"])
+                return
```
joeyorlando, line 149: main change (TODO: comment why this is like this)

engine/apps/slack/tasks.py
```diff
@@ -303,14 +303,14 @@ def post_slack_rate_limit_message(integration_id):
         return
 
     default_route = integration.channel_filters.get(is_default=True)
-    if (slack_channel_id := default_route.slack_channel_id_or_org_default_id) is not None:
+    if (slack_channel := default_route.slack_channel_or_org_default) is not None:
```
joeyorlando, line 306: related to [this change](https://github.com/grafana/oncall/pull/5314/files#diff-5b78eae626bcf20076ff2f5f25daa3f5703ff8dd9336e5d3e3dbbd7aa827d6b2R175)

engine/apps/slack/scenarios/distribute_alerts.py
joeyorlando: going to try reproducing this theory locally just to be sure. Since I've been mucking around a lot in this file lately, decided to add more comprehensive test coverage in [engine/apps/slack/tests/scenario_steps/test_distribute_alerts.py](https://github.com/grafana/oncall/pull/5314/files#diff-4878c162db9a602b78ec80d84d36d1e936f12bad0b26a02b40d7b1910c9e92cb)

engine/apps/slack/scenarios/distribute_alerts.py
```diff
@@ -191,75 +263,6 @@ def process_signal(self, alert: Alert) -> None:
...
-        except SlackAPIChannelNotFoundError:
-            alert_group.reason_to_skip_escalation = AlertGroup.CHANNEL_ARCHIVED
-            alert_group.save(update_fields=["reason_to_skip_escalation"])
-            logger.info("Not delivering alert due to channel is archived.")
-        except SlackAPIRestrictedActionError:
-            alert_group.reason_to_skip_escalation = AlertGroup.RESTRICTED_ACTION
-            alert_group.save(update_fields=["reason_to_skip_escalation"])
-            logger.info("Not delivering alert due to workspace restricted action.")
-        finally:
-            alert.save()
```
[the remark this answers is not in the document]
joeyorlando: (I also just moved all of the logic that was in `_post_alert_group_to_slack` inside of `process_signal`, as it was more straightforward to do all of the exception handling there directly 🙂

engine/apps/slack/tests/scenario_steps/test_distribute_alerts.py
joeyorlando, line 1: since I've been mucking around a lot lately in `engine/apps/slack/scenarios/distribute_alerts.py`, I felt an obligation to add full test coverage for `IncomingAlertStep.process_signal`.. voila

engine/apps/slack/scenarios/distribute_alerts.py
```diff
@@ -128,26 +129,102 @@ def process_signal(self, alert: Alert) -> None:
...
-                if alert_group.channel_filter
-                # if channel filter is deleted mid escalation, use default Slack channel
-                else alert_receive_channel.organization.default_slack_channel
-            )
+            # if channel filter is deleted mid escalation, use the organization's default Slack channel
+            slack_channel = channel_filter.slack_channel if channel_filter else organization.default_slack_channel
+
+            # slack_channel can be None if
+            # - the channel filter is deleted mid escalation
+            # - AND the organization does not have a default slack channel set
```
Ferril, line 137: If I'm not mistaken, `channel_filter.slack_channel` also can be None if slack channel was not specify for the channel filter :thinking: We should probably use `slack_channel_or_org_default` here instead:
```suggestion
            slack_channel = channel_filter.slack_channel_or_org_default if channel_filter else organization.default_slack_channel

            # slack_channel can be None if
            # - the channel filter is deleted mid escalation
            # - AND the organization does not have a default slack channel set
```

engine/apps/slack/scenarios/distribute_alerts.py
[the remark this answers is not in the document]
Ferril: The main part makes a lot of sense to me 👍 
I'd propose to recheck the part with updating `slack_message_sent` field. We reset it to False _only_ if we reraise the exception 🤔 I think the idea here is to prevent attempts to post a message to slack in case when we are sure it's not possible (e.g. slack token error), because we call this step with every new alert in the alert group. So with every next alert in the alert group `num_updated_rows` should return `0`, and we should skip "post the first message" part. Does it make sense?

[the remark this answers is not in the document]
joeyorlando: good point, thanks for catching this 👍 updated in https://github.com/grafana/oncall/pull/5314/commits/62b5a85523baa559f1e349d85562cc94875481d9 (and updated the tests to assert that this value is set to the proper value, for all of the various `Exception`s that we handle here)

PR #5321, opened 2024-12-03
feat: add support for mattermost chatops

matiasb:
Related to https://github.com/grafana/oncall/issues/96


grafana-plugin/src/models/mattermost/mattermost_channel.ts
```diff
@@ -21,6 +26,28 @@ export class MattermostChannelStore extends BaseStore {
     this.path = '/mattermost/channels/';
   }
 
+  @action.bound
+  async updateMattermostChannels() {
+    const response = await makeRequest(this.path, {});
```
brojd, line 31: It would be good to provide the type in here to so we have a typed response `await makeRequest<MattermostChannel[]>`

grafana-plugin/src/models/mattermost/mattermost_channel.ts
```diff
@@ -21,6 +26,28 @@ export class MattermostChannelStore extends BaseStore {
...
+      {}
+    );
+
+    runInAction(() => {
+      this.items = {
+        ...this.items,
+        ...items,
+      };
+
+      this.currentTeamToMattermostChannel = response.map((mattermostChannel: MattermostChannel) => mattermostChannel.id);
```
brojd, line 47: The name of this property / observable is misleading imho. What it stores is a list of channel ids from the last request.

But anyway, it seems we're just using it to find out whether there are any channels already fetched, like that:
```
mattermostChannelStore.currentTeamToMattermostChannel?.length > 0;
```

The same can be done by checking if `Object.keys(MattermostChannelStore).length > 0`. So we probably don't need this observable at all?

grafana-plugin/src/models/mattermost/mattermost_channel.ts
```diff
@@ -65,6 +92,21 @@ export class MattermostChannelStore extends BaseStore {
...
+
+  async startAutoUpdate() {
+    this.autoUpdateTimer = setInterval(this.updateMattermostChannels.bind(this), 3000);
+  }
+
+  async stopAutoUpdate() {
+    if (this.autoUpdateTimer) {
+      clearInterval(this.autoUpdateTimer);
+    }
+  }
```
brojd, line 108: Those 3 methods, although I can see they also exist in msteams and telegram, are not used. We can remove them

grafana-plugin/src/models/mattermost/mattermost_channel.ts
```diff
@@ -75,4 +117,8 @@ export class MattermostChannelStore extends BaseStore {
   async deleteMattermostChannel(id: MattermostChannel['id']) {
     return super.delete(id);
   }
+
+  async getMattermostChannels() {
+    return super.getAll();
+  }
```
brojd, line 123: It seems that this one is also not used anywhere

grafana-plugin/src/pages/integration/Integration.helper.ts
```diff
@@ -68,8 +68,7 @@ export const IntegrationHelper = {
     const hasTelegram =
       store.hasFeature(AppFeature.Telegram) && store.telegramChannelStore.currentTeamToTelegramChannel?.length > 0;
     const isMSTeamsInstalled = Boolean(store.msteamsChannelStore.currentTeamToMSTeamsChannel?.length > 0);
-    const connectedChannels = store.mattermostChannelStore.getSearchResult();
-    const isMattermostInstalled = store.hasFeature(AppFeature.Mattermost) && Boolean(connectedChannels && connectedChannels.length)
+    const isMattermostInstalled = Boolean(store.mattermostChannelStore.currentTeamToMattermostChannel?.length > 0);
```
brojd, line 71: Can be just `store.mattermostChannelStore.currentTeamToMattermostChannel?.length > 0` (without `Boolean` part) and the result will be always the same

matiasb:
Thanks for the great work here @ravishankar15! :rocket:

PR #5347, opened 2024-12-07
Add mattermost alert group integration flow

ravishankar15:
Add comment

# What this PR does

Related to [issue link here]

## Checklist

### DIFF
--- a/engine/apps/mattermost/models/channel.py
+++ b/engine/apps/mattermost/models/channel.py
@@ -65,7 +65,10 @@ class MattermostChannel(models.Model):
         channel_id = alert_group.channel_filter.notification_backends[MattermostBackend.backend_id].get("channel")
         enabled = alert_group.channel_filter.notification_backends[MattermostBackend.backend_id].get("enabled")
 
-        if not enabled or not channel_id:
+        if not enabled:
+            return None
+
+        if not channel_id:
             return default_channel
 
         channel = cls.objects.filter(
--- a/engine/apps/mattermost/models/channel.py
+++ b/engine/apps/mattermost/models/channel.py
@@ -49,11 +49,33 @@ class MattermostChannel(models.Model):
 
     @classmethod
     def get_channel_for_alert_group(cls, alert_group: AlertGroup) -> typing.Optional["MattermostChannel"]:
+        from apps.mattermost.backend import MattermostBackend  # To avoid circular import
+
         default_channel = cls.objects.filter(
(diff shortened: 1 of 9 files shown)


engine/apps/mattermost/models/channel.py
```diff
@@ -49,11 +49,33 @@ class Meta:
...
+            alert_group.channel_filter is None
+            or not alert_group.channel_filter.notification_backends
+            or not alert_group.channel_filter.notification_backends.get(MattermostBackend.backend_id)
+        ):
+            return default_channel
+
+        channel_id = alert_group.channel_filter.notification_backends[MattermostBackend.backend_id].get("channel")
+        enabled = alert_group.channel_filter.notification_backends[MattermostBackend.backend_id].get("enabled")
+
+        if not enabled or not channel_id:
```
matiasb, line 68: Shouldn't we avoid sending a Mattermost notification if the setting is not enabled? So you need to explicitly enable and choose a channel (or leave the channel empty to use default).

engine/apps/mattermost/tests/test_backend.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize(
+    "input_data,expected_data",
+    [
+        ({"enabled": True}, {"enabled": True}),
+        ({"enabled": False}, {"enabled": False}),
+        ({"channel": None, "enabled": True}, {"channel": None, "enabled": True}),
+        ({"channel": None}, {"channel": None}),
```
matiasb, line 71: `input_data` and `expected_data` are always the same? I think you are missing some case to cover when you make changes to input during validation? (e.g. passing `"enabled": 1`? some of the other exceptional cases?)

matiasb (commented):
LGTM, minor comments added.

engine/apps/mattermost/tests/test_backend.py
ravishankar15: Yeah the idea of this test case is to cover the toggle option flow and the option to clear the channel once it is selected using the `X` on the dropdown. For other exception cases I have added separate tests [ref](https://github.com/grafana/oncall/pull/5347/files#diff-b9c01c665a1c5d4873783e707a541cab535839629a294638c1cde5d894ab8ca3R88). Do you see anyother case that needs to be covered in the tests ?

matiasb:
Thanks!

PR #5348, opened 2024-12-09
Inbound email: download from S3 + convert HTML to plaintext

vadimkerr:
# What this PR does

* Make `AmazonSESValidatedInboundWebhookView` able to download emails from S3 by providing AWS credentials via env variables
* Convert HTML to plaintext when there's only `text/html` available

Related to https://github.com/grafana/oncall-private/issues/2905

## Checklist


engine/apps/email/inbound.py
```diff
@@ -25,6 +27,15 @@ class AmazonSESValidatedInboundWebhookView(amazon_ses.AmazonSESInboundWebhookVie
...
 
+    def __init__(self):
+        super().__init__(
+            session_params={
+                "aws_access_key_id": settings.INBOUND_EMAIL_AWS_ACCESS_KEY_ID,
+                "aws_secret_access_key": settings.INBOUND_EMAIL_AWS_SECRET_ACCESS_KEY,
+                "region_name": settings.INBOUND_EMAIL_AWS_REGION,
+            },
+        )
+
```
vadimkerr, line 38: adding AWS creds so OnCall can download emails from S3

engine/apps/email/inbound.py
```diff
@@ -74,11 +85,10 @@ def dispatch(self, request):
         if request.method.lower() == "head":
             return HttpResponse(status=status.HTTP_200_OK)
 
-        integration_token = self.get_integration_token_from_request(request)
-        if integration_token is None:
+        if self.integration_token is None:
```
vadimkerr, line 88: a little bit of refactoring: instead of having a `self.get_integration_token_from_request(request)` method, it's now just a property `self.integration_token`. no functional changes, just looks better IMO

engine/apps/email/inbound.py
```diff
@@ -167,16 +179,105 @@ def check_inbound_email_settings_set(self):
...
-        subject = subject.strip()
-        message = email.text or ""
-        message = message.strip()
-        sender = self.get_sender_from_email_message(email)
+    @classmethod
+    def get_alert_payload_from_email_message(cls, email: AnymailInboundMessage) -> EmailAlertPayload:
+        if email.text:
+            message = email.text.strip()
+        elif email.html:
+            message = cls.html_to_plaintext(email.html)
```
vadimkerr, line 187: this is the main change of the PR. If there's no plaintext representation for an email, we'll convert it from HTML to plaintext in the `html_to_plaintext` method below

engine/apps/email/inbound.py
```diff
@@ -167,16 +179,105 @@ def check_inbound_email_settings_set(self):
...
+            message = ""
+
+        return {
+            "subject": email.subject.strip() if email.subject else "",
+            "message": message,
+            "sender": cls.get_sender_from_email_message(email),
+        }
+
+    @staticmethod
+    def html_to_plaintext(html: str) -> str:
```
joeyorlando, line 198: yikes, converting HTML to plaintext looks complicated! 😬

joeyorlando (approved):

engine/apps/email/inbound.py
```diff
@@ -167,16 +179,105 @@ def check_inbound_email_settings_set(self):
...
+        """
+        Converts HTML to plain text. Renders links as "text (href)" and removes any empty lines.
+        Converting HTML to plaintext is a non-trivial task, so this method may not work perfectly for all cases.
+        """
```
vadimkerr, line 202: There are ESPs such as Mailgun that provide the same `HTML -> plaintext` functionality but there are also ESPs such as AWS SES that don't do it. In order to keep things _relatively_ backward compatible when migrating from Mailgun to AWS SES, we want to convert HTML to plaintext in OnCall (and do it at least somewhat similarly to how Mailgun does it).

engine/apps/email/inbound.py
vadimkerr: Thank you for the super quick review @joeyorlando ❤️ 
Yes converting HTML to plaintext is non-trivial because HTML is basically a "picture", not text. I added this mostly for backward compatibility so I tried to keep it similar to how Mailgun does it. Also see https://github.com/grafana/oncall/pull/5348#discussion_r1890326476

PR #5360, opened 2024-12-12
feat: update service account auth not to require rbac enabled org

matiasb:

Also allow empty value for users in sync (instead of returning a 400 response).


engine/apps/api/permissions.py
```diff
@@ -175,7 +175,7 @@ def user_is_authorized(user: "User", required_permissions: LegacyAccessControlCo
     `required_permissions` - A list of permissions that a user must have to be considered authorized
     """
     organization = user.organization
-    if organization.is_rbac_permissions_enabled:
+    if organization.is_rbac_permissions_enabled or user.is_service_account:
```
joeyorlando, line 178: 👍

joeyorlando (approved):

PR #5415, opened 2025-01-14
feat: add filter affected services internal endpoint

matiasb:
Related to https://github.com/grafana/oncall-private/issues/2977

e.g.

`GET /api/plugins/grafana-oncall-app/resources/alertgroups/filter_affected_services?service=service-a&service=service-b`

```
[
  {"name": "service-a",
   "service_url": "http://localhost:3000/a/grafana-slo-app/service/service-a",
   "alert_groups_url": "http://localhost:3000/a/grafana-oncall-app/alert-groups?status=0&status=1&started_at=now-7d_now&label=service_name:service-a"}
]
```

### DIFF
--- a/engine/apps/api/tests/test_alert_group.py
+++ b/engine/apps/api/tests/test_alert_group.py
@@ -2413,3 +2413,44 @@ def test_filter_default_started_at(
     )
     assert response.status_code == status.HTTP_200_OK
     assert response.json()["pk"] == old_alert_group.public_primary_key
+
+
+@pytest.mark.django_db
+def test_alert_group_affected_services(
+    alert_group_internal_api_setup,
+    make_user_for_organization,
+    make_user_auth_headers,
+    make_alert_group_label_association,
+):
+    _, token, alert_groups = alert_group_internal_api_setup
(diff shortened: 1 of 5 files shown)


vstpme (approved):
LGTM

matiasb:
> More out of curiosity, was does "affected" mean in this context? It just means Alert Groups that have a `service_name` label, right?
> 
> Or is it more _robust_ than that and has to do w/ service dependencies? (ie. if Service A depends on Service B and Service B is "degraded", return me Service A).
> 
> Also, isn't `filter_affected_services` _mostly_ just doing label key:value filtering? (if so, don't we have an existing endpoint for this?)

Original implementation would find out upstream service dependencies for `service-a` and only return those with recent active alert groups (which could be related to a given alert group belonging to `service-a`). In our first implementation though, frontend will get all service dependencies and this endpoint will filter from there the ones which are affected only (meaning there are recent alert groups with active alert groups), besides returning some additional data (URLs).

Makes sense? Happy to rename as needed.

joeyorlando:
> frontend will get all service dependencies and this endpoint will filter from there the ones which are affected only

ahh okay, makes sense 🙂 thanks!

PR #5440, opened 2025-02-06
feat: add personal webhook form to user profile settings

mthorning:
Adds the personal webhook form to the user profile settings.

Related to https://github.com/grafana/irm/issues/332

recording

### DIFF
--- a/grafana-plugin/src/containers/UserSettings/parts/tabs/PersonalWebhookInfo/PersonalWebhookInfo.tsx
+++ b/grafana-plugin/src/containers/UserSettings/parts/tabs/PersonalWebhookInfo/PersonalWebhookInfo.tsx
@@ -13,7 +13,7 @@ import { ActionKey } from 'models/loader/action-keys';
 import { WebhookTriggerType } from 'models/outgoing_webhook/outgoing_webhook.types'
 import { useStore } from 'state/useStore';
 
-function useWebhooksOptions(): [boolean, { label: string, value: string }[]] {
+function useWebhooksOptions(): [boolean, Array<{ label: string, value: string }>] {
   const { outgoingWebhookStore, loaderStore } = useStore();
   const [isLoading, setIsLoading] = useState(true)
   const isLoadingWebhooks = loaderStore.isLoading(ActionKey.FETCH_WEBHOOKS)
--- a/grafana-plugin/src/containers/UserSettings/parts/tabs/PersonalWebhookInfo/PersonalWebhookInfo.tsx
+++ b/grafana-plugin/src/containers/UserSettings/parts/tabs/PersonalWebhookInfo/PersonalWebhookInfo.tsx
@@ -84,8 +84,10 @@ export const PersonalWebhookInfo = observer(() => {
 
   const user = userStore.items[userStore.currentUserPk];
   const selectedWebhook = watch('webhook');
-  const hasConnectedWebhook = user.messaging_backends.WEBHOOK != null;
   const [isLoadingOptions, webhookOptions] = useWebhooksOptions()
+
+  const hasConnectedWebhook = user.messaging_backends?.WEBHOOK != null;
+  const hasSelectedValidWebhook = webhookOptions.some(option => option.value === selectedWebhook)
 
   useEffect(() => {
     (async () => {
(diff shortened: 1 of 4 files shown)


grafana-plugin/src/containers/UserSettings/parts/tabs/PersonalWebhookInfo/PersonalWebhookInfo.tsx
```diff
@@ -1,27 +1,202 @@
...
+  }, [isLoadingWebhooks]);
+
+  useEffect(() => {
+    if (!isLoadingWebhooks && hasRegisteredLoadingWebhooks) {
+      setIsLoading(false);
+    }
+  }, [isLoadingWebhooks])
+
+  return isLoading;
+}
```
mthorning, line 40: `loaderStore.isLoading(ActionKey.FETCH_WEBHOOKS)` returns false on the first couple of renders which results in a flicker as the form is rendered briefly before showing the loading placeholder. I wrote this hook to prevent that but it feels like I shouldn't have to so perhaps I'm missing something here.

grafana-plugin/src/containers/UserSettings/parts/tabs/PersonalWebhookInfo/PersonalWebhookInfo.tsx
```diff
@@ -1,27 +1,202 @@
...
+
+  useEffect(() => {
+    (async () => {
+      await userStore.updatePersonalWebhook();
+      await outgoingWebhookStore.updateItems(
+        {
+          trigger_type: WebhookTriggerType.PersonalNotification.key,
+        },
+        true
+      );
```
mthorning, line 76: I wasn't sure about a lot of what I am doing with the stores. Is this the right thing to do here? I think I will be updating the items in the webhook store with only webhooks that are of the "personal notification" type. Could that cause issues to other pages?

grafana-plugin/src/models/user/user.ts
```diff
@@ -30,6 +30,7 @@ export class UserStore {
   notifyByOptions: any = [];
   currentUserPk?: ApiSchemas['User']['pk'];
   usersCurrentlyBeingFetched: { [pk: string]: boolean } = {};
+  personalWebhook: { webhook: string, context: object };
```
mthorning, line 33: I really wasn't sure if I should add this here or if it should be in a `useState` in PersonalWebhookInfo.

grafana-plugin/src/models/user/user.ts
```diff
@@ -304,4 +305,28 @@ export class UserStore {
...
+    await this.updatePersonalWebhook();
+  }
+
+  async updatePersonalWebhook() {
+    const response = await makeRequest('/webhooks/current_personal_notification');
+
+    runInAction(() => {
+      this.personalWebhook = response;
+    });
+  }
```
mthorning, line 324: I also wasn't sure if I should fetch this in `loadCurrentUser` or only when the user visits the webhook tab in the settings.

grafana-plugin/src/containers/UserSettings/parts/tabs/PersonalWebhookInfo/PersonalWebhookInfo.tsx
```diff
@@ -1,27 +1,208 @@
...
+      isValid,
+      errors,
+    },
+    watch,
+    control,
+    getValues,
+    handleSubmit,
+    reset,
+  } = useForm<FormFields>({
+    mode: 'onBlur',
```
brojd, line 81: I think our standard mode is 'onChange'

grafana-plugin/src/containers/UserSettings/parts/tabs/PersonalWebhookInfo/PersonalWebhookInfo.tsx
brojd: Maybe we can chat about this one, I'll ping u

grafana-plugin/src/models/user/user.ts
brojd: I think you need to assign the default value `... = {}`, otherwise we experienced that MobX doesn't create an observable

brojd (approved):

PR #5456, opened 2025-02-16
update hashing functions from md5 to sha256

kranurag7:
# What this PR does

updates all instances of md5 to sha256.

N/A (there were no open issues about this) 

## Checklist

### DIFF
--- a/engine/apps/alerts/models/alert.py
+++ b/engine/apps/alerts/models/alert.py
@@ -276,7 +276,7 @@ class Alert(models.Model):
             group_distinction = cls.insert_random_uuid(group_distinction)
 
         if group_distinction is not None:
-            group_distinction = hashlib.md5(str(group_distinction).encode()).hexdigest()
+            group_distinction = hashlib.sha256(str(group_distinction).encode()).hexdigest()
 
         if (
             resolve_condition_template := template_manager.get_attr_template("resolve_condition", alert_receive_channel)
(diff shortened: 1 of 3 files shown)


mamccorm:
+1

Konstantinov-Innokentii:
Hi! Why this change is needed? md5 is not used to hash anything sensitive, like passwords or something. It hashes only alert group id, which is used solely for grouping.
Also, It looks like this change will break grouping for open alert groups, because new alert, which might belong to it, will have different group id, because of different hash func.

maxgio92:
I think FIPS compliant environment with strict enforcement in OpenSSL would benefit from this change

Konstantinov-Innokentii:
> I think FIPS compliant environments with strict enforcement in OpenSSL would benefit from this change, as `hashlib.md5()` would fail otherwise
Thanks for the clarification. Is it possible to implement it under the feature flag or something, so users who do not need it will still use md5 & they will not receive unexpected pages during the update?

kranurag7:
> Additionally, this change seems likely to break grouping for open alert groups. Since the new hash function would generate different group distinction, new alerts that should belong to an existing group might end up in a separate one

Hey @Konstantinov-Innokentii I would like to understand more of why it's a breaking change, changing to sha256 should be safe here right. I am thinking more on the line that md5 gives you random keys and sha256 will give you same randomness but a little longer. Is there something more detailed that I need to understand here in case I missed something here.

Konstantinov-Innokentii:
Hey @kranurag7!
So, the initial state of the system is:
1. OnCall is running with md5 hash used for grouping
2. There is an open AlertGroup; let's assume it's grouped by field **payload.system**
So, the PR is merged, we released a new oncall version and oncall instance version is updated
1. New alert with same value of **payload.system** is received.
2. The new version of the code calculated group distinction using sha256 instead of md5.
3. These values are different despite the **payload.system** being the same. So, instead of grouping the alert into the open group, OnCall will create a new group. Which may cause some unexpected paging.

That's why I'm asking why this change is needed & if it's for some security compliance – is it possible to use sha256 only under feature flag.

PR #5495, opened 2025-03-12
feat: add opsgenie to migrator script

joeyorlando:
This PR adds support for migrating data from OpsGenie to Grafana IRM.

Closes https://github.com/grafana/irm/issues/1179


tools/migrators/lib/tests/pagerduty/resources/test_services.py
joeyorlando, line 1: moved from `tools/migrators/lib/tests/pagerduty/test_services.py`

tools/migrators/lib/pagerduty/migrate.py
joeyorlando, line 1: just some refactoring. filtering functions moved to their respective file under `lib/pagerduty/resouces`

tools/migrators/lib/session.py
joeyorlando, line 1: related to [this change](https://github.com/grafana/oncall/pull/5495/files#diff-f80b7dc844e6877cf041fcdf9594833110a983259326e64cb8858a171d0c8f63R20)

joeyorlando (commented):
main changes:
- OpsGenie migration capabilities. capabilities + caveats have been documented + unit-tested
- `User-Agent` header now set when communicating with OnCall API (anonymous usage data)

related to the changes in `lib/pagerduty`, there is also a bit of shuffling-around of some code in this PR. That is simply shuffling-around/de-duplication, no functional difference on the PagerDuty migrator side of things.

tools/migrators/lib/opsgenie/resources/integrations.py
```diff
@@ -0,0 +1,63 @@
...
+    if integration["oncall_integration"]:
+        OnCallAPIClient.delete(
+            f"integrations/{integration['oncall_integration']['id']}"
+        )
+
+    # Create new integration
+    payload = {
+        "name": integration["name"],
+        "type": integration["oncall_type"],
+        "team_id": None,
```
matiasb, line 57: Just to be sure, teams are not migrated, right? (so users and resources are not really linked to a team after migration, except by the "name convention")

matiasb (approved):
I haven't done an extensive review but at first look things look ok.

joeyorlando:
@matiasb correct. at the moment none of the migrator script variants migrate teams due to the sync (e.g. you create the Grafana team, but the team isn't immediately available in OnCall).

in a future PR we might be able to implement this by:
1. modifying the public API to trigger a synchronous team sync by looking at the newly added migrator User-Agent header (similar to the terraform behaviour)
2. we would also need to then fetch the list of teams and build a map of OnCall -> <PD|Splunk|OpsGenie> team, and pass that into the migrate functions which reference teams, to know which OnCall team to assign to the newly created objects

tools/migrators/lib/opsgenie/resources/integrations.py
joeyorlando: @matiasb correct. at the moment none of the migrator script variants migrate teams due to the sync (e.g. you create the Grafana team, but the team isn't immediately available in OnCall).

in a future PR we might be able to implement this by:
1. modifying the public API to trigger a synchronous team sync by looking at the newly added migrator User-Agent header (similar to the terraform behaviour)
2. we would also need to then fetch the list of teams and build a map of OnCall -> <PD|Splunk|OpsGenie> team, and pass that into the migrate functions which reference teams, to know which OnCall team to assign to the newly created objects

PR #5524, opened 2025-05-01
chore: remove "expensive" e2e tests

joeyorlando:
These CI workflow steps are never invoked (we no longer run the "expensive" e2e tests), doing some clean-up


willgallego-grafana (approved):
👍

PR #5551, opened 2025-06-13
fix: Improve normalization of service_name

musculman:
# What this PR does

Improves the normalization of `service_name`

### DIFF
--- a/tools/migrators/lib/pagerduty/resources/services.py
+++ b/tools/migrators/lib/pagerduty/resources/services.py
@@ -450,12 +450,13 @@ def _transform_service(
     is_technical = isinstance(service, TechnicalService)
     service_type = "service" if is_technical else "business_service"
 
-    service_name = (
-        re.sub("([^-a-zA-Z0-9 .])", "-", service.name)
-        .removesuffix("-")
-        .removeprefix("-")
-        .replace(" ", "-")
-    )
+    service_name = re.sub("[^-a-zA-Z0-9.]", "-", service.name)
+    service_name = re.sub(
+        "-+", "-", service_name
+    )  # Collapse multiple dashes to single dash
+    service_name = re.sub(
+        "^-+|-+$", "", service_name
+    )  # Remove leading and trailing dashes
     # Create the base component structure
     component = {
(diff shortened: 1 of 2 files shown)


tools/migrators/lib/pagerduty/resources/services.py
```diff
@@ -450,14 +450,18 @@ def _transform_service(
...
+    )
     # Create the base component structure
     component = {
         "apiVersion": "servicemodel.ext.grafana.com/v1alpha1",
         "kind": "Component",
         "metadata": {
-            "name": service.name.lower().replace(
-                " ", "-"
-            ),  # Convert to k8s-friendly name
+            "name": service_name.lower(),  # Convert to k8s-friendly name
```
willgallego-grafana, line 464: This can technically end in a number. If the k8s friendly part is that the service name is made a k8s label, I think that might fail? Likewise, do we have a character limit?

I'm not 100% sure on this and it might be slightly beyond the scope of the PR.

tools/migrators/lib/pagerduty/resources/services.py
```diff
@@ -450,14 +450,18 @@ def _transform_service(
     is_technical = isinstance(service, TechnicalService)
     service_type = "service" if is_technical else "business_service"
 
+    service_name = (
+        re.sub("([^-a-zA-Z0-9 .])", "-", service.name)
+        .removesuffix("-")
+        .removeprefix("-")
+        .replace(" ", "-")
```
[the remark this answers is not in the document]
willgallego-grafana: Note: I have `service_name` in places where it should be `service.name`, like in the first regex replace. The above is example code

tools/migrators/lib/pagerduty/resources/services.py
joeyorlando, line 1: _nit_ maybe add a quick test case [here](https://github.com/grafana/oncall/blob/dev/tools/migrators/lib/tests/pagerduty/resources/test_services.py#L278-L324)?

jdbaldry (commented):
The changes to the workflow look fine

PR #5576, opened 2025-09-24
chore: update faro to 1.19.0

eskirk:
upgrading faro deps to 1.19.0

covering our bases to see if this is due to some plugins with old faro deps caused [this escalation](https://grafanalabs.enterprise.slack.com/archives/C09GWJ20FHU)


mthorning (approved):
LGTM

PR #5611, opened 2025-12-16
chore: refactoring and upgrade telegram/urllib3 deps

matiasb:


engine/apps/telegram/utils.py
```diff
@@ -1,14 +1,98 @@
...
+        )
+
+    # Use asyncio.run() which properly manages event loop lifecycle
+    # With nest_asyncio applied, this will work even if there's a loop in the thread
+    try:
+        return asyncio.run(coro)
+    except RuntimeError as e:
+        error_str = str(e)
+        # Handle "Event loop is closed" or similar errors
+        if "closed" in error_str.lower() or "Event loop" in error_str:
```
willgallego-grafana: I don't have a better solution but this does feel a bit brittle

